PR #41, opened 2021-11-30
Finish initial Sidecar and improve responsiveness

ashleemboyer:
🔎  [Preview link](https://dev-portal-git-ambimprove-sidecar-and-responsiveness-hashicorp.vercel.app/waypoint/docs)
🎟️  Asana tickets:
- [Collapse Sidecar in narrower views](https://app.asana.com/0/1201325088736660/1201363188349590/f)
- [Finish Sidecar component](https://app.asana.com/0/1201325088736660/1201265683986463/f)

## What

- Added a `DeviceSizeProvider` and `useDeviceSize` hook for consuming a state variable that stores the following booleans based off two `matchMedia` queries:
  - `isDesktop`
  - `isMobile`
  - `isTablet`
- Added 2 CSS variables for breakpoint variables:
  - `--mobile-width-breakpoint`
  - `--tablet-width-breakpoint`
- Added 2 helper classes for hiding content on mobile or tablet:
  - `.hide-on-mobile`
  - `.hide-on-tablet`
- Added `50vh` padding to the bottom of the `<main>` element to improve the viewing experience at the bottom of the page and make it possible to scroll to the last heading
- Added `use-active-section` hook that comes from https://github.com/hashicorp/react-components/pull/325 (thank you @BRKalow for helping me get this code in here! 👏🏻)
  - This hook helps us figure out which heading is currently being viewed so we can give that item an active state in `Sidecar`
  - The `isEnabled` parameter prevents the code from running as an optimization since we have to instantiate this hook even when we know we're going to hide the sidebar on smaller widths. `Sidecar` uses `useDeviceSize` to pass the parameter.
- Updated styles to match the latest designs (e.g. default, active, hover, & focus states)
- Added a `getTruncatedTitle` helper that truncates the `Sidecar` text on full words with a maximum of 25 characters. This character count is a number Brayden helped me determine and we agree it's a good start that can be tweaked later.


src/pages/_app.tsx
```diff
@@ -11,10 +12,12 @@ export default function App({ Component, pageProps }) {
   const Layout = Component.layout ?? BaseLayout
 
   return (
-    <ErrorBoundary FallbackComponent={Error}>
-      <Layout {...pageProps?.layoutProps}>
-        <Component {...pageProps} />
-      </Layout>
-    </ErrorBoundary>
+    <DeviceSizeProvider>
```
ashleemboyer, line 15: TODO: move this to be a child of `ErrorBoundary`

src/components/sidecar/use-active-section.ts
```diff
@@ -0,0 +1,100 @@
...
+              return headings[newIndex].slug
+            })
+          }
+        }
+
+        if (currentY) {
+          previousY.current = currentY
+        }
+      },
+      { rootMargin: '0% 0% -75% 0%' }
```
ashleemboyer, line 83: I think we need to tweak this number some more, here's one example where the next heading is highlighted after clicking on a link: https://dev-portal-git-ambimprove-sidecar-and-responsiveness-hashicorp.vercel.app/waypoint/docs#getting-started

src/layouts/docs/docs-layout.module.css
```diff
@@ -46,6 +46,7 @@
 .main {
   max-width: 804px;
   padding: 0 24px;
+  padding-bottom: 50vh;
```
alexcarpenter, line 49: I see the comment about the reasoning to have this but this much empty space makes it feel like something isn't rendering as expected here.

src/components/sidecar/use-active-section.ts
alexcarpenter: Noticing Getting started doesn't get hit here. Not sure if that'd be related/fixed in a root margin change 

https://user-images.githubusercontent.com/825855/146246232-d8ebbd87-85a7-497a-a686-c7e3a52cc007.mp4

alexcarpenter (commented):
CSS and markup look good, will leave the functionality review to bryce. Just had a comment about the usage of `50vh`. Wonder if there is anything we can pull from turbo repo. I see they are doing similar functionality and now its open source. https://turborepo.org/docs/features/pipelines

BRKalow:
Here's the turbo intersection logic: https://github.com/vercel/turborepo/blob/f160c418c9e564659d462f84fa2721ec1823e0a4/docs/nextra-theme-docs/misc/theme.js#L104

src/components/sidecar/use-active-section.ts
ashleemboyer: Yes, that should be fixed with the root margin change! It's related at the very least

src/layouts/docs/docs-layout.module.css
ashleemboyer: Hadn't thought about that. My personal reasons for liking it is because I like having content at the top of my screen when reading. But I'd really like to hear from more people on this and if they've seen this addressed in some other way!

ashleemboyer:
> Here's the turbo intersection logic: https://github.com/vercel/turborepo/blob/f160c418c9e564659d462f84fa2721ec1823e0a4/docs/nextra-theme-docs/misc/theme.js#L104.
> 
> It looks like they put the intersection observer onto the heading element using a custom MDX component 💭 , but overall it looks pretty similar.

@BRKalow I saw that they have `aria-hidden` on their span that renders a `#`, I added that to the active indicator in my code! Now to read the other code we're focused on lol

src/contexts/device-size.tsx
```diff
@@ -0,0 +1,97 @@
...
+    const isTablet = tabletMediaQueryListObject?.matches
+    const isDesktop = !(isMobile || isTablet)
+    return {
+      isDesktop,
+      isMobile,
+      isTablet,
+    }
+  }
+
+  const [value, setValue] = useState<DeviceSize>(() => getDerivedState())
```
BRKalow, line 65: Just calling out that if we do use this for hiding components instead of CSS we'll get hydration mismatches, e.g.:
Screen Shot 2021-12-16 at 10 25 36 AM

(I just added `if (!isDesktop) return null` to validate the above ☝️, we aren't using it in this way as part of this PR though)

ashleemboyer: This would be perfect to note in a Usage section of a README for this hook! I'm going to make an Asana ticket for documenting the the Context/hook and will link to this comment so we don't lose track of it.

src/layouts/docs/docs-layout.module.css
BRKalow: This is a fair point by Alex. I think it's probably worth raising to the broader group. If we do want to keep it around, maybe there's some way we can make it feel more intentional 💭

BRKalow (approved):
LGTM! We can tweak the active state behavior and other things as a follow-up

ashleemboyer:
Here's a link to the Asana task I created for following up on the active state behavior: https://app.asana.com/0/0/1201508675437098/f

src/layouts/docs/docs-layout.module.css
ashleemboyer: I also remembered this morning that there will be a footer component added directly below the content/sidecar portion of the page, I wonder if this will help alleviate what we're observing with empty space? Link to beta footer design here: https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Dev-Portal?node-id=2835%3A60851

PR #42, opened 2021-12-08
feat: refine preview modes, add sentinel website

zchsh:
This PR fixes an issue with partials location resolution in `docs-page`. It also refactors redirects, such that they're loaded from the remote product repositories (as we've decided to retain redirect config in product repos, since authors edit them quite often - eg when moving docs pages around).

This PR also makes other preview mode refinements, such as updating env variables to match the recent MKTG-040 RFC.

And, this PR adds the `sentinel` website into the mix - which will be served from https://test-st.hashi-mktg.com.


zchsh:
Power merging this in, still proving our preview modes, and want to get a https://test-st.hashi-mktg.com demo up ASAP 👍

PR #43, opened 2021-12-10
Initial ProductSwitcher

ashleemboyer:
🔎  [Preview link](https://dev-portal-git-ambinitial-product-switcher-hashicorp.vercel.app/waypoint/docs)
🎟️  [Asana task](https://app.asana.com/0/1201325088736660/1201247589988629/f)

## What, in general

- Added a new file for common types: `src/common/types.ts`
  - This is [an approach I found from basarat](https://stackoverflow.com/questions/36633033/how-to-organize-typescript-interfaces/36636368#36636368) that I'm fond of. It's a nice place to have the most used types defined in, such as types for the entities that exist in an app.
- Updated `ProductIcon` (`src/components/icons/product-icon/index.tsx`) to have more icons and use `@hashicorp/flight-icons` instead of `@hashicorp/mktg-logos`
- Abstracted the repetitive focus styles (border + box-shadow - outline) into a helper class called `.focus-border-and-box-shadow`, which I placed in `src/pages/style.css`

## Specific `ProductSwitcher` functionality

These are the kind of things I would like to detail in documentation, but not sure if we should include a README for now since this hasn't been design reviewed yet and could very well change dramatically.

- Opens & closes on click
- Closes when there is a click on the page outside of the component
  - Does not close if you click somewhere else in the window (`onBlur` + keyboard nav is just a little more complicated than what I've implemented here but willing to consider the idea if we need the switcher to close in this case)
- Closes when `Escape` is pressed & re-focuses the activator button
- When the activator button is focused and either `Space` or `Enter` are pressed:
  - the switcher opens
  - the first option is focused
- Keyboard navigation within the list of options:
  - If the last item is focused, `Tab` moves focus to the first option. Otherwise, `Tab` moves focus to the next option.
  - If the first item is focused, `Shift+Tab` moves focus to the last option. Otherwise, `Shift+Tab` moves focus to the previous option.
  - Notes on this approach:
    - I've chosen to stick with `Tab` & `Shift+Tab` instead of implementing arrow key keyboard navigation because this is a navigation list, which users are more than likely used to navigating with the `Tab` & `Shift+Tab` keys. If this was a select input, I would opt for implementing arrow keys for navigation between options.
    - I've chosen to make the focus within options wrap because that was easier to implement IMO than making the switcher close when `Tab`-ing/`Shift+Tab`ing away or only preventing default behavior to trap but not wrap focus in the option list.

## ARIA details

- General note: the following were created to make referenced IDs programmatic instead of manual. This makes intentions clear and reduces the possibility for regressions if there are refactors.
  - `OPTION_LIST_ID`
  - `OPTION_ID_PREFIX`
  - `generateSwitcherOptionIdFromProduct`
- Setting `aria-current` on the anchor element associated with the current product docs being viewed
  - This conveys to screen readers that the link is for the current page
  - We can use the selector in CSS for special styling
- Setting `aria-expanded` conveys to screen readers whether or not the switcher is open when the activator button is focused
- `aria-controls`connects the activator button with the option list element that it controls the visibility of. It's not clear if we need this since it's most useful when the two elements are further apart from each other in the document, and because it's purpose is to enable the user to automatically move their focus to the controlled element, which we are doing in the JavaScript on `Space/Enter` on the activator button.
- `aria-labelledby` gives a label to the activator button based on the anchor elements in the option list. Their accessible labels are calculated to be the name of the product they're for. I think we can improve upon this label to make it more clear that this component is a navigation element, but this seems like a good starting point.

## Questions

In addition to question TODOs I've added in the code:

- What logos are we supposed to use for Sentinel and HashiCorp Cloud Platform in `ProductSwitcher`? For now I've used the HashiCorp logo to have something in place.

## Relevant Resources

- [Link + Disclosure Widget Navigation](https://adrianroselli.com/2019/06/link-disclosure-widget-navigation.html) - _Adrian Roselli_
- [Using the aria-controls attribute](https://tink.uk/using-the-aria-controls-attribute/) - _Léonie Watson_


src/components/icons/product-icon/index.tsx
```diff
@@ -1,12 +1,24 @@
...
 interface ProductIconProps {
-  product: 'vault'
+  product: ProductCode
 }
 
-const productNamesToIcons = {
-  vault: require('@hashicorp/mktg-logos/product/vault/logomark/white.svg?include'),
+// TODO: what logos for hcp & sentinel?
+const productNamesToIcons: { [key in ProductCode]: any } = {
+  boundary: require('@hashicorp/mktg-logos/product/boundary/logomark/color.svg?include'),
```
alexcarpenter, line 12: Product logos are now included in Flight icons https://github.com/hashicorp/flight/pull/341

ashleemboyer: Oh nice!! I'll work on using these instead 😊

src/components/product-switcher/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+  return `${OPTION_ID_PREFIX}${product.code}`
+}
+
+const getAllAnchorElements = () => {
+  const listElement = document.getElementById(OPTION_LIST_ID)
+  const anchorElements = listElement.querySelectorAll('a')
+  return anchorElements
+}
+
+const ProductSwitcher: React.FC = () => {
```
BRKalow, line 82: Just thinking out loud here...there's a lot of (good) code here to ensure we've got all of our accessibility requirements handled, did we look assess any of the Reach components to see if we might be able to get some of that out of the box? This one looks like it could potentially be valuable: https://reach.tech/menu-button

I see your comment about not using `role="menu"`, so maybe that's why we're doing this custom 💭

src/components/product-switcher/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+        return
+      }
+
+      const isTabbingForward = !e.shiftKey && e.key === 'Tab'
+      const isTabbingBackward = e.shiftKey && e.key === 'Tab'
+      if (!(isTabbingForward || isTabbingBackward)) {
+        return
+      }
+
+      const anchorElements = getAllAnchorElements()
```
BRKalow, line 174: It looks like we are only concerned with focusing the first and last elements in the list. We could use two static refs for this, and that way we wouldn't need to query the DOM when we do need to put focus elsewhere.

```tsx
const firstAnchorRef = useRef()
const lastAnchorRef = useRef()

// ...

let ref = null
if (isFirstItem) ref = firstAnchorRef
if (isLastItem) ref = lastAnchorRef

// ...

<a
  ref={ref}
  // ...
```

src/components/product-switcher/style.module.css
```diff
@@ -0,0 +1,104 @@
...
+*/
+.switcherOption {
+  &:not(:last-child) {
+    margin-bottom: 3px;
+  }
+}
+
+/*
+Two items to note:
+  - transparent border on default is to prevent layout shift on focus
```
BRKalow, line 67: Nice comment ⭐

src/components/sidecar/style.module.css
```diff
@@ -30,17 +30,12 @@
...
-    padding: 0 10px;
-
-    &:focus {
-      border: 1px solid var(--blue-vibrant);
-      box-shadow: 0 0 0 3px #a5c2f3; /* TODO: replace w/ variable when available */
-      outline: none;
-    }
-  }
+.sidecarListItemAnchor {
+  composes: focus-border-and-box-shadow from global;
```
BRKalow, line 36: 💅

src/components/product-switcher/style.module.css
```diff
@@ -0,0 +1,104 @@
...
+  margin-right: 4px;
+}
+
+/*
+FIXME: box-shadow value is copied from Figma, but it's not rendering in the
+browser the same as it does in Figma.
+*/
+.switcherOptionList {
+  background-color: white;
+  border-radius: 5px;
```
BRKalow, line 42: Design observation...the button radius is 4px, while this is 5px...I wonder if we could align on 4 for this element 💭

src/components/product-switcher/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+import classNames from 'classnames'
+import { IconCaret16 } from '@hashicorp/flight-icons/svg-react/caret-16'
+import { Product } from 'common/types'
+import ProductIcon from 'components/icons/product-icon'
+import s from './style.module.css'
+
+const OPTION_LIST_ID = 'product-chooser-option-list'
+const OPTION_ID_PREFIX = 'product-chooser-list-item-'
+
+// TODO: should we put this somewhere else for easier reuse?
```
BRKalow, line 18: We do have a `config/` folder, which might be an appropriate spot for something like this.

BRKalow (commented):
Looking good. As we start to add components with more complex interaction / accessibility concerns, it might help us in the long-run to ensure we're starting to add tests to cover the core behavior.

src/common/types.ts
```diff
@@ -0,0 +1,29 @@
...
+  | 'boundary'
+  | 'consul'
+  | 'hcp'
+  | 'nomad'
+  | 'packer'
+  | 'sentinel'
+  | 'terraform'
+  | 'vagrant'
+  | 'vault'
+  | 'waypoint'
```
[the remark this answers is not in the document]
ashleemboyer: I like that! `code` was a totally random decision on my part. Glad to use something more people are familiar with. 😊

src/components/product-switcher/index.tsx
ashleemboyer: Yep! That's exactly why I didn't use that component. 👍🏻

src/components/product-switcher/index.tsx
alexcarpenter: Helpful hook for managing active indexes that wrap https://github.com/souporserious/use-roving-index

ashleemboyer: This is such a great idea! 🙌🏻  I'll work on making this change today.

src/components/product-switcher/style.module.css
ashleemboyer: Good catch, I'll ask about this and double check that this is what we want!

src/components/product-switcher/index.tsx
ashleemboyer: Thoughts on putting in a new file like `products.ts`? 🤔

src/components/product-switcher/index.tsx
ashleemboyer: Ha, @alexcarpenter and I replied at almost the same time. 😆  My previous comment was directed at @BRKalow, but I will definitely take a look at that hook Alex! I've not seen it before. :til:

alexcarpenter: @ashleemboyer ha no problem, just sharing as a reference. Managing indexes can be common, so a shared hook might be a helpful abstraction. No worries if not.

src/components/product-switcher/index.tsx
alexcarpenter: Not sure if `@hashicorp/platform-product-meta` is helpful to build off of here? 
CleanShot 2021-12-13 at 12 52 01@2x

src/components/product-switcher/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+        aria-expanded={isOpen}
+        aria-labelledby={generateSwitcherOptionIdFromProduct(currentProduct)}
+        className={s.switcherButton}
+        onClick={() => {
+          setIsOpen(!isOpen)
+        }}
+        onKeyDown={handleKeyDown}
+        ref={buttonRef}
+      >
+        <div className={s.switcherOptionContainer}>
```
alexcarpenter, line 225: Can we use a span here? `div`'s are not recommended as children of `button` elements. Notice `ProductIcon` returns with wrapping div as well.
CleanShot 2021-12-13 at 16 56 51@2x

src/components/product-switcher/index.tsx
ashleemboyer: The only thing I can think of there is that it doesn't include `"sentinel"`. Should we update that to include it?

src/components/product-switcher/index.tsx
ashleemboyer: Oof! Wasn't even thinking about this. Great catch. 👀

src/components/product-switcher/index.tsx
ashleemboyer: Oh, also not seeing a HashiCorp Cloud Platform which we need to have in the list of products: https://github.com/hashicorp/web-platform-packages/blob/main/packages/product-meta/index.tsx. I incorrectly assumed that's what the `"hashicorp"` one was for.

alexcarpenter: You can extend it if you'd like, something like: 
CleanShot 2021-12-14 at 09 15 39@2x

src/components/product-switcher/index.tsx
```diff
@@ -0,0 +1,208 @@
...
+        </div>
+        <IconCaret16 className={s.switcherCaret} />
+      </button>
+      <ul
+        className={s.switcherOptionList}
+        id={OPTION_LIST_ID}
+        onKeyDown={handleKeyDown}
+        style={{ display: isOpen ? 'block' : 'none' }}
+      >
+        {products.map(renderProductGroup)}
```
alexcarpenter, line 202: Looks like we're missing wrapping `li` around each group? Nested `ul's contain default margin 
CleanShot 2021-12-15 at 16 51 30@2x

ashleemboyer: Sorry for the confusion! This isn't ready for review yet anymore. I've converted it back to a draft. I'm currently working on converting this to dark mode and also to include horizontal rules.

src/components/product-switcher/index.tsx
ashleemboyer: Fixed! And double-checked with this validator. 😊

src/components/product-switcher/index.tsx
ashleemboyer: Two updates:

- Used some of this TS beauty @alexcarpenter  shared in `src/common/types.ts` :ts-da:
- Added a `config/products.ts` file (uses TS things from `src/common/types.ts`

src/components/product-switcher/style.module.css
ashleemboyer: I asked about this and we do indeed want consistent values! The value is supposed to be `5px` and I've made the updates.

src/components/product-switcher/index.tsx
ashleemboyer: Well, since I've added groups within the the list now, I don't think the hook is going to help us anymore. But out of curiosity and to make sure I'm understanding it correctly, it seems like it's most useful when we have to manage the focus of items that don't have default handling? In this instance I just want to override the handling of two items and don't have to manage how focus works for all of them.

ashleemboyer: Also, I've updated the code to work with 2 refs now instead of querying the document for anchors!

ashleemboyer:
@BRKalow @alexcarpenter I've pretty much re-done both the styling and the functionality, so it's probably worth doing a whole new code review when you get the chance. 🙏🏻

ashleemboyer:
> Looking good. As we start to add components with more complex interaction / accessibility concerns, it might help us in the long-run to ensure we're starting to add tests to cover the core behavior.

@BRKalow I definitely want to test and document things that get complex like this. It's tricky to know when to do that since we could end up re-doing the code completely. What are your thoughts on when we should add tests? As these components for January 5th get reviewed, I don't think they'll be changing much at all. I think if the components don't change incredibly after our first round of feedback from that January 5th release, that's when we should implement tests/docs to make sure we maintain functionality.

src/common/types.ts
```diff
@@ -0,0 +1,12 @@
+import { ProductMeta, Products } from '@hashicorp/platform-product-meta'
```
BRKalow, line 1: Bike-shedding a bit here....but I would maybe prefer a `src/types` directory and then expose meaningful submodules, e.g. for this stuff `src/types/products.ts` feels more intuitive.

Some immediate thoughts that led me here: What else would go in common? What if this types file gets really big? What if we want to create separate types files just to make it a little more discoverable?

ashleemboyer: I'm cool with this, I'll make the change! I really like your point about discoverability.

BRKalow (approved):

PR #45, opened 2021-12-14
Initial main navigation links

ashleemboyer:
🔎  [Preview link](https://dev-portal-git-ambinitial-main-nav-links-hashicorp.vercel.app/waypoint/docs)
🎟️  [Asana ticket](https://app.asana.com/0/1201325088736660/1201247589988627/f)

## What

- Added a `config/main-navigation.ts` file for housing the list of navigation links rendered.
- Renamed & added CSS variables to match the design token variable names. Hoping this will ease the transition into use the tokens when they're ready. 🤞🏻(Also, added a comment about this in the file.)
- Updated some styles that had become out-of-date.
- Rendering the links with either `Link` or just `<a>` (see `NavLink` component & comment) and styling the current link using the `aria-current='page'` selector


ashleemboyer:
TODO: add focus styles to the links. Designs: https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Dev-Portal?node-id=2683%3A55352

src/components/navigation-header/style.module.css
```diff
@@ -33,20 +33,36 @@
   list-style: none;
   margin-bottom: 0;
   margin-left: 0;
-  margin-right: 56px;
+  margin-right: 72px;
   margin-top: 0;
 
   & li {
-    padding: 8px 12px;
+    padding: 6px 8px;
```
alexcarpenter, line 40: Looks like our padding is off a bit from design. Looks a little shorter on the y axis.

CleanShot 2021-12-15 at 16 36 45@2x

CleanShot 2021-12-15 at 16 36 18@2x

alexcarpenter (commented):
Just noticing some height differences then what I am seeing in designs.

src/components/navigation-header/style.module.css
ashleemboyer: Ah, I was looking at the wrong page in designs. I'll address this!

config/main-navigation.ts
```diff
@@ -0,0 +1,37 @@
...
+    label: 'Plugins',
+  },
+  {
+    href: 'https://www.waypointproject.io/downloads',
+    id: 'downloads',
+    label: 'Downloads',
+  },
+]
+
+export { navigationLinks }
```
BRKalow, line 37: This will change per product I'm fairly sure, here's where we've been putting per-product data, as well as subnav items, which is what these would map to I think: https://github.com/hashicorp/dev-portal/blob/main/src/data/waypoint.json#L27

ashleemboyer: Hmm true. I'll work on moving this

ashleemboyer: @BRKalow Here's the change I made: https://github.com/hashicorp/dev-portal/pull/45/commits/415ace3beda06ff40130d4e21c41c06fd5028796. What do you think?

src/components/navigation-header/style.module.css
ashleemboyer: Addressed! https://github.com/hashicorp/dev-portal/pull/45/commits/73753bc5ecf8cf626fe70554e6f575fdfa490adf

src/components/navigation-header/style.module.css
```diff
@@ -33,20 +33,36 @@
...
-      line-height: 16px;
+      font-weight: 500;
+      line-height: 20px;
+      padding: 3px 8px;
 
       &:hover {
-        cursor: pointer;
-        text-decoration: underline;
+        color: var(--token-color-palette-neutral-300);
+      }
```
BRKalow, line 54: nit: It'd be a nice touch if there was a very short transition on the hover color change here ✨

A .1s makes transition makes it feel fancy 😍

src/components/navigation-header/index.tsx
```diff
@@ -1,9 +1,37 @@
...
 import ProductChooser from 'components/product-chooser'
+// TODO: we'll need a programatic way to get this data when there are more products
+import waypointData from 'data/waypoint.json'
 import s from './style.module.css'
 
+/**
+ * This is a temporary component. Once all of the nav links point to internal paths,
+ * it won't be needed anymore. Not planning to move into a separate file.
+ */
+const NavLink: React.FC<{
```
BRKalow, line 14: nit: `React.FC` implicitly makes `children` a valid prop in the component's type definition, so if we don't do anything with `children`, it might not be what we want. I see this is a temporary component though, so not a big deal.

ref: https://www.harrymt.com/blog/2020/05/20/react-typescript-react-fc.html

src/components/navigation-header/index.tsx
```diff
@@ -1,9 +1,37 @@
+import Link from 'next/link'
 import HashiCorpLogo from '@hashicorp/mktg-logos/corporate/hashicorp/logomark/white.svg?include'
-import HeaderSearchInput from 'components/header-search-input'
 import InlineSvg from '@hashicorp/react-inline-svg'
+import HeaderSearchInput from 'components/header-search-input'
 import ProductChooser from 'components/product-chooser'
+// TODO: we'll need a programatic way to get this data when there are more products
+import waypointData from 'data/waypoint.json'
```
BRKalow, line 7: :nod:

src/components/navigation-header/style.module.css
```diff
@@ -33,20 +33,36 @@
...
-      color: var(--navigation-header-link-color);
+      border: 1px solid transparent;
+      border-radius: 5px;
+      color: var(--token-color-palette-neutral-400);
+      display: block;
       font-size: 14px;
-      line-height: 16px;
+      font-weight: 500;
+      line-height: 20px;
+      padding: 3px 8px;
```
BRKalow, line 50: nit: most of our spacing is base 4, I wonder if it'd be feasible to adjust this to `padding: 4px 8px` 💭 

No big deal if not, just an observation!

src/components/navigation-header/index.tsx
```diff
@@ -12,20 +40,18 @@ const NavigationHeader: React.FC = () => (
...
-          </li>
-          <li>
-            <a>API</a>
-          </li>
-          <li>
-            <a>Downloads</a>
-          </li>
+          {waypointData.subnavItems.map((navLink, index) => (
+            <li key={navLink.id}>
+              {/* TODO: we'll use the router to determine the current link once they're all internal */}
```
BRKalow, line 46: 👍

src/styles/color.css
```diff
@@ -1,13 +1,19 @@
...
+  /* Neutrals */
+  --token-color-palette-neutral-700: #0c0c0e;
+  --token-color-palette-neutral-600: #3b3d45;
+  --token-color-palette-neutral-500: #656a76;
+  --token-color-palette-neutral-400: #8c909c;
+  --token-color-palette-neutral-300: #c2c5cb;
+  --token-color-palette-neutral-200: #dedfe3;
+  --token-color-palette-neutral-100: #f1f2f3;
+  --token-color-palette-neutral-50: #fafafa;
+  --token-color-palette-neutral-0: #fff;
```
BRKalow, line 16: I like these ✨

BRKalow (approved):
Nothing blocking!

src/components/navigation-header/style.module.css
ashleemboyer: So this is because of some math I had to do to make sure that a focus border doesn't add layout shift. I've had to do this in a lot of places across my PRs, so I think I'll do a follow-up PR for cleaning these things up with variables & `calc` to make it explicit that some math had to happen. What do you think?

src/styles/color.css
ashleemboyer: I will like them even more when get them from somewhere else! 😎

src/components/navigation-header/style.module.css
BRKalow: That's fine! A comment explaining why we're using odd values could be helpful too, if you don't wanna introduce a new approach 👍

src/components/navigation-header/style.module.css
ashleemboyer: Done! Great idea

src/components/navigation-header/index.tsx
ashleemboyer: Didn't think about that! Very good point. Updated. 😊

src/components/navigation-header/style.module.css
ashleemboyer: I took a combined approach and added a TODO with a reason. 😊

PR #46, opened 2021-12-14
feat: migrate vault's io site

BRKalow:
Migrating vault's io site using the existing migration pattern laid out by @zchsh


.gitignore
```diff
@@ -4,3 +4,4 @@ node_modules
 out
 .env*.local
 .vercel
+_temp-migrations/
```
zchsh, line 7: 👍 great call

zchsh (approved):
🎉 Yay, ran it locally and seems good to me! Apologies for the messiness in the migration scripts 🙇‍♂️ 

Had a minor nit with the favicon, pushed a commit for that 😬 For some reason the one in the Vault repo is a Consul favicon, and there's just direct-from-dato loading that looks pretty unique 🤔

scripts/migrate-io/migrate-vault.js
```diff
@@ -221,7 +221,8 @@ async function migrateVaultIo() {
   //
   const assetsToCopy = [
     // meta images
-    '/img/favicons/favicon.ico',
+    // '/img/favicons/favicon.ico', NOTE: actually a Consul favicon
```
BRKalow, line 224: wtf

PR #48, opened 2021-12-16
Add docs content styles

ashleemboyer:
🔎  [Preview link](https://dev-portal-git-ambadd-content-styles-hashicorp.vercel.app/waypoint/docs)
🎟️  [Asana ticket](https://app.asana.com/0/0/1201490161340316/f)

---

This PR adds styles for content that falls under `DocsLayout`. The styles are all defined in `src/layouts/docs/docs-layout.module.css`. This is a temporary measure for a few reasons:

- design system tokens and helpers aren't ready to consume yet, so we've got some hardcoded values
- there are a lot of caveats and interesting content structures where these styles will likely have to be tweaked, so having them in one place makes it easy to find this code

Once these two items are less of a reason, I think we can abstract these styles in a cleaner way.

This PR also:

- Handles rendering our default MDX components with `defaultMdxComponents` from `@hashicorp/platform-docs-mdx`.
- Adds a new `DocsPage` component for abstracting the use of `defaultMdxComponents` on the waypoint docs, commands, and plugins pages. This is a great step towards abstracting things across those three pages!

Open question:

- How do we set up system fonts for `--font-monospace` like we've done for `--font-body` and `--font-display`? I've left a TODO comment near the relevant code


ashleemboyer:
FYI I am actively going over this with Brayden today/tomorrow!

src/components/docs-page/index.tsx
```diff
@@ -0,0 +1,16 @@
...
+import Placement from 'components/author-primitives/shared/placement-table'
+import NestedNode from 'components/author-primitives/waypoint/nested-node'
+
+const additionalComponents = defaultMdxComponents('waypoint', {
+  Placement,
+  NestedNode,
+})
+
+const DocsPage = (props): ReactElement => {
+  return <MDXRemote {...props} components={additionalComponents} />
```
BRKalow, line 13: 💭 We may or may not want the ability to pass in custom components per product. Should be easy enough to add if we do want to continue supporting that, but I don't think we need to adjust anything righ tnow.

src/pages/style.css
```diff
@@ -39,12 +39,16 @@
...
+
   --font-body: system-ui, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto,
     Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji',
     'Segoe UI Symbol';
   --font-display: system-ui, -apple-system, BlinkMacSystemFont, 'Segoe UI',
     Roboto, Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji',
     'Segoe UI Symbol';
+
+  /* TODO: what should this be? */
+  --font-monospace: Menlo;
```
ashleemboyer, line 51: For more context, I think we need to be using system monospace fonts? But I'm not sure how to do it.

src/components/docs-page/index.tsx
```diff
@@ -0,0 +1,16 @@
+import { ReactElement } from 'react'
+import { MDXRemote } from 'next-mdx-remote'
+import defaultMdxComponents from '@hashicorp/platform-docs-mdx'
+import Placement from 'components/author-primitives/shared/placement-table'
+import NestedNode from 'components/author-primitives/waypoint/nested-node'
+
+const additionalComponents = defaultMdxComponents('waypoint', {
```
BRKalow, line 7: 💭 Stating the obvious here, but this will eventually need to be configurable by product.

src/layouts/docs/docs-layout.module.css
```diff
@@ -25,31 +25,261 @@
   height: 100%;
   justify-content: center;
   overflow-y: auto;
-  padding: 24px 24px 100px 24px;
+  padding-bottom: 100px;
+  padding-left: 48px;
+  padding-right: 24px;
+  padding-top: 32px;
```
BRKalow, line 31: Is there a reason we expanded this to the long-form syntax?

src/layouts/docs/docs-layout.module.css
```diff
@@ -25,31 +25,261 @@
...
+      }
+
+      & p {
+        &:first-child {
+          margin-top: 0;
+        }
+
+        &:last-child {
+          margin-bottom: 0;
+        }
```
BRKalow, line 73: What are these doing? Is it to avoid stacked margin from margin applied to `p`? Would margin collapsing come into play here? https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing

src/layouts/docs/docs-layout.module.css
```diff
@@ -25,31 +25,261 @@
...
-.versionSwitcher {
-  height: 18px;
-  width: 346px;
-}
-
 /* TODO: these styles will be changing very soon, this is just to get us started */
 .main {
+  color: #3b3d45;
+  flex-grow: 1;
+  flex-shrink: 1;
```
BRKalow, line 39: What are these doing?

BRKalow (commented):
Looking good! 😄

I'll be interested to hear @zchsh's thoughts here as well, as I know we've had a couple conversations about adjusting our base content styles to be applied by overriding the default MDX components.

Either way, what you've got here is great!

src/layouts/docs/docs-layout.module.css
ashleemboyer: These + the `width` (although I _think_ `flex-shrink` isn't doing anything, need to double check) make the `<main>` fill the space between the sidebar and the sidecar rather than grow too wide for the space. When there are images in the content, the main is too wide and not fully visible. I will add a comment when I check if `flex-shrink` is doing anything!

src/components/docs-page/index.tsx
ashleemboyer: Agreed, I'll add a TODO so that anyone who stops by will know we're thinking of it 😊

src/pages/style.css
zchsh: When I was skimming the `hashicorp/design-system-tokens` repo I noticed some tokens for this - https://github.com/hashicorp/design-system-tokens/blob/6406550e15588de305a01edd3b6bcf3a47f4a36b/dist/products/css/tokens.css#L298 . TLDR; `SF Mono, Consolas, Ubuntu Mono, monospace;` looks like the system monospace stack they're using at the DS level ✨

src/layouts/docs/docs-layout.module.css
```diff
@@ -25,31 +25,261 @@
...
+    color: white;
+    margin: 24px 0;
+
+    & code {
+      background-color: transparent;
+      border: none;
+      border-radius: 0;
+    }
+  }
+
```
zchsh, line 221: Bit of a nit, but I _think_ we can actually omit these styles completely, since we expect to always be swapping in our `@hashicorp/react-code-block` component, which has its own styles.

src/layouts/docs/docs-layout.module.css
```diff
@@ -25,31 +25,261 @@
...
+      color: #a737ff;
+    }
   }
 
   & img {
     max-width: 100%;
   }
+
+  /* inline code elements */
+  & code {
```
zchsh, line 199: In [our `react-content` styles](https://github.com/hashicorp/react-components/blob/1b3791abfef10b96b05495598ce313524be9213e/packages/content/styles/code-inline.module.css#L3) we do ~ `& :not(pre) code` to avoid these styles bleeding into our swapped-in-via-MDX code block. Not sure if these are necessary, but would recommend as it can act as a light guarantee we won't clobber custom styles (and may also remove the need for the `& pre code` "reset"-ish styles below).

zchsh (approved):
😍💖 Looking great! Had a few small nits, but nothing felt blocking 💯

src/layouts/docs/docs-layout.module.css
zchsh: ☝️I think this also solves for an edge case I ran into testing out the product landing - if the content in `main` is very narrow (eg a short heading and like "Click something in the sidebar" as the only paragraph of text), with existing styles, `main` doesn't grow to fill the available space, and the content ends up floating in the middle. I think these styles will fix that case too 👍

src/layouts/docs/docs-layout.module.css
ashleemboyer: I did it for readability. Personally, it's easier for me to parse this mentally than it is the shorthand!

src/layouts/docs/docs-layout.module.css
ashleemboyer: So if these weren't here, then the alert boxes would be too tall like this:

image

When these are present, the alert boxes look like this with the appropriate padding:

image

I'm not sure if we have any multi-paragraph alert boxes, but this approach allows multiple paragraphs to have the correct spacing between them without adding too much spacing at the top & bottom of the alert box. I can add a comment to explain since it is definitely interesting to see!

src/layouts/docs/docs-layout.module.css
ashleemboyer: This is a great point, I'll make this change!

src/layouts/docs/docs-layout.module.css
ashleemboyer: I believe you're right! These are leftover from my first pass before I put our custom component in, I think I forgot to remove them. 😎

src/layouts/docs/docs-layout.module.css
zchsh: ☝️ Will +1 that sometimes I find the shorthand higher effort than separate properties.

I am all for consistency though... one related thought here is whether we should adopt something like https://github.com/hudochenkov/postcss-sorting. This way we could have clearly encoded, and automatically applied, conventions for things like this (much like we have for JS already!). Also wary that such a plugin would be a great way to just bikeshed for a while... but feels like CSS is kind of a second class citizen in terms of linting & auto-formatting of CSS properties. Will add this as a task for discussion on the platform board!

src/layouts/docs/docs-layout.module.css
ashleemboyer: @zchsh I think I know what you're saying, but do double check: what you're referring to is different than the main & sidecar sticking together and centering like this, right?

image

src/layouts/docs/docs-layout.module.css
BRKalow: Oh man I totally missed this was nested within the alert styles, my bad!

src/layouts/docs/docs-layout.module.css
ashleemboyer: I'm working on an agenda-like thing for a code style jam session (just adding links for things that are relevant, not working too hard on it since there are endless links haha), I'll add this to that list too!

src/layouts/docs/docs-layout.module.css
ashleemboyer: Never mind actually! I think I found what you were talking about. 😊

image

src/layouts/docs/docs-layout.module.css
ashleemboyer: No worries! 😊

src/layouts/docs/docs-layout.module.css
ashleemboyer: I wasn't successful in making this change, and since we'll be styling code blocks differently at some point I'm not super concerned for now. We can come back to this if needed!

src/layouts/docs/docs-layout.module.css
ashleemboyer: Gave it a try, see this comment: https://github.com/hashicorp/dev-portal/pull/48/files#r774789613

src/pages/style.css
ashleemboyer: I took another peek at [the WIP HDS Product Figma](https://www.figma.com/file/oQsMzMMnynfPWpMEt91OpH/?node-id=1436%3A10434) (also there's the link if you haven't seen it yet!), and I'm seeing that `ui-monospace` is a thing? So here's what I'm going with (a combination of what you've found too):

```
--font-monospace: ui-monospace, Menlo, Consolas, Ubuntu Mono, monospace;
```

ashleemboyer:
@BRKalow @zchsh I actually had to make some structural changes here, so I won't merge until we get back from the break that way y'all have a chance to review! Summary of changes:

- The anchor styles I was setting in `docs-layout.module.css` within the `main` element were overwriting the anchor styles in the breadcrumb bar (guess since the selectors were more specific in that file than in the breadcrumb bar style file). But I was able to fix this by moving the anchor styles into our newly created `docs-anchor` component! (`src/components/docs-anchor`). In that file, I created a little `passableProps` object that appends our custom class to the existing list of classnames.
- The way I was using `defaultMdxComponents` was very incorrect, so I fixed that. I thought it accepted 2 parameters but it actually accepts an object with `product` and `additionalComponents` properties. It fully works now. 😊

src/components/docs-page/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+   * TODO: this will eventually be configurable by all products, not just waypoint
+   */
+  const additionalComponents = defaultMdxComponents({
+    product: 'waypoint',
+    additionalComponents: {
+      Placement,
+      NestedNode,
+      a: makeDocsAnchor('waypoint', ['docs', 'plugins', 'commands']),
+    },
+  })
```
BRKalow, line 23: Can we move this out of the component function body for now? Just to ensure we aren't re-creating components during render, which might cause React to re-mount instead of re-render these components. Alternatively we could wrap it in a `useMemo` call!

BRKalow (approved):
Approving pending my latest comment, feel free to merge once addressed!

src/components/docs-page/index.tsx
ashleemboyer: Moving it out seems easiest for the time being since we aren't currently passing any more components through the props. I'll do that now!

ashleemboyer: Done! ✅ 
https://github.com/hashicorp/dev-portal/pull/48/commits/d3b71cb88e6f348a9e6a039f5c35f2dfdb4c104b

ashleemboyer:
I would normally wait for @zchsh to be back in to merge, but I'd like to give him as much time as possible tomorrow to work on his tasks for Wednesday and I'd also like to make sure I have time available to review his code for those tasks. 😎

PR #50, opened 2021-12-17
Make nav links internal and fix double scrollbars

ashleemboyer:
🔎  [Preview link](https://dev-portal-git-ambmake-nav-links-internal-hashicorp.vercel.app/waypoint)

🎟️  Asana tickets:
- [Update NavigationHeader links to point internally](https://app.asana.com/0/0/1201540744015400/f)
- [Address double scrollbar with new global banner](https://app.asana.com/0/0/1201540744015404/f)

## Additions

- Fixes the double scrollbar issue
- New hook & tests: `useCurrentPath`
  - Gets the current path based on `next/router`'s `router.asPath`. This is needed in order to prevent hydration issues. Using `window.location` in this helper will cause hydration issues. Since `router.asPath` includes a URL's hash and search if they're present, the helper has options for excluding those if desired: `excludeHash` & `excludeSearch`.
  - The tests cover all the cases I could think of. Using `test.each` makes the tests easier to write, but it's not _extremely_ readable IMO, so if there are suggestions for improving readability please send 'em! 🙏🏻
  - `@testing-library/react-hooks` was added to make testing the new hook easier using `renderHook` (needed because of `useRouter` is being called).
  - Helpful resource for how to mock `useRouter`: https://github.com/vercel/next.js/issues/7479
- New layout: `BaseNewLayout`
  - As @BRKalow and I were working on the double scrollbar issue earlier today, he suggested that we place the new `AlertBanner` directly next to `NavigationHeader`, which was not previously only rendered in `DocsLayout` and needed to be rendered in more pages that don't use `DocsLayout`. This new layout helps use these components in more pages smoothly.

## Other updates, by file

- `src/components/navigation-header/index.tsx`
  - There was a temporary `NavLink` component that handled sometimes wrapping links in `Link`, but this is no longer needed with every link in the header now being internal. They're all rendered in `next/link` now.
  - An `isCurrentPage` helper was added to keep the markup clean since there's some string splitting happening and two special path format cases to check. There's a comment explaining its purpose.
- `src/components/product-switcher/index.tsx`
  - `useLayoutEffect` was replaced with `useEffect` to remove the following warning from the terminal. This change had no effect on the code since the effect runs when `isOpen` changes, which happens after everything has been rendered on the page.
    > Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://reactjs.org/link/uselayouteffect-ssr for common fixes.
- `src/components/sidebar/components/sidebar-nav/sidebar-nav-menu-item/index.tsx`
  - the `getPath` was updated to build the path for a menu item based on the current path's page, which is needed since the Sidebar is used on three waypoint subpages (`commands`, `docs`, `plugins`) as of this PR
- `src/components/sidebar/index.tsx`
  - replaced the `getCurrentPathWithoutParamsOrAnchors` helper with usage of the new `useCurrentPath` hook
- `src/data/waypoint.json`
  - Replaced the `href` properties in the data with internal `path`s
- `src/pages/waypoint/commands/[[...page]].tsx` and `src/pages/waypoint/plugins/[[...page]].tsx`
  - updated to work with `DocsLayout` 🎉
- `src/pages/waypoint/downloads/index.tsx`
  - Changed to use TypeScript, replaced the content with placeholder content, and set the layout to `BaseNewLayout`
- src/pages/waypoint/index.tsx
  - Replaced the content with different placeholder content, and set the layout to `BaseNewLayout`


src/hooks/__tests__/use-current-path.test.ts
```diff
@@ -0,0 +1,140 @@
+import { renderHook } from '@testing-library/react-hooks'
+import useCurrentPath from '../use-current-path'
+
+describe('useCurrentPath', () => {
```
BRKalow, line 4: noice 👍

src/hooks/use-current-path.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Gets the current path similar to Next's router but optionally excludes the
+ * hash and/or search portion of the path. Uses `window.location`.
+ */
+const useCurrentPath = (options: UseCurrentPathOptions = {}): string => {
+  const router = useRouter()
+  const { excludeHash = false, excludeSearch = false } = options
+  const { hash, pathname, search } = new URL(
+    router.asPath,
+    // TODO: replace this with an environment variable soon
+    'http://www.test.com'
```
BRKalow, line 18: If we're only ever gonna return relative links, it's probably okay to put a dummy URL here...but maybe `https://www.hashicorp.com`? 😄

BRKalow (approved):
LGTM!

src/hooks/__tests__/use-current-path.test.ts
ashleemboyer: Thanks! 😊

src/hooks/use-current-path.ts
ashleemboyer: Done 😁

PR #51, opened 2021-12-20
feat: add dev popover, use for search

zchsh:
👀 [Preview](https://dev-portal-git-zsadd-dev-tooltip-hashicorp.vercel.app/waypoint/docs)

This PR adds a `dev-popover` component, and uses it for the header `<Search />` component. This is intended as a minimum viable approach to showing search functionality during internal demos.

This PR also adds components to our Swingset instance:
- [`dev-popover`](https://dev-portal-git-zsadd-dev-tooltip-hashicorp.vercel.app/swingset/components/devpopover), for use in development
- [`header-search-input`](https://dev-portal-git-zsadd-dev-tooltip-hashicorp.vercel.app/swingset/components/headersearchinput), not implemented, but now shows a dev popover
- [`popover`](https://dev-portal-git-zsadd-dev-tooltip-hashicorp.vercel.app/swingset/components/popover), a generic popover, swiped from a recent change to `@hashicorp/react-logo-grid`

## Preview

CleanShot 2021-12-20 at 12 44 56@2x

> Note: I started off by trying out a `tooltip` rather than `popover`, but felt the response time for tooltips and other issues meant they weren't a good fit for these types of alerts. I have some related work-in-progress in https://github.com/hashicorp/dev-portal/pull/53 which I intend to abandon for now (but may revisit if / when we need tooltips for some other purpose).


zchsh:
@ashleemboyer For context, I'm starting off the `Search` implementation with the _very_ minimum of adding a popover we can show for demo purposes. Intent here is not to have something ship-able (but very open to any & all feedback regardless!)

global.d.ts
```diff
@@ -1,7 +1,7 @@
 /// <reference types="@hashicorp/platform-types" />
 
 declare module 'swingset/page' {
-  const createPage: () => $TSFixMe
+  const createPage: (opts: $TSFixMe) => $TSFixMe
```
zchsh, line 4: Required so we can pass custom `{ components }` for use in Swingset pages. Specifically, I'm showing `@hashicorp/react-search` on the [`header-search-input`](https://dev-portal-git-zsadd-dev-tooltip-hashicorp.vercel.app/swingset/components/headersearchinput) page for functionality demo purposes (and as a starting point for upcoming implementation work, which will definitely leverage the package, including `<SearchProvider />`).

package.json
```diff
@@ -48,10 +48,13 @@
     "@hashicorp/react-inline-svg": "^6.0.3",
     "@hashicorp/react-min-100-layout": "^2.0.1",
     "@hashicorp/react-product-downloads-page": "^2.5.2",
+    "@hashicorp/react-search": "^6.4.0",
```
zchsh, line 51: Added for functionality demo in Swingset.

package.json
```diff
@@ -48,10 +48,13 @@
...
     "@hashicorp/react-section-header": "^5.0.4",
     "@hashicorp/react-subnav": "^9.1.3",
     "@hashicorp/react-vertical-text-block-list": "^7.0.0",
     "@hashicorp/remark-plugins": "^3.3.1",
+    "@reach/popover": "^0.16.2",
+    "@reach/portal": "^0.16.2",
```
zchsh, line 57: Added for `popover` component, and by extension the `dev-popover` used for the demo note on Search.

src/components/dev-popover/docs.mdx
```diff
@@ -0,0 +1,43 @@
+---
+componentName: 'DevPopover'
+---
+
+**🚧 DO NOT USE IN PRODUCTION, FOR WORK IN PROGRESS ONLY 🚧**
```
zchsh, line 5: Added some docs here since I figured they might be helpful during dev, if there are other features similar to Search that we want to annotate.

src/components/header-search-input/index.tsx
```diff
@@ -1,16 +1,43 @@
...
-  <div className={s.headerSearchInput}>
-    <div>
-      <IconSearch16 className={s.searchIcon} />
-      <p>Search</p>
-    </div>
-    <IconSlashSquare16 />
-  </div>
-)
+interface HeaderSearchInputProps {
+  theme?: 'light' | 'dark'
```
zchsh, line 8: I think we'll want to address theming concerns soon, but i feel confident that we'll need some kind of prop interface. This component could be extended in the future to consume theme Context, at which point this `theme` prop would be a way to override context rather than the only way to set the component's theme.

For context, one alternative might be a superglobal toggling of CSS property values, but this can wreak havoc if we have "mixed" theme elements, such as this "dark" themed header and search input in an otherwise "light" themed view.

src/components/navigation-header/index.tsx
```diff
@@ -61,7 +61,7 @@ const NavigationHeader: React.FC = () => (
             </li>
           ))}
         </ul>
-        <HeaderSearchInput />
+        <HeaderSearchInput theme="dark" />
```
zchsh, line 64: Previously this component inherited `color` from `navigation-header`. Have implemented a super basic `theme` within `header-search-input` to make things a lil more encapsulated.

src/components/popover/docs.mdx
```diff
@@ -0,0 +1,82 @@
+---
+componentName: 'Popover'
+---
+
+**🚧 WORK IN PROGRESS 🚧**
+
+This is a rough cut at a Popover component. This implementation is largely duplicative of the tooltip partial in [`@hashicorp/react-logo-grid`](https://github.com/hashicorp/react-components/tree/main/packages/logo-grid/partials/popover-tooltip).
```
zchsh, line 7: To add some colour to this, I *think* this `Popover` component may be useful outside of just `dev-popover`. Could certainly use some iteration, though I think it's at least in a decent spot it terms of a11y etc thanks to the underlying `@reach/popover` library.

src/components/popover/icons/x.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
zchsh, line 1: I should probably replace this with a Flight icon... will make that change now!

src/lib/hooks/use-on-click-outside.js
```diff
@@ -0,0 +1,29 @@
+import { useEffect } from 'react'
```
zchsh, line 1: Added some hooks as used in `popover`. Added to `/lib` as it feels like they may be useful elsewhere as well.

src/components/popover/icons/x.svg
zchsh: ✅ Done

jescalan:
😍 omg this looks amazing!

jescalan (approved):

src/components/dev-popover/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+}: {
+  children: React.ReactNode
+  note: React.ReactNode
+  title?: string
+}): React.ReactElement {
+  const [showDialog, setShowDialog] = useState(false)
+  const triggerRef = useRef(null)
+  return (
+    <>
+      <button
```
ashleemboyer, line 24: Love that a `<button>` is rendered here! Covers some usability cases that might otherwise be tricky to handle. 👏🏻

src/components/dev-popover/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+        style={{
+          border: 'none',
+          background: 'none',
+          fontSize: 'inherit',
+          fontFamily: 'inherit',
+          padding: 0,
+          margin: 0,
+          display: 'block',
+          cursor: 'pointer',
+          color: 'inherit',
```
ashleemboyer, line 34: Are these styles inline so that they're hard to overwrite?

src/components/header-search-input/index.tsx
ashleemboyer: I think we only have designs for dark mode in header components, but having this prop in place and ready to go will make it easier to implement a light theme design when it comes! 👏🏻

src/components/header-search-input/style.module.css
```diff
@@ -1,7 +1,17 @@
 .headerSearchInput {
+  &.theme-light {
+    --border-color: var(--token-color-palette-neutral-300);
+    --text-color: var(--token-color-palette-neutral-700);
+  }
+
+  &.theme-dark {
+    --border-color: var(--token-color-palette-neutral-500);
+    --text-color: var(--token-color-palette-neutral-0);
+  }
```
ashleemboyer, line 10: ✨

src/components/popover/docs.mdx
ashleemboyer: Side question: do you know what the difference is between the Reach UI popover and tooltip? I can't find any docs on their site about popover.

src/lib/hooks/use-on-click-outside.js
ashleemboyer: This is a really neat hook! Seems like it would be useful [in `ProductSwitcher`](https://github.com/hashicorp/dev-portal/blob/assembly-ui-v1/src/components/product-switcher/index.tsx#L67-L76) as well, what do you think? Not asking you to implement there, mainly asking to make sure I understand the code here fully. 😊

src/lib/hooks/use-on-focus-outside.js
```diff
@@ -0,0 +1,22 @@
+import { useEffect } from 'react'
+
+/**
+ * Hook that runs a callback on focus outside the ref
+ */
+function useOnFocusOutside(ref, handler) {
```
ashleemboyer, line 6: So cool! ✨

src/components/dev-popover/index.tsx
zchsh: Naw, just inline cause it's a dev component and I was too lazy to split them out 😅 I'll patch this up now!

zchsh: ✅ Fixed this up!

zchsh:
@ashleemboyer Only change since your last review was inline-ing the `dev-popover` styles - just wanted to double-check that you're okay with this being merged 🙇

BRKalow:
Observation 👀  when I open the popover, the width of it animates to here:
Screen Shot 2021-12-21 at 9 09 49 AM

src/lib/hooks/use-on-click-outside.js
BRKalow, line 1: nit: Wanna move to `src/hooks`? Per https://github.com/hashicorp/dev-portal#component-organization

src/components/dev-popover/docs.mdx
```diff
@@ -0,0 +1,43 @@
...
+  <div
+    style={{
+      border: '1px solid rebeccapurple',
+      padding: '0.5rem 1rem',
+      borderRadius: '4px',
+    }}
+  >
+    This is a work in progress element! Focus and activate it, or click it for
+    info.
+  </div>
```
BRKalow, line 41: nit: nesting a `div` inside of a button is technically not valid HTML ([ref](https://stackoverflow.com/questions/12982269/is-it-semantically-incorrect-to-put-a-div-or-span-inside-of-a-button))

zchsh:
@BRKalow thanks for the feedback, great calls re: `src/hooks` and avoiding `button > div`.

> Observation 👀 when I open the popover, the width of it animates to here:

That screenshot definitely doesn't look right to me, the popover is supposed to keep itself `8px` from the viewport edge, and should be able to occupy more width than shown there. Tried to reproduce on Safari & Firefox but wasn't able to - will download Chrome and try that. Let me know if there was some other repro step for that width-animating state!

BRKalow:
@zchsh I'm seeing it in both FF and Chrome and all I did was click on it 😅 . Happy to screenshare if you can't reproduce.

src/lib/hooks/use-on-click-outside.js
ashleemboyer: @zchsh I was wondering because I think we might want to name this `useClickOutside` instead of including `Dropdown` in the name, but that's a _nit_ and I don't feel super strongly about needing to rename now

ashleemboyer (approved):
Left a nit comment, good to go from me on whichever you decide to do! 🙌🏻

src/components/popover/docs.mdx
zchsh: Yeah, it's pretty strange they don't have docs on `popover`. IIRC i was working on `logo-grid` when this came up.

From my understanding on the difference - `tooltip` is shown on `hover` (and `focus` i think?), where `popover` is shown on click or tap (or activation via keyboard, not sure what the short name for that is?).

src/lib/hooks/use-on-click-outside.js
zchsh: Great call, have renamed - I think the name was just a holdout from lifting this function out of a context where it was for a dropdown, even though none of the code is really dropdown specific.

Which probably answers the `ProductSwitcher` question as well - that seems like a great use case!

zchsh: Done, thanks for flagging!

src/components/dev-popover/docs.mdx
zchsh: Ah, great callout 💖  kind of took a quick and dirty approach with this. Have refactored so that the `button` element within `DevPopover` is a sibling of the passed `children` rather than a wrapper 👍

zchsh:
@BRKalow I think I've got most feedback addressed, except the animating-width issue! I downloaded Chrome, and tried messing around in there along with messing around in Safari & Firefox, but still haven't been able to reproduce...

I was able to get some wonky behaviour and see some width animations at smaller viewport sizes (due to the search bar being rendered offscreen), and by doing the weird "overscroll" thing in Safari. I'd be very down to look at a screenshare or even pair on this to try to sort it out (maybe Thursday?) 🙇

zchsh:
Update here: synced with Bryce on the issues. TLDR; the issues are fixed, but thought I might expand a bit as it's making me rethink a little piece of "cover parent element" CSS I often re-use.

---

## The "not working" issue

The "not working at all" issue started happening after I changed the approach to avoid nesting content within the `button` trigger (so I felt like I had tested in other browsers, but actually that had been before I made this change).

I'd tried to use a `position: absolute; top: 0; left: 0; right: 0; bottom: 0` approach to get the `button` element to cover its parent. But, oddly enough, this only works in Safari for `button` elements (have got [a minimal repro in a code pen](https://codepen.io/zchsh/pen/gOGxZpx)). An alternate approach, which up until now I considered equivalent, is `position: absolute; top: 0; left: 0; width: 100%; height: 100%`, and this seems to work in all browsers:

### Safari

CleanShot 2021-12-21 at 20 23 25@2x

### Chrome

CleanShot 2021-12-21 at 20 23 36@2x

### Firefox

CleanShot 2021-12-21 at 20 23 49@2x

## The "animating" issue

Seems to be gone with the new approach that no longer nests content in `button`, and uses more stable "cover parent div" code as above.

zchsh:
@BRKalow ☝️ popover issues should be fixed now, hopefully! 🤞 Let me know if it looks alright for you!

BRKalow (approved):
Nice work 👍

PR #54, opened 2021-12-20
feat: add edit this page link

zchsh:
👀 [Preview](https://dev-portal-git-zsadd-edit-github-link-hashicorp.vercel.app/waypoint/docs)

This PR adds an `Edit this page on GitHub` link to the docs view. It also makes a minor tweak to the `getTruncatedTitle` function.

## Preview

as-designed


src/components/edit-on-github-link/index.tsx
```diff
@@ -0,0 +1,22 @@
+import InlineSvg from '@hashicorp/react-inline-svg'
+import svgGithub from '@hashicorp/flight-icons/svg/github-16.svg?include'
+import s from './style.module.css'
+
+function EditOnGithubLink({
```
zchsh, line 5: Split out this component, as I thought it might be useful to test placement options depending how prominent we want this link to be. I think previously we put it at the bottom of the page in our designs simply because we didn't have a nice sidebar to put it in.

src/components/edit-on-github-link/style.module.css
```diff
@@ -0,0 +1,20 @@
+.editLink {
+  align-items: center;
+  color: var(--blue-vibrant);
+  cursor: pointer;
+  display: inline-flex;
+  font-family: inherit;
+  font-size: 14px;
```
zchsh, line 7: I think there may be a point soon where we'll want to identify the type styles used in the Figma doc and create some abstraction for them. Maybe this is a `<Type />` component... or perhaps it's global utility classes as we're using in other projects.

The latter appeals to me because of simplicity, and because through all our time using global type utility classes, the only issue we've run into is nitpicking performance (which could be marginally better if we included ONLY the type classes used on a page, rather than all of them all the time). I think we'll have a lot hairier issues to tackle if we try to adopt a component pattern.

That being said, this is probably outside the scope of this PR - mostly just thinking out loud and putting this idea out in the universe 🎇

src/components/sidecar/get-truncated-title.test.ts
```diff
@@ -0,0 +1,45 @@
+import getTruncatedTitle from './get-truncated-title'
+
+describe('getTruncatedTitle', () => {
+  it('leaves a short title unmodified', () => {
+    const title = 'Hello world'
+    expect(getTruncatedTitle(title)).toBe(title)
+  })
+
+  it('does not add ... to a title of exact limit length', () => {
```
zchsh, line 9: I split this out and added a test file because I noticed the very specific `contrib Tools and Helpers...` issue, and wanted to debug.

src/components/sidecar/get-truncated-title.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    const title = 'contrib Tools and Helpers'
+    const expected = 'contrib Tools and...'
+    expect(getTruncatedTitle(title)).toBe(expected)
+  })
+
+  /* @TODO
```
zchsh, line 15: This is a bit long-winded, but I think we should try out a CSS solution to truncation - I think it might be less brittle (ie, adapt to available pixel space rather than requiring specific max character counts per use case), and might also give us some nice polish (eg splitting words at "natural" locations, since browser use the same logic for truncation as they do for hyphenation, IIRC).

> This is assuming we haven't already taken a shot at this, if we have, pls ignore! 🙇

src/components/sidecar/get-truncated-title.ts
```diff
@@ -0,0 +1,27 @@
...
+    const words = fullTitle.split(' ')
+    const wordsToInclude = []
+    words.forEach((word, idx) => {
+      const wordLength = word.length
+      if (characterCount + wordLength < MAX_CHARACTERS) {
+        wordsToInclude.push(word)
+        characterCount += wordLength
+        // count spaces as well, since these are counted
+        // in the fullTitle.length comparison
+        if (idx !== words.length) characterCount++
```
zchsh, line 19: Fixes a nitpicky issue I noticed on the [current UI assembly preview](https://dev-portal-git-assembly-ui-v1-hashicorp.vercel.app/waypoint/docs). The title `contrib Tools and Helpers` ends up being "truncated" into the `contrib Tools and Helpers...`, I think counting spaces consistently fixes this 👍.

Very related: I bumped up the `MAX_CHARACTERS` from 25 to 26 to counteract the above change, as `contrib Tools and Helpers` felt like it should pass our "short enough to not truncate" bar.

src/components/sidecar/get-truncated-title.ts
```diff
@@ -0,0 +1,27 @@
...
+      }
+    })
+    truncatedTitle = wordsToInclude.join(' ') + `…`
```
zchsh, line 22: Changed from three periods `...` to ellipsis `…`. On macOS at least these end up looking identical, but figure this might be relevant if we ever use this function with eg. a monospace font.

src/layouts/docs/docs-layout.module.css
```diff
@@ -27,6 +27,10 @@
   overflow-y: auto;
   padding: 24px 24px 100px 24px;
   position: relative;
+
+  & :global(.__permalink-h) {
```
zchsh, line 31: Quick fix to hide permalinks as I wanted to take some screenshots for design review. I'll backlink to this comment in the `content styles` Asana task so this doesn't get buried 👍

src/components/edit-on-github-link/index.tsx
```diff
@@ -0,0 +1,22 @@
+import InlineSvg from '@hashicorp/react-inline-svg'
+import svgGithub from '@hashicorp/flight-icons/svg/github-16.svg?include'
```
BRKalow, line 2: We can use the `<IconGithub16 />` component here

```ts
import { IconGithub16 } from '@hashicorp/flight-icons/svg-react/github-16'
```

zchsh: @BRKalow Glad you brought this up and curious to get your take on this - we talked a while back during Learn dev work (with Will) about the merits of rendering a React-ified SVG component, vs setting the SVG as inline HTML.

Previous concern with the React-ified SVG was performance... which kinda made sense to me, since React would have to care about eg each `path` element rather than treat the `svg` as a fixed, just-render-it thing. But that's with a pretty limited knowledge of the inner workings of React.

Do you think that perf concern is worth using React-ified SVGs only when we actually need re-render-within-the-SVG functionality? (Say, so we can set the `title` from outside the SVG for a117 purposes). Or is the performance concern so minor that we might as well just always use React-ified SVGs, and we'll be fine even if there are dozens or hundreds rendering on a page?

BRKalow: I'd be curious to see at what point we would start experiencing performance issues, I'd be surprised if we ever encountered any with the number of icons we use 💭  I tried to do some initial research when setting up this approach, and there was nothing glaring that I ran into where folks encountered noticeable performance issues rendering SVGs as React elements vs. inlining them.

One thing to consider with the inline/string approach is that those blobs can't be minified as well, so ultimately it likely results in larger JS bundles. I think if we're trying to squeeze out performance, something which uses SVG `use` ([ref](https://css-tricks.com/svg-use-with-external-reference-take-2/)) would be the way to go.

zchsh:
@ashleemboyer I needed something small to do as my day was winding down, so jumped on the GitHub link. I *really* like the sidecar component, so I did a quick experiment placing the `Edit this page` link there rather than at the bottom of the content (latter is what our current design specifies):

sidecar-suggestion

Intent here was:
- Increased prominence ~= increased interaction rates, and this is something where we want more people to interact & contribute (I think?). Felt like placement at the bottom of the page was a bit buried.
- Outside of contributions, might also better signal the open-source, community-oriented stance of the docs and of HashiCorp generally. So kind of an affirmation of core values / brand stuff as well as a more direct & practical thing.


Very open to thoughts, feedback, hot takes etc. Floated this by Brayden as well (via Slack, wasn't sure of GitHub handle). Might also be the kind of tiny detail we can call out as a practitioner-oriented improvement during demos (We have tons more of these already, of course, but figured why not try to add to the pile).

src/components/edit-on-github-link/index.tsx
zchsh: Oh awesome, I didn't realize you'd already looked into this 💯 Since the performance concern seems pretty minor / non-noticeable, will swap out for the React component, definitely feels better in terms of consistency & ease of use 👍

src/components/sidecar/get-truncated-title.test.ts
```diff
@@ -0,0 +1,45 @@
+import getTruncatedTitle from './get-truncated-title'
```
BRKalow, line 1: nit: What do you think about putting tests in a `__tests__` folder? This makes for a cleaner separation between test files and source files.

src/layouts/docs/docs-layout.module.css
BRKalow: Maybe a `TODO` here would be helpful context too?

src/components/sidecar/get-truncated-title.test.ts
BRKalow: (I like seeing tests here!)

BRKalow (approved):

src/components/sidecar/get-truncated-title.test.ts
zchsh: This is a great question, and tbh i'm not sure what I think ahah 😅

On the one hand, I've used co-located `.test` files before and have liked that it's immediately clear when contributing to source whether a function has a test file, and that it kind of raises the visibility that tests should exist. Co-locating test files also seems to create a convention that there should be one test file per source file, which I also kinda like (but now that I think about it, must have its downsides as well).

On the other hand, I've never really worked in a project that uses a `__tests__` folder outside the source files so I don't have a lot to go off of (I guess I have seen these in some more recent projects, I think? But this wasn't an explicit convention in anything I've worked in yet, IIRC).  I am curious about the perceived benefit of making that cleaner separation... but given I haven't tried it, I might as well try it now -  will move this into a `__tests__` folder! 👍 💖

src/components/edit-on-github-link/style.module.css
ashleemboyer: Yes! As far as I know, there will be helper classes for different types coming from the design systems tokens. And they're supposed to match up with the types defined in Figma. They've been changing very frequently though, so I haven't created variables or classes at this point since even the names weren't fully worked out yet. I think we regroup on this after the holiday/first preview date and see if the naming of the tokens is more stable so we can start creating CSS variables at the very least?

src/components/sidecar/get-truncated-title.test.ts
ashleemboyer: AFAIK it's not possible to truncate text on full words with CSS. Have you done it or seen it done before?

src/components/sidecar/get-truncated-title.test.ts
BRKalow: Yeah I guess it's largely personal preference. I've almost always used `__tests__`, so maybe it's just a familiarity thing haha. I do like that it clearly separates the test files, but I see your point about the clear 1:1 mapping as well. Try it out and lemme know what ya think!

src/layouts/docs/index.tsx
```diff
@@ -21,6 +23,12 @@ const DocsLayout: React.FC<DocsLayoutProps> = (props) => (
           {/* <div className={s.versionSwitcher}>VERSION SWITCHER</div> */}
           <main className={s.main} id="main">
             {props.children}
+            {props.githubFileUrl ? (
+              <EditOnGithubLink
+                url={props.githubFileUrl}
+                label="Edit this page on GitHub"
+              />
+            ) : null}
```
ashleemboyer, line 31: Thoughts on using `&&` instead of a ternary with a `null` portion?

```suggestion
            {props.githubFileUrl && (
              <EditOnGithubLink
                url={props.githubFileUrl}
                label="Edit this page on GitHub"
              />
            )}
```

src/components/sidecar/get-truncated-title.test.ts
zchsh: @BRKalow  oh dang i just realized that for some reason I had this in my head as the `__tests__` option:

- `src/__tests__/components/breadcrumb-bar/get-breadcrumb-links.test.ts`

and realizing now that you wrote `"a"` `__tests__` folder, not `"the"` tests, folder, so maybe what i should have been picturing is:

- `src/components/breadcrumb-bar/__tests__/get-breadcrumb-links.test.ts`

Is that accurate? If so, zero objections and fully on board with that, really dig the balance between separation and co-location.

src/components/edit-on-github-link/style.module.css
```diff
@@ -0,0 +1,17 @@
+.editLink {
```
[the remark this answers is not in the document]
zchsh: Ah, great callout re: font-weight. Looks like the Figma also now has no underline, so will make those tweaks 👍

TLDR; on `inline-flex` is cause then the link behaves like other text would, and we can `text-align` it. This only came up cause of previous positioning concerns when I was experimenting with placement in the `<Sidecar />` - it allowed the whole `<EditOnGithubLink />` stick to the right of the `<Sidecar />` like everything else because of the existing `text-align` setting. The `flex-end` was actually a holdover from having `display: flex` on there and trying to come up with an alternate solution for that right alignment (removed now!).

I dig the `display: flex` & `width: fit-content` - seems like a great alternative since we might not need the link to obey `text-align` if it's at the bottom of the page. I only recently learned about `fit-content` so that's why it didn't pop in my mind first! ☺️

src/layouts/docs/index.tsx
zchsh: Thanks for calling this out 💯 This is a holdover from personal habits 😬 have always done this cause I'm paranoid the initial condition will unexpectedly be `0` or `NaN` or something and get rendered (instead of the intended "render nothing"). But with Typescript I guess I can rest a lot easier and enjoy the good live of less ternaries 😌🎉

src/components/edit-on-github-link/style.module.css
zchsh: ☝️ Sounds good! Yeah, I'm on board with regrouping on this after break 💯 

It seems like the structure at least is stable (eg some "body small" -ish style will exist, even if it's not called "body small"), so maybe we could even jump the gun a bit for the sake of our own developer experience in these initial phases and set up classes for those (and then run a global find-replace if the names change later). 

Also just noticed there's already an `npm` package published, version 0 i guess... definitely curious on how stable that feels since it seems that as we get into more varied UI work with stuff other than grays, it'd be really convenient to be able to pull in all the tokens at once via `npm` rather than having to recreate the CSS files in `dist` 🤔

src/components/sidecar/get-truncated-title.test.ts
zchsh: @ashleemboyer Ah, great point, I haven't seen truncation on full words before in CSS. The intent here was mainly just making things less brittle - for example if the sidecar changes width, or if the text size changes due to user settings.

I think there might be small benefit to truncating not just on full words as well - eg if a heading contains a long word like "infrastructure", we can show the first part of the word:

Frame 2323

That being said, there are definitely downsides to the CSS approach as well, so probably should have emphasized "try out". One downside with the CSS approach that I've noticed is that the current active states have a font-weight shift, and this can cause the truncation to vary in order to prevent line wrapping....

CleanShot 2021-12-21 at 21 51 12

But, this font-weight shift can also cause issues with the custom truncation approach as well - this is a bit of a convoluted example, but the 22-character title `Wow Waypoint wwwwwwwww` is a decent example of both how character counts can be a bit brittle since letter width varies, and on the potential jank caused by `font-weight` shifting (which is maybe even worse since it results in line wrapping, but MUCH less likely to happen so perhaps not as relevant):

CleanShot 2021-12-21 at 22 18 13

All that being said, depending on folks opinions, the mid-word truncation might be a downside as well, maybe enough to outweigh potential brittleness.

zchsh: ☝️ Tested out the CSS approach in https://github.com/hashicorp/dev-portal/pull/56, let me know what you think!

src/components/sidecar/get-truncated-title.ts
```diff
@@ -0,0 +1,27 @@
+const MAX_CHARACTERS = 26
+
+function getTruncatedTitle(fullTitle: string): string {
+  let truncatedTitle: string
+
+  if (fullTitle.length < MAX_CHARACTERS) {
```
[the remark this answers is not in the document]
zchsh: I'm not sure -  apart from counting the spaces, I think we'd need to keep `fullTitle.length < MAX_CHARACTERS` and `characterCount + wordLength < MAX_CHARACTERS` consistent (ie both `<` or both `<=`). This might make `MAX_CHARACTERS` more intuitive though (ie it's an inclusive max), so applying it and will update the other comparison as well 👍

src/components/sidecar/get-truncated-title.ts
```diff
@@ -0,0 +1,27 @@
...
+    truncatedTitle = fullTitle
+  } else {
+    let characterCount = 0
+    const words = fullTitle.split(' ')
+    const wordsToInclude = []
+    words.forEach((word, idx) => {
+      const wordLength = word.length
+      if (characterCount + wordLength < MAX_CHARACTERS) {
```
zchsh, line 14:
```suggestion
      if (characterCount + wordLength <= MAX_CHARACTERS) {
```

src/components/sidecar/get-truncated-title.test.ts
zchsh: ☝️  Updated with a `__tests__` folder within `components/sidecar`, and moved the source file so now we've got:

- Source: `src/components/sidecar/utils/get-breadcrumb-links.test.ts`
- Test: `src/components/sidecar/__tests__/get-breadcrumb-links.ts`

Let me know if this seems reasonable / matches the `__tests__` approach we're going for 🙇‍♂️

src/components/edit-on-github-link/style.module.css
ashleemboyer: I agree it'd be more convenient to pull in all the tokens, but that's not what's been recommended by the design systems team. If you find that you have time to add some variables and CSS helper classes you need in passing I say go for it, but we're so constrained on time that I personally am not prioritizing that until we're past January 5th. I've avoided adding any extra variables that aren't in the designs I've worked with because the tokens aren't stable yet and don't want to lose too much time to creating and re-creating variables.

src/components/sidecar/get-truncated-title.test.ts
ashleemboyer: I think you bring up great points on both sides and I think we need to have a discussion between engineering and design to decide how to move forward. We consciously decided to go with truncating text on full words with JavaScript for our first preview, so I think we need to table the discussion until after then since I know design is a little pressed for time and we already made a short term decision on this. I think if we're lucky we might also get some feedback on this from our January 5th previewers and hopefully that can also help us make an informed decision! Can we hold on to the #56 experiment to use as an artifact in the larger discussion?

ashleemboyer:
> @ashleemboyer I needed something small to do as my day was winding down, so jumped on the GitHub link. I _really_ like the sidecar component, so I did a quick experiment placing the `Edit this page` link there rather than at the bottom of the content (latter is what our current design specifies):
> 
> sidecar-suggestion
> 
> Intent here was:
> 
> * Increased prominence ~= increased interaction rates, and this is something where we want more people to interact & contribute (I think?). Felt like placement at the bottom of the page was a bit buried.
> * Outside of contributions, might also better signal the open-source, community-oriented stance of the docs and of HashiCorp generally. So kind of an affirmation of core values / brand stuff as well as a more direct & practical thing.
> 
> Very open to thoughts, feedback, hot takes etc. Floated this by Brayden as well (via Slack, wasn't sure of GitHub handle). Might also be the kind of tiny detail we can call out as a practitioner-oriented improvement during demos (We have tons more of these already, of course, but figured why not try to add to the pile).

I like that the link is more prominent in the Sidecar! I've found myself several times wishing that links like this were easier to find in other documentation sites. There's two things I can think of here to note that influence my short-term take of not putting the link in the sidecar for now:

- We have a few very long sidecars due to the structure of some docs pages. We're not addressing that for January 5th, so it's unclear in that aspect if we'll get the benefit of this link being more prominent at the end of the Sidecar.
- Some sidecars only have one heading for the title of a short docs page and we've weighed not showing the Sidecar in those instances. That's another topic that's been tabled until after January 5th, but if we did decide to hide the Sidecar sometimes we'd also need to think about whether or not this link should be hidden. It looks great with the populated Sidecar but I'm not sure how the link would look by itself.

src/components/edit-on-github-link/style.module.css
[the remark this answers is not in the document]
ashleemboyer: Thank you for elaborating! I super appreciate the context and totally makes sense. 😊

src/components/sidecar/get-truncated-title.test.ts
ashleemboyer: Here's an example of what I've seen a lot in my experience. I've used 4-space indents to hopefully make the hierarchy more prominent. I had a hard time seeing it myself with just 2 spaces. 😅

- all component files are co-located in the component's folder (i.e. no __tests__ folder since there's typically only one test file based on the next point)
- utils a component uses aren't co-located to the component, but in a `utils` or similarly named folder
- utils and utils-like folders (i.e. a hooks folder) have a single `__tests__` folder and don't have any other subfolders or nesting according to how the code inside them are used

```
src/
    components/
        breadcrumb-bar/
            index.tsx
            breadcrumb-bar.module.css
            breadcrumb-bar.test.ts
    hooks/
        __tests__/
            use-count.test.ts
            use-query-string.test.ts
        use-count.ts
        use-query-string.ts
    utils/
        __tests__/
            add-some-numbers.test.ts
            get-breadcrumb-links.test.ts
        add-some-numbers.ts
        get-breadcrumb-links.ts
    ...
```

That's what I had in mind, but I think this also depends on how complex a component is and if it's being packaged up for use in other projects? If a component has a few utils that are helpful or needed to consume the component, then the structure might be a little different but I think I've mostly seen that handled by having a `packages` root folder and then basically sub-project folders within there for each thing getting packaged up.

BRKalow: Yeah a colocated __tests__ folder is what I had in mind here, sorry if that wasn't clear! So anywhere we might put tests, we would add a __tests__ folder and put tests/test related code there. 👍

src/components/sidecar/get-truncated-title.test.ts
zchsh: ☝️ Sounds great - super on board with tabling the discussion. Will leave #56 up and will leave this PR as-is with the Javascript word-based truncation 💯

src/components/sidecar/get-truncated-title.test.ts
zchsh: @ashleemboyer @BRKalow I think since I'm still getting familiar with the approach here, I might need specific feedback on the test files in this PR 😅 Are ya'll good with the setup I have as-is:

- Source: `src/components/sidecar/utils/get-truncated-title.test.ts`
- Test: `src/components/sidecar/__tests__/get-truncated-title.ts`

If not, where should I move those two files?

BRKalow: Re `__tests__` or `*.test.ts`, I would advocate for one or the other, but not both.

If we're going with `__tests__` I would say put the test file at: `src/components/sidecar/utils/__tests__/get-truncated-title.test.ts`

ashleemboyer: In this case since `get-truncated-title` can be reused beyond sidecar, I would move these files to:

- `src/utils/get-truncated-title.ts`
- `src/utils/__tests__/get-truncated-title.test.ts`

We have a `lib` folder already, is that the same as `utils` just a different name? If so, `lib` is fine with me

ashleemboyer: @BRKalow does webpack / next know to ignore files inside of a `__tests__` folder? I thought that file `.test.` file extensions were always needed to ensure test files aren't included in a bundle?

BRKalow: If they aren't imported elsewhere there should be no problem. I usually also use the `.test.ts` extension, even if in a `__tests__` folder.

PR #55, opened 2021-12-21
feat: add breadcrumb bar

zchsh:
👀 [Preview](https://dev-portal-git-zsadd-breadcrumb-bar-hashicorp.vercel.app/waypoint/docs/intro/vs/helm)

This PR adds a `<BreadcrumbBar />` component, and adds it to the `/waypoint/docs` route.

## Implementation details

- The structure is modelled off of the [the WAI-ARIA Authoring Practices Breadcrumb widget](https://www.w3.org/TR/wai-aria-practices-1.1/#breadcrumb)
- Implements a helper function to transform Docs `navData` into breadcrumb links.
    - There are some interesting edge cases for "Overview" pages, and for "index-less category pages" to note here, so this helper is unfortunately not as trivial as one might hope.
- Adds the component to the `/waypoint/docs` route, which includes some manual additions to the top-end of the breadcrumb bar (to allow linking back to the product home and to the docs root, as in Figma mocks).
- Adds [Swingset page for `<BreadcrumbBar />`](https://dev-portal-git-zsadd-breadcrumb-bar-hashicorp.vercel.app/swingset/components/breadcrumbbar). This page has some light documentation on both the display portion of the component, and on the `get-docs-breadcrumbs` utility function.

## Preview

Common use case:

CleanShot 2021-12-22 at 09 30 46@2x

Parsing titles from "Overview" pages:

CleanShot 2021-12-22 at 09 30 28@2x

Inserting unlinked breadcrumb for index-less category pages:

CleanShot 2021-12-22 at 09 21 08@2x

## Caveats

- 📱🚫 This implementation  **is not** styled to handle mobile viewports very intentionally. The Figma designs specify more complex behaviour for small viewports, with "extra" breadcrumb links that would otherwise overflow being hidden. For now, we simply `flex-wrap` overflowing items onto another line:

CleanShot 2021-12-22 at 09 35 11@2x


src/components/breadcrumb-bar/get-breadcrumb-links.ts
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * Given a basePath, navData, and array of path parameters,
+ * Return an array of breadcrumb links leading to this page.
+ *
+ * Note that the returned array does not include links beyond
+ * the basePath. So, in many use cases, consumers will likely
+ * want to prefix this array to provide a more complete
+ * list of breadcrumb links.
+ */
+function getBreadcrumbLinks({
```
zchsh, line 12: There's some complexity here due to the flexibility in our `navData` structures. Namely:
- Some breadcrumb parts correspond to `{ title, path }` nodes with useful titles. These are easy.
- Some breadcrumb parts correspond to "index" nodes, which often have the title "Overview". If we naively match these nodes and use their `title` values, we end up with breadcrumbs like `Waypoint > Docs > Overview > Overview > Helm`. So we implement some logic to detect these "index" nodes, and instead use the parent category's title, giving us something like `Waypoint > Docs > Introduction > Waypoint vs. Other Software > Helm`
- Some breadcrumb parts correspond to index-less categories. An example of an index-less category is Waypoint's `kubernetes` category - [/docs/kubernetes/install](https://www.waypointproject.io/docs/kubernetes/install) is a valid page, but [/docs/kubernetes](https://www.waypointproject.io/docs/kubernetes) is a 404 (as `Kubernetes` is a category with no index page). In these cases, we want to show the index-less category's title in the breadcrumb, but not have that breadcrumb part link anywhere (as otherwise, for the `/docs/kubernetes/install` breadcrumb we end up with `Waypoint > Docs > Install`, which is misleading).

src/components/breadcrumb-bar/style.module.css
```diff
@@ -0,0 +1,37 @@
...
+  align-items: center;
+}
+
+.breadcrumbText {
+  display: block;
+  padding: 0;
+  composes: ds-type-body-small;
+  color: var(--token-color-palette-neutral-500);
+
+  &[href]:hover {
```
zchsh, line 27: We only want to apply hover styles if we have a valid `href`.

src/layouts/docs/docs-layout.module.css
```diff
@@ -25,7 +25,7 @@
   height: 100%;
   justify-content: center;
   overflow-y: auto;
-  padding: 24px 24px 100px 24px;
+  padding: 32px 24px 100px 24px;
```
zchsh, line 28: Tweaked this padding per a comment in the Figma file.

src/pages/waypoint/docs/[[...page]].tsx
```diff
@@ -52,12 +53,23 @@ export async function getStaticProps({ params }) {
     remarkPlugins: [[anchorLinks, { headings }]],
   })
 
+  const breadcrumbLinks = [
```
Comment (zchsh) [inline src/pages/waypoint/docs/[[...page]].tsx:56]: This could possibly be abstracted even further, for example we could generated the two first links based on the product slug (`waypoint`) and `basePath` (mapping eg `docs` > `Docs`, `commands` > `Commands`, etc consistently for all products). But this felt a little early to try to tackle in detail, and felt like this bit of manual specificity probably isn't so harmful anyways.

src/components/breadcrumb-bar/utils/get-docs-breadcrumbs.ts
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * Given a basePath, navData, and array of path parameters,
+ * Return an array of breadcrumb links leading to this page.
+ *
+ * Note that the returned array does not include links beyond
+ * the basePath. So, in many use cases, consumers will likely
+ * want to prefix this array to provide a more complete
+ * list of breadcrumb links.
+ */
+function getDocsBreadcrumbs({
```
zchsh, line 13: There's some complexity here due to the flexibility in our `navData` structures. Namely:
- Some breadcrumb parts correspond to `{ title, path }` nodes with useful titles. These are easy.
- Some breadcrumb parts correspond to "index" nodes, which often have the title "Overview". If we naively match these nodes and use their `title` values, we end up with breadcrumbs like `Waypoint > Docs > Overview > Overview > Helm`. So we implement some logic to detect these "index" nodes, and instead use the parent category's title, giving us something like `Waypoint > Docs > Introduction > Waypoint vs. Other Software > Helm`
- Some breadcrumb parts correspond to index-less categories. An example of an index-less category is Waypoint's `kubernetes` category - [/docs/kubernetes/install](https://www.waypointproject.io/docs/kubernetes/install) is a valid page, but [/docs/kubernetes](https://www.waypointproject.io/docs/kubernetes) is a 404 (as `Kubernetes` is a category with no index page). In these cases, we want to show the index-less category's title in the breadcrumb, but not have that breadcrumb part link anywhere (as otherwise, for the `/docs/kubernetes/install` breadcrumb we end up with `Waypoint > Docs > Install`, which is misleading).

src/components/breadcrumb-bar/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+          return (
+            <li key={`${title}_${url}`} className={s.listItem}>
+              <Elem
+                className={s.breadcrumbText}
+                href={url}
+                aria-current={isCurrentPage ? 'page' : undefined}
+              >
+                {title}
+              </Elem>
+            </li>
```
BRKalow, line 50: What do you think about using `next/link` here? All the links should be internal, so I don' think there's a downside!

src/components/breadcrumb-bar/utils/get-docs-breadcrumbs.ts
```diff
@@ -0,0 +1,132 @@
...
+  // able to render the breadcrumb bar.
+  // We take this fallback approach in production contexts,
+  // but throw an error in other envs to flag the navData issue.
+  if (process.env.HASHI_ENV === 'production') {
+    return matches[0]
+  } else {
+    throw new Error(
+      `Ambiguous breadcrumb path: Found ${matches.length} matches for "${pathString}". Please ensure there is exactly one node or index-less category with the path "${pathString}" in the provided navData.`
+    )
+  }
```
BRKalow, line 129: This _kind of_ feels like it doesn't belong in the breadcrumbs logic, but I don't feel too strongly about that 💭

src/pages/waypoint/docs/[[...page]].tsx
Comment (BRKalow) [reply src/pages/waypoint/docs/[[...page]].tsx:56]: Keeping in mind that this will expand out to other product pages, I think anything we can do to reduce boilerplate would be a win 👍

BRKalow (commented):
Just an observation, but I think we've got similar logic which determines the active links in the side nav.
Screen Shot 2021-12-22 at 9 23 26 AM

(via [this method](https://github.com/hashicorp/dev-portal/blob/assembly-ui-v1/src/components/sidebar/index.tsx#L33))

It would be handy to have a "canonical" spot where this data is stored. A context feels like it could be appropriate? Curious to hear your thoughts, this is definitely something that could be a future optimization, but it seems worth calling out!

src/components/breadcrumb-bar/utils/get-docs-breadcrumbs.ts
zchsh: +1 I'm on the fence here as well. As I was writing it, I was kind of like.... I feel like this should be guaranteed by the incoming `navData`. And feels duplicative of the validation checks we already have; if those are in place there should be no possibility of ambiguous or missing matches.

Rather than throw errors here, we could silently just resolve these ambiguous matches. Or I could just make this a simpler error throw, eg if `matches.length != 1` then throw a more generic error about invalid navData. Let me know what you think!

zchsh:
@BRKalow Re: active states, we've also got https://github.com/hashicorp/react-components/blob/main/packages/docs-sidenav/utils/flag-active-nodes.js 🤔 That's in use client-side in our current `docs-sidenav`... and handles some annoying edge cases, such as duplicate items - https://github.com/hashicorp/react-components/pull/269.

Maybe [the relevant effect](https://github.com/hashicorp/react-components/blob/4e9a08e34a7e9ac357b02851f9b0b4ed8c4a6d1a/packages/docs-sidenav/index.js#L89) could be moved into a context provider, that would provide the `navData` with `__isActive` added to each node?

Let me know what you think - in general agreed that that additional context probably falls outside the scope of this PR, ie better for a follow-up than to try to handle here (slash before our first demo).

src/pages/waypoint/docs/[[...page]].tsx
Comment (zchsh) [reply src/pages/waypoint/docs/[[...page]].tsx:56]: Sounds good - I'll try out the "pretty name" mapping for `basePath` values, which should let us move the `getDocsBreadcrumbs` call into the docs `layout` rather than having to include it on every docs path 👍

Comment (zchsh) [reply src/pages/waypoint/docs/[[...page]].tsx:56]: err wait will actually split into a separate util, as then we can still call it in `getStaticProps` rather than client-side 👍

src/components/breadcrumb-bar/utils/get-docs-breadcrumbs.ts
BRKalow: Maybe dev-warnings would be appropriate here, and we can work to ensure that our nav data validation is up-to-par? With the eventual goal being these code paths _should never happen_.

Side note, it'd be really nice to have some validation that runs on PRs which comments back to the PR with any errors if the nav data validation fails. I feel like that'd be a really nice feedback loop for contributors, as opposed to parsing through Vercel build logs (which many don't always do).

BRKalow:
@zchsh Yeah, totally fine to defer this. I wonder if as we build out additional components and features we'll have more need for access to our "internal" navigation state, spitballing:

* current product
* active nav items
* version (?)
* sub-product (?) (can of worms in itself, but thinking like Consul's suite of products and/or Terraforms suite of products)
* more? 🤷 

Definitely feels like something we'll want to address in the not-too-distant-future, but right now it doesn't seem critical.

src/components/breadcrumb-bar/style.module.css
```diff
@@ -0,0 +1,44 @@
+/* Note: the ds- type style comes from the global design system.
+   Once we have a way to consume these styles from a more
+   central location, we should switch to that approach. */
+.ds-type-body-small {
+  font-size: 13px;
+  line-height: 18px;
+}
```
ashleemboyer, line 7: Want to add this to a `typography.css` file [in `src/styles`](https://github.com/hashicorp/dev-portal/tree/assembly-ui-v1/src/styles)?

src/layouts/docs/index.tsx
```diff
@@ -20,6 +22,9 @@ const DocsLayout: React.FC<DocsLayoutProps> = (props) => (
           {/* TODO: implement version switcher (ref: https://app.asana.com/0/1201010428539925/1201342966970641/f) */}
           {/* <div className={s.versionSwitcher}>VERSION SWITCHER</div> */}
           <main className={s.main} id="main">
+            {props.breadcrumbLinks ? (
+              <BreadcrumbBar links={props.breadcrumbLinks} />
+            ) : null}
```
ashleemboyer, line 27: _nit_ 😊
```suggestion
            {props.breadcrumbLinks && (
              <BreadcrumbBar links={props.breadcrumbLinks} />
            )}
```

src/pages/waypoint/docs/[[...page]].tsx
Comment (zchsh) [reply src/pages/waypoint/docs/[[...page]].tsx]: @BRKalow Tried it out, ended up shying away from the mapping, it felt odd, figured it's easier to pass in the desired "pretty" name for the basePath. This means a bit more boilerplate in the docs `[[...page]]` file though (extra arg to pass). I think the best "boilerplate reduction" solution here may be to move the breadcrumb link generation into `docs-page`'s `generateStaticProps`.

With getting things ready for the demo in mind, have further abstracted the `getDocsBreadcrumbs` function, and added a `TODO` comment noting that we may want to move it into `docs-page` in the future. Let me know what you think! 🙇

Comment (BRKalow) [reply src/pages/waypoint/docs/[[...page]].tsx]: Sounds good!

Comment (ashleemboyer) [reply src/pages/waypoint/docs/[[...page]].tsx]: In my PR for making the main nav links internal, I was able to determine the product slug and what I called the subpage (docs, commands, plugins) here: https://github.com/hashicorp/dev-portal/pull/50/files#diff-4943d44a4e7e2fcf35e1eb0eee646918c1247641ca15621f5b0a28544f336456R14-R20. It's safe to have in that area since those are those three subpages are the only ones that use the sidebar, but it's probably something good to abstract soonish! It's worth noting that we will definitely have a context in the future that tracks the current product slug, it'll be updated by the `ProductSwitcher`. Maybe when we first create that context it pulls the current product slug from the router path?

src/components/breadcrumb-bar/utils/get-docs-breadcrumbs.ts
zchsh: I dig the idea of dev-warnings here, and really dig the validation on PRs at the source (content repos) that flag any issues with `navData` 💯 

To clarify re: `dev-warning` - what do you think the best way to detect dev environments would be?

I'm thinking we may be able to rely on [Vercel's System Environment variables](https://vercel.com/docs/concepts/projects/environment-variables#system-environment-variables), specifically `VERCEL_ENV`.... Though this would require us to remove the currently-hard-coded `VERCEL_ENV="production"` from our `.env` file in this project. Have taken that approach for now, so I've got:

```js
const IS_DEV = process.env.VERCEL_ENV !== 'production'

// If navData has ambiguous matches, warn in development.
  // We can fallback to returning the first match, and this should
  // be fine from a visitor perspective. Less urgent to fix these types of issues.
  if (matches.length > 1) {
    if (IS_DEV) {
      console.warn(
        `Ambiguous breadcrumb path: Found ${matches.length} matches for "${pathString}". Please ensure there is exactly one node or index-less category with the path "${pathString}" in the provided navData.`
      )
    }
    return matches[0]
  }
  // Otherwise, we have zero matches, which would mean a breadcrumb with missing parts.
  // This would feel broken to a site visitor, so should be a blocking
  // issue we need to fix, so we throw an error.
  throw new Error(
    `Missing breadcrumb path: Found zero matches for "${pathString}". Please ensure there is a node (or index-less category) with the path "${pathString}" in the provided navData.`
  )
```

Let me know if that looks good... I guess another option would be using `HASHI_ENV` instead of `VERCEL_ENV`, but unsure on why those both exist 🤔

BRKalow: I think in this case we could just do `NODE_ENV !== 'production'`. Agreed we should clarify and/or cleanup `HASHI_ENV`/`VERCEL_ENV`

src/pages/waypoint/docs/[[...page]].tsx
Comment (zchsh) [reply src/pages/waypoint/docs/[[...page]].tsx]: ☝️Nice, that seems like a good plan! On board with that as a near-future abstraction 💯

src/components/breadcrumb-bar/style.module.css
zchsh: ✅ Done, thanks for flagging! I've added a `typography.css` file with a lil todo there about pulling in the `--tokens` from the `hashicorp/design-system-tokens` `npm` package, but figure we leave that outside the scope of this PR to avoid yak-shaving. Let me know what you think!

src/components/breadcrumb-bar/style.module.css
```diff
@@ -0,0 +1,36 @@
+/* Note: mobile styles have not been fully implemented.
+   For now, the breadcrumb links wrap to additional
+   lines on small viewports. This does not match our
+   current design spec, but felt like an acceptable
+   stopgap. */
```
ashleemboyer, line 5: Forgot to note earlier, but agree 👍🏻

src/styles/typography.css
```diff
@@ -0,0 +1,31 @@
+/*
```
ashleemboyer, line 1: Looks good! Only note is that it's not clear what we're doing with `rem`s yet. I believe we're all on board with using them (sure hope so! 😁) but we haven't touched on how to handle larger fonts yet. I'm sure many things will need addressing!

ashleemboyer (approved):

PR #57, opened 2021-12-22
feat: adds io site deploy preview build

BRKalow:
Adding a node build script for io site deploy previews. Largely just a lift and land of @zchsh existing script, but in JS for a little added flexibility.

Can see it implemented in waypoint here: https://github.com/hashicorp/waypoint/pull/2850


scripts/capture-build-metrics.ts
```diff
@@ -47,7 +47,7 @@ function captureMetric({
 }
 
 async function main() {
-  if (!process.env.CI) return
+  if (!process.env.CI || process.env.DEV_IO) return
```
BRKalow, line 50: We don't want to collect build metrics for io sites atm I don't think...

scripts/content-repo-preview/build.ts
```diff
@@ -0,0 +1,99 @@
...
+   */
+  if (fs.existsSync(path.join(cwd, '..', '.next'))) {
+    console.log('.next folder found, moving into website-preview...')
+    await execFile('mv', ['../.next', './.next'])
+
+    if (fs.existsSync(path.join(cwd, '.next', 'cache', 'node_modules'))) {
+      console.log('Found cached node_modules, moving...')
+      await execFile('mv', ['./.next/cache/node_modules', './node_modules'])
+    }
+  }
```
BRKalow, line 30: If cache is found, move it to the correct spots in the nested `website-preview` directory.

scripts/content-repo-preview/build.ts
```diff
@@ -0,0 +1,99 @@
...
+      console.log(`🧹 removing pages for ${dir.name}`)
+      await fs.promises.rm(path.join(proxiedIoPagesDir, dir.name), {
+        recursive: true,
+      })
+    }
+  }
+
+  /** Install deps */
+  console.log('📦 Installing dependencies')
+  execFileSync('npm', ['install', '--include=dev'], { stdio: 'inherit' })
```
BRKalow, line 82: using `npm install` instead of `npm ci` here as `ci` will actually remove `node_modules` if it already exists.

scripts/content-repo-preview/build.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  /** Build */
+  execFileSync('npm', ['run', 'build'], { stdio: 'inherit' })
+
+  // Copy the .next folder to the top-level so the site can be deployed and we can take advantage of the cache
+  // We are using cp here because the Vercel build is still running in the context of the nested website-preview directory,
+  // and so it needs the .next folder available to complete the deployment
+  await execFile('cp', ['-R', '.next', '../.next'])
```
BRKalow, line 90:
```suggestion
  execFileSync('cp', ['-R', '.next', '../.next'], { stdio: 'inherit' })
```

scripts/content-repo-preview/build.ts
```diff
@@ -0,0 +1,99 @@
+import { promisify } from 'util'
+import { execFileSync } from 'child_process'
+import path from 'path'
+import fs from 'fs'
+
+async function main() {
+  // imports
+  const execFile = promisify((await import('child_process')).execFile)
```
thiskevinwang, line 8: ❔ What's the thought behind `execFile` vs `execFileSync` here?

BRKalow: 🤷‍♂️  I had initially started with this, but I realized I needed stdio to come through in the parent process, so pivoted to `execFileSync`, which promisified `execFile` doesn't support as neatly. I'll probably switch everything to just use `execFileSync`.

scripts/content-repo-preview/build.ts
```diff
@@ -0,0 +1,95 @@
...
+  /** Build */
+  execFileSync('npm', ['run', 'build'], { stdio: 'inherit' })
+
+  // Copy the .next folder to the top-level so the site can be deployed and we can take advantage of the cache
+  // We are using cp here because the Vercel build is still running in the context of the nested website-preview directory,
+  // and so it needs the .next folder available to complete the deployment
+  execFileSync('cp', ['-R', '.next', '../.next'], { stdio: 'inherit' })
+
+  // Put node_modules into .next/cache so we can retrieve them on subsequent builds
+  // We are using mv here because it is more efficient than cp, and node_modules is a big chungus
```
thiskevinwang, line 89: ✨

thiskevinwang (approved):
👍 I'm seeing this script for the first time, so I don't have too much input aside from the "execFileSync" comment!

Who or what will be the main executor of this repo's `build:deploy-preview` script?

ex. [Authors in other repos?](https://github.com/hashicorp/waypoint/pull/2850/files#diff-283e018acbc584f3d90abcb941b581710a8a227626b72df5fd38525d007772aeR12)

BRKalow:
@thiskevinwang This script will likely only be called during a deploy preview Vercel build. We'll have an accompanying script for local previews, which will optionally be called by authors locally directly, or indirectly via a docker setup. Zach wrote up some details on the implementation here: https://docs.google.com/document/d/1tvEhrLF0YyRimgR-Ibd_lo7sqTvw0TFAi77jbgjROVk/edit#heading=h.4f853qau69eb

scripts/content-repo-preview/build.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * exclude any imports in the global CSS file which rely on other products
+   */
+  const globalCSSFileContents = await fs.promises.readFile(
+    globalCSSFile,
+    'utf-8'
+  )
+
+  const newContents = globalCSSFileContents
```
zchsh, line 40: 🤩 ✨ nicely done, so much better than the manual specification 💯 also a perfect example of a thing that i would have a very hard time doing in a bash script, but an easy time doing in a node script. Thanks for setting this up as the latter 🙇

scripts/content-repo-preview/build.ts
zchsh: 🐰

scripts/content-repo-preview/build.ts
zchsh: TIL `--include=dev` 👍

I could only find sparse docs on `--include` (under [omit](https://docs.npmjs.com/cli/v7/commands/npm-install#omit)). Is this ~ equivalent to `--production=false`, in the sense that it forces devDepencies even if `NODE_ENV == "production"`?

scripts/content-repo-preview/build.ts
```diff
@@ -0,0 +1,95 @@
+import { execFileSync } from 'child_process'
+import path from 'path'
+import fs from 'fs'
+
+async function main() {
+  // TODO: check for necessary env vars?
```
zchsh, line 6: Would the idea here be to throw an error if any of the required env vars aren't set? I'd be on board with that! (Would also be on board with merging this as-is, if we want).

zchsh (approved):
💖  This looks great to me! MUCH better than the bash, and stoked that we're now caching the `node_modules` from the `dev-portal` install 💯 

Only nit I could find is the `TODO` on checking for env vars. But IMO not blocking to fill that TODO in 👍

scripts/content-repo-preview/build.ts
BRKalow: Yeah that was my thought! We need REPO, DEV_IO and IS_CONTENT_PREVIEW, right?

scripts/content-repo-preview/build.ts
BRKalow: I think it would be equivalent yeah. I found it in a SO answer I think. Happy to switch to --production=false if we think that's clearer!

scripts/content-repo-preview/build.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  return true
+}
+
+async function main() {
+  if (!checkEnvVars()) {
+    process.exit(1)
+    return
+  }
```
BRKalow, line 27: @zchsh  I've added the env var check here. I went with checking that all three are defined, `REPO`, `DEV_IO`, and `IS_CONTENT_PREVIEW`. If one/some of these aren't needed we can adjust later!

BRKalow:
Merging for now, thanks for the reviews!

PR #58, opened 2021-12-22
Another pass at sidebar

ashleemboyer:
🔎 [Preview link](https://dev-portal-git-ambanother-pass-at-sidebar-hashicorp.vercel.app/waypoint/docs)
🎟️ Asana tickets
  - [Add missing pieces to Sidebar](https://app.asana.com/0/0/1201490161340314/f)
  - [Address Sidebar QA feedback from 11/29/2021](https://app.asana.com/0/0/1201477739256213/f)

---

## What

- Added `DevPopover` and focus styles on `SidebarBackToLink`
- Added dividers back in
- Added `fullPath`, `hasChildrenMatchingFilter`, and `id` properties to the `MenuItem` interface, which are all added to items in `addItemMetadata` of `Sidebar`
- Started using the new properties listed in the previous point
- Updated styles in `SidebarBackToLink`, `ProductSwitcher`, `SidebarFilterInput`, `SidebarNavMenuItem`, "Skip to main content" link to match the latest designs
- Fixed small `length-zero-no-unit` stylelint issues
- New external links added to the bottom of the sidebar manually

## Why

- `SidebarBackToLink` still needs some of it's functionality determined, so we're using the `DevPopover` as a placeholder for now.
- As for dividers, we need them for the time being because replacing them with section titles will require some extra planning, discussion, and content changes to make happen.
- The `fullPath`, `hasChildrenMatchingFilter`, and `id` properties are used in the following ways:
  - `fullPath`: used by `SidebarNavLink` to determine whether or not a link is internal or external. If an item has the `fullPath` property defined, then it is the path of an internal link. If this property isn't defined, then it renders just a semantic anchor element (no `next/link` component).
  - `hasChildrenMatchingFilter`: this is used for determining if a submenu item should be expanded because it has a child element that matches the filter text typed in the input at the top of the sidebar.
  - `id`: used for `key` props for elements rendered within `map` loops and also for the `aria-controls` and associated `<ul>` element used in submenu items.

## How

- For the `SidebarBackToLink` focus styles, I added a `buttonClassName` prop to `DevPopover` that gets passed to the `<button>` element in the component, and I also added a `containerClassName` component to make it clear that different classnames are able to be passed to specially style the popover on an individual basis.
- New `MenuItem` properties:
  - `fullPath` is comprised of three parts:
    - `currentProductSlug` (`waypoint` in this case)
    - `currentProductSubpage` (`cli`, `docs`, or `plugins` in the case of Waypoint)
    - `item.path` (comes from Waypoint data)
  - `hasChildrenMatchingFilter`: determined by recursively iterating over all menu items, submenu items, and their children
  - `id`: there are three kinds of ids based on the type of menu item. There is some regex in place for "slugifying" properties to turn into IDs and remove duplicate characters in the generated ID
    - submenu item ID - prefixed with `submenu-` and generated from the item title
    - internal menu item - prefixed with `menu-item-` and generated from the `fullPath` property
    - external link item - prefixed with `external-url-` and also generated from the item title

## Test Steps

- [ ] You should see dividers in the sidebar in the same places they are currently in Waypoint's docs
- [ ] You should see three external links at the bottom of the sidebar going to the correct place and showing an external link icon
- [ ] Tab to the "Back to" link
- [ ] You should see the common focus styles we use on elements on light backgrounds
- [ ] Click the "Back to" link
- [ ] You should see a dev popover as a placeholder for the link functionality
- [ ] In the filter input, try searching for a deeply nested menu item
- [ ] The correct submenu(s) should be open and have the correct open state styles applied
- [ ] Search "overview" in the filter input
- [ ] You should see many submenus open with the correct open state styles applied

## Anything else?

Nope!


src/components/sidebar/components/sidebar-nav/sidebar-nav-menu-item/index.tsx
```diff
@@ -40,7 +47,18 @@ const SidebarNavSubmenu: React.FC<{
...
+  )
+
+  /**
+   * Without this effect, the menu items aren't re-rerendered to be open. Seems
+   * to be because the item prop sent to the component don't change. Might work
+   * if we pass the props needed instead of just the item object?
+   */
+  useEffect(() => {
+    setIsOpen(item.hasActiveChild || item.hasChildrenMatchingFilter)
+  }, [item.hasActiveChild, item.hasChildrenMatchingFilter])
```
BRKalow, line 61: I think this would still be necessary, because this component also acts as a "controlled" component, where its open state could be changed at any point by its props, not just the initial props. What we've got here is the right approach!

BRKalow (commented):
A behavior observation from current production:

When I filter the sidebar, if a **parent item matches**, its immediate children are displayed, regardless of whether or not they match the filter text. This seems like it could be a valuable behavior to bring over. Example:
Screen Shot 2022-01-03 at 12 54 41 PM
Screen Shot 2022-01-03 at 12 54 12 PM

ashleemboyer:
> When I filter the sidebar, if a parent item matches, its immediate children are displayed, regardless of whether or not they match the filter text. This seems like it could be a valuable behavior to bring over.

@BRKalow good catch!! I didn't notice this functionality was in place but I agree that it's needed. Should be straightforward to get that working.

src/components/sidebar/components/sidebar-back-to-link/style.module.css
```diff
@@ -11,6 +12,15 @@
...
 }
 
+/*
+TODO: This is a temporary class. It should be removed when the development
+popover is removed.
+*/
+.popoverButton {
+  composes: focus-border-and-box-shadow-light from global;
+  border-radius: 5px;
+}
```
zchsh, line 22: This seems great, apologies for missing better `:focus` state in the original popover PR. Bit of a nit, but IMO we could move this into `dev-popover/style.module.css`, and remove both the `buttonClassName` and `containerClassName` from the `dev-popover` props 👍 Let me know what you think!

src/components/sidebar/index.tsx
```diff
@@ -45,14 +47,31 @@ const addActiveStateMetadata = (
...
-      const result = addActiveStateMetadata(currentPath, item.routes)
+    }
+
+    if (item.routes) {
+      const result = addItemMetadata(currentPath, item.routes)
       foundActiveItem = result.foundActiveItem
       itemCopy.routes = result.itemsWithMetadata
       itemCopy.hasActiveChild = result.foundActiveItem
-    } else {
+      itemCopy.id = `submenu-${itemCopy.title
```
zchsh, line 57: Nit: I've seen [`slugify` ](https://www.npmjs.com/package/slugify) in use in a few of our projects, might be useful here.

zchsh (approved):
🎉 This looks great to me! Had a couple really minor nits, nothing felt non-blocking, especially with the demo approaching 👍

src/components/sidebar/index.tsx
ashleemboyer: I actually tried that one! But it doesn't remove `/` so I decided not to use it.

src/components/sidebar/components/sidebar-back-to-link/style.module.css
ashleemboyer: I was wondering about this too, I think that'll be a good change to make. Mind if I do it in a follow-up PR to get this one merged faster?

ashleemboyer:
Due this being a little bit of a blocker for Zach, and me not being able to make this last fix for a few hours, I've gotten the pass from @zchsh and @BRKalow to merge this and do a fast follow of the fix later today. Thanks to you both!

PR #59, opened 2021-12-23
feat: add demo home & product landing

zchsh:
👀 [Preview of `/` main landing](https://dev-portal-git-zsscaffold-home-page-hashicorp.vercel.app) 
👀 [Preview of `/waypoint` product landing](https://dev-portal-git-zsscaffold-home-page-hashicorp.vercel.app/waypoint)

This PR implements the current Figma mocks for both the top-level project home page `/`, and for the product landing page `/waypoint`.


src/components/sidebar/components/sidebar-nav/index.tsx
```diff
@@ -18,14 +23,29 @@ const SidebarNav: React.FC<SidebarNavProps> = ({ menuItems, title }) => (
...
+        // TODO: One option, as below, is to make basePath optional,
+        // TODO: and if it is present, then generate the href from the
+        // TODO: basePath rather than try to determine the necessary
+        // TODO: item.path prefix from the window location (as currently
+        // TODO: done in getItemPath() if item.href is not present).
+        const href =
+          item.path && basePath ? basePath + '/' + item.path : item.href
+        return (
+          <SidebarNavMenuItem
+            item={{ ...item, href }}
```
ashleemboyer, line 42: As a heads up, `SidebarNavMenuItem` will be using the `path` and `href` properties to distinguish between internal and external links, respectively. (Working on this now). So if we introduce a `basePath` here, then I think it needs to be passed through to `SidebarNavItem` and `getItemPath` needs to be updated to use `basePath` if it exists instead of the current route.

zchsh: 👍 Thanks for flagging! Will see if I can get that working (just noticed the changes I've made so far don't seem to be working, ahah 😭 )

src/components/navigation-header/index.tsx
```diff
@@ -20,43 +20,51 @@ const isCurrentPage = (pagePath: string, currentPath: string): boolean => {
...
-                      {navLink.label}
-                    </a>
-                  </Link>
-                </li>
-              )
-            })}
-          </ul>
-          <HeaderSearchInput theme="dark" />
-        </div>
+        {currentProductSlug == 'waypoint' && (
```
zchsh, line 46: We don't want to show these `subnavItems` links on the homepage, and right now the code is specifically using `waypointData`, so added a conditional here.

src/components/product-switcher/index.tsx
```diff
@@ -53,6 +55,9 @@ const ProductSwitcher: React.FC = () => {
     })
   )
 
+  /* @TODO: handle case where there is no
```
zchsh, line 58: Have taken a VERY rough shot at this, but it doesn't feel like a proper fix (just one oriented at getting the demo done), so have added this `TODO`.

src/views/waypoint/home/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+    text: string
+  }
+  columns?: number
+  cards?: CardInterface[]
+}
+
+type Block =
+  | (HeadingProps & BlockBase)
+  | (GetStartedProps & BlockBase)
+  | (CardProps & BlockBase)
```
zchsh, line 46: I've been trying to get this "mixed type" card to work, but can't seem to avoid Typescript errors without the duplicative properties declared in the `BlockBase` type. Going to keep looking into this and try to sort it out 👍

CC @BRKalow @ashleemboyer in case either of y'all have some TypeScript wizardry or pattern suggestions to spare.

zchsh:
@BRKalow @ashleemboyer Roughing in the `/` and `/waypoint` pages in a very demo-oriented way. Very open to feedback here, and hot takes on how much we want to clean up & refactor this before vs after the demo 🙇

judithpatudith:
Is it possible to display the full text of truncated H2 headings in the sidecar on hover? This would be a nice bandaid for headers that are too long to be useful at their current character length. [Here are some examples](https://dev-portal-git-zsscaffold-home-page-hashicorp.vercel.app/waypoint/docs/troubleshooting#exec-into-the-application-container).

It would also be amazing to have some guidance on the new character limits on H2s and a list of headers that exceed that so we know what we're up against in terms of content changes.

src/components/product-card-grid/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+  nomad: IconNomadColor16,
+  packer: IconPackerColor16,
+  vagrant: IconVagrantColor16,
+  waypoint: IconWaypointColor16,
+  boundary: IconBoundaryColor16,
+}
+
+const iconDict = {
+  hcp: IconHashicorp16,
+}
```
ashleemboyer, line 29: This is very similar to what `ProductIcon` does, are you able to use that component here?

zchsh: @ashleemboyer Ah, nice! Thanks for calling this out 💯  I should be able to, might need to tweak a couple things so I can pass `className`. Will give that a shot, and will add `ProductIcon` to swingset 👍

zchsh:
@judithpatudith Thanks for the great feedback re: heading truncation in the `<Sidecar />` 💖  I've created [a task for us to address that issue](https://app.asana.com/0/1100423001970639/1201603303114657/f), I think we'll need to bring that feedback back to design. cc @braydenlove

src/components/product-card-grid/index.tsx
zchsh: @ashleemboyer Have updated to use `ProductIcon`, and added [some swingset docs for ProductIcon](https://dev-portal-git-zsscaffold-home-page-hashicorp.vercel.app/swingset/components/producticon) as well. Let me know what you think!

src/views/waypoint/home/index.tsx
zchsh: Have updated this to avoid the duplicative optional properties, but have had to case `block` as specific types in an `if else` block. This feels a lot better, but I suspect it isn't the truly correct way of doing things. Will keep digging 👀

src/views/waypoint/home/components/maybe-internal-link/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+  href: string
+  children: React.ReactNode
+} & React.HTMLProps<HTMLAnchorElement>): React.ReactElement {
+  const Elem = isAbsoluteUrl(href) ? InternalLink : 'a'
+  return (
+    <Elem href={href} {...rest}>
+      {children}
+    </Elem>
+  )
+}
```
BRKalow, line 19: I could definitely see this being useful elsewhere! 💭

braydenlove:
@judithpatudith @zchsh thanks for the feedback! I actually explored a few ideas around this earlier but ultimately felt that the shifting/wrapping of longer titles made the sidecar hard to use. I would like to avoid line wrapping or any shifting on this component if possible. Also, it is just one simple click away to see the full title via the anchor link scroll. I think over engineering an interaction like this can be a bit dangerous if we can try to solve for it from the content side first. That being said an ideal character limit is an awesome place to start and we can totally take a look at this again in future versions.

src/views/waypoint/home/components/heading/types.ts
```diff
@@ -0,0 +1,5 @@
+export interface HeadingProps {
+  __heading_slug: string
+  heading: string
+  level: number
```
BRKalow, line 4: This might allow us to avoid the `as React.ElementType` in the component itself:
```suggestion
  level: 1 | 2 | 3 | 4 | 5 | 6
```

src/views/waypoint/home/components/heading/index.tsx
```diff
@@ -0,0 +1,13 @@
+import { HeadingProps } from './types'
+
+function Heading({
+  __heading_slug,
+  heading,
+  level,
+}: HeadingProps): React.ReactElement {
+  const Component = `h${level}` as React.ElementType
+  return <Component id={__heading_slug}>{heading}</Component>
```
BRKalow, line 9: nit: Could this just be `id` or `slug`? 💭

src/views/waypoint/home/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+        if (type === 'heading') {
+          const { heading, __heading_slug, level } = block
+          return <Heading key={idx} {...{ __heading_slug, heading, level }} />
+        } else if (type === 'get_started') {
+          const { heading, iconSvg, text, link } = block
+          return <GetStarted key={idx} {...{ iconSvg, heading, text, link }} />
+        } else if (type === 'cards') {
+          const { columns, cards } = block
+          return <Cards key={idx} {...{ columns, cards }} />
+        }
```
BRKalow, line 42: I would say this or a `switch` statement is the correct way to go with a discriminated union type like you've got!

BRKalow (commented):
Looking really good! Nothing major stands out to me.

src/views/waypoint/home/components/heading/index.tsx
zchsh: Nice, yep, definitely could be - updated to `slug` 👍

src/views/waypoint/home/components/maybe-internal-link/index.tsx
zchsh: Sweet - I've ended up lifting this up into `components` to get client-side links in the `sidebar` going 👍

src/views/waypoint/home/components/cards/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+function Cards({ columns, cards }: CardProps): React.ReactElement {
+  return (
+    <div className={s.root}>
+      <div className={classNames(s.cards, s[`columns-${columns}`])}>
+        {cards.map((card, idx) => {
+          const { iconSvg, heading, text, tags } = card
+          return (
+            <div
+              // eslint-disable-next-line react/no-array-index-key
+              key={idx}
```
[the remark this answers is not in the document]
zchsh: Great callout.... tbh on the fence about this, as I'm worried `heading` might not be specific enough. Eg what if there are multiple sub-headings under broader heading sections, eg `## This one thing >> ### Install` & `## This other thing > ### Install` in the same `.mdx` doc... then I think we'd end up with identical `key` values, which I think is more dangerous than index-as-key? (esp since we're not expecting to re-render or re-order this content, I think?)

Let me know if that makes any sense, I feel like I've ran into this issue with not-guaranteed-unique keys before, but not sure 😬 🙇

src/components/navigation-header/index.tsx
zchsh: Great callout - added 👍

src/components/breadcrumb-bar/__tests__/get-docs-breadcrumbs.test.ts
```diff
@@ -14,6 +14,10 @@ describe('getDocsBreadcrumbs', () => {
       navData: waypointNavData,
     }
     const expected = [
+      {
+        title: 'Developer',
```
zchsh, line 18: Adds top-level `Developer` link to home page, per Figma mockups.

src/views/waypoint/home/server.ts
```diff
@@ -0,0 +1,86 @@
...
+  CONTENT.blocks = CONTENT.blocks.map((block) => {
+    const { type, heading } = block
+    if (type !== 'heading') return block
+    return {
+      ...block,
+      slug: slugify(heading, { lower: true }),
+    }
+  })
+
+  const navData = [
```
zchsh, line 31: Customized `navData` for the `/waypoint` product landing, again to match Figma (was previously the full `docs` nav data). Have added `TODO` comments above to signal that we need to refine content sourcing here.

Definitely expecting this to change (esp the `_demoExternalLink` part), this is very much targeted at getting things ready to demo.

src/views/waypoint/home/components/heading/style.module.css
```diff
@@ -0,0 +1,5 @@
+.h2 {
+  font-size: var(--token-typography-heading-2-font-size) !important;
+  color: var(--token-color-palette-neutral-700) !important;
+  margin: 24px 0 24px 0 !important;
```
zchsh, line 7: Needed to update these heading styles after merging in upstream work, as in comment.

src/components/sidebar/components/sidebar-back-to-link/style.module.css
```diff
@@ -1,4 +1,6 @@
 .backToLink {
+  composes: ds-token-typography-body-base from global;
+  font-weight: var(--ds-token-font-weight-medium);
```
zchsh, line 3: Tweaked styles here to match Figma.

src/components/sidebar/components/sidebar-filter-input/style.module.css
```diff
@@ -47,8 +47,10 @@
 .filterInput {
   composes: focus-border-and-box-shadow-light from global;
   background-color: transparent;
-  border: 1px solid white;
+  border-color: var(--token-elevation-inset-border-color);
   border-radius: 5px;
+  border-style: solid;
+  border-width: var(--token-elevation-inset-border-width);
```
zchsh, line 53: Tweaked styles here to match Figma, which shows border on filter input even on inactive state. Used these `--token-elevation-inset` styles because that seems to be the Effect style in use in Figma.

src/data/waypoint.json
```diff
@@ -1,29 +1,30 @@
 {
   "subnavItems": [
     {
-      "id": "reference-docs",
```
zchsh, line 4: Updated to match latest work in Figma.

src/layouts/docs/index.tsx
```diff
@@ -12,12 +13,16 @@ interface DocsLayoutProps {
   navData: MenuItem[]
   breadcrumbLinks?: BreadcrumbLink[]
   githubFileUrl?: string
+  backToLink?: {
+    text: string
+    url: string
+  }
```
zchsh, line 19: Added `backToLink` so that Figma mocks, which show different "Back to" links on `/waypoint` and `/waypoint/docs`, could be implemented.

src/components/sidebar/components/sidebar-nav/sidebar-nav-menu-item/index.tsx
```diff
@@ -84,8 +86,17 @@ const SidebarNavSubmenu: React.FC<{
...
   // TODO: the designs don't currently show a divider
+  // TODO: Update 2022-01-03: product home page, eg /waypoint, does
+  // TODO: show divider. Including a basic implementation for now,
+  // TODO: this is something we likely need to revisit with design.
   if (item.divider) {
-    return null
+    return <hr className={s.divider} />
+  }
+
+  // TODO: 2022-01-03: designs show a heading on the product home page,
```
zchsh, line 96: Have added `TODO`s here to signal that these features were implemented for the demo, so not with the usual rigour.

src/components/sidebar/components/sidebar-nav/index.tsx
zchsh: @ashleemboyer with the latest Figma mocks, which have custom content in the `/waypoint` view sidebar, was able to sidestep having to add `basePath` for now, and have gone with `href` for demo purposes. Have shimmed in an external link icon functionality separate to `href`, but definitely open to completely undoing that post-demo.

Have mulled this over a bit, and am thinking a longer-term solution here might be to update our server-side docs page or layout functions so that the returned `navData` already has the correct routes built in to `path` (ie `navData` contains paths like `/waypoint/docs/intro` rather than `/intro`). This way we could avoid having to complete the paths on the client-side using router utilities etc, and the `sidebar-nav` component might become a more flexible display component rather than having route-specific logic.... not a fully formed idea yet, but tossing it out there anyways ☺️

src/components/sidebar/components/sidebar-back-to-link/index.tsx
```diff
@@ -1,12 +1,16 @@
...
+  text?: string
+  url?: string
+}> = ({ text = 'Back to lorem ipsum', url = '/' }) => (
+  <MaybeInternalLink className={s.backToLink} href={url}>
     <IconArrowLeft16 className={s.icon} />
-    <span>Back to lorem ipsum</span>
-  </a>
+    <span>{text}</span>
+  </MaybeInternalLink>
 )
```
ashleemboyer, line 14: It was decided last week via Slack to use a placeholder for this component, I've implemented it in my sidebar PR here: https://github.com/hashicorp/dev-portal/pull/58/files#diff-969313d21e13fab2f5de2cb624c051abb51b3ab45018f0bedff88df293fbf37a

src/components/sidebar/components/sidebar-back-to-link/style.module.css
ashleemboyer: I've updated the styles for this component in my sidebar PR, could we undo the changes here?

src/components/sidebar/components/sidebar-filter-input/style.module.css
ashleemboyer: Similar comment, I've updated these styles in my sidebar PR. Could we undo these changes in this PR?

src/components/sidebar/components/sidebar-nav/sidebar-nav-menu-item/index.tsx
```diff
@@ -3,6 +3,8 @@ import { MenuItem } from 'components/sidebar'
 import useCurrentPath from 'hooks/use-current-path'
```
ashleemboyer, line 3: My sidebar PR implements external link icons, the Link component for internal links, dividers, and updated styles. Can we undo changes related to those item here?

src/data/waypoint.json
ashleemboyer: Could you link to the Figma designs you're referencing? I didn't know there were changes and I just want to make sure I'm up to date. 😊

src/pages/waypoint/index.tsx
```diff
@@ -1,15 +1,7 @@
...
-WaypointHomePage.layout = BaseNewLayout
-
-export default WaypointHomePage
+import HomePage from 'views/waypoint/home'
+import server from 'views/waypoint/home/server'
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export async function getStaticProps() {
+  return await server.getStaticProps()
+}
+export default HomePage
```
ashleemboyer, line 7: Just requesting some extra whitespace here for readability.

```suggestion
import HomePage from 'views/waypoint/home'
import server from 'views/waypoint/home/server'

// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
export async function getStaticProps() {
  return await server.getStaticProps()
}

export default HomePage
```

src/views/waypoint/home/components/get-started/style.module.css
```diff
@@ -0,0 +1,81 @@
...
+}
+
+.heading {
+  color: var(--token-color-palette-neutral-700);
+  /**
+    * NOTE: !important is necessary here, as higher specificity styles
+    * are bleeding in from docs layout styles. One path to resolve this
+    * might be to use MDX custom components, rather than raw-element-targeting styles,
+    * in order to styles docs contents
+    **/
```
ashleemboyer, line 37: Agree with this, I think a good first test will be with the headings and permalink icons!

src/views/waypoint/home/components/get-started/types.ts
```diff
@@ -0,0 +1,9 @@
+export interface GetStartedProps {
+  iconSvg: string
+  heading: string
+  text: string
+  link: {
+    url: string
+    text: string
+  }
+}
```
ashleemboyer, line 9: Do you prefer to put component props interfaces in separate files than the component?

src/views/waypoint/home/content.json
```diff
@@ -0,0 +1,125 @@
...
..."<svg width=\"96\" height=\"96\" viewBox=\"0 0 96 96\" fill=\"none\" xmlns=\"http://www.w3.org/2000/svg\">\n<path d=\"M0 6C0 2.68629 2.68629 0 6 0H90C93.3137 0 96 2.68629 96 6V90C96 93.3137 93.3137 96 90 96H6C2.68629 96 0 93.3137 0 90V6Z\" fill=\"url(#paint0_linear_219_677)\"/>\n<path d=\"M6 1H90V-1H6V1ZM95 6V90H97V6H95ZM90 95H6V97H90V95ZM1 90V6H-1V90H1ZM6 95C3.23858 95 1 92.7614 1 90H-1C-1 93.866 2.13401 97 6 97V95ZM95 90C95 92.7614 92.7614 95 90 95V97C93.866 97 97 93.866 97 90H95ZM90 1C92.7614 1 95 3.23858 95 6H97C97 2.13401 93.866 -1 90 -1V1ZM6 -1C2.13401 -1 -1 2.13401 -1 6H1C1 3.23858 3.23858 1 6 1V-1Z\" fill=\"#CBF1F3\"/>\n<path d=\"M79.6579 28.2935L71.7513 41.9868L63.8446 28.2935H79.6579Z\" fill=\"#14C6CB\"/>\n<path d=\"M37.4487 37.4357H32.1731L45.3598 60.2801L40.0842 69.4179L16.342 28.2935H42.7242L55.9153 51.1423L58.5509 46.5735L47.9998 28.2935H58.5509L63.8309 37.4357L69.1065 46.5735L55.9153 69.4179L37.4487 37.4357Z\" fill=\"#14C6CB\"/>\n<defs>\n<linearGradient id=\"paint0_linear_219_677\" x1=\"96\" y1=\"96\" x2=\"0\" y2=\"0\" gradientUnits=\"userSpaceOnUse\">\n<stop stop-color=\"#E0FCFF\"/>\n<stop offset=\"1\" stop-color=\"#F6FEFF\"/>\n</linearGradient>\n</defs>\n</svg>\n",
```
ashleemboyer, line 12: I think it will be nice in the future (once we know for sure our custom icon set) to create an Icon component and a map of icon slugs/names to their svgs, then we could replace this svg content with the slug/name and have the code that renders these svgs use the component. I think that would improve the readability of this file but definitely not needed for this PR. 😊

src/views/waypoint/home/server.ts
ashleemboyer: The code in my sidebar PR renders an external link icon when the `href` prop is present, so I think this should "just work" without the `_demoExternalLink` property once this is working off of those changes.

ashleemboyer (commented):
A few things here conflict with my sidebar PR, if we need to jump on a call to work those out and discuss I'm happy to do so!

src/components/sidebar/components/sidebar-back-to-link/style.module.css
zchsh: For sure! Will review #58 now, and happy to wait till that's merged, then rebase this work. With the `href` > `External link icon` functionality in #58, I might need to add back in the `basePath` work I had here previously, but that should fine for demo purposes, I think 👍

src/data/waypoint.json
zchsh: For sure - I believe it's the same `Dev Portal` file ([link](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Dev-Portal?node-id=536%3A39227)) and page (`Page Templates`). I saw some work happening in there yesterday, and when I ran the preview links from this PR by design, there was some QA related to recent changes.

src/components/sidebar/components/sidebar-back-to-link/index.tsx
zchsh: Ah okay - most recent feedback I have from design is to do `← Developer` on `/waypoint` and `← Waypoint` on `/waypoint/docs`. Will wait till #58 merges, and will double-check with design what they'd like to demo before merging this PR 👍

src/views/waypoint/home/content.json
zchsh: ☝️ 💯  For sure, icon dictionaries seem like a good call. Was even thinking this content (including the icon source files) may live outside the codebase (eg Dato or something), so was aiming at trying to figure out which parts should be author-able and split those out.

src/components/sidebar/components/sidebar-back-to-link/index.tsx
ashleemboyer: Here's the latest thread I have: https://hashicorp.slack.com/archives/C028LTF7TJ5/p1640279174192700?thread_ts=1640273232.181900&cid=C028LTF7TJ5. This is definitely confusing! Can we try to elicit that feedback in `#proj-dev-portal-eng` that way everyone will be aware of the latest? I'm also having a working session later today with Brayden where I'll be getting some QA feedback, I'll make sure we what we cover documented somewhere like Slack or Asana or a Google doc.

src/views/waypoint/home/content.json
ashleemboyer: Good points, Jimmy is/was working on something similar for dot-com right? I wonder if he has some helpful insights he's come across in that endeavor!

src/components/sidebar/components/sidebar-back-to-link/style.module.css
zchsh: Have rebased! I think we still need the `composes: body-base` and `font-weight` updates here to match the latest mockups, so have left those in.

src/components/sidebar/components/sidebar-filter-input/style.module.css
```diff
@@ -47,7 +47,7 @@
 .filterInput {
   composes: focus-border-and-box-shadow-light from global;
   background-color: transparent;
-  border: 1px solid white;
+  border-color: var(--token-elevation-inset-border-color);
```
zchsh, line 50: Adds border to filter input, to match mockups.

ashleemboyer: So two things here:

- The border of the input is being handled by the `box-shadow` below, and that came from the Figma designs
- the 1px white border is needed in the input's default state so that it doesn't shift when receiving focus

src/views/waypoint/home/server.ts
```diff
@@ -0,0 +1,85 @@
...
+    {
+      title: 'Community Forum',
+      href: 'https://discuss.hashicorp.com',
+      _demoExternalLink: true,
+    },
+    {
+      title: 'Support',
+      href: 'https://support.hashicorp.com',
+      _demoExternalLink: true,
+    },
```
ashleemboyer, line 57: Not sure if you've seen this yet or not, but it looks like these links are showing at the top of the sidebar in the docs, plugins, and cli pages after merging in my sidebar changes.

image

src/components/sidebar/index.tsx
```diff
@@ -22,49 +22,51 @@ export interface MenuItem {
...
-      itemCopy.isActive = foundActiveItem
-      itemCopy.fullPath = `/${currentProductSlug}/${currentProductSubpage}/${item.path}`
+      // Note: if an active item has already been found,
+      // we do not flag this node as active.
+      itemCopy.isActive = !foundActiveItem && currentPath.endsWith(item.path)
       itemCopy.id = `menu-item-${itemCopy.fullPath
         .replace(/\//g, '-')
         .toLowerCase()}`.replace(/-+/g, '-')
+      // Flag if we've found an active item
+      foundActiveItem = itemCopy.isActive || foundActiveItem
```
ashleemboyer, line 69: Can we talk about why this is being changed so much? Is something broken?

src/components/sidebar/components/sidebar-filter-input/style.module.css
zchsh: Ah nice, seeing this now, reverted this 👍 thank you! 💖

src/views/waypoint/home/components/get-started/types.ts
zchsh: No strong preference, have seen a separate `types.ts` file a few places. Happy to move this inline in the file if that feels more conventional 👍

src/data/waypoint.json
ashleemboyer: I'm seeing three different things there so I'm not sure which is the right one 😆

- ["Reference Docs"](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Dev-Portal?node-id=1377%3A50800)
- ["Docs"](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Dev-Portal?node-id=3241%3A65842)
- ["Reference"](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Dev-Portal?node-id=3274%3A66112)

src/components/sidebar/components/sidebar-back-to-link/index.tsx
ashleemboyer: I went back to that Slack thread to ask for clarification for what we want to show tomorrow, and the vote is for showing a placeholder for now: https://hashicorp.slack.com/archives/C028LTF7TJ5/p1641330820226500?thread_ts=1640273232.181900&cid=C028LTF7TJ5

src/components/sidebar/index.tsx
zchsh: @ashleemboyer  For sure, apologies for not reviewing this yet, was planning on it. There are two main changes I'm suggesting here:

1. Removing the `fullPath` creation, and instead having this be the responsibility of the consumer of `sidebar`. This lets us fill in the `fullPath` values only once, at build time rather than on the client. More importantly, this enables the flexibility we need to show custom, non-external `sidebar` items on the `/waypoint` product view. i don't necessarily mean to suggest this as a long-term solution (though I think it might be a viable path!), mainly just focused on getting the product view ready to demo. Without this change, I don't think we can properly show the links needed on the `/waypoint` page sidebar (ie would need to use `href` but that would show external link icons which would not match design... or we could pass down `basePath` as suggested earlier, but I think that might add complexity compared to this solution).
2. The `foundActiveItem` refactoring is less important, happy to drop these changes... but I believe the `if (foundActiveItem)` conditional might not be functioning as expected... it seems to set an item's `hasActiveChild` or `isActive` to `false` in that first `if` block, but then in the subsequent `if / else` block, I think that would be undone by the `itemCopy.hasActiveChild` or `itemCopy.isActive` assignments. Refactored because after making the `fullPath` changes, for some reason I was having issues getting the category items to expand. I think longer-term, we may want to leverage something like the [existing utility in docs-sidenav](https://github.com/hashicorp/react-components/blob/main/packages/docs-sidenav/utils/flag-active-nodes.js) to retain existing behaviour of also highlighting `href` values, and handling some rare "alias" edge cases.

src/views/waypoint/home/components/get-started/types.ts
ashleemboyer: Feel free to leave it for now, I'll add it as an agenda item for our code style jam next week.

src/components/sidebar/index.tsx
zchsh: ☝️ Let me know if that's helpful - happy to adjust the approach here 🙇‍♂️

src/components/product-switcher/index.tsx
```diff
@@ -33,7 +33,9 @@ const getLastProduct = (products: Product[][]) => {
 
 const ProductSwitcher: React.FC = () => {
   const router = useRouter()
-  const currentProductSlug = router.asPath.split('/')[1] as ProductSlug
+  const currentProductSlug = router.asPath
+    .split('/')[1]
+    .split('#')[0] as ProductSlug
```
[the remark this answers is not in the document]
zchsh: Great call, will update this!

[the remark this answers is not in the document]
zchsh: ✅ Updated

src/components/product-switcher/index.tsx
```diff
@@ -197,7 +202,9 @@ const ProductSwitcher: React.FC = () => {
       <button
         aria-controls={OPTION_LIST_ID}
         aria-expanded={isOpen}
-        aria-labelledby={generateSwitcherOptionIdFromProduct(currentProduct)}
+        aria-labelledby={generateSwitcherOptionIdFromProduct(
+          currentProduct || getFirstProduct(products)
```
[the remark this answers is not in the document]
zchsh: Ah nice! 💯 Have updated, thanks for calling this out 🙇

src/views/waypoint/home/server.ts
zchsh: Thanks for calling this out! Have updated slightly, ended up being a little less dry, but have the `/waypoint` page sidebar contents displaying 👍

src/styles/_temp-ds-tokens.css
```diff
@@ -0,0 +1,318 @@
+/**
```
ashleemboyer, line 1: I know this file comes from somewhere else, but I think it might need changes? So I just want to make a comment so we hopefully don't forget to raise it as a question later.

- I believe the non-mono fonts need to be this:
  ```
  system-ui, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol'
  ```
- I believe the mono font families need to be this:
  ```
  ui-monospace, Menlo, Consolas, Ubuntu Mono, monospace
  ```

src/components/product-switcher/index.tsx
```diff
@@ -32,8 +32,8 @@ const getLastProduct = (products: Product[][]) => {
 }
 
 const ProductSwitcher: React.FC = () => {
-  const router = useRouter()
-  const currentProductSlug = router.asPath.split('/')[1] as ProductSlug
+  const currentPath = useCurrentPath()
```
ashleemboyer, line 35: If I'm understanding your intention here, I think you also need to send two options in this `useCurrentPath` call:

```suggestion
  const currentPath = useCurrentPath({ excludeHash: true, excludeSearch: true })
```

src/components/product-switcher/index.tsx
```diff
@@ -197,7 +200,9 @@ const ProductSwitcher: React.FC = () => {
       <button
         aria-controls={OPTION_LIST_ID}
         aria-expanded={isOpen}
-        aria-labelledby={generateSwitcherOptionIdFromProduct(currentProduct)}
+        aria-labelledby={generateSwitcherOptionIdFromProduct(
+          currentProduct || firstProduct
+        )}
```
ashleemboyer, line 205: I didn't catch this when I reviewed earlier, but using `firstProduct` here will actually break the screen reader presentation of this component on the `/` page.

Here's a screenshot of the HTML for reference:

image

So what happens when the `ProductSwitcher` comes into focus on the home page, the screen reader says "Terraform, collapsed, button." This is fine when we're on the terraform page (mostly fine, I think we could improve it but it's a good start), but not helpful from the homepage.

Short term fix I'm proposing for this PR:

- update `aria-labelledby` to be conditionally set like this:
  ```
  aria-labelledby={currentProduct ?
    generateSwitcherOptionIdFromProduct(currentProduct) :
    undefined
  }
  ```
- add an `aria-label` prop, and set it conditionally as well:
  ```
  aria-label={currentProduct ?
    undefined :
    'Product switcher'
  }
  ```

I don't like this solution long term, but it does ensure we have the correct screen reader presentation for now!

src/components/sidebar/components/sidebar-back-to-link/index.tsx
```diff
@@ -1,33 +1,16 @@
...
-}
+// TODO: design is still planning this functionality & states
+const SidebarBackToLink: React.FC<{
+  text?: string
+  url?: string
+}> = ({ text = 'Back to [Page]', url = '/' }) => (
+  <MaybeInternalLink className={s.backToLink} href={url}>
+    <IconArrowLeft16 className={s.icon} />
+    <span>{text}</span>
+  </MaybeInternalLink>
```
ashleemboyer, line 13: Is it possible to make this conditionally render the popover when the props aren't provided? I think that's a good in-between for the pages this doesn't work on yet?

src/components/sidebar/components/sidebar-back-to-link/style.module.css
ashleemboyer: 👍🏻

src/components/sidebar/components/sidebar-filter-input/style.module.css
```diff
@@ -47,7 +47,7 @@
 .filterInput {
   composes: focus-border-and-box-shadow-light from global;
   background-color: transparent;
-  border: 1px solid white;
+  border: 1px solid transparent;
```
ashleemboyer, line 50: 👍🏻

src/components/sidebar/components/sidebar-nav/sidebar-nav-menu-item/index.tsx
ashleemboyer: I think this'll stay long term, or a version of it will, because it's possible we'll be replacing dividers with text headers. It'll definitely help us if that's the case!

src/components/sidebar/index.tsx
```diff
@@ -1,4 +1,5 @@
 import { useMemo, useState } from 'react'
+import slugify from 'slugify'
```
ashleemboyer, line 2: FYI this removes `/`, doesn't replace them with `-`, which is why I didn't end up using it. I think we need to discuss this but okay to do that outside of this PR, just want to discuss soon after.

src/layouts/docs/utils/add-full-paths-to-nav-data.ts
```diff
@@ -0,0 +1,48 @@
...
+  path: string
+}
+
+function isNavBranch(value: NavNode): value is NavBranch {
+  return value.hasOwnProperty('routes')
+}
+
+function isNavLeaf(value: NavNode): value is NavLeaf {
+  return value.hasOwnProperty('path')
+}
```
ashleemboyer, line 26: I currently don't agree that we should be importing from `react-docs-sidenav` because my current understanding is that the code we're writing is meant to replace that code, but let's add that as another topic to a discussion we have following this PR. Not a blocker for merging. 👍🏻

src/layouts/docs/utils/add-full-paths-to-nav-data.ts
```diff
@@ -0,0 +1,48 @@
...
+    // For nodes with routes, add fullPaths to all routes
+    return { ...node, routes: addFullPathsToNavData(node.routes, basePaths) }
+  } else if (isNavLeaf(node)) {
+    // For nodes with paths, add fullPath to the node
+    return { ...node, fullPath: `/${basePaths.join('/')}/${node.path}` }
+  } else {
+    // Otherwise return the node unmodified
+    return node
+  }
+}
```
ashleemboyer, line 46: Another topic to add to a follow-up discussion: in the same spirit of these changes, I think we should also generate the menu item `id`s at build time since they're based off of data that's available at that time. Something to be mindful of when we do that: make sure we are only iterating over the menu items once, so we will probably need one function that adds full paths & ids and then some other functions for generating those properties.

src/views/waypoint/home/server.ts
```diff
@@ -0,0 +1,87 @@
+import slugify from 'slugify'
```
ashleemboyer, line 1: Would like to discuss the introduction of this page, mainly to understand your decision making for context reasons. But that can come after the PR since this works!

src/views/waypoint/home/server.ts
```diff
@@ -0,0 +1,87 @@
...
+    { title: 'Introduction', fullPath: '/waypoint/docs/intro' },
+    {
+      title: 'Getting Started',
+      fullPath: '/waypoint/docs/getting-started',
+    },
+    { divider: true },
+    { heading: 'Resources' },
+    {
+      title: 'Releases',
+      href: 'https://releases.hashicorp.com',
```
ashleemboyer, line 40: Should this link to one of these two links?

- https://releases.hashicorp.com/waypoint/
- https://releases.hashicorp.com/waypoint-entrypoint/

src/views/waypoint/home/server.ts
```diff
@@ -0,0 +1,87 @@
...
+      _demoExternalLink: true,
+    },
+    {
+      title: 'HashiCorp Learn',
+      href: 'https://learn.hashicorp.com',
```
ashleemboyer, line 45: I think this is supposed to be: https://learn.hashicorp.com/waypoint

src/views/waypoint/home/server.ts
```diff
@@ -0,0 +1,87 @@
...
+      _demoExternalLink: true,
+    },
+    {
+      title: 'Community Forum',
+      href: 'https://discuss.hashicorp.com',
```
ashleemboyer, line 50: I think this is supposed to be: https://discuss.hashicorp.com/c/waypoint/51

ashleemboyer (changes requested):
A couple of changes requested before merging, but many other comments that note some things I'd like to pair over tomorrow via Zoom if possible!

src/components/sidebar/index.tsx
ashleemboyer: I'd like to have non-textual discussion about this since there are a few different things at play, but looks like it's working right now so no blocker here for merging. 👍🏻

config/products.ts
```diff
@@ -29,7 +29,7 @@ const productSlugsToUrls: { [key in ProductSlug]: string } = {
   terraform: 'https://www.terraform.io/docs/index.html',
   vagrant: 'https://www.vagrantup.com/docs',
   vault: 'https://www.vaultproject.io/docs',
-  waypoint: '/waypoint/docs',
+  waypoint: '/waypoint',
```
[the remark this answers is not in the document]
ashleemboyer: Yep, this is what we want now that we have a Waypoint homepage with content!

src/styles/_temp-ds-tokens.css
zchsh: ☝️Great call - this file comes from https://github.com/hashicorp/design-system-tokens/blob/main/dist/products/css/tokens.css . I think this is something we should flag with the design system team! 💯

src/components/product-switcher/index.tsx
zchsh: Perfect, thank you! 💯

src/layouts/docs/utils/add-full-paths-to-nav-data.ts
zchsh: ☝️ yeah this is a great call - I can try to stub something out quickly to add `id`s. 

Is the concern around iterating over menu items only once performance? Or is there something else?

src/views/waypoint/home/server.ts
zchsh: Honestly not sure - https://releases.hashicorp.com/waypoint/ certainly seems like a better bet than the current https://releases.hashicorp.com, will update with that! 👍

src/views/waypoint/home/server.ts
zchsh: Sounds good, very on board with that! 💯  To clarify, `page` as in the `/waypoint` page, or `package` as in `slugify`?

src/components/product-switcher/index.tsx
zchsh: Ah this is a great catch - really appreciated 💯  Have updated with your short term fix, and agreed on exploring what else we could do longer term 👍

src/layouts/docs/utils/add-full-paths-to-nav-data.ts
zchsh: Sounds good - yeah just reaching for what's available, very open to updating with new code 👍

src/components/sidebar/components/sidebar-back-to-link/index.tsx
zchsh: Excellent idea, love it 💯 have taken a shot at implementing it 👍

src/views/waypoint/home/server.ts
zchsh: ✅ Updated, ty!

src/views/waypoint/home/server.ts
zchsh: ✅ Updated, ty!

src/layouts/docs/utils/add-full-paths-to-nav-data.ts
zchsh: Took a shot at this with a single iteration 👍

src/views/waypoint/home/server.ts
ashleemboyer: I didn't word that well! 😅  Please replace the word page `page` with `file`. The `src/views/waypoint/home/server.ts` file in general is what I'd like to learn more about. 😁

ashleemboyer (approved):

src/views/waypoint/home/server.ts
zchsh: ah sweet! Yeah for sure. Zoom's probably best, IIRC i did this cause of the combination of our `views / pages` organization with the NextJS requirement to split server-only imports into separate files, but would definitely love to dig into it a bit more and see if there's a different pattern we could adopt 🧐

braydenlove:
> Is it possible to display the full text of truncated H2 headings in the sidecar on hover? This would be a nice bandaid for headings that are too long to be useful at their current character length. [Here are some examples](https://dev-portal-git-zsscaffold-home-page-hashicorp.vercel.app/waypoint/docs/troubleshooting#exec-into-the-application-container).
> 
> It would also be amazing to have some guidance on the new character limits on H2s and a list of headings that exceed that so we know what we're up against in terms of content changes.

Hey @judithpatudith! Thanks for your thoughtful feedback. Showing the full text on hover would cause content to jump which is something we are trying to avoid based on technical constraints and UX norms. We are also trying to avoid having these wrap to multiple lines as the purpose of this component is to help folks quick navigate deeper into the page. That being said, having some idea of what they are clicking on is a must which is why we went the truncate route. It is tricky as I don't want to limit headers as they are intentionally written but don't think they are totally necessary or useful in the sidecar as its purpose is to help folks navigate quickly. We also have some technical constraints around the sidecar's sticky property... What is a good solution here? Perhaps we can allow folks to write a shorter sidecar title that gives the user enough context to what the anchor link is? Seeing the full h2 content is one simple interaction away but we need to be sure the titles are helpful. Open to feedback here!

heitzke:
I appreciate the conversation and trying to build the best solution possible. I do think we're over-indexing on this one a little bit and think we should set line-length guidance, truncate the text and then readdress it all later if this component ends up causing any confirmed heartburn or usability problems. Ultimately is more of a "nice to have" component and not likely to be a primary path of wayfinding.

andreadaniels1:
Just curious.... What is the character limit set to?

PR #60, opened 2021-12-23
feat: Add DocsAnchor to fix internal links

BRKalow:
👀 [Preview](https://dev-portal-git-brkfeat-docs-anchor-hashicorp.vercel.app/waypoint/docs/intro)

Adding `DocsAnchor` to rewrite internal links so that they have the right product path segment. In the future it'd be ideal if we had some sort of `NavigationContext` which contained the current product and relevant base paths so that we didn't have to pass them in.

Usage:
```ts
import makeDocsAnchor from 'components/docs-anchor'

const components = {
  a: makeDocsAnchor('waypoint', ['docs', 'commands', 'plugins'])
}
```

### DIFF
--- a/src/components/docs-anchor/index.tsx
+++ b/src/components/docs-anchor/index.tsx
@@ -1,0 +1,36 @@
+/* eslint-disable jsx-a11y/anchor-has-content -- We're just passing through props here, so the links should be well formatted */
+import { ProductSlug } from 'types/products'
+import Link from 'next/link'
+
+export default function makeDocsAnchor(
+  product: ProductSlug,
+  basePaths: string[]
+) {
+  const DocsAnchor: React.FC<JSX.IntrinsicElements['a']> = ({
+    href,
+    ...rest
+  }) => {
+    if (!href) return <a {...rest} />
+
+    let adjustedHref = href
+    // TODO: infer this condition dynamically
+    if (basePaths.some((basePath) => href.startsWith(`/${basePath}`))) {
+      // TODO: infer the product dynamically
+      adjustedHref = `/${product}${href}`
+    }
+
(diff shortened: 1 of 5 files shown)


src/components/docs-anchor/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+    ...rest
+  }) => {
+    if (!href) return <a {...rest} />
+
+    let adjustedHref = href
+    // TODO: infer this condition dynamically
+    if (basePaths.some((basePath) => href.startsWith(`/${basePath}`))) {
+      // TODO: infer the product dynamically
+      adjustedHref = `/${product}${href}`
+    }
```
BRKalow, line 20: I think in an ideal world we could take advantage of something like a custom `NavigationContext`, which we've talked in a few other PRs, to get access to `product` and valid base paths.

src/components/docs-anchor/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export default function makeDocsAnchor(
+  product: ProductSlug,
+  basePaths: string[]
+) {
+  const DocsAnchor: React.FC<JSX.IntrinsicElements['a']> = ({
+    href,
+    ...rest
+  }) => {
+    if (!href) return <a {...rest} />
```
zchsh, line 13: Super nitty but instead of disabling `anchor-has-content`, I think you could do:

```suggestion
    href,
    children,
    ...rest
  }) => {
    if (!href) return <a {...rest}>{children}</a>
```

☝️  Not actually sure this is preferable, just wanted to put the idea out there.

BRKalow: It worked! Great call.

zchsh (approved):

PR #61, opened 2021-12-23
fix: restore VERCEL_ENV to fix local dev experience

zchsh:
This PR restores `VERCEL_ENV="production"`. Without this, devs working locally would have to manually set this variable in order for Waypoint's remote content to work as expected.

### DIFF
--- a/.env
+++ b/.env
@@ -1,4 +1,9 @@
 ENABLE_VERSIONED_DOCS=true
+# TODO: docs-page/server code requires VERCEL_ENV="production" to be set
+# TODO: in order to load remote content. We may want to update the server
+# TODO: code to rely only on the ENABLE_VERSIONED_DOCS env var instead,
+# TODO: as setting this manually for local dev feels inaccurate.
+VERCEL_ENV="production"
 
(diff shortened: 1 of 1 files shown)


.env
```diff
@@ -1,4 +1,9 @@
 ENABLE_VERSIONED_DOCS=true
+# TODO: docs-page/server code requires VERCEL_ENV="production" to be set
+# TODO: in order to load remote content. We may want to update the server
+# TODO: code to rely only on the ENABLE_VERSIONED_DOCS env var instead,
+# TODO: as setting this manually for local dev feels inaccurate.
+VERCEL_ENV="production"
```
BRKalow, line 6: I think we should be able to mitigate this by migrating to the loaders pattern, which allows us to bypass the implicit toggling based on environment variables. I can look at making this switch soon! For now this seems reasonable 👍

BRKalow (approved):

.env
kylemac: 🙇‍♂️ yes, a more durable fix here is the way to go but recognize this is a totally valid band-aid

PR #62, opened 2021-12-23
feat: Dockerfile for local preview

BRKalow:
Adding a Dockerfile which will serve local preview.

Validating in waypoint with:
```
REPO=waypoint docker run -it --publish "3000:3000"  \
  --workdir "/app" \
  --volume "$(pwd)/content:/app/content" \
  --volume "$(pwd)/public:/app/public" \
  --volume "$(pwd)/data:/app/data" \
  --volume "$(pwd)/redirects.js:/app/redirects.js" \
  --volume "next-dir:/app/website-preview/.next" \
  --volume "$(pwd)/.env:/app/.env" \
  -e REPO \
  dev-portal-local
```

I've also added a new GH workflow which will publish a docker image to Docker Hub on push to master.


Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM docker.mirror.hashicorp.services/node:14.17.0-alpine AS deps
```
BrandonRomano, line 1: Didn't know we had a mirror, this is cool.

BrandonRomano (approved):
Looks good!

.github/workflows/docker_publish.yml
```diff
@@ -0,0 +1,87 @@
+name: Docker Publish
```
BRKalow, line 1: Largely adapted from: https://github.com/actions/starter-workflows/blob/main/ci/docker-publish.yml

zchsh (commented):
@BRKalow This looks good to me!

Tried to build the Docker image locally, but received `optipng`-related errors during the `npm i` step that only went away if I added `--platform=linux/amd64` after the `FROM`s in the `Dockerfile` (I think cause M1?). Once I got the image building, tried to run it on a local copy of `waypoint` using the command as in the PR description, but kept getting `start.sh: applet not found`.

Are there modifications I'd need to make in the `waypoint` repo before running the `docker run` command? (I feel almost certain this is just user error on my end - happy to just merge this rather than iron out my own setup; also happy to keep trying so I can provide a more reassuring review ahah ☺️ )

BRKalow:
@zchsh Thanks for testing it out! As a quick sanity check, can you try it on this waypoint branch? `zs.add-dev-portal-preview`

PR #63, opened 2022-01-05
fix: patch code render in breadcrumbs

zchsh:
👀 [Preview](https://dev-portal-git-zspatch-breadcrumb-html-issue-hashicorp.vercel.app/waypoint/docs/waypoint-hcl/build)

This PR fixes an issue with `<code>` element rendering in the `<BreadcrumbBar />` component.

## Before

CleanShot 2022-01-05 at 11 22 10@2x

## After

CleanShot 2022-01-05 at 11 30 07@2x

<details>
<summary>Alternate treatment</summary>

This treatment renders the `<code>` element rather than stripping it out. It felt like more of a departure from the appearance of the rest of the breadcrumb bar, so was dropped in favour of stripping out HTML tags from titles when rendering them in the breadcrumb bar.

CleanShot 2022-01-05 at 11 23 06@2x

</details>

### DIFF
--- a/src/components/breadcrumb-bar/index.tsx
+++ b/src/components/breadcrumb-bar/index.tsx
@@ -44,9 +44,10 @@ function BreadcrumbBar({
                 className={s.breadcrumbText}
                 href={url}
                 aria-current={isCurrentPage ? 'page' : undefined}
-              >
-                {title}
-              </Elem>
+                // Note: HTML setting is necessary here, as some titles
+                // are formatted as "<code>Some title</code>"
+                dangerouslySetInnerHTML={{ __html: title }}
+              />
             </li>
           )
         })}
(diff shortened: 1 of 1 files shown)


zchsh:
@braydenlove This fixes the issue with `<code>` rendering into the breadcrumb bar. Let me know if this looks like the right solution, also explored an alternate solution (as in PR description) but figured the visually simpler option was preferable 👍

ashleemboyer (approved):
Great fix! 🥇

zchsh:
@braydenlove Going to merge this in as-is since it'll at least make things look non-buggy - definitely happy to adjust & iterate further if we want a different appearance for `<code>` in BreadcrumbNav 👍

braydenlove:
@zchsh looks great, thanks so much!

PR #65, opened 2022-01-06
Fix product switcher non-clickable edges

ashleemboyer:
🔎 [Preview link](https://dev-portal-git-ambfix-product-switcher-hover-hashicorp.vercel.app/)
🎟️ [Asana ticket](https://app.asana.com/0/1201010428539925/1201602267333025/f)

## What

This PR fixes an issue where it was possible to hover over a `ProductSwitcher` option around its edges but not be able to click on it. This was because of the padding within the `<li>` elements containing the clickable `<a>` elements.

## How

I removed the padding within the `<li>` elements, increased the padding of the `<a>` elements to achieve the desired height, added a `<span>` directly within the `<a>` element for each switcher option, and then when the `<a>` receives focus that `<span>` gets the custom focus outline applied to match the "inset" style that's been specified in the designs.

## Testing

- [ ] Nothing should have changed visually about the default, hover, or focus states of switcher options
- [ ] Each switcher option should be entirely clickable now

### DIFF
--- a/src/components/product-switcher/index.tsx
+++ b/src/components/product-switcher/index.tsx
@@ -80,11 +80,15 @@ const ProductSwitcher: React.FC = () => {
   }, [isOpen])
 
   /**
-   * It would be more optimal to set onKeyDown for the containing element, but that is not allowed
-   * on a <div> with no role. We do not want to use a menu role (see https://adrianroselli.com/2017/10/dont-use-aria-menu-roles-for-site-nav.html)
-   * so there is no role currently set on the containing <div>. If we find an appropriate role, then we can
-   * change the approach. For now, it's most appropriate to set onKeyDown on the <button> and <ul>,
-   * hence the `KeyboardEventHandler<HTMLButtonElement | HTMLUListElement>` definition.
+   * It would be more optimal to set onKeyDown for the containing element, but
+   * that is not allowed on a <div> with no role. We do not want to use a menu
+   * role (see Adrian Roselli link below) so there is no role currently set on
+   * the containing <div>. If we find an appropriate role, then we can change
+   * the approach. For now, it's most appropriate to set onKeyDown on the
+   * <button> and <ul>, hence this defintion:
+   * `KeyboardEventHandler<HTMLButtonElement | HTMLUListElement>`.
+   *
+   * https://adrianroselli.com/2017/10/dont-use-aria-menu-roles-for-site-nav.html
    */
   const handleKeyDown: KeyboardEventHandler<
     HTMLButtonElement | HTMLUListElement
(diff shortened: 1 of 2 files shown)


zchsh (approved):
✅ Looks great, love the idea of a  `Testing` section in the PR description 😍 💯  Tested based on those criteria, seems great!

ashleemboyer:
> ✅ Looks great, love the idea of a `Testing` section in the PR description 😍 💯 Tested based on those criteria, seems great!

@zchsh thank you! I'm going to do a quick PR later for adding a description template since that's also relevant to some other stuff I want to try automating. 😊

zchsh:
@ashleemboyer Ack wait I think I missed something testing `:focus` styles - actually not seeing those show up, I think? Let me know if you can repro!

ashleemboyer:
@zchsh that's expected, have you tried the `:focus-visible` option? I'm assuming you're using the browser dev tools to change the state of the anchor elements?

zchsh:
@ashleemboyer I tried at first just with keyboard nav, and didn't see it, and yep tried to confirm in browser dev tools with `:focus` but didn't see it there either (though maybe that latter confirmation isn't meaningful)

ashleemboyer:
Interesting, could you check this Keyboard setting and see if it's enabled on your machine?

> Use keyboard navigation to move focus between controls

image

zchsh:
@ashleemboyer It does seem to be set... here's a screencap of this branch:

https://user-images.githubusercontent.com/4624598/148457259-c6214803-7071-4226-841e-6f7a59aff0d8.mp4

And a screencap of upstream:

https://user-images.githubusercontent.com/4624598/148457267-7d5efde4-c142-425b-b6e8-258e35243052.mp4

I do notice that upstream, setting `:focus` in dev tools shows the same focus styles. This might just be me not being as familiar with `:focus-visible`!

ashleemboyer:
Looks like there's another setting specifically for Safari that needs to be enabled as well. Here's an article with the steps: https://www.a11yproject.com/posts/macos-browser-keyboard-navigation/

ashleemboyer:
`focus-visible` still doesn't seem to be working correctly in Safari, so I created a task for it: https://app.asana.com/0/1201010428539925/1201626190223621/f

zchsh:
@ashleemboyer I think I've got that toggled on as well:

CleanShot 2022-01-06 at 16 57 40@2x

CleanShot 2022-01-06 at 16 57 53@2x

I do find that if I view the examples on https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible, I don't see the orange border styles in 
Safari:

safari

But I am seeing them in Chrome:

chrome

So, seems like just a Safari issue, not a contents-of-this-PR issue 👍

ashleemboyer:
@zchsh Thanks for digging in with me!! I'll go ahead and merge if that's cool with you? 😊

zchsh:
@ashleemboyer  Apologies for the derailment there - since it is a Safari issue, which we'll likely need to discuss outside the scope here, IMO merge away 🚀

PR #64, opened 2022-01-05
Final sweep for initial preview

ashleemboyer:
_There is no Asana task for this PR_

🔎  [Preview link](https://dev-portal-git-ambsweep-for-initial-preview-hashicorp.vercel.app/)

## What

- Updates Sidebar filtering implementation to show submenu children if the submenu's title matches the filter text. This is an issue @BRKalow identified in #58 as something that was missing from what the current .io site sidebars do when filtering.
- Updates the Vault icon used in the `ProductSwitcher` to be white. We'd like it to be yellow, but that is not available yet from flight-icons. Making the icon white makes it visible in the `ProductSwitcher` for the time being.
- Improves the `font-size` and `font-weight` of `<code>` elements within heading elements. They were a bit jarring being so small, they're less jarring with these changes.
- Updates h4-h6 font colors (recent design request).
- Adds the use of the [`:focus-visible`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) pseudo-class so that most focus styles no longer show on elements clicked by a mouse.
- Updates the main nav links to use the `focus-border-and-box-shadow-dark` helper class.
- Updates the `.alert-info` class to have smaller padding (recent design request).

## Testing

- [ ] In the sidebar filter input, type a submenu's title
- [ ] The submenu should be open and it's children should show in the results
- [ ] Open the `ProductSwitcher`
- [ ] The Vault icon should be white
- [ ] Go to a page with a `<code>` element in the title (example: [waypoint.hcl `app` Stanza](https://dev-portal-git-ambsweep-for-initial-preview-hashicorp.vercel.app/waypoint/docs/waypoint-hcl/app))
- [ ] The `font-size` and `font-weight` of the `<code>` should be larger and heavier than the `<code>` elements in the page content
- [ ] Go to the [Upgrading Waypoint page](https://dev-portal-git-ambsweep-for-initial-preview-hashicorp.vercel.app/waypoint/docs/upgrading)
- [ ] Inspect one of the `h4` elements and check that the color is correct (`token-color-palette-neutral-700`
- [ ] Inspect one of the `.alert-info` elements and ensure that the padding is correct (`16px`)
- [ ] Click the ProductSwitcher
- [ ] It **should not** show a focus outline
- [ ] Bring the Product Switcher into focus withe keyboard
- [ ] It **should** show a focus outline
- [ ] Click a sidebar submenu
- [ ] It should open and **not** show a focus outline
- [ ] Bring a sidebar submenu into focus with the keyboard
[...]

PR #66, opened 2022-01-06
Adding PR template

ashleemboyer:
## Relevant links

- A preview link is not needed for this PR, there are no changes to the project code
- There is no Asana task for this work

## What

- Adds a template for PRs 😊 

## Why

- In my experience I've found that automating PR content is a large time and effort saver by preventing everyone from having to remember all the details we'd like to see in PRs.
- The goal of the `"Ready for Review" checklist` section is to ensure we're giving updates across communication channels for folks outside of the engineering team.
- @BRKalow and @zchsh have both complimented the `Testing` section I've included in recent PRs ([this one](https://github.com/hashicorp/dev-portal/pull/64) and [this one](https://github.com/hashicorp/dev-portal/pull/65)) so it seems like a suitable time to add a template. 😊

## How

This template is based off of one I wrote in my previous role, which was based off research of templates and guides publicly posted by other engineering organizations and teams. I personally found it useful for writing PRs and I also received positive feedback from others that used it for writing PRs and/or reviewed my PRs written in this format. There weren't many people involved (small company & team) and of course all people are different, so I'm excited to hear feedback on this. 😊

## Testing

Nothing to test!

## Anything else?

Nope

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,38 @@
+## "Ready for Review" checklist
+
+- [ ] The Vercel preview link has been added to this PR's description
+- [ ] The Asana task has been added to this PR's description
+- [ ] This PR's link has been added to the "📐 Pull Request" field on the Asana task
+- [ ] This Vercel preview link has been added to the "🔍 Preview" field on the Asana task
+
+---
+
+## Relevant links
+
+- [Preview link](url) 🔎
+- [Asana task](url) 🎟️
+
+## What
+
+_Briefly list out the changes proposed in this PR._
+
+## Why
+
+_Describe why the changes proposed are needed. Some examples: new feature requested, refactor to make things easier later, styling tweaks requested by design, etc._
+
+## How
+
+_Dive into the approach you took, list resources you referenced, detail other approaches you tried but didn't end up going with, etc._
+
+## Testing
+
+_Create a checklist for going through how to test your proposed changes. If there is anything to configure before interacting with the project in a browser, such as toggling feature flags, changing machine settings, or simulating behavior in browser dev tools, list those steps first._
+
+- [ ] Step 1
+- [ ] Step 2
+- [ ] Step 3
+- [ ] ...
+
+## Anything else?
+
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,39 @@
...
+---
+
+## Relevant links
+
+- [Preview link](url) 🔎
+- [Asana task](url) 🎟️
+
+## What
+
+_Briefly list out the changes proposed in this PR._
```
BRKalow, line 18: What do you think about making the section descriptions comments? That way we don't need to delete them as we're filling out the PR. Just one less point of friction!

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,39 @@
+## "Ready for Review" checklist
+
+- [ ] The Vercel preview link has been added to this PR's description
+- [ ] The Asana task has been added to this PR's description
```
BRKalow, line 4: Do you think it would be acceptable to not include these? As they are pre-defined in the relevant links section, it seems like a potentially unnecessary step.

BRKalow (commented):
I like this!

.github/PULL_REQUEST_TEMPLATE.md
ashleemboyer: Great idea! I'll make this change

.github/PULL_REQUEST_TEMPLATE.md
ashleemboyer: I'd like to keep these in mainly because I've seen in other repos where there are placeholder links for the preview and Asana ticket that come from a PR template that don't get updated with real links, and I think having this checklist right up front will help folks remember to update those links. The placeholder are there primarily so it's clear where they belong. I forget to add the links even after having done it many times, so I also personally find it helpful for these items to be listed. 😊

BRKalow: Makes sense to me!

BRKalow (approved):
Whoop! I'm down to move forward with this. We can always iterate if needed, but this is a great start 😄

zchsh (approved):
🎉 This is awesome! Ty for putting this together 💖

PR #69, opened 2022-01-10
feat: add vault landing

zchsh:
👀 [Preview](https://dev-portal-git-zsadd-vault-landing-hashicorp.vercel.app/vault)
🎟️ [Asana task](https://app.asana.com/0/1100423001970639/1201625161290269/f)

This PR implements the `/vault` landing page.

## Details

- Uses existing work on the `/waypoint` landing page as a starting point.
- Abstracts out a new `views/product-landing`, now used for both `/waypoint` & `/vault`
    - Intent here is to better match [our intended views pattern](https://github.com/hashicorp/dev-portal#component-organization)

---
  - https://app.asana.com/0/0/1201631159784198


src/lib/clamp.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Given a number, a minimum value, and a maximum value,
+ * Return the minimum value if the number is less than the minimum,
+ * or the maximum value if the number is greater than the maximum,
+ * or the number itself if it is greater than the minimum or less than the maximum.
+ */
+export default function clamp(
```
zchsh, line 7: Added this function to ensure `heading` blocks are rendered with a valid style. See `views/product-landing/components/heading/index.tsx` for more info.

zchsh:
@ashleemboyer @BRKalow This page roughs in the `/vault` landing page, as detailed in Figma. We have a sync setup Wednesday with Matt Ryan to discuss content, and want to get something up for discussion.

Tagged you both as I wanna make sure the `views` pattern here aligns with our guidelines - I think it does, but figured couldn't hurt to double check.

src/components/sidebar/index.tsx
```diff
@@ -29,6 +29,8 @@ export interface MenuItem {
 
 interface SidebarProps {
   menuItems: MenuItem[]
+  /** productName to be shown as the title of the sidebar */
+  productName: string
```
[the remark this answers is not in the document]
BRKalow: Probably a good candidate for leveraging @ashleemboyer's WIP context refactor

src/views/product-landing/components/heading/index.tsx
```diff
@@ -0,0 +1,23 @@
+import { HeadingProps } from './types'
+import s from './style.module.css'
+
+function Heading({
+  slug,
+  heading,
+  level,
+  style_as,
```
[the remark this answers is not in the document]
BRKalow: nit: consider camelCase here for prop name consistency.

src/views/product-landing/components/heading/types.ts
```diff
@@ -0,0 +1,8 @@
+export interface HeadingProps {
+  slug: string
+  heading: string
+  /** Set the HTML heading level. */
+  level: 1 | 2 | 3 | 4 | 5 | 6
+  /** Optionally use a different heading-x style than the default for the semantic HTML heading level. Note that only 0-4 are supported, as these are the styles we have available from the design system. */
+  style_as?: 0 | 1 | 2 | 3 | 4
```
BRKalow, line 7: Very interesting that this is zero-indexed 😄

src/layouts/docs/index.tsx
```diff
@@ -11,6 +11,7 @@ interface DocsLayoutProps {
   children: React.ReactNode
   headings: SidecarHeading[]
   navData: MenuItem[]
+  productName: string
```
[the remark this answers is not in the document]
BRKalow: Might become unnecessary with @ashleemboyer's context!

src/views/product-landing/components/heading/types.ts
BRKalow: Bike shedding alert, feel free to ignore: would `size` be a good(er) name here? I think `styleAs` is just fine, but offering up `size` as I've seen it used in this way before.

BRKalow (approved):
The views pattern usage looks on point to me and this looks pretty swell from a code perspective!

src/views/product-landing/components/heading/types.ts
zchsh: yeah +1 that the zero indexing is a little odd... i dig `size`, will update! 👍

src/views/product-landing/components/heading/index.tsx
```diff
@@ -0,0 +1,24 @@
+import { HeadingProps } from './types'
+import clamp from 'lib/clamp'
+import s from './style.module.css'
+
+function Heading({
+  slug,
+  heading,
+  level,
+  size,
```
ashleemboyer, line 9: Can we talk about this `size` prop a little more? At the moment, I feel like it enables too much mixing and matching of heading elements and heading font styles. I personally think that if someone wants to mix and match styles, they should have to really work for it via sending in a `className` -- I think a manual approach can help ensure someone is doing something with well thought out intentions. I do realize the trade off with that approach is that a ton of things become adjustable away from the styles specified by design systems.

src/views/product-landing/components/heading/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+}: HeadingProps): React.ReactElement {
+  const Component = `h${level}` as React.ElementType
+  return (
+    <Component
+      id={slug}
+      data-style-as={size}
```
ashleemboyer, line 15: What's this property used for?

src/views/product-landing/components/heading/index.tsx
```diff
@@ -0,0 +1,24 @@
+import { HeadingProps } from './types'
+import clamp from 'lib/clamp'
+import s from './style.module.css'
+
+function Heading({
```
ashleemboyer, line 5: [Not to be addressed in this PR]: What do you think about moving this component to the main components folder and using it as a custom component for the docs content like our new docs-anchor component? 💭

src/views/product-landing/server.ts
```diff
@@ -1,14 +1,19 @@
...
+async function getStaticProps({
+  product,
+  contentJsonFile,
+}: {
+  product: { slug: string; name: string }
+  contentJsonFile: string
+}): Promise<$TSFixMe> {
+  // TODO: need to discuss from whence we should
+  // TODO: source content down the road. For now,
+  // TODO: sourcing from JSON for demo purposes.
```
ashleemboyer, line 14: Want to link the Asana task here for easy reference? https://app.asana.com/0/1100423001970639/1201631159784193/f

ashleemboyer (changes requested):

ashleemboyer:
This is out of scope of this PR, but I think we need to update the landing page content to use flight icons instead of svgs, such as this one: https://flight-hashicorp.vercel.app/?query=terminal-screen. This will ensure we're using the correct icon sizes defined by the design system. I've added an Asana task for us to discuss this during the weekly DevDot Eng Sync (or asynchronously if we prefer): https://app.asana.com/0/1201010428539925/1201639881308087/f.

src/components/sidebar/index.tsx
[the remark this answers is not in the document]
zchsh: Great calls on both fronts - will update this to `title`, & maybe there's a way we can leverage product context if `title` isn't provided 👍

src/data/waypoint-landing.json
```diff
@@ -1,10 +1,36 @@
 {
+  "navData": [
```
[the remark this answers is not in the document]
zchsh: I dig this idea, thanks for mentioning! 💯

src/lib/clamp.ts
zchsh: ✅  Great call - added a short test suite to cover the cases, realized we probably want to throw an error if the provided `min` value is greater than `max` 👍

src/pages/vault/index.tsx
```diff
@@ -0,0 +1,11 @@
+import ProductLandingView from 'views/product-landing'
+import server from 'views/product-landing/server'
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export async function getStaticProps() {
+  const contentJsonFile = 'src/data/vault-landing.json'
+  const product = { name: 'Vault', slug: 'vault' }
+  return await server.getStaticProps({ product, contentJsonFile })
```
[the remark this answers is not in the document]
zchsh: Great call, no not really! Maybe just not having to re-declare `revalidate` (I assume we'll want it the same for all pages?). Have updated to use the `generateStaticProps` pattern for consistency 👍

src/views/product-landing/components/heading/index.tsx
zchsh: Nothing, thanks for catching this - was cruft from earlier debugging 😅 Removed! ✅

src/views/product-landing/components/heading/index.tsx
zchsh: At a high level, on board with it! I think one challenge here that I'm always wary of is whether this holds true for design... ie will design be 100% consistent with these heading styles across landing pages and docs pages, or will they maybe want more flexibility on landing pages. However I think I need to temper that wariness a bit now that we have this stuff coming from the DS team, with the clear intent of being fully consistent 🙌 So very open to moving this into the main components folder in a follow-up PR! 💯

src/views/product-landing/components/heading/index.tsx
zchsh: Caught up a bit on the Slack chatter on this - my impression so far is we need something like this `size` prop to implement current designs without breaking HTML semantics (since in current design, the `Explore Vault Documentation` feels semantically like an `h2` but is styled with `heading-1`, which the remainder of the headings on the page also feel semantically like `h2`s and are styled with `heading-2`).

VERY on board with your suggestion of changing the style names so they don't sound exactly like HTML elements 🙌 💯 even the short description above makes my brain hurt a lil.

zchsh:
@ashleemboyer Thanks for the thorough review 💖 🙇‍♂️ Have gone through and addressed the feedback items. Let me know what you think!

src/views/product-landing/components/heading/index.tsx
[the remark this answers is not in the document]
ashleemboyer: 🐍

src/lib/clamp.ts
ashleemboyer: Thank you! And the error message is a great addition, nicely done ✨

src/pages/vault/index.tsx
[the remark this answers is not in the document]
ashleemboyer: I think a future iteration could be to also return `revalidate` from the `generateStaticProps`, and maybe even accept an optional parameter to the function for revalidate? But I'm with you, I assume it'll be the same for every page

src/views/product-landing/components/heading/index.tsx
ashleemboyer: I think we do need this for now, and if they go with your suggestion of renaming to use `100`-ish names, then this `size` prop is probably the appropriate naming. Happy for this to be left as is for now since we do need _something_ for the time being to allow mixing and matching.

src/views/product-landing/components/heading/index.tsx
ashleemboyer: TIL that cruft is a word!

ashleemboyer (approved):
Looks great! 😊🥇

PR #70, opened 2022-01-10
feat: implement footer, ConsentManager

zchsh:

---

## Relevant links

- [Preview link](https://dev-portal-git-zsdemo-footer-hashicorp.vercel.app) 🔎
- [Asana task](https://app.asana.com/0/1100423001970639/1201625161290264/f) 🎟️

## What

This PR implements the `footer` component, as reflected in current designs. To implement the `Cookie Manager` functionality of the `footer`, this PR also adds in `ConsentManager` at the app level.

## Why
These changes are needed because the current `footer` implementation does not match designs. The placeholder footer on the home page, and the lack of footer on `/docs` pages, can be distracting when reviewing and QA-ing designs.

## How

- Add `ConsentManager` to `_app.jsx`
- Add `footer` to `BaseNewLayout`, as opt-in option, used on `/` "homepage homepage"
- Add `footer` to `DocsLayout`

## Testing

For ConsentManager:

- [ ] Visit any page, such as [the home page](https://dev-portal-git-zsdemo-footer-hashicorp.vercel.app)
- [ ] You should see a ConsentManager prompt
- [ ] If you accept preferences, you should see a `hashi-consent-preferences` cookie set reflecting preferences.

For Footer:

Note that the footer has two distinct layout contexts:

1. on full-width pages, such as [the home page](https://dev-portal-git-zsdemo-footer-hashicorp.vercel.app)
2. on sidebar-layout pages, such as [`/waypoint/docs`](https://dev-portal-git-zsdemo-footer-hashicorp.vercel.app/waypoint/docs)

- [ ] Check the footer renders as expected, and matches designs on all viewports
- [ ] Check that all links are working
- [ ] Check that ConsentManager opener works as expected.

### Conflicts with docs styles

Continued conflicts with styles introduced in #48 mean there are additional `!important`s introduced in this PR. I've created [an Asana task](https://app.asana.com/0/1201010428539925/1201653123138729/f) that proposes avoiding conflicts by using styled MDX custom components to implement docs content styles, rather than global-ish HTML-primitive-targeting CSS.

### ConsentManager

Note that we may need to adjust app-level use of `ConsentManager` in the future, when co-locating with `.io` website working. To avoid the wrong abstraction too early, and because it's not an immediate concern, I have left that issue out for now. I've created [an Asana task](https://app.asana.com/0/1100423001970639/1201646681395160/f) to ensure ConsentManager use works when merging with upstream `.io` work on `main`.


src/pages/_app.tsx
```diff
@@ -1,10 +1,15 @@
 import '@hashicorp/platform-util/nprogress/style.css'
+import createConsentManager from '@hashicorp/react-consent-manager/loader'
 import { ErrorBoundary } from '@hashicorp/platform-runtime-error-monitoring'
 import useAnchorLinkAnalytics from '@hashicorp/platform-util/anchor-link-analytics'
 import { DeviceSizeProvider } from 'contexts'
-import BaseLayout from 'layouts/base'
+import BaseLayout from 'layouts/base-new'
 import './style.css'
 
+const { ConsentManager, openConsentManager } = createConsentManager({
```
zchsh, line 9: As mentioned in [this Asana task](https://app.asana.com/0/1100423001970639/1201646681395160/f), we'll likely need to rethink this when merging with `main` and co-locating `dot-ios`, but for now, keeping it simple.

src/components/footer/index.tsx
```diff
@@ -1,4 +1,7 @@
 import React from 'react'
+import svgHashicorpLogo from '!!raw-loader!./svg/hashicorp-logo.svg'
```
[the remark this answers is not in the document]
BRKalow: @zchsh Does https://github.com/hashicorp/mktg-logos have what we need?

src/components/footer/style.module.css
```diff
@@ -1,9 +1,79 @@
...
+.links {
+  position: relative;
+  width: 100%;
+  display: flex;
+  flex-wrap: wrap;
+  list-style: none;
+  /* !important needed to prevent list styles from leaking in
+     from docs layout content CSS */
+  padding: 0 !important;
+  margin: 0 !important;
```
BRKalow, line 48: Am I imagining it, or has this scenario popped up in a few recent PRs? It feels like we should try to scope the content CSS if possible so we don't need to do this.

src/components/footer/index.tsx
```diff
@@ -15,10 +18,48 @@ function Footer({
 }: FooterProps): React.ReactElement {
   return (
     <footer className={classNames(s.root, className)}>
-      Dev Portal Footer Placeholder{' '}
-      {typeof openConsentManager == 'function' && (
-        <button onClick={openConsentManager}>Open Consent Manager</button>
-      )}
+      <InlineSvg className={s.logo} src={svgHashicorpLogo} />
+      <ul className={s.links}>
+        <li className={s.linkListItem}>
```
BRKalow, line 23: For consideration: many of the other places where we render links like this, the link data is declared as JSON/data somewhere. Do we think it might make sense to do the same for the footer links?

src/components/footer/style.module.css
```diff
@@ -1,9 +1,79 @@
...
+    content: '\200B';
+    height: 1px;
+    left: 0;
+    left: 50%;
+    overflow: hidden;
+    position: absolute;
+    top: 0;
+    transform: translateX(-50%);
+    width: 100%;
+    width: 200vh;
```
BRKalow, line 24: The first declaration shouldn't be necessary I think.
```suggestion
    width: 200vh;
```

src/layouts/base-new/index.tsx
```diff
@@ -2,8 +2,16 @@ import AlertBanner from 'components/alert-banner'
 import NavigationHeader from 'components/navigation-header'
 import Footer from 'components/footer'
 import s from './base-new-layout.module.css'
+interface BaseNewLayoutProps {
+  withFooter?: boolean
```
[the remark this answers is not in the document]
BRKalow: 💭  feel free to ignore: Do you think `showFooter` would be clearer than `withFooter`? I'm not sure, but just offering it up for consideration!

BRKalow (commented):
It might be good to address the root cause of the `!important` use if possible, and a small note on `mktg-logos`, looks good though.

src/components/footer/style.module.css
zchsh: Not imaging it at all - it has come up since #48 merged. One option here would be to try to write some clever CSS to scope the content CSS - eg one thing I've done in the past in `react-content` is like `ul:not([class])`. But that has pitfalls, and feels like a stopgap...

I think a longer-term solution here would be using MDX custom components with accompanying CSS module styles, rather than targeting HTML primitives [as we do now](https://github.com/hashicorp/dev-portal/blob/assembly-ui-v1/src/layouts/docs/docs-layout.module.css).

Not sure when we can fit this in - but definitely agree we should address ASAP. Have created [a task for this](https://app.asana.com/0/1201010428539925/1201653123138729/f) since we don't have one yet 👍 Happy to switch gears on that and hold on this PR if we want to avoid introducing these new `!important`s.

src/components/footer/index.tsx
zchsh: I think it could, but held off for now due to the `Cookie manager` button wedged in the middle. Definitely open to coming up with a data structure that accounts for that if we feel this is a priority 👍

src/layouts/base-new/index.tsx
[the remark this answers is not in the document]
zchsh: Sweet, yeah I dig `showFooter` - will update!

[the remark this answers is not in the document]
zchsh: ✅

src/components/footer/index.tsx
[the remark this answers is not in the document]
zchsh: ✅ It does indeed! 🤦 Thanks for catching this - have updated and deleted the local file 👍

src/components/footer/style.module.css
BRKalow: I think for now we could move the HTML selectors into a nested class that only wraps the rendered MDX so they aren't bleeding into footer 💭 
Screen Shot 2022-01-12 at 11 50 00 AM

src/components/footer/index.tsx
```diff
@@ -1,4 +1,7 @@
 import React from 'react'
+import svgHashicorpLogo from '!!raw-loader!@hashicorp/mktg-logos/corporate/hashicorp/primary/black.svg'
```
ashleemboyer, line 2: What does the `!!raw-loader!` prefix on here do? It's not something I did for the logo in navigation-header, so I'm curious if that needs to be updated or not. Haven't seen this yet!

src/layouts/docs/index.tsx
```diff
@@ -39,6 +41,10 @@ const DocsLayout: React.FC<DocsLayoutProps> = (props) => (
                 label="Edit this page on GitHub"
               />
             )}
+            <Footer
+              className={s.footer}
+              openConsentManager={props.openConsentManager}
+            />
```
ashleemboyer, line 47: Did you already try showing the footer by passing props to the `BaseNewLayout` this layout renders above?

src/pages/index.tsx
```diff
@@ -2,21 +2,17 @@ import { ReactElement } from 'react'
...
+    <div className={s.root}>
+      <HomepageHero className={s.hero} />
+      <h2 className={s.cardGridHeading}>Explore Product Documentation</h2>
+      <ProductCardGrid className={s.cardGrid} />
+    </div>
   )
 }
 
-Homepage.layout = BaseNewLayout
+Homepage.layout = (p) => <BaseNewLayout {...p} showFooter />
```
ashleemboyer, line 17: Have you already weighed the pros/cons of the approach you've taken compared to exporting a `getStaticProps` function that sets `layoutProps` like this?

  return {
    props: {
      layoutProps: {
        showFooter
      },
    },
  }
}

ashleemboyer (changes requested):
Got a few comments here and also noticed that this branch is out of date with `assembly-ui-v1`. Could you merge that into here when you get the chance?

src/components/footer/style.module.css
ashleemboyer: I'd be happy to make that Asana task my next piece of code work! My load is a little lighter until I start the downloads view, and I can fit this task in before I start that especially since that page will be affected by this issue. I'm spending the rest of my day working on writing Asana task for the downloads view and then I should be able to start the task tomorrow. How does that sound?

src/components/footer/index.tsx
zchsh: Does the same as `?include`, but doesn't require fancy webpack stuff. I think we should be consistent with the former, this is an old habit, so have updated to use `?include` 👍

src/layouts/docs/index.tsx
zchsh: Yep - it ends up rendering the footer below the content area, rather than within it as spec'd in the designs:

CleanShot 2022-01-12 at 18 17 18@2x

Have asked design about this since I think it would be nice to take a more consistent approach to footer (#77 implements a way for us to easily switch between em, if needed). I think there's a decent possibility that we may want a more robust, site-mappy footer like the one we have on [hashicorp.com](https://www.hashicorp.com), in which case I think we'll something closer to the approach you're suggesting 👍

src/components/footer/style.module.css
zchsh: On board with both of these - for now, added a quick container in order to avoid introducing new `!important`s in `footer` 👍

src/layouts/docs/docs-layout.module.css
```diff
@@ -289,3 +287,7 @@ heading with only a couple words, the element can become too narrow.
 .editOnGithubLink {
   margin-top: 64px;
 }
+
+.footer {
+  margin-top: 360px;
```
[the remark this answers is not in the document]
zchsh: Note: probably going to switch this around again in #77, but leaving as-is for now. Seems preferable (and more feasible) to get this PR merged first.

zchsh:
@ashleemboyer @BRKalow Have rebased and addressed comments. I think it makes sense to get this merged in before #77 , and have further footer-in-layout discussions on that PR - let me know what you think 🙇

src/layouts/docs/index.tsx
ashleemboyer: Ah, I see why you took your approach now! I didn't fully notice that the footer is moved over and next to the sidebar on these pages. My take right now (and out of scope of this PR since there'd need to be a wide discussion) is that the footer should take up the full width for the same reasons main navigation headers take up the full width, but what you've got here is great and to-spec. ✨

src/pages/index.tsx
ashleemboyer: Right on, we can address this over there no problem!

ashleemboyer (changes requested):
Code looks good, just a couple of items I noticed:

1. When the nav list goes vertical, there's too much space between the items. [The designs I'm looking](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/?node-id=2835%3A61038) at show 16px of space between the items and I'm seeing 32px in the preview link.
  image
2. On hover, I'm seeing [in the designs](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/?node-id=3403%3A70017) that there should be an underline on the footer links. Not seeing that underline in the preview link.
3. The HashiCorp logo on the left side of the footer is supposed to be a link to hashicorp.com according to this design note: [link](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/?node-id=1498%3A43240).
4. (Feel free to ignore this one in a follow-up if it takes too long, seems like an edge case) At 1324px on a docs page, I'm seeing the nav links stay horizontal but wrap under the logo like this:
  image

zchsh:
@ashleemboyer Thanks for catching those, all great calls! Have tried to address:

1. Ack, this was a gotcha for me, margin doesn't collapse in `flex` layouts 🤦 Have adjusted, spacing should match designs now!
2. Added the underline on hover, also removed a `transition` for color, which I noticed is not in the design spec (not sure why i sprinkled that on in the first place 😅 )
3. Linked this, added some visually hidden text so the link has more meaningful content when announced.
4. Very much related to the concern I think we both share around sometimes showing `footer` full width (on home page), sometimes very constrained width (on `docs` pages). Longer term would like to address this with design; for now, have added a more conservative media query that avoids laying out the links horizontally until `1400 px`.

Let me know if there are any further revisions you think we should make! 🙇

src/components/footer/index.tsx
```diff
@@ -18,7 +19,10 @@ function Footer({
 }: FooterProps): React.ReactElement {
   return (
     <footer className={classNames(s.root, className)}>
-      <InlineSvg className={s.logo} src={svgHashicorpLogo} />
+      <a href="https://www.hashicorp.com/">
+        <InlineSvg className={s.logo} src={svgHashicorpLogo} />
+        <VisuallyHidden>HashiCorp home page</VisuallyHidden>
+      </a>
```
ashleemboyer, line 25: I would opt for using `aria-label` here on the anchor element instead of visually hidden text, is there a particular reason you opted for the latter? We get a lot of things `VisuallyHidden` does for free by using `aria-label`.
- Secondary comment, I'd also add a little bit to the front of the text you have here -- "_Go to_ HashiCorp home page" -- because the additional "go to" text that expresses exactly what will happen when the presented element is activated.

---

💭 Ideally, we'd be able to only add an `alt` prop to `InlineSvg` (no `aria-label`), but that doesn't behave as I immediately expected it to. It applies the `alt` prop to the containing `div` and I'm not sure exactly how we'd pass that prop down to the `svg`. There's two things that providing an `alt` prop would do for us here on an a11y front:

- give an alt text on the image (svg specifically in this case)
- give discernible text to the link

ashleemboyer (changes requested):
Just one more item + a bonus ramble for handling in the future 💭

src/components/footer/index.tsx
zchsh: Sweet, will go with `aria-label` - greatly appreciate the input here 🙇‍♂️

Re: passing `alt` to the SVG, this is always something that makes my head spin a bit (eg when thinking about setting up a repo for logos, & whether it makes sense to include `title` in logo SVGs). It seems like an accurate description of the logo SVG (`HashiCorp logo` or something) would be in tension with an accurate description of the link (I think `Go to HashiCorp home page` is great as you suggested). To clarify, if we did have the `alt` prop working the way we want, would we do `alt="Go to HashiCorp home page"`? Or `alt="HashiCorp logo"`? Or something else entirely 🤔

ashleemboyer: You're on exactly the right track! An accurate description of the logo SVG _would_ be in tension with an accurate description oof the link. This is a good example of where we would not want to "automate" `alt` or `title` in a logos repo, unless we wanted to set up the repo to expose two kinds of components:

- A plain svg/img component where the `alt` is generated to be  `"<Product name> logo"`
- A svg/img component within a link that could maybe look something like this in order to generate a useful alt text:
  ```jsx
  const SvgLogoLink = ({ href, product, pageName ...stuff }) => {
    return (
      <a href={href}>
        {/* src would be based off of the product prop */}
        <svg alt={`Go to ${product} ${pageName}`} {...stuff} />
      </a>
    )
  }
  ```

My default stance is not jumping to automate things like this too quickly, because both the `alt` text and component API deserve great care. Your head-spinning feeling is totally valid! It's a complex issue that takes some time to get used to.

ashleemboyer (approved):

src/components/footer/index.tsx
ashleemboyer: I've added docs in our Accessibility Notion page! Link here for posterity. 😊 

https://www.notion.so/hashicorp/Accessibility-4c9cc7245b09498cbedee2e1dde99727#3e0cc537c160489c97aa1c4d05b4764f

PR #72, opened 2022-01-11
feat: refactor docs server code to use RemoteContentLoader

BRKalow:

---

## Relevant links

- [Preview link](url) 🔎
- [Asana task](url) 🎟️

## What

Refactor the docs layout server code to use `RemoteContentLoader`.

## Why

Migrating off of the deprecated `generaticStatic*` functions. This aligns with what other sites are using and should make future enhancements / maintenance easier going forward.

## How

I didn't use the exported `getStaticGenerationFunctions` method from `@hashicorp/react-docs-page/server` because of our use of the `anchorLinks` remark plugin, which requires us to be able to pass a reference to the `headings` array on each request, so I imported `RemoteContentLoader` directly and wrapped that.

## Testing

- [ ] Visit any docs page
- [ ] Validate that the content is rendering as expected
- [ ] Validate that the nav data is correct
- [ ] Validate that the sidecar data is correct

### DIFF
--- a/src/layouts/docs/server.ts
+++ b/src/layouts/docs/server.ts
@@ -1,108 +1,119 @@
-import {
-  generateStaticPaths as _generateStaticPaths,
-  generateStaticProps as _generateStaticProps,
-} from '@hashicorp/react-docs-page/server'
+import { getStaticGenerationFunctions as _getStaticGenerationFunctions } from '@hashicorp/react-docs-page/server'
+import RemoteContentLoader from '@hashicorp/react-docs-page/server/loaders/remote-content'
 import { anchorLinks } from '@hashicorp/remark-plugins'
 
 import prepareNavDataForClient from 'layouts/docs/utils/prepare-nav-data-for-client'
 import getDocsBreadcrumbs from 'components/breadcrumb-bar/utils/get-docs-breadcrumbs'
 
-// because some of the util functions still require param arity, but we ignore
-// their values when process.env.ENABLE_VERSIONED_DOCS is set to true, we'll
-// just use this string to make it clear by using this k/v
-const temporary_noop = 'im just for show'
+const BASE_REVALIDATE = 10
 
-async function generateStaticPaths({
+/**
+ * Returns static generation functions which can be exported from a page to fetch docs data
+ *
+ * Example usage:
+ *
+ * ```ts
+ * const { getStaticPaths, getStaticProps } = getStaticGenerationFunctions({
+ *   product,
+ *   basePath,
+ * })
+ *
+ * export { getStaticPaths, getStaticProps }
+ * ```
+ */
+export function getStaticGenerationFunctions({
   product,
   basePath,
 }: {
-  basePath: string
-  product: { name: string; slug: string }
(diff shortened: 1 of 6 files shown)


.env
```diff
@@ -1,10 +1,3 @@
-ENABLE_VERSIONED_DOCS=true
-# TODO: docs-page/server code requires VERCEL_ENV="production" to be set
-# TODO: in order to load remote content. We may want to update the server
-# TODO: code to rely only on the ENABLE_VERSIONED_DOCS env var instead,
-# TODO: as setting this manually for local dev feels inaccurate.
-VERCEL_ENV="production"
```
BRKalow, line 6: These shouldn't be necessary anymore with the usage of `RemoteContentLoader`

src/layouts/docs/server.ts
```diff
@@ -1,108 +1,119 @@
...
-      navData: navDataWithFullPaths,
-      breadcrumbLinks,
-      githubFileUrl,
-      backToLink: {
-        text: `Back to ${product.name}`,
-        url: `/${product.slug}`,
-      },
+      const breadcrumbLinks = getDocsBreadcrumbs({
+        productPath: product.slug,
+        productName: product.name,
```
BRKalow, line 92: It'd be nice to read this from the upcoming product context so we don't have to thread the product name/slug through here 💭

ashleemboyer: I think that's tricky since these things are all happening server side and the Context can only be used client side, right?

ashleemboyer (approved):
I think this looks good, should we wait to merge until Zach's merged #69?

src/layouts/docs/server.ts
BRKalow: Correct, the way its currently implemented all the breadcrumb data is generated here. It would require a larger refactor in how breadcrumbs are rendered most likely

BRKalow:
Yeah we can defer this! Don't wanna create extra work for already prioritized tasks.

src/layouts/docs/server.ts
ashleemboyer: Ah, okay! Wasn't sure if that's the kind of refactor you were thinking about. Right now my hot take is that if the breadcrumbs can all be generated ahead of time, then we doing so on the server side might be a valuable optimization? But I think it'd be good to weigh the pros and cons of each approach and I don't hold my current opinion very strongly. 😊

BRKalow: Yeah, that's reasonable! I think there's probably a middle ground here somewhere. We have to iterate over and render the breadcrumb data even with it generated on the server, so we could conceivably access the context at the point and change the URL, maybe swapping out a placeholder or something.

```ts
{ title: '{productName}', url: '/{productPath}' }

// ...

const product = useCurrentProduct()

<a href={url.replace('{productPath}', product.slug)}>{title.replace('{productName}', product.name)}</a>
```

I'm a big fan of trying to avoid passing down arguments/props multiple levels if possible, but there are certainly trade-offs to both approaches.

zchsh (approved):
🎉 💯 Looks great to me

PR #67, opened 2022-01-07
Implement and use CurrentProductContext & useCurrentProduct hook

ashleemboyer:
## "Ready for Review" checklist

- [x] The Vercel preview link has been added to this PR's description
- [x] The Asana task has been added to this PR's description
- [x] This PR's link has been added to the "📐 Pull Request" field on the Asana task
- [x] This Vercel preview link has been added to the "🔍 Preview" field on the Asana task
- [x] The description template has been filled in below

---

## Relevant links

- [Preview link](https://dev-portal-git-ambcurrent-product-context-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1201010428539925/1201602267333006/f) 🎟️

### Main additions

- Added `CurrentProductProvider` and renders/hydrates it in `src/pages/_app.tsx`
- Added `useCurrentProduct` hook for consuming the value of `CurrentProductContext`
- Added a test suite for `CurrentContextProvider` (`src/contexts/__tests__/current-product.test.tsx`)

### Supplemental additions

- Added JSON files in `src/data` for Boundary, Consul, HashiCorp Cloud Platform, Nomad, Packer, Sentinel, Terraform, Vagrant, and Vault
- Added placeholder home pages for Boundary, Consul, HashiCorp Cloud Platform, Nomad, Packer, Sentinel, Terraform, and Vagrant
- Added `NavigationHeaderItem` interface for use in the `Product` interface, among other useful properties

### Refactors

- Updated `config/products.ts` to "generate" product urls from their slugs (just prepends `/` to the slug) now that we have internal placeholder pages to point to
- Updated `src/components/docs-anchor/index.tsx` to just be a component (no makeDocsAnchor) and leverage `useCurrentProduct`, and updated the tests
- Updated `DocsPage` to leverage `useCurrentProduct` and memoize `additionalComponents`
- Updated `NavigationHeader` to leverage `useCurrentProduct`
- Updated `ProductSwitcher` to leverage `useCurrentProduct` (and fixed the Vault icon when it is the current product being viewed)
- Moved Vault's nav data from `src/data/vault-landing.json` to `src/data/vault.json`
- Moved Waypoint's nav data from `src/data/waypoint-landing.json` to `src/data/waypoint.json`
- Updated `src/layouts/docs/server.ts` and `src/views/product-landing/server.ts` to build navData from `data` files and additionally return `product` in `finalProps`
- Updated `src/pages/vault/index.tsx` to read Vault's data from `data/vault.json`
[...]

PR #73, opened 2022-01-11
feat: add vault docs views

zchsh:

---

## Relevant links

- [Preview link](https://dev-portal-git-zsadd-vault-docs-views-hashicorp.vercel.app/vault/docs) 🔎
- [Asana task](https://app.asana.com/0/1100423001970639/1201631159784199/f) 🎟️

## What

This PR adds the Vault documentation views:
- [/vault/docs](https://dev-portal-git-zsadd-vault-docs-views-hashicorp.vercel.app/vault/docs)
- [/vault/api-docs](https://dev-portal-git-zsadd-vault-docs-views-hashicorp.vercel.app/vault/api-docs)

## Why

We want to get docs content in place early, to enable stakeholders to get a feel for the new user experience on the unified dev-portal site as they review and contribute feedback.

## How

- Copies and pastes existing `docs/[[...page]].tsx` views.
- Makes modifications to our custom MDX components setup, so that we have a core set of components shared across all pages, and so that each page can pass in `additionalComponents`
- Localizes `@hashicorp/platform-docs-mdx` into `dev-portal`, for ease in iteration, and since existing consuming projects (docs & learn) will be added in `dev-portal` in the future.
- Adds a `adjustLinkUrls` remark plugin, which allows us to fix `/api` links in Vault docs, which should instead link to `/api-docs`.

## Testing

Basic rendering:

- [ ] Visit [/vault/docs](https://dev-portal-git-zsadd-vault-docs-views-hashicorp.vercel.app/vault/docs), should render as expected
- [ ] Visit [/vault/api-docs](https://dev-portal-git-zsadd-vault-docs-views-hashicorp.vercel.app/vault/api-docs), should render as expected
- [ ] Visit [/vault](https://dev-portal-git-zsadd-vault-docs-views-hashicorp.vercel.app/vault), cards that link to internal documentation (ie on `/vault/docs` or `/vault/api-docs`) should work.

Custom Vault MDX components should work:

- [ ] `Column` and `Tag` MDX components should render as expected
    - These "additional components" are only used on [/vault/docs/plugin-portal](https://dev-portal-git-zsadd-vault-docs-views-hashicorp.vercel.app/vault/docs/plugin-portal)
    - Compare to [live](https://www.vaultproject.io/docs/plugin-portal). `Columns` means multi-column lists of links are rendering as expected. `Tag` is rendered eg as `External` beside `Oracle` in the [database section](https://www.vaultproject.io/docs/plugin-portal#database).
- [ ] Internal Vault docs links should work
    - Example on [/vault/docs/secrets/ad](https://dev-portal-git-zsadd-vault-docs-views-hashicorp.vercel.app/vault/docs/secrets/ad).
    - Note that links have been corrected from `/api` to `/api-docs` using a `remark` plugin. This is intended to imitate [existing rewrite config on the Vault.io site](https://github.com/hashicorp/vault/blob/e014065c642fc56259d0aa912d1aa589152f06ec/website/next.config.js#L20) (which is only partially working in a sense, since [we don't have corresponding redirects](https://app.asana.com/0/1100423001970639/1201655704127120/f).

Custom Waypoint MDX components should not be broken:

- [ ] Internal Waypoint docs links should work
    - Example on [/waypoint/docs/projects](https://dev-portal-git-zsadd-vault-docs-views-hashicorp.vercel.app/waypoint/docs/projects)
- [ ] `Placement` MDX component, eg on [/waypoint/docs/waypoint-hcl/config](https://dev-portal-git-zsadd-vault-docs-views-hashicorp.vercel.app/waypoint/docs/waypoint-hcl/config), should continue to work

### Major issues

🎟️ [Get remotely sourced redirects working in assembly-ui-v1](https://app.asana.com/0/1201010428539925/1201662082096106/f)

Redirects still need to be address - Vault content contains `/intro/*` links which should be redirected to Learn. For now, these have (as a first step) been configured to be rewritten to `/vault/intro/* links (via `basePaths` config in `data` and `DocsAnchor`). We should handle redirects in a more robust way in follow-up work, ideally correcting broken and stale links in the source content rather than continually adding redirects.

### Minor Issues

The `Column` and `Tag` components were added ~ 2 years ago for a one-off use case (https://github.com/hashicorp/vault/pull/9590). In theory they both seem useful and re-usable, but in practice they are used on exactly one page - [/vault/docs/plugin-portal](https://dev-portal-git-zsadd-vault-docs-views-hashicorp.vercel.app/vault/docs/plugin-portal). Seems possible to tackle this issue as a later enhancement - have a pretty [generic Asana task](https://app.asana.com/0/1100423001970639/1201648132084613) that I think covers this.


src/components/author-primitives/vault/columns/docs.mdx
```diff
@@ -0,0 +1,20 @@
+---
+componentName: 'Columns'
+---
+
+> Note: this component was added ~ 2 years ago for a one-off use case.
+
+An _experimental_ component currently used only on `/api-docs/plugins-directory` -- currently only works correctly when applied to `ul` elements.
```
zchsh, line 7: Intended for use in `swingset`, but currently not rendering since our components folder isn't flat.

src/components/navigation-header/index.tsx
```diff
@@ -4,11 +4,18 @@ import HashiCorpLogo from '@hashicorp/mktg-logos/corporate/hashicorp/logomark/wh
...
-// TODO: we'll need a programatic way to get this data when there are more products
-// TODO: currently we only show Waypoint navigation data
+// TODO: we could likely improve the dictionary approach here.
+// TODO: eg, subnavItems could be a NavigationHeader prop,
+// TODO: and we could pass it in at the layout level.
 import waypointData from 'data/waypoint.json'
+import vaultData from 'data/vault.json'
 import s from './style.module.css'
 
+const subnavItemsDict = {
```
zchsh, line 14: This doesn't feel great, but feels like a reasonable stopgap as we add content in. Very open to other suggestions here 🙇‍♂️

src/data/vault-landing.json
```diff
@@ -2,11 +2,11 @@
   "navData": [
     {
       "title": "Introduction",
-      "href": "https://www.vaultproject.io/docs/what-is-vault"
+      "fullPath": "/vault/docs/what-is-vault"
     },
     {
       "title": "Getting Started",
-      "href": "https://www.vaultproject.io/docs/install"
+      "fullPath": "/vault/docs/install"
```
zchsh, line 9: Updated links to be internal since content is now rendering internally 🎉

src/data/vault.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "Vault",
```
zchsh, line 2: Added this since we seem to have settled on `data` as our one home for product data. Def happy to wait for upstream changes when #67 merges.

src/layouts/docs/utils/mdx-components.ts
```diff
@@ -0,0 +1,48 @@
+import Columns from 'components/author-primitives/vault/columns'
+import InlineTag from 'components/author-primitives/vault/inline-tag'
+import Placement from 'components/author-primitives/shared/placement-table'
+import NestedNode from 'components/author-primitives/waypoint/nested-node'
+import makeDocsAnchor from 'components/docs-anchor'
+import defaultMdxComponents from '@hashicorp/platform-docs-mdx'
+
+// Asana task to discuss MDX custom components in dev-portal:
+// https://app.asana.com/0/1100423001970639/1201648132084613/f
```
zchsh, line 9: I think the issue of how to handle custom MDX components is a pretty big one, and one we should try to tackle sooner rather than later. I have some suggestions in comments below on the possibility of removing the need to call `defaultMdxComponents` with different `product` options. I think this would be a great start, and would be happy to tackle it as part of this PR to try to mitigate this problem early.

Let me know what you think of the proposed comments below @ashleemboyer @BRKalow 🙇

src/pages/vault/api-docs/[[...page]].tsx
```diff
@@ -0,0 +1,33 @@
+import { generateStaticPaths, generateStaticProps } from 'layouts/docs/server'
```
Comment (zchsh) [inline src/pages/vault/api-docs/[[...page]].tsx:1]: These page files are not too big, but they are super repetitive - the only diff is the `basePath` and the `mdx-components` import. I think we can probably further abstract, but also think it makes sense to tackle the hairier problem of MDX custom components first.

(Hoping to get to both in this PR, but IMO may make sense to drop this further abstraction task if resolving the MDX custom component challenges drags out).

src/pages/vault/downloads/index.tsx
```diff
@@ -0,0 +1,16 @@
+import BaseNewLayout from 'layouts/base-new'
+import { ReactElement } from 'react'
+
+const VaultDownloadsPage = (): ReactElement => {
```
zchsh, line 4: Stubbed this in simply so there's something to link to in the top nav (as has been done for Waypoint).

src/pages/waypoint/commands/[[...page]].tsx
```diff
@@ -1,7 +1,8 @@
 import { generateStaticPaths, generateStaticProps } from 'layouts/docs/server'
 import waypointConfig from '../../../../config/waypoint.json'
 import DocsLayout from 'layouts/docs'
-import DocsPage from 'components/docs-page'
+import { MDXRemote } from 'next-mdx-remote'
```
Comment (zchsh) [inline src/pages/waypoint/commands/[[...page]].tsx:4]: As far as I could tell, `DocsPage` only role was setting up Waypoint-specific MDX components, so for now have replaced with MDXRemote directly (and removed `components/docs-page`). Definitely expecting to refactor this during work to reduce complexity in MDX custom component config.

zchsh:
@BRKalow @ashleemboyer Would love early review here. Open to any aspects, though there's one specific issue I'm pretty certain we'll want to address, which is the complexity in configuring MDX custom components per-product.

As mentioned in some `TODO` comments, I think it might be worth it to make a couple changes:
1. make upstream changes to `platform-mdx` in https://github.com/hashicorp/web-platform-packages in order to remove the need to pass `product`.
2. write a (likely pretty basic) remark plugin to prefix MDX links (aka build the `adjustedHref`) at build time (eg `/some-route` becomes `/docs/some-route`). Perceived benefit being we remove the need for `product` context when building `defaultMdxComponents`. Coupled with the above change, means we can build one set of `defaultMdxComponents` to share across all products. Another benefit is we move a bit of work we're currently doing client-side to build-time. We'd still retain `DocsAnchor` for the internal-`Link` purposes, of course 👍

Let me know what you think of those two suggestions 🙇 

> Note: I think there is a broader question here of how we'll make the transition from current linking behaviour, ie links do NOT reflect the actual routes on dev-portal, to potential future behaviour, where authors might write docs that properly reflect actual dev-portal routes. By moving the link-adjusting work to build-time, I think there's potential to make this transition... at some version cutoff per-product, we fix all the links in the upstream MDX, and then only apply the link adjustment to MDX from versions older than the cutoff. Bit of a secondary concern, but felt relevant enough to mention in a footnote.

src/layouts/docs/utils/mdx-components.ts
```diff
@@ -0,0 +1,48 @@
...
+    // TODO: apply makeDocsAnchor as a remark plugin instead?
+    // TODO: combined with above change, this could allow us
+    // TODO: to use a shared set of base components for all products,
+    // TODO: and only need to pass additionalComponents, on a per-route
+    // TODO: basis. (eg Columns & Tags are only needed on /vault/docs,
+    // TODO: not /vault/api-docs, and in fact are only needed on
+    // TODO: a single page - /vault/docs/plugin-portal.)
+    // TODO: Longer-term, we may be able to deprecate some MDX components...
+    // TODO: although this will admittedly be pretty difficult, since
+    // TODO: we need to support all past versions of documentation!
```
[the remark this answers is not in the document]
BRKalow: Acknowledging that we can do something similar with `EnterpriseAlert`, it seems like we are just about to the point where we can pass in `additionalComponents` on a per-product basis without additional work.

I think the remark plugin would also require us to pass in `product` to it, which seems to kind of not solve the issue and just move the dependency on product elsewhere?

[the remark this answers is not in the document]
BRKalow: It feels like this ☝️ would be similar to how we're handling it in `DocsPage` currently. Do you see issues with that approach?

src/pages/vault/api-docs/[[...page]].tsx
Comment (BRKalow) [reply src/pages/vault/api-docs/[[...page]].tsx:1]: I think the boilerplate gets reduced a bit further with this: https://github.com/hashicorp/dev-portal/pull/72

src/components/navigation-header/index.tsx
BRKalow: Is this something that's solved with @ashleemboyer's current product context? 🤔

zchsh: Possibly! I think we'd still need the dictionary of data, but would be leveraging product context as a more standard way of getting `currentProductSlug` (rather than `useRouter` as is now).

src/pages/vault/api-docs/[[...page]].tsx
Comment (zchsh) [reply src/pages/vault/api-docs/[[...page]].tsx:1]: Oh, nice! 💯 I'll rebase off that work 👍

src/layouts/docs/utils/mdx-components.ts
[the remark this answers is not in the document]
BRKalow: I could see an authoring benefit to _not_ requiring the product prefix in the MDX source. Without it, internal links still intuitively match with the source content file structure, and links across "boundaries" would still be possible and clear when looking at the source. We'd need some sort of validation to make sure that authors are adding the product prefix, so we're shifting the complexity we're handling here now to the authors, which I'm not sure is the best tradeoff. So I think my gut take is that not having the product prefix isn't the root of the issue 😅 

Expanding on brittle-ness a bit...I guess it depends on what the remark plugin is doing. If we're just visiting link nodes and changing the url itself, I agree that's not brittle. If we're using a remark plugin to swap links out with a component/different markup that makes me a bit nervous. Having looked into the future a bit with MDX v2, the AST structure has changed there so that is something we'd need to change when we get to that migration. I think I was being unfair there though, in general remark plugins are pretty effective!

If we don't move towards requiring the product prefix in source, the remark plugin would always need the product slug and the product's basePaths. If we apply the remark plugin during the content upload, we'll need to coordinate with the .io cutover as the product-prefixed URLs would NOT work with our current setup.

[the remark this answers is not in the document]
BRKalow: It looks like `basePaths` is included in the context, so it should all be okay. https://github.com/hashicorp/dev-portal/pull/67/files#diff-0bf9bafd60b21f0b8784b6574b02bdd59a37ad5882e2c262e48c776b118a3358R4

No objections to the remark plugin if you wanna spike it out! Just thinking out loud that we'll likely need to pass a custom link component either way to wrap them with `next/link` where appropriate.

package.json
```diff
@@ -60,13 +60,15 @@
...
+    "mdast": "^3.0.0",
     "next": "^11.1.2",
     "next-mdx-remote": "^3.0.5",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
     "react-intersection-observer": "^8.32.1",
     "slugify": "^1.6.4",
-    "swingset": "^0.10.2"
+    "swingset": "^0.10.2",
+    "unist-util-visit": "^4.1.0"
```
zchsh, line 71: I've added `unist-util-visit` to create a `remark` plugin to fix the `/api` links in Vault content, and have added `mdast` for clearer typing of that plugin.

src/data/vault.json
```diff
@@ -31,5 +32,27 @@
         "href": "https://support.hashicorp.com"
       }
     ]
-  }
+  },
+  "navigationHeaderItems": [
```
zchsh, line 36: Adds a nav for Vault 👍

src/layouts/docs/server.ts
```diff
@@ -24,9 +25,13 @@ const BASE_REVALIDATE = 10
 export function getStaticGenerationFunctions({
   product,
   basePath,
+  baseName,
+  additionalRemarkPlugins = [],
 }: {
   product: Product
   basePath: string
+  baseName: string
+  additionalRemarkPlugins?: RemarkPluginEntry[]
```
zchsh, line 34: Have added an option to pass additional remark plugins, which lets us pass the `vaultUrlAdjuster` for `/vault` docs views in order to correct the `/api` link issue.

src/layouts/docs/utils/adjust-link-urls-plugin.ts
```diff
@@ -0,0 +1,14 @@
+import { visit } from 'unist-util-visit'
+import type { Root } from 'mdast'
+
+export default function adjustLinkUrlPlugin({
```
zchsh, line 4: This plugin can be used to adjust the `url` values of links in markdown content. Currently only used for Vault's `/api` link issue.

src/layouts/docs/utils/product-url-adjusters.ts
```diff
@@ -0,0 +1,26 @@
+import { RemarkPlugin, RemarkPluginOptions } from 'types/remark'
+import adjustLinkUrlPlugin from './adjust-link-urls-plugin'
+
+const vaultUrlAdjuster: [RemarkPlugin, RemarkPluginOptions] = [
+  adjustLinkUrlPlugin,
```
zchsh, line 5: Leverages the `adjustLinkUrlPlugin` to fix the `/api` link issue. We could, if we want, modify this to also handle the `href` adjustment currently happening in `DocsAnchor` - though the latter is already working fine, so might be preferable to leave that question out of this PR.

src/pages/waypoint/commands/[[...page]].tsx
```diff
@@ -3,19 +3,27 @@ import waypointData from 'data/waypoint.json'
 import { Product } from 'types/products'
 import { getStaticGenerationFunctions } from 'layouts/docs/server'
 import DocsLayout from 'layouts/docs'
-import DocsPage from 'components/docs-page'
+import DocsView from 'views/docs-view'
+import Placement from 'components/author-primitives/shared/placement-table'
+import NestedNode from 'components/author-primitives/waypoint/nested-node'
 
 const basePath = 'commands'
+const baseName = 'Commands'
```
Comment (zchsh) [inline src/pages/waypoint/commands/[[...page]].tsx:11]: This is somewhat unrelated to this PR's contents, but the lack of configuration option for `baseName` means the breadcrumb bar gets confusing - for example [the `/commands` page shows "Docs" in the breadcrumb bar](https://developer.hashi-mktg.com/waypoint/commands).

src/layouts/docs/utils/_local_platform-docs-mdx/README.md
```diff
@@ -0,0 +1,6 @@
+# `@hashicorp/platform-docs-mdx`
+
+🚨 _PLEASE NOTE_: the files in this folder have been localized from our [`@hashicorp/platform-docs-mdx` package](https://github.com/hashicorp/web-platform-packages/tree/main/packages/docs-mdx). The rationale for localizing these files, rather than making upstream changes in the platform-mdx package, is:
```
zchsh, line 3: Have added this README to record the rationale for localizing this work before modifying it. Open to thoughts here 🙇

src/types/remark.ts
```diff
@@ -0,0 +1,23 @@
+import type { Root } from 'mdast'
```
zchsh, line 1: Added this type file to avoid doing `$TSFixMe` for the `additionalRemarkPlugins` option in `docs/server`.

src/pages/vault/docs/[[...page]].tsx
```diff
@@ -0,0 +1,34 @@
...
+import DocsLayout from 'layouts/docs'
+import DocsView from 'views/docs-view'
+import Columns from 'components/author-primitives/vault/columns'
+import InlineTag from 'components/author-primitives/vault/inline-tag'
+import { vaultUrlAdjuster } from 'layouts/docs/utils/product-url-adjusters'
+
+const basePath = 'docs'
+const baseName = 'Docs'
+const product = vaultData as Product
+const additionalComponents = {
```
Comment (zchsh) [inline src/pages/vault/docs/[[...page]].tsx:14]: Our `docs/server` utils now include a base of shared MDX components only. This allows us to add custom components per-route, to match existing configuration (and to avoid conflicts & page bloat where custom components are not used).

For example, `/vault/docs` requires additional custom components, but `/vault/api-docs` does not.

src/types/remark.ts
BRKalow: `unified` exports a `Pluggable` type, which we might be able to use. ref: https://github.com/hashicorp/next-mdx-remote/blob/main/src/types.ts#L12

src/layouts/docs/utils/mdx-components.ts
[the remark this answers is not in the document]
zchsh: Update here: have left the existing `DocsAnchor` `adjustedHref` logic in place, since it's already working 🎉 Did end up adding a `remark` plugin to adjust link `url` values, but for the related-but-separate issue of Vault's `/api` links needing to be corrected to `/api-docs`. Let me know if that seems like a good approach (for now at least!) 🙇

zchsh:
@BRKalow @ashleemboyer This feels ready for review - major changes from earlier WIP noted in the PR description. Let me know what you think! 🙇

src/layouts/docs/utils/_local_platform-docs-mdx/README.md
BRKalow: For consideration: acknowledging that this will most likely live long-term in this repo, we could refactor it a bit to drop the dead code and call it something else?  I think it'd also be fine to defer that to not increase the scope of the PR.

src/layouts/docs/utils/adjust-link-urls-plugin.ts
BRKalow: nit: this might be easier to discover if it were closer to the `src/` root...maybe `src/remark-plugins` or `src/lib/remark-plugins`?

src/layouts/docs/utils/product-url-adjusters.ts
```diff
@@ -0,0 +1,26 @@
...
+      // (Note that further modification happens client-side,
+      // via our DocsAnchor custom MDX component)
+      return url
+    },
+  },
+]
+
+export { vaultUrlAdjuster }
+// eslint-disable-next-line import/no-anonymous-default-export
+export default { vaultUrlAdjuster }
```
BRKalow, line 26: nit: I don't think we need the default export here

BRKalow:
Looking great overall! Clicked through the preview with your testing plan and all looks good.

src/layouts/docs/utils/_local_platform-docs-mdx/README.md
zchsh: Great call, will at least drop the dead code now 👍 Going to defer the name change and keep this readme in place for now (so the origin is clear as we migrate stuff).

src/types/remark.ts
zchsh: ✅  Great call - updated to use `Plugin` & `Pluggable` from `unified`, deleted this file 👍

src/layouts/docs/utils/_local_platform-docs-mdx/README.md
zchsh: ✅ Dropped the dead code

src/pages/waypoint/plugins/[[...page]].tsx
```diff
@@ -3,19 +3,27 @@ import waypointData from 'data/waypoint.json'
...
 const product = waypointData as Product
+const additionalComponents = {
+  Placement,
+  NestedNode,
+}
 
 // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
-const WaypointPluginsPage = ({ mdxSource }): ReactElement => {
-  return <DocsPage {...mdxSource} />
+function WaypointPluginsPage({ mdxSource }): ReactElement {
```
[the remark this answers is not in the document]
Comment (zchsh) [reply src/pages/waypoint/plugins/[[...page]].tsx]: ✅ Switched to `const` & arrow function 👍

src/layouts/docs/utils/product-url-adjusters.ts
zchsh: ✅ Removed

src/layouts/docs/utils/product-url-adjusters.ts
```diff
@@ -0,0 +1,26 @@
...
+      if (isBadApiUrl) url = url.replace(/^\/api/, '/api-docs')
+      // Return the modified URL
+      // (Note that further modification happens client-side,
+      // via our DocsAnchor custom MDX component)
+      return url
+    },
+  },
+]
+
+export { vaultUrlAdjuster }
```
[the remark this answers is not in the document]
zchsh: ✅ Applied & removed the default export 👍

src/views/docs-view/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+  scope: Record<string, unknown>
+  lazy?: boolean
+}): ReactElement => {
+  const components = defaultMdxComponents({ additionalComponents })
+  return (
+    <MDXRemote {...{ compiledSource, scope, lazy }} components={components} />
+  )
+}
+
+export default DocsPage
```
BRKalow, line 22: nit:
```suggestion
const DocsView = ({
  additionalComponents = {},
  compiledSource,
  scope,
  lazy,
}: {
  additionalComponents: Record<string, ReactElement>
  compiledSource: string
  scope: Record<string, unknown>
  lazy?: boolean
}): ReactElement => {
  const components = defaultMdxComponents({ additionalComponents })
  return (
    <MDXRemote {...{ compiledSource, scope, lazy }} components={components} />
  )
}

export default DocsView
```

BRKalow (approved):
From a code perspective and a quick walkthrough of the preview, this looks good to me. Thanks for addressing my earlier comments. I'll leave my approval, but kinda leaning on @ashleemboyer to do a more thorough review of the changes as it might relate to any finer UI concerns.

Good job!

ashleemboyer (changes requested):
In addition to the comments below, found something else. I believe this is out of scope for the PR but seems necessary to discuss to start coming up with a plan for!
- On [/vault/docs](https://dev-portal-git-zsadd-vault-docs-views-hashicorp.vercel.app/vault/docs), there's a `Getting Started` guide link. It points to `/vault/intro/getting-started/install` which isn't a page here. When I click this link in the current Vault site, I think it's _redirecting_ to https://learn.hashicorp.com/tutorials/vault/getting-started-install. Is that the case? How do we know all the links that need this to happen? Do we know?

src/views/docs-view/index.tsx
```diff
@@ -0,0 +1,22 @@
+import { ReactElement } from 'react'
```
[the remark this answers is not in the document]
zchsh: @ashleemboyer This was based on my rough interpretation of [the component organization section in the README](https://github.com/hashicorp/dev-portal#component-organization) - felt like this fit better as `Componentry which represents a full site "view."` rather than `Shareable, smaller components`, and felt like the term `page` is usually applied to a specific route while `view` is applied to components that get rendered on multiple pages.

Wasn't aware of the plan to move server code into the `components` folder - happy to move this back into `components/docs-page` if my interpretation is off 👍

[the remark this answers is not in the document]
ashleemboyer: Thank you for elaborating! 💖

It should be possible to move that server code to the new home in `views`, right? If so I think that sounds okay! As long as we're able to continue moving code over from the react-components `DocsPage`, then I'm good with these changes. 😎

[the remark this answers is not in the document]
zchsh: Great Q, I think there should be no problem moving server code into `views` 👍 I don't think NextJS has any special treatment for `src/components` or `src/views` (Or really anything much other than `src/pages`?) so I believe we have a great deal of flexibility here 💪 🎉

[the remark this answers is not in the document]
ashleemboyer: Sweet! Let's do it ⚡

zchsh:
@ashleemboyer Great callout re: the `/intro` links - that should have merited a section in `Anything else?`, adding now! TLDR; is there are [redirects in place](https://github.com/hashicorp/vault/blob/main/website/redirects.next.js) for intro links, and also [intro content that's still rendering even though we redirect away from it](https://app.asana.com/0/563192436488770/1200137410929768/f), and generally it's kinda it's own whole mess.

I figured it would be very hard to redirect accurately unless we prefixed with `/vault`, so added `intro` to `basePaths` as well as part of this PR, but definitely need to follow-up on redirects generally (left commented out for now in `next.config`).

src/pages/vault/downloads/index.tsx
```diff
@@ -0,0 +1,28 @@
+import { ReactElement } from 'react'
+import { GetStaticProps } from 'next'
+import vaultData from 'data/waypoint.json'
```
ashleemboyer, line 3: OMG I led you wrong. So sorry! Copy-paste error 🤣

```suggestion
import vaultData from 'data/vault.json'
```

zchsh:
@ashleemboyer Thanks for catching that last import, helped me catch another lil fix I need to make as well 💯

ashleemboyer:
@zchsh Looks great!! 🎉 🚀

PR #77, opened 2022-01-12
feat: refactor sidebar layout

zchsh:

---

## Relevant links

- [Preview link](https://dev-portal-git-zsrefactor-sidebar-layout-hashicorp.vercel.app/waypoint) 🔎
- [Asana task](https://app.asana.com/0/1100423001970639/1201625161290271/f) 🎟️

## What

This PR refactors the `docs-layout` so that page content is located in the main document's scrolling area (rather than in a separate scroll container as in the current implementation).

## Why

NextJS's client-side routing does not play nicely with scrollable content not in the main document's scrolling area. For example, we do not see the content jump back to the top of the page on navigation as we'd expect.

## How

A new partial, `sidebar-sidecar-layout`, has been created within `docs-layout`. This partial handles layout positioning concerns only.

A demo of this partial can be seen at [/_layout-dev](https://dev-portal-git-zsrefactor-sidebar-layout-hashicorp.vercel.app/_layout-dev). (This demo will be removed before merging this PR; it felt useful for dev and review purposes).

## Testing

- [ ] Test out the simplified demo example of `sidebar-sidecar-layout` at [/_layout-dev](https://dev-portal-git-zsrefactor-sidebar-layout-hashicorp.vercel.app/_layout-dev), toggling the various content lengths in each of the `sidebar`, `main`, and `sidecar` areas. Ensure the behaviour is what you expect.
- [ ] Visit pages that use the layout, such as [`/waypoint`](https://dev-portal-git-zsrefactor-sidebar-layout-hashicorp.vercel.app/waypoint) and [`/waypoint/docs`](https://dev-portal-git-zsrefactor-sidebar-layout-hashicorp.vercel.app/waypoint/docs) pages. Ensure the layout is behaving as expected with real docs content.

## Anything else?

To reduce brittleness in `SidebarSidecarLayout`, we could observe the height of the sticky `header` in `BaseNewLayout`.  This would eliminate the need for global CSS variables that try to statically capture the height of potentially dynamic content. Have noted this caveat in code comments, and deferring this work until later iterations.


src/components/dev-alert-banner/index.tsx
```diff
@@ -0,0 +1,22 @@
+import AlertBanner from 'components/alert-banner'
+
+function DevAlertBanner(): React.ReactElement {
```
zchsh, line 3: Added `DevAlertBanner` to decouple the chained `docs-layout` >> `base-new-layout` while not repeating this code in both layouts.

src/components/sidebar/style.module.css
```diff
@@ -1,15 +1,5 @@
...
-    0 12px 28px rgba(101, 106, 118, 0.25);
-  height: 100%;
-  left: 0;
-  overflow-x: hidden;
-  overflow-y: auto;
-  padding: var(--sidebar-padding);
-  position: absolute;
-  top: 0;
-  width: var(--sidebar-width);
+  padding: 0 24px 24px 24px;
```
zchsh, line 2: `box-shadow` and other sidebar layout styles are now set in `sidebar-sidecar-layout`.

src/components/sidecar/style.module.css
```diff
@@ -1,11 +1,5 @@
 .sidecar {
-  margin-left: 24px;
-  max-width: 220px;
-  min-width: 220px;
-  position: sticky;
   text-align: right;
-  top: 0;
-  width: 220px;
```
zchsh, line 8: Positioning and width styles no longer needed, `sidebar-sidecar-layout` handles these, and we just need to drop in content.

src/layouts/base-new/index.tsx
```diff
@@ -1,27 +1,13 @@
...
-            rel="noopener noreferrer"
-            target="_blank"
-          >
-            We&apos;d love to hear your feedback
-          </a>
-          !
-        </p>
-      </AlertBanner>
+    <>
+      <DevAlertBanner />
```
zchsh: Abstracted `DevAlertBanner` so it could be re-used in `DocsLayout` without having to use this layout as the underlying layout.

src/layouts/base-new/base-new-layout.module.css
```diff
@@ -1,9 +1,4 @@
-.baseNewLayout {
-  display: flex;
-  flex-direction: column;
-  min-height: 100vh;
```
zchsh, line 4: Have used `Min100Layout` as it implements this same functionality. Added `.header` class to ensure header bars are sticky in this layout.

src/layouts/docs/docs-layout.module.css
```diff
@@ -1,57 +1,14 @@
...
+  padding-bottom: 50vh;
+  /* Remainder of styles here are for docs content.
+    We likely want to convert these to MDX components.
+    Asana task: https://app.asana.com/0/1201010428539925/1201653123138729/f
+    */
   color: var(--token-color-palette-neutral-600);
-  flex-grow: 1;
   font-size: 16px;
   line-height: 22px;
-  max-width: 804px;
```
zchsh, line 51: The content width and positioning are now handled by `sidebar-sidecar-layout`, this `main` content is responsible for content only (and for its semantic role, since `sidebar-sidecar-layout` just uses divs).

src/layouts/docs/partials/sidebar-sidecar-layout/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+  /** Sidebar slot, displayed in an elevated area at the left of the viewport. If the content in the sidebar is very tall, it will overflow. */
+  sidebar: React.ReactNode
+  /** Optional sidecar slot, displayed closely alongside the centered content area. If the content in the sidecar is very tall, it will extend the height of the page. */
+  sidecar?: React.ReactNode
+  /** Optional footer slot, to display below the sidebar + content + sidecar area. If the page content is shorter than 100vh, the footer will still be displayed at the bottom of the viewport.
+   * Note: Not used in current designs, though I think there's a possibility it will be used in the future, so wanted to ensure this implementation can accommodate it. */
+  footer?: React.ReactNode
+}
+
+const SidebarSidecarLayout: React.FC<SidebarSidecarLayoutProps> = ({
```
zchsh, line 15: We could, if we wanted to, roll this layout right into the `docs-layout` component, since it isn't exactly meant to be used on its own. I found it very helpful to split it out when developing though, so figure it might make sense to keep it split it for maintenance (especially while we're still figuring out docs-layout content styling).

src/layouts/docs/index.tsx
```diff
@@ -20,36 +22,45 @@ interface DocsLayoutProps {
...
+const DocsLayout: React.FC<DocsLayoutProps> = ({
+  productName,
+  navData,
+  backToLink,
+  headings,
+  breadcrumbLinks,
+  children,
+  githubFileUrl,
+}) => (
+  <SidebarSidecarLayout
```
zchsh, line 34: Have simplified `docs-layout` a bit by leveraging a new `sidebar-sidecar-layout` component, which handles the layout positioning concerns.

src/pages/_layout-dev/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+      {name} Content
+      <br />
+      <button type="button" onClick={() => setIsLong(!isLong)}>
+        Demo {isLong ? 'short' : 'long'} {name} Content
+      </button>
+    </div>
+  )
+}
+
+function DemoLayoutDev(): React.ReactElement {
```
zchsh, line 19: Intention is to delete this, started this as a skateboard version of getting the layout working the way I think we want it to. Leaving it up for now to make it easier to isolate and debug layout arrangement concerns from docs page things.

src/pages/_layout-dev/style.module.css
```diff
@@ -0,0 +1,48 @@
+/* DEMO CONTENT */
```
zchsh, line 1: As mentioned in accompanying `.tsx` file, intention is to delete this once we're done reviewing and have the layout work approved. Though also considering trying to move this into Swingset or something (though not exactly sure how I'll do that, yet).

src/pages/index.module.css
```diff
@@ -3,6 +3,11 @@
   width: 100%;
   max-width: 1604px;
   margin: 0 auto;
+
+  /* TODO: remove this min-height,
+  just demo-ing the sticky header in
+  BaseNewLayout */
+  min-height: 150vh;
```
zchsh, line 10: Intent is to remove this as mentioned in the TODO, just wanted to demo the `BaseNewLayout` working with longer content (so sticky header behaviour could be confirmed)

src/pages/index.tsx
```diff
@@ -2,19 +2,15 @@ import { ReactElement } from 'react'
 import BaseNewLayout from 'layouts/base-new'
 import ProductCardGrid from 'components/product-card-grid'
 import HomepageHero from 'components/homepage-hero'
-import Footer from 'components/footer'
```
zchsh, line 5: `Footer` now handled by `BaseNewLayout`. I think I might've handled this in the Vault `/docs` PR as well, but hoping to merge this PR first, I think, so will rebase the `/docs` work onto this.

src/pages/style.css
```diff
@@ -10,7 +10,6 @@
 /* Local global style helpers */
 @import 'styles/elevation.css';
 @import 'styles/sizing.css';
-@import 'styles/spacing.css';
```
zchsh, line 13: File is now empty, so baleeted.

src/styles/sizing.css
```diff
@@ -1,7 +1,9 @@
 :root {
   --input-height: 38px;
   --input-icon-size: 16px;
+  /* TODO: remove the need for --navigation-header-height
```
zchsh, line 4: As this comment mentions, the one thing I think is objectively worse about the `body`-as-scrolling layout approach is that it requires we know the height of the sticky nav in order to implement the sidebar behaviour design has given us`*`. The simple-but-not-robust solution to that problem is to have hard-coded CSS vars as we have here.

This hard-coded approach kind of forces us to make our layouts brittle in ways that will affect users who change their default font-size. I think a more robust solution here would be to observe the sticky bar height and dynamically update the calculated height, to avoid that brittleness (as mentioned in the `sidebar-sidecar-layout/style.module.css`).

But, holding off on the not-hard-coded enhancement for now - want to get a higher-level go-ahead that this layout approach makes sense first! 🙇 

> `*` CAVEAT - seems required as far as I can tell - we had to run this kind of layout in Learn and ran into a very similar conundrum... in that case we've just stuck with magic static numbers for the nav bar heights, I think, but it's always kinda worried me.

src/components/sidebar/components/sidebar-nav/style.module.css
```diff
@@ -1,6 +1,9 @@
 .sidebarNavElement {
   padding-top: 8px;
   position: relative;
+  /* necessary as otherwise the skipToMainContent
+     link causes x-overflow */
+  overflow: hidden;
```
BRKalow, line 6: 👀  seems the focus outline is getting clipped:
Screen Shot 2022-01-12 at 5 22 20 PM

BRKalow:
I'd like to see if we could continue retaining the layout composition of `BaseNewLayout -> SidebarSidecarLayout`. Right now it looks like we've got duplicate styles across both, as well as duplicative implementation of `<NavigationHeader />` and `<DevAlertBanner />`. Curious to hear why you opted to decouple from the base layout!

zchsh:
@BRKalow Happy to give the re-coupling a shot - opted to decouple from the base layout as a WIP step while working through the complexities of the `sidebar-sidecar-layout`.

In order to keep the `BaseNewLayout -> SidebarSidecarLayout` composition, I think `SidebarSidecarLayout` will need to be aware of the sticky header height (hopefully not through magic numbers, but maybe that's okay?), and we'll therefore need to have the height observer logic in `BaseNewLayout`. 

Do you have any thoughts on the height observer issue? Eg are the magic `--alert-bar-height` & `--navigation-header-height` feel acceptable, or does implementing a height observer for the sticky bar seem preferable?

BRKalow:
I think in the interest of iteration it'd okay to use static numbers for now! We can introduce the height observer down the line if it feels like the right move.

zchsh:
@BRKalow sounds good - will keep the global CSS vars then, and re-couple to `BaseNewLayout` 👍 (and make the footer optional in `BaseNewLayout` as it was before).

src/components/sidebar/components/sidebar-nav/style.module.css
zchsh: Ah dang... when I don't have `overflow: hidden`, we end up with a massive scroll issue due to how `Skip to main content` is positioned by default.

@ashleemboyer Would an acceptable resolution path here to be position the `Skip to main content` at `top: 4; left: 4` (when focused) in order to avoid this cutoff?

src/pages/index.module.css
zchsh: ✅ Removed

zchsh:
@BRKalow Have refactored so that `SidebarSidecarLayout` leverages `BaseNewLayout`, and have left in the static CSS vars. @ashleemboyer tagged you in the `SkipToMainContent` positioning issue, hopefully that fix looks okay 🙇

BRKalow (approved):
This looks solid to me. Preview looks good, I tested the scroll behavior and all seems to be working like we would expect! Thanks for the detailed comments in the CSS!

ashleemboyer (changes requested):
In addition to the individual comments I've left, I have some other concerns that were harder to pinpoint where to place comments. There were some margin, padding, and width changes made that don't match [the grids defined in Figma](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/?node-id=1531%3A40889), such as some spacing that was built from margin + padding was combined into just margin and the max-width of the main content area was changed. I think it'll be easier to chat about these things either on Slack or Zoom, so let me know when you want to do that @zchsh!

src/components/sidebar/components/sidebar-nav/style.module.css
zchsh: Absolutely - here's a screen recording showing the horizontal scroll issue when `overflow: hidden` is removed:

https://user-images.githubusercontent.com/4624598/151180142-d2f2c5e8-338c-4a8a-af2b-2f75a6a95677.mp4

Screen recording is in Safari, but I'm seeing the issue in Firefox and Chrome as well. The `top` & `left` set to `0` is minor compared to the overflow issue - was done as a fix for the cutoff Bryce mentioned above _after_ adding `overflow: hidden` to s

overflow-hidden-issue
olve for the horizontal scrolling issue:

src/layouts/docs/partials/sidebar-sidecar-layout/style.module.css
```diff
@@ -0,0 +1,65 @@
...
+  must be used for sticky positioning in addition to
+  padding the content area. So, we create a CSS var
+  to make it reusable and express this intent. */
+  --main-area-top-padding: 32px;
+}
+
+.sidebar {
+  background-color: white;
+  box-shadow: 0 2px 3px rgba(101, 106, 118, 0.1),
+    0 12px 28px rgba(101, 106, 118, 0.25);
```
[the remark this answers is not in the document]
zchsh: I think they do have something to do with the layout - the `sidebar` part of the layout creates a scroll container and positions itself to take up the expected height portion of the screen, even if its contents aren't that tall. The shadow etc need to be applied at that level, I think, rather than within the scroll container. Moving the `box-shadow` & `background-color` into `components/sidebar` would mean we'd need to break that model, ie rather than just having the sidebar contents, `components/sidebar` would then need to take on some of these other layout concern styles as well (`max-height`, `width`, etc).

Let me know if that layout vs content rationale seems reasonable 🙇‍♂️

src/layouts/docs/partials/sidebar-sidecar-layout/style.module.css
```diff
@@ -0,0 +1,65 @@
...
+  flex-shrink: 0;
+  margin-right: 48px;
+  /* Sidebar should occupy the full available screen height.
+  We have sticky nav bars, so the full height is equal to
+  100vh minus the combined height of the sticky nav bars. */
+  max-height: calc(100vh - var(--sticky-bars-height));
+  overflow-y: auto;
+  padding-top: 24px;
```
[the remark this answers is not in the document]
zchsh: Great call - yep, will update! 👍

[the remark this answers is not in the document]
zchsh: ✅ Updated

src/layouts/docs/index.tsx
```diff
@@ -22,40 +22,41 @@ interface DocsLayoutProps {
...
+      {breadcrumbLinks && <BreadcrumbBar links={breadcrumbLinks} />}
+      <div className={s.tempContentStyles}>{children}</div>
+      {githubFileUrl && (
+        <EditOnGithubLink
+          className={s.editOnGithubLink}
+          url={githubFileUrl}
+          label="Edit this page on GitHub"
+        />
+      )}
+      <Footer className={s.footer} openConsentManager={openConsentManager} />
```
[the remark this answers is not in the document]
zchsh: Thanks for catching this - ✅ have moved adjacent to `main` 👍

src/layouts/docs/partials/sidebar-sidecar-layout/style.module.css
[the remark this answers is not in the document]
ashleemboyer: I considered the `background-color` and `box-shadow` to be properties that are set regardless of the sidebar's positioning in the layout, things that are more "constant" across usage rather than dependent on the layout. Thinking some more on it `box-shadow`, would be different if in some future layout we put the sidebar on the right side of the page, so that probably is layout dependent! `background-color` seems like a constant though, and like something that should be delegated to the sidebar component to set separate from the layout.

src/components/sidebar/components/sidebar-nav/style.module.css
ashleemboyer: Ah! I see now! So there are a couple of things we can do here instead of `overflow: hidden` in this class.
- Add `overflow-x: hidden` to the `.sidebar` class in `src/layouts/docs/partials/sidebar-sidecar-layout/style.module.css` (it used to be in the `.sidebar` class defined in `src/components/sidebar/style.module.css`).
- Change the default value of `right` in the `.skipToMainContent` class to be `10000px` instead of `-10000px`. This places the skip link on the left side of the sidebar content instead of the right.

I like the second option, and I think it's what I meant to when I implemented this originally!

ashleemboyer:
One thing I forgot to mention yesterday in my review: the heading links aren't quite working when you try to navigate to them directly or when you try to go to them from the sidebar. Here's an example link: https://dev-portal-git-zsrefactor-sidebar-layout-hashicorp.vercel.app/waypoint#featured-reference-docs. I think this has something to do with the page content now going behind the sticky header?

src/layouts/docs/partials/sidebar-sidecar-layout/style.module.css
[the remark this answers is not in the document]
zchsh: @ashleemboyer Sounds good, thanks for the added thoughts - I've moved `background-color` to `sidebar` 👍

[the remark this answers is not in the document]
ashleemboyer: Awesome! I'll mark this as resolved 😊

zchsh:
@ashleemboyer Great catch re: scroll offset. Would it be alright with you if we handle that concern in a separate PR? I think this is [an issue on our existing sites as well](https://www.vaultproject.io/docs/install#linux-package), just wary of having the scope of this PR expand further and get stuck. Let me know what you think! 🙇

ashleemboyer:
> @ashleemboyer Great catch re: scroll offset. Would it be alright with you if we handle that concern in a separate PR? I think this is [an issue on our existing sites as well](https://www.vaultproject.io/docs/install#linux-package), just wary of having the scope of this PR expand further and get stuck. Let me know what you think! 🙇

I'm not seeing an offset when I go to that URL, and I think it depends if we should solve here or not. There isn't currently an offset issue so it's something about the refactor that introduced it. I'm not sure if we should merge the regression if we don't have a clear idea of what to try to fix it. What do you think?

zchsh:
@ashleemboyer Pushed up an offset fix, I thought this was something we'd done on the docs sites, but when I visit https://www.vaultproject.io/docs/install#linux-package I'm seeing the `Linux package` heading hidden under the top sticky navs (should have been more clear about this - if you're not seeing the same let me know, maybe there's a fix we have in place that I'm somehow missing!).

Going to get those upstream changes in to fix the conflicts since #79 merged, and then I _think_ this should be ready for re-review - I believe one outstanding item is how we're structuring or naming `docs-layout` vs `sidebar-sidecar-layout`.

My intent with the current `sidebar-sidecar` naming, and separate partial, is that we may be able to re-use it for the `Install` view. That view seems to follow the same layout, but with different sidecar contents. Definitely open to refining the layout organization further in further PRs. Let me know what you think! 🙇

ashleemboyer:
> @ashleemboyer Pushed up an offset fix, I thought this was something we'd done on the docs sites, but when I visit https://www.vaultproject.io/docs/install#linux-package I'm seeing the `Linux package` heading hidden under the top sticky navs (should have been more clear about this - if you're not seeing the same let me know, maybe there's a fix we have in place that I'm somehow missing!).

Weird! I'm not seeing that. Here's a screen recording of what I see. Maybe it's my screen size or that I'm using Chrome?

https://user-images.githubusercontent.com/43934258/151254323-abac4833-52a7-4a82-a9ed-02a4b07e2676.mov

> Going to get those upstream changes in to fix the conflicts since #79 merged, and then I _think_ this should be ready for re-review - I believe one outstanding item is how we're structuring or naming `docs-layout` vs `sidebar-sidecar-layout`.
>
> My intent with the current `sidebar-sidecar` naming, and separate partial, is that we may be able to re-use it for the `Install` view. That view seems to follow the same layout, but with different sidecar contents. Definitely open to refining the layout organization further in further PRs. Let me know what you think! 🙇

Thanks for elaborating! That makes sense. My plan for `Sidecar` with that new view (which I haven't brought up in any discussions yet) is to refactor `Sidecar` to accept children instead of only rendering one thing. Then I'm thinking to create a `TableOfContents` component (or something of that name) that handles rendering the `<nav>` element currently in the `Sidecar` and passing that as a child. With this refactor in mind, I think renaming `docs-layout` to `sidebar-sidecar-layout` is a good approach, then everything is abstracted at a nice level and everything (except `Sidebar` lol) is essentially just a container rendered in the correct place on the page by `sidebar-sidecar-layout`.

zchsh:
@ashleemboyer Ah neat, yep, I'm seeing the same thing in Chrome and Firefox (heading visible, which is nice), just seems like Safari where I'm seeing the hidden heading.

Does the fix for heading offsets work for you? Eg does https://dev-portal-git-zsrefactor-sidebar-layout-hashicorp.vercel.app/waypoint#featured-reference-docs work? (Seems to work for me in Safari, Chrome, & Firefox 🤞. Also want to acknowledge it's not the best long-term fix; my hope is we can iterate on it as part of the general `.__permalink-h` refactor work 👍)

Re: `DocsLayout` renaming, do you mind if we handle that in your Install view PR? Definitely open to the rename etc, just having trouble figuring out how all those pieces fit together without the extra Install page in place (eg I assumed we'd pull out `sidebar-sidecar-layout` from `layouts/docs/partials`, and leave eg `layouts/docs/server` in place; I think those concerns will be easier to work through with more concrete reasons to move things around in front of us). Let me know if that's alright! 🙇‍♂️

zchsh:
Not sure what's going on with the Bundle Analysis check here 🤔 Seems unrelated to this PR I think?

src/components/sidebar/components/sidebar-nav/style.module.css
ashleemboyer: @zchsh looks like this still needs to be addressed! 🙏🏻

src/layouts/docs/docs-layout.module.css
```diff
@@ -83,6 +47,30 @@ heading with only a couple words, the element can become too narrow.
...
+    &::before {
+      display: block;
+      content: '\200B';
+      height: calc(var(--navigation-header-height) + var(--alert-bar-height));
+      margin-top: calc(
+        -1 * calc(var(--navigation-header-height) + var(--alert-bar-height))
+      );
+      visibility: hidden;
+    }
+  }
```
ashleemboyer, line 72: This is really neat! Nicely done ✨

ashleemboyer (changes requested):
In addition to the things we've already chatted about via Slack, just seeing that one more thing needs addressed! We can defer the renaming of the layout, potentially even when I work on refactoring Sidecar.

src/layouts/docs/docs-layout.module.css
BRKalow: for consideration: I think something similar to this is how we're handling it on the current docs sites: https://github.com/hashicorp/nomad/blob/main/website/pages/style.css#L33-L49

src/components/sidebar/components/sidebar-nav/style.module.css
zchsh: Ah thanks for flag, lost track of this conversation! Just adjusted with the second option 💯 ✅

ashleemboyer (changes requested):
Latest updates are looking great! It uncovered one more bug I wasn't noticing before. The `<nav>` in sidecar didn't used to have a parent container, and it has two classes on it that handle making the element go away on narrower views: `hide-on-mobile` and `hide-on-tablet`. Those classes need to be moved to the new containing `<div>` around that `<nav>` or we see something like this, an empty div that's `220px` wide:

image

zchsh:
@ashleemboyer Thanks for catching the sidecar bug - should be fixed now! 🙇

ashleemboyer (approved):
Woo! 🥇

PR #80, opened 2022-01-23
feat: add IconTile and IconTileLogo

zchsh:

---

## Relevant links

- [Preview link - `IconTile`](https://dev-portal-git-zsadd-icon-tile-hashicorp.vercel.app/swingset/components/icontile) 🔎
- [Preview link - `IconTileLogo`](https://dev-portal-git-zsadd-icon-tile-hashicorp.vercel.app/swingset/components/icontilelogo) 🔎
- [Asana task - `IconTile`](https://app.asana.com/0/1100423001970639/1201654639085754/f) 🎟️
- [Asana task - `IconTileLogo`](https://app.asana.com/0/1100423001970639/1201654639085755/f) 🎟️

## What

This PR adds `IconTile` and `IconTileLogo` components. It includes documentation for these components in Swingset, but does not use the components on any pages.

## Why

This change is being made both to support upcoming Downloads aka Install page work, as well as to support cleanup of content authoring on `/{product}` pages.

## How

As outlined in the Asana task:
- `IconTile` accepts `size` and `brandColor` props, and accepts `@hashicorp/flight-icon` icons rendered through the `children` prop.
- `IconTileLogo` leverages `IconTile` under the hood, as well as the existing `ProductIcon` dictionary-like component, in order to provide a simplified props API. Consumers provide a `product` slug and the `IconTileLogo` renders the appropriate product icon, using the slug for the `brandColor` to ensure the theme and icon match.

## Testing

- [ ] Visit the Swingset pages, and try out the components using `LiveComponent`
- [ ] Review the new component code (hopefully the props API and styling make sense given our use cases and the designs we need to implement)

## Anything else?

- Note that `IconTile` is an incomplete implementation relative to the HashiCorp Design System spec - it does not include the `secondary` icon feature.
- Note that `IconTileLogo` differs in look from the HashiCorp Design System spec. This is because the designs we have from Figma differ in look. We will likely need to revisit this in the future (tweaking either the dev-portal designs, or our implemented styles), but for now our focus is on continuing to ship the specs we receive from the dev-portal design team.

---
  - https://app.asana.com/0/0/1201654639085755
  - https://app.asana.com/0/0/1201654639085754


didoo:
hey @zchsh I was lurking the PRs in the repo and I noticed this one, that refers to an `IconTile` and `IconTileLogo` and I thought to give you more context for the IconTile component in the design system, expecially now that the work is still in draft.

What you see in the Figma UI Kit ([link](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/?node-id=1377%3A11992)), that is two "components" - `IconTile` and `IconTileLogo` - is because of a limitation of Figma and a simplifications the designers decided to have, so it's easier for designers to use the component.

But in the actual implementation in code, the component is a single one and is called just `IconTile`. You can see the docs here https://design-system-components-hashicorp.vercel.app/components/icon-tile and the implementation in Ember code here: https://github.com/hashicorp/design-system-components/tree/main/addon/components/hds/icon-tile

My suggestion is to create a single component also in React (you can mimic the code implementation) and if you can keep the APIs/props the same (or very similar) that would be gold (and if you can't maybe we can discuss, and we can change them on the Ember side).

Essentially, what controls the two cases is just if the user passes an `icon` or a `logo` prop.

PS: wicked idea, you may consider to use the Sass directly somehow?! https://github.com/hashicorp/design-system-components/blob/main/app/styles/components/icon-tile.scss

src/components/icon-tile-logo/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Use the IconTileLogo component when representing the product in our interfaces.
+ *
+ * Renders a product icon in a brand-colored tile. Note that this IconTileLogo
+ * is currently styled to match dev-portal designs, and differs in size and
+ * background coloration from the IconTileLogo documented in the
+ * HashiCorp Design System.
+ *
+ * ref: https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/HDS-Product---Components-%5BWIP%5D?node-id=1377%3A11992
```
zchsh, line 18: Added notes on the `IconTile` & `IconTileLogo` implementations, to acknowledge they're not 1:1 with the HashiCorp Design System Figma yet.

src/components/icon-tile-logo/docs.mdx
```diff
@@ -0,0 +1,52 @@
...
+Renders a product icon in a brand-colored tile. Note that this IconTileLogo
+is currently styled to match dev-portal designs, and differs in size and
+background coloration from [the IconTileLogo documented in the
+HashiCorp Design System](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/HDS-Product---Components-%5BWIP%5D?node-id=1377%3A11992).
+
+<LiveComponent>{`
+    <IconTileLogo product="waypoint" />
+`}</LiveComponent>
+
+## Examples
```
zchsh, line 16: Have added what I think is an exhaustive set of Examples here. Happy to add more, or add any other info here, if it feels like anything is missing 👍

src/components/icon-tile/docs.mdx
```diff
@@ -0,0 +1,30 @@
...
+not fully implement [the IconTile component documented in the
+HashiCorp Design System](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/HDS-Product---Components-%5BWIP%5D?node-id=1377%3A11992), as it does not support "secondary" icons.
+
+<LiveComponent>{`
+    <IconTile size="large"><svg id="flight-smile-16" viewBox="0 0 16 16" width="100%" height="100%"><g><path d="M8 12c-1.01 0-1.782-.504-2.267-.945a4.72 4.72 0 01-.564-.614 3.31 3.31 0 01-.212-.305.75.75 0 011.284-.775 3.214 3.214 0 00.5.584c.341.31.769.555 1.259.555.49 0 .918-.246 1.258-.555a3.214 3.214 0 00.5-.584.75.75 0 011.285.775l-.212.305c-.128.167-.317.39-.564.614C9.782 11.495 9.01 12 8 12zM5 6a1 1 0 011-1h.007a1 1 0 010 2H6a1 1 0 01-1-1zM10 5a1 1 0 100 2h.007a1 1 0 100-2H10z"></path><path fill-rule="evenodd" d="M8 0a8 8 0 100 16A8 8 0 008 0zM1.5 8a6.5 6.5 0 1113 0 6.5 6.5 0 01-13 0z" clip-rule="evenodd"></path></g></svg></IconTile>
+`}</LiveComponent>
+
+## Examples
+
+Due to limitations in our ability to import external packages into Swingset documentation pages, it's unfortunately not practical to fully demonstrate the use of `@hashicorp/flight-icons` with `IconTile`. An example of use should look something like:
```
zchsh, line 18: Feels a bit harder to show useful examples of `IconTile`, since I think I'd have to inline the SVG as in the `LiveComponent`. Have opted for a code block example here.

> The way this is rendering isn't great IMO - making me think about getting some better default content styles in place for Swingset. But that definitely feels out of scope for this PR.

src/components/icon-tile/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+  | 'neutral'
+  | Exclude<Exclude<ProductSlug, 'sentinel'>, 'hcp'>
+
+interface IconTileProps {
+  /** Pass a single child, which should be a Flight icon. For 'small' and 'medium' size, pass the 16px icon size; for other sizes pass the 24px icon size. Note that non-"color" icons will be colored using the "brandColor". */
+  children: React.ReactNode
+  /** Note: the "extra-large" option is not documented in the design system. It's being used for the IconTileLogo component, as used on the /{product} view pages. */
+  size: 'small' | 'medium' | 'large' | 'extra-large'
+  /** Optional product slug to use for brand color theming. If not provided, defaults to "neutral". Note that "sentinel" and "hcp" are not supported. */
+  brandColor?: ProductBrandColor
```
zchsh, line 15: Have gone with the props API suggested in the task 👍 Feels pretty perfect to me so far! 💯

src/components/product-icon/index.tsx
```diff
@@ -1,6 +1,6 @@
 import { IconBoundaryColor16 } from '@hashicorp/flight-icons/svg-react/boundary-color-16'
 import { IconConsulColor16 } from '@hashicorp/flight-icons/svg-react/consul-color-16'
-import { IconHashicorp16 } from '@hashicorp/flight-icons/svg-react/hashicorp-16'
+import { IconHashicorpColor16 } from '@hashicorp/flight-icons/svg-react/hashicorp-color-16'
```
zchsh, line 3: Made this update - icons are equivalent, except the `Color` icon won't change color via `currentcolor`, which is necessary for use in `IconTileLogo`.

zchsh:
@didoo  Thanks for this, I appreciate the extra context here! ☺️

An important note here, I think, is that our scope for these icons is just `dev-portal` for now - we don't mean for these to be the React "reference" implementations for the HashiCorp Design System (yet!), and do not intend to use them outside of `dev-portal`. Hopefully that helps contextualize the approach we're taking (ie not perfectly one-to-one with the design system in every aspect... yet!). Will add some further notes below...

---

Re: `IconTile` and `IconTileLogo` being separate components, this is intentional on our part - we want to ensure that `IconTileLogo` matches the design specs we have (which currently only have one `96px` square size of `IconTileLogo`), and that, for example, the `boundary` icon is used only with the `boundary` colour theme. We could change the React props API as you mentioned (or simply not implement the separate `IconTileLogo` component) but I think this would be a reduction in the quality of the dev experience.

Happy to adjust this in future PRs if there are strong opinions on this - our focus right now is on implementing the specs we receive from the dev-portal design team 🎨 

Re: Using the Sass directly, that's a neat suggestion! It's something we could potentially explore, but for now, we're using a different solution to the problems Sass solves in the form of PostCSS. So unfortunately we can pluck much from Sass files. As mentioned though, definitely open to revisiting this in future iterations to ensure the React and Ember implementations are more one-to-one.

src/components/icon-tile/docs.mdx
ashleemboyer: > making me think about getting some better default content styles in place for Swingset

I've been feeling similarly! Maybe we can get some table styles in too? 😊 (agree, not in the scope of this PR)

src/components/icon-tile/index.tsx
```diff
@@ -0,0 +1,46 @@
+import s from './style.module.css'
+import classNames from 'classnames'
+import { ProductSlug } from 'types/products'
+
+type ProductBrandColor =
+  | 'neutral'
+  | Exclude<Exclude<ProductSlug, 'sentinel'>, 'hcp'>
```
ashleemboyer, line 7: I _believe_ you can simplify this "double Exclude" with a `|` like this:

```suggestion
type ProductBrandColor = 'neutral' | Exclude<ProductSlug, 'sentinel' | 'hcp'>
```

src/components/product-icon/index.tsx
ashleemboyer: This change breaks this icon in `ProductSwitcher`:

image

Can you add a special class in there like what's done for Vault to set the color to white in the switcher?

ashleemboyer (changes requested):

src/components/product-icon/index.tsx
didoo: For context:
- as Ashlee noticed, the icon for HCP should not be the "color" one (the official "brand" color for this icon is black) but the normal one that can be tinted (in this case white, I imagine).
  - even better, you could use the `hcp-16` icon, that exists exactly for HCP (https://flight-hashicorp.vercel.app/?query=hcp) 
- the vault icon has been recently released (see https://github.com/hashicorp/flight/issues/372) with the correct yellow: https://flight-hashicorp.vercel.app/?query=vault-color

src/components/icon-tile-logo/index.tsx
```diff
@@ -0,0 +1,31 @@
+import ProductIcon from 'components/product-icon'
+import { ProductSlug } from 'types/products'
+import IconTile from 'components/icon-tile'
+
+type ProductWithIconSlug = Exclude<ProductSlug, 'sentinel'>
+interface IconTileLogoProps {
+  product: ProductWithIconSlug
+}
```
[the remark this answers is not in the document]
zchsh: These are great suggestions, ty 💯 Have applied both! ✅

src/components/product-icon/index.tsx
zchsh: Thanks for catching this @ashleemboyer , and thanks for the extra context @didoo ! I've taken the approach Ashlee suggested - we need the "color" icon in place for now to avoid having the color change in `IconTileLogo` contexts.

Have added a `TODO` for us to revisit this in the future - eg perhaps we could expand `ProductIcon` to have a `color={true | false}` option to switch between "color" and tintable icons in the future.

src/components/icon-tile/type.ts
```diff
@@ -0,0 +1,14 @@
+import { ProductSlug } from 'types/products'
+
+export type ProductBrandColor =
+  | 'neutral'
+  | Exclude<Exclude<ProductSlug, 'sentinel'>, 'hcp'>
```
ashleemboyer, line 5: Not sure what happened here, but the double `Exclude` came back!

ashleemboyer (changes requested):
Double `Exclude` came back in, and I also noticed that [I misled you on a file name](https://github.com/hashicorp/dev-portal/pull/80#discussion_r796057489) with my own typo! 😭 Could you rename the two new `type.ts` files to `types.ts`? Sorry about that miscommunication!

ashleemboyer (approved):
🥇

PR #83, opened 2022-01-25
[🚧 WIP] Add Card and CardLink components

ashleemboyer:

---

## Relevant links

- [Preview link](url) 🔎
- [Asana task](url) 🎟️


didoo:
@ashleemboyer if it can help, this is the `Card` component in Ember: https://design-system-components-hashicorp.vercel.app/components/card

I don't know if it's the same component you're building ([this is the design in Figma](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/?node-id=2%3A9)) but if it is you may want to consider more use cases (likely the designers will use them, sooner or later) like with/without border, or with a different background, plus overflow (needed in case the content or the background is not the same as the one the card is on top.

ashleemboyer:
> @ashleemboyer if it can help, this is the `Card` component in Ember: https://design-system-components-hashicorp.vercel.app/components/card
> 
> I don't know if it's the same component you're building ([this is the design in Figma](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/?node-id=2%3A9)) but if it is you may want to consider more use cases (likely the designers will use them, sooner or later) like with/without border, or with a different background, plus overflow (needed in case the content or the background is not the same as the one the card is on top.

@didoo thanks for the comment! Yes, that's the component we referenced for the planning of this `Card` component. We have an internal team RFC where we discussed the designs in the Figma you linked, current DevDot designs, potential props, and the use cases we are building for currently. For the time being, we are building this component to go with the designs we have already implemented and the ones we plan to implement in the next couple of weeks. As we see more use cases in our DevDot designs, we will iterate on these components and their APIs as needed.

didoo:
> @didoo thanks for the comment! Yes, that's the component we referenced for the planning of this `Card` component. 

@ashleemboyer not to nitpick, but for sharing: in theory the Card component should not support the "low" elevation, according to designs. I imagine the DevDot designers have taken this decision, to use also "low" too, but wanted to ping @heatherlarsen on this, so she knows (and we may consider less strict rules for the designers).

ashleemboyer:
> @ashleemboyer not to nitpick, but for sharing: in theory the Card component should not support the "low" elevation, according to designs. I imagine the DevDot designers have taken this decision, to use also "low" too, but wanted to ping @heatherlarsen on this, so she knows (and we may consider less strict rules for the designers).

Yes, so the reason that option is supported here is because that is the elevation specified in our designs. Here's an example: https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/?node-id=3241%3A65835.

braydenlove:
@ashleemboyer, Cris and I need to update our card styles to be aligned with DST. We made these cards before the documentation came out and haven't had a chance to update them yet.

ashleemboyer:
> @ashleemboyer, Cris and I need to update our card styles to be aligned with DST. We made these cards before the documentation came out and haven't had a chance to update them yet.

@braydenlove thanks for the heads up, that's fine! Just let us know when y'all have updated and we can make another PR that updates the code to match.

src/components/card-link/types.ts
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * NOTE: "base" is excluded from `elevation` because that elevation should only
+ * be used with non-interactive Cards, according to the design system.
+ */
+interface CardLinkProps {
+  children: ReactNode
+  className?: string
+  elevation?: Exclude<CardProps['elevation'], 'base'>
+  href: string
+}
```
BRKalow, line 13: nit: Sometimes I find it helpful to add docs comments above properties if it makes sense. So for example, I might move the comment above the interface to right above `elevation`.

```suggestion
interface CardLinkProps {
  children: ReactNode
  className?: string
  /**
   * NOTE: "base" is excluded from `elevation` because that elevation should only
   * be used with non-interactive Cards, according to the design system.
   */
  elevation?: Exclude<CardProps['elevation'], 'base'>
  href: string
}
```

src/components/card/types.ts
```diff
@@ -0,0 +1,9 @@
+import { ReactNode } from 'react'
+
+interface CardProps {
+  children: ReactNode
+  className?: string
+  elevation?: 'base' | 'low' | 'mid' | 'high'
+}
+
+export type { CardProps }
```
BRKalow, line 9: This could be changed to the following without any impact, and I think aligns with what we do elsewhere:
```suggestion
export interface CardProps {
  children: ReactNode
  className?: string
  elevation?: 'base' | 'low' | 'mid' | 'high'
}
```

BRKalow (commented):
Exciting!

src/components/card/types.ts
ashleemboyer: Ah, okay! I think I went with `export type` because of something about cyclical exports? Not sure if I applied that correctly, but I remember you mentioning it during pairing some time ago. Do you remember what we were talking about?

src/components/card-link/types.ts
ashleemboyer: So I thought about doing that, but went against it because while it still looks okay when there's just one comment above a property, I find it harder to read overall when every property has comments above it. Maybe we should try to establish some guidelines here? 🤔

BRKalow: I'd say the comment above the property is preferable so that it shows up in intellisense:
Screen Shot 2022-01-27 at 1 38 33 PM

src/components/card/types.ts
BRKalow: Hmm, I do not 😅 

I don't think cyclical imports/exports would be a problem here, I think these two are functionally equivalent. cyclical imports could come into play if we were, say, importing something from the index file in the types file, and also importing something from the types file in the index file.

ashleemboyer: That makes sense, thanks for elaborating! The update has been made in https://github.com/hashicorp/dev-portal/pull/83/commits/f49491e9c9b3d5445131c4f1252d0b3fbf50af7a. 😊

src/components/card-link/types.ts
ashleemboyer: Great point, I've updated and added a comment for each property in https://github.com/hashicorp/dev-portal/pull/83/commits/b30dfa09ecf2a5af16035bfdabdde43b27cb68d0!

src/components/card-link/card-link.module.css
```diff
@@ -0,0 +1,34 @@
+/*
+NOTE: `color: black` is needed because the color is otherwise set by a global
+style applied to anchor elements.
+*/
+.root {
+  color: black;
+
+  /* Only enable animation if query is supported and no-preference on motion */
+  @media (prefers-reduced-motion: no-preference) {
+    transition: background-color 0.2s;
```
BRKalow, line 10: This would make it ✨ fancy
```suggestion
    transition: background-color 0.2s, box-shadow 0.2s;
```

src/components/card-link/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+const CardLink = ({
+  children,
+  className,
+  href,
+}: CardLinkProps): ReactElement => {
+  const classes = classNames(s.root, className)
+
+  return (
+    <MaybeInternalLink href={href}>
```
BRKalow, line 16: At least in swingset, I'm seeing the anchor element expand to fit the full width of its containing element, so you can click in the empty space to the right of the card and you'll hit the link.

ashleemboyer: Uh oh! 😨  Will get right on that.

src/components/card-link/card-link.module.css
ashleemboyer: Agree!! Didn't think to add this in after adding the hover state changes. 😊

src/components/card-link/index.tsx
ashleemboyer: Just pushed a fix in https://github.com/hashicorp/dev-portal/pull/83/commits/6069a1fdf3b27ea5c9c3e4c97870554e2fc6f987!

BRKalow (approved):
👏

src/components/product-card-grid/style.module.css
```diff
@@ -107,26 +107,16 @@
...
 .sectionBodyCardWrapper {
-  display: block;
   flex: 1 1 0;
   min-width: 120px;
   padding: 12px;
   width: min-content;
 }
 
 .sectionBodyCard {
-  composes: hds-surface-low from global;
```
zchsh, line 118: Love how much simpler this is from a "i wanna make a layout with cards" perspective 💯

zchsh (approved):
🙌 💯 💖 Apologies for the delayed review here - this looks great to me!

Love the Swingset pages as well 👍 Excited to iterate on those... I know the table styles don't look the greatest at the moment 😅 But props table still feels super useful 🎉

src/components/product-card-grid/style.module.css
ashleemboyer: Agreed! It was satisfying deleting the code now handled by the card. 😊

PR #97, opened 2022-02-02
Improve sidecar scrolling

ashleemboyer:

---

## Relevant links

- [Preview link](https://dev-portal-git-ambimprove-sidecar-scrolling-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1201010428539925/1201508675437098/f) 🎟️

## What

- Updates `Sidecar` to completely ignore headings with level 3 or greater
- Tweaks the `rootMargin` and `threshold` values in `IntersectionObserver` in `use-active-section`

## Why

- Ignoring headings with level 3 or greater is something we only did halfway at the time it was added. We stopped rendering the headings in `Sidecar` but we didn't have `use-active-section` ignoring them. It was setting headings that weren't visible in the `Sidecar` as active, making it look like there was no active heading in some pages while scrolling.
- The `rootMargin` and `threshold` tweaks are hopefully refining the accuracy of detecting active sections. 🤞🏻

## How

- Used `Array.filter` on the `headings` passed into `Sidecar` to get only the headings we're interested in seeing: `h1` and `h2`, and used `useMemo` to create that filtered list so it's only calculated when the `headings` array changes.
- The `rootMargin` and `threshold` values are just some values I decided to try that appear to be working.

## Testing

It's tedious, but we need to step through _many_ pages under the following sections of the app to make sure we're covering as many cases as possible with these changes:

- [ ] https://dev-portal-git-ambimprove-sidecar-scrolling-hashicorp.vercel.app/waypoint
- [ ] https://dev-portal-git-ambimprove-sidecar-scrolling-hashicorp.vercel.app/waypoint/docs
- [ ] https://dev-portal-git-ambimprove-sidecar-scrolling-hashicorp.vercel.app/waypoint/commands
- [ ] https://dev-portal-git-ambimprove-sidecar-scrolling-hashicorp.vercel.app/waypoint/plugins
- [ ] https://dev-portal-git-ambimprove-sidecar-scrolling-hashicorp.vercel.app/vault
- [ ] https://dev-portal-git-ambimprove-sidecar-scrolling-hashicorp.vercel.app/vault/docs
- [ ] https://dev-portal-git-ambimprove-sidecar-scrolling-hashicorp.vercel.app/vault/api-docs

## Anything else?

Some helpful resources:

- [Now You See Me: How To Defer, Lazy-Load And Act With IntersectionObserver](https://www.smashingmagazine.com/2018/01/deferring-lazy-loading-intersection-observer-api/)
- [The Intersection Observer API - Medium Article](https://blog.arnellebalane.com/the-intersection-observer-api-d441be0b088d)

### DIFF
--- a/src/components/sidecar/use-active-section.ts
+++ b/src/components/sidecar/use-active-section.ts
@@ -1,5 +1,7 @@
 import { useEffect, useRef, useState } from 'react'
 import { SidecarHeading } from './types'
+import { useCurrentProduct } from 'contexts'
+import useCurrentPath from 'hooks/use-current-path'
 import getCSSVariableFromDocument from 'lib/get-css-variable-from-document'
 
 /**
@@ -34,6 +36,13 @@ export function useActiveSection(
   const visibleHeadings = useRef<Set<string>>(new Set())
   const [activeSection, setActiveSection] = useState<SidecarHeading['slug']>()
   const previousY = useRef<number>()
+
+  // isProductLanding is needed to determine the IntersectionObserver threshold
+  // because the headings on product landing pages are smaller than the on docs
+  // pages
+  const currentPath = useCurrentPath({ excludeHash: true, excludeSearch: true })
+  const currentProduct = useCurrentProduct()
+  const isProductLanding = currentPath === `/${currentProduct.slug}`
 
   useEffect(() => {
     // Need to clear this when all headings change
@@ -115,7 +124,10 @@ export function useActiveSection(
           previousY.current = currentY
         }
       },
-      { rootMargin: `-${getFullNavHeaderHeight()}px 0% -60% 0%`, threshold: 1 }
+      {
+        rootMargin: `-${getFullNavHeaderHeight()}px 0% -60% 0%`,
+        threshold: isProductLanding ? 0 : 1,
+      }
     )
 
     headings.forEach((section) => {
@@ -133,7 +145,7 @@ export function useActiveSection(
         if (el) observer.unobserve(el)
       })
     }
-  }, [headings, isEnabled])
+  }, [headings, isEnabled, isProductLanding])
 
   return activeSection
 }
--- a/src/components/sidecar/use-active-section.ts
+++ b/src/components/sidecar/use-active-section.ts
@@ -80,7 +80,7 @@ export function useActiveSection(
           previousY.current = currentY
         }
       },
-      { rootMargin: '0% 0% -75% 0%' }
+      { rootMargin: '0% 0% -65% 0%', threshold: 1 }
     )
 
     headings.forEach((section) => {
--- a/src/components/sidecar/use-active-section.ts
+++ b/src/components/sidecar/use-active-section.ts
@@ -67,8 +67,8 @@ export function useActiveSection(
           ? findMatchingSectionIndex(entries[0].target.id)
           : -1
 
-        // Find the heading closest to the top
         if (visibleHeadings.current.size > 0) {
+          // Find the heading closest to the top of the viewport
           let shortestDistance
           let closestHeading
           visibleHeadings.current.forEach((headingId) => {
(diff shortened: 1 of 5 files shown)


ashleemboyer:
📌 Found an issue with this heading: https://dev-portal-git-ambimprove-sidecar-scrolling-hashicorp.vercel.app/waypoint/docs/getting-started#installing-waypoint

BRKalow (approved):
Clicked and scrolled through a bunch of pages. This seems like a definite improvement!

ashleemboyer:
📌 Another, the bottom section isn't "reachable" in the active state detection: https://dev-portal-git-ambimprove-sidecar-scrolling-hashicorp.vercel.app/waypoint/docs/kubernetes/helm-deploy#reference-documentation

BRKalow:
Same as above ☝️ bottom section doesn't highlight as it doesn't have enough content (FWIW, this seems common with a component like this and I wouldn't consider it blocking): https://dev-portal-git-ambimprove-sidecar-scrolling-hashicorp.vercel.app/waypoint/docs/server/auth#revoke-inspect-etc

BRKalow:
Heading below the linked one is highlighted: https://dev-portal-git-ambimprove-sidecar-scrolling-hashicorp.vercel.app/waypoint/docs/server/run#upgrading

ashleemboyer:
All of the identified issues are fixed with the latest changes! I'm going to keep poking around and see if I find any more problems but I think a bunch of things are fixed now! 🎉

src/components/sidecar/use-active-section.ts
```diff
@@ -42,22 +67,32 @@ export function useActiveSection(
...
+          let closestHeading
+          visibleHeadingIds.forEach((headingId) => {
+            const headingEntry = visibleHeadings.current[headingId]
+            const distance = headingEntry.getBoundingClientRect().bottom
+            if (!closestHeading || distance < shortestDistance) {
+              closestHeading = headingId
+              shortestDistance = distance
+            }
+          })
+          setActiveSection(closestHeading)
```
BRKalow, line 83: Thinking out loud...I wonder this could be accomplished by tweaking the root margin instead? If we increase the margin to something like `-80`, there would in theory be less collisions where a heading lower down the page would intersect. The root margin viz in this article was helpful for my mental model: https://blog.arnellebalane.com/the-intersection-observer-api-d441be0b088d

src/components/sidecar/use-active-section.ts
```diff
@@ -25,14 +48,16 @@ export function useActiveSection(
...
-        const visibleHeadings = []
 
         entries.forEach((entry) => {
+          const entryId = entry.target.id
           currentY = window.scrollY
           scrollTrend = previousY.current < currentY ? 'down' : 'up'
 
           if (entry.isIntersecting) {
-            visibleHeadings.push(entry.target.id)
+            visibleHeadings.current[entryId] = entry.target
```
BRKalow, line 58: Storing references to DOM nodes in this way is _potentially_ dangerous, in that react might recreate them at any moment. An alternative might be to query by the ID when we need the element, but maybe the elements are stable enough in our situation.

ashleemboyer: Great point, I'll make this adjustment!

src/components/sidecar/use-active-section.ts
ashleemboyer: I agree with this, I think we can actually start refining this value now that we've got a better hold on tracking visible headings! I'll up that bottom rootMargin value and we can test it out. 😊

ashleemboyer: Gave some numbers a try, and `-60%` on the bottom seems like the safest bet. `-80%` was too large with the top value now being set, and `-65%` seemed like the smallest value that would work with the very last headings on a page. I think `-60%` gives us a little more room for those bottom headings.

src/components/sidecar/use-active-section.ts
```diff
@@ -1,18 +1,53 @@
...
+  const visibleHeadings = useRef<Set<string>>(new Set())
   const [activeSection, setActiveSection] = useState<SidecarHeading['slug']>()
   const previousY = useRef<number>()
 
+  // isProductLanding is needed to determine the IntersectionObserver threshold
+  // because the headings on product landing pages are smaller than the on docs
+  // pages
+  const currentPath = useCurrentPath({ excludeHash: true, excludeSearch: true })
+  const currentProduct = useCurrentProduct()
+  const isProductLanding = currentPath === `/${currentProduct.slug}`
```
ashleemboyer, line 45: I don't _love_ this, and maybe this calls for revisiting why headings have different sizes between the product landing pages and docs-y content pages, but I was getting really tired of testing out a bunch of values to only have them work on either a product landing page or it's docs-y content.

src/layouts/docs/docs-layout.module.css
```diff
@@ -61,12 +61,18 @@
     */
     & a:global(.__target-h)::before,
     &::before {
+      --permalink-scroll-offset: 24px;
```
ashleemboyer, line 64: This just feels like a nice addition to have so that headings aren't right at the top of the viewport when going to a permalink.

BRKalow (approved):
I think this is a great improvement! Nice job here. I wonder if there's opportunity to further simplify some of it now that we are looking for the "top" intersecting element, instead of the last one. 💭Regardless, let's ship this for now and passively keep an eye on it for any further refinement. Nice work👏

PR #104, opened 2022-02-09
[🚧 WIP] Add tabs component

ashleemboyer:

---

## Relevant links

- [Preview link](url) 🔎
- [Asana task](url) 🎟️

### DIFF
--- a/src/components/dev-dot-tabs/docs.mdx
+++ b/src/components/dev-dot-tabs/docs.mdx
@@ -1,0 +1,25 @@
+---
+componentName: DevDotTabs
+---
+
+<LiveComponent>{`
+<DevDotTabs
+  tabs={[
+    {
+      id: 'tab-1',
+      label: 'Tab 1',
+      content: <p>I am in Tab 1</p>
+    },
+    {
+      id: 'tab-2',
+      label: 'Tab 2',
+      content: <p>I am in Tab 2</p>
+    },
+    {
+      id: 'tab-3',
+      label: 'Tab 3',
+      content: <p>I am in Tab 3</p>
+    },
+  ]}
+/>
+`}</LiveComponent>
--- a/src/components/dev-dot-tabs/index.tsx
+++ b/src/components/dev-dot-tabs/index.tsx
@@ -1,0 +1,49 @@
+import { ReactElement, ReactNode } from 'react'
+
+interface DevDotTabsProps {
+  tabs: {
+    content: ReactNode
+    id: string
+    label: string
+  }[]
+}
+
+const DevDotTabs = ({ tabs }: DevDotTabsProps): ReactElement => {
+  return (
+    <div>
+      {/* TODO: label prop? or aria-labelledby? either/or?*/}
+      <div role="tablist">
+        {tabs.map((tab, index) => {
+          // TODO: also check the upcoming activeIndex prop
+          const isActive = index === 0
+
+          return (
+            <button
+              aria-controls={`${tab.id}-tabpanel`}
+              aria-selected={isActive}
+              id={`${tab.id}-tab`}
+              key={tab.id}
+              role="tab"
+              tabIndex={isActive ? 0 : -1}
+              type="button"
+            >
+              {tab.label}
+            </button>
+          )
+        })}
+      </div>
+      {tabs.map((tab) => (
+        <div
+          aria-labelledby={`${tab.id}-tab`}
+          id={`${tab.id}-tabpanel`}
+          key={tab.id}
+          tabIndex={0}
+        >
+          {tab.content}
+        </div>
+      ))}
+    </div>
+  )
+}
+
+export default DevDotTabs
--- a/src/pages/waypoint/downloads/index.tsx
+++ b/src/pages/waypoint/downloads/index.tsx
@@ -7,6 +7,7 @@ import SidebarSidecarLayout from 'layouts/sidebar-sidecar'
 import Card from 'components/card'
 import Text from 'components/text'
 import s from './downloads-page.module.css'
+import DevDotTabs from 'components/dev-dot-tabs'
 
 const LOREM_IPSUM =
   'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus vitae leo id nunc convallis euismod et vel erat. Fusce vel velit turpis. Vivamus fringilla consequat metus, vitae euismod sem eleifend in. Morbi in ullamcorper dui. Quisque rutrum auctor tristique. Vivamus ac turpis non arcu fringilla interdum. Aliquam feugiat lectus ipsum, eu tincidunt mi tristique id. Aliquam sodales eros semper pharetra molestie. Mauris porta, nunc in tempor eleifend, metus massa sagittis nisi, non maximus quam mauris a erat. Duis nec risus diam. Aenean auctor accumsan ipsum. Interdum et malesuada fames ac ante ipsum primis in faucibus. Fusce et sagittis nunc. Cras vel eros id purus sollicitudin lobortis. Vivamus hendrerit volutpat nulla.'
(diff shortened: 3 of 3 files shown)


ashleemboyer:
> @ashleemboyer Did you assess if `@reach/tabs` would give us some of this out of the box? https://reach.tech/tabs

@BRKalow yes I did assess `@reach/tabs`! My original reason for not using it was because I think it does way more than we need, but today I'm finding that they do not implement `keyup` listeners, which I am also learning are necessary for accessibility reasons. I'm looking for shareable resources on this, but what it comes down to is what happens in the component when a key is held down. `keydown` events are triggered repeatedly when a key is held down, and `keyup` events are not. Handling this correctly is important for people that press a key long enough for multiple events to trigger, maybe due to something like a mobility issue. Does that all make sense? 😊

src/pages/_app.tsx
```diff
@@ -26,18 +27,20 @@ export default function App({ Component, pageProps }) {
...
-            {...pageProps?.layoutProps}
-            openConsentManager={openConsentManager}
-          >
-            <Component {...pageProps} />
-          </Layout>
-        </CurrentProductProvider>
-      </DeviceSizeProvider>
-      <ConsentManager />
-    </ErrorBoundary>
+    <SSRProvider>
```
ashleemboyer, line 30: Do we want this to wrap everything?

src/layouts/sidebar-sidecar/utils/_local_platform-docs-mdx/index.jsx
```diff
@@ -96,17 +96,11 @@ function EnterpriseAlert(props) {
...
-      <TabsBase className="g-tabs" defaultTabIdx={defaultTabIdx}>
-        {children}
-      </TabsBase>
-    </span>
-  )
+/**
+ * Because the `defaultTabIdx` prop is being renamed to `initialActiveIndex`,
+ * this wrapper ensures any use of `defaultTabIdx` is still supported in older
+ * versions of docs.
+ */
```
ashleemboyer, line 103: I'd like to rename the prop because `initialActiveIndex` is a little more clear that the index is only used on the initial load. The `default` portion of `defaultTabIdx` doesn't convey that same meaning, and I also want to remove the `Idx` abbreviation.

src/pages/_app.tsx
zchsh: Based on [the docs](https://react-spectrum.adobe.com/react-aria/SSRProvider.html), just a +1 that this looks right to me!

ashleemboyer:
TODO in the morning: write a unit test for the following claim

> Hitting the Tab key when a tab button is focused should bring focus to the active panel

ashleemboyer:
@BRKalow @zchsh FYI - just pushed two commits that update some styles using the `focus-ring-from-box-shadow` global helper class and updated `Tabs` to use it. The reason I hadn't used it before was because I couldn't compose it in the CSS class with the selectors on `role="tab"` within the `.root` block. This morning I remembered that I could just pass the global class name as a string to the component in the JSX. 😊

ashleemboyer:
> TODO in the morning: write a unit test for the following claim
> 
> > Hitting the Tab key when a tab button is focused should bring focus to the active panel

This is done now, PR is fully ready for review! ✅

src/components/tabs/tabs.test.tsx
```diff
@@ -0,0 +1,339 @@
...
+          const firstTabPanel = screen.queryByRole('tabpanel', {
+            name: testData[0].heading,
+          })
+          expect(firstTabPanel).toBeInTheDocument()
+        })
+      })
+
+      /**
+       * NOTE: `fireEvent.key(document.activeElement, { key: 'Tab' })` after
+       * focusing the first tab button does not actually hange focus from the
```
BRKalow, line 158:
```suggestion
       * focusing the first tab button does not actually change focus from the
```

BRKalow (approved):
👏

src/components/tabs/components/tab/index.tsx
```diff
@@ -0,0 +1,9 @@
+import { ReactElement } from 'react'
+import { TabProps } from './types'
+
+const Tab = (props: TabProps): ReactElement => {
```
zchsh, line 4: nit, but from what I can tell (and I might have the wrong idea), we never really render this component, right? If that's the case, maybe we can add a quick comment here:

```suggestion
/**
 * Note that we never actually render the `Tab` component,
 * we've set up this component for authoring convenience in MDX.
 */
const Tab = (props: TabProps): ReactElement => {
```

... or if I'm off on this, definitely feel free to drop this suggestion & resolve 👍

src/components/tabs/docs.mdx
```diff
@@ -0,0 +1,113 @@
...
+  <Tab heading="Mac OS">
+    <p>The family of Macintosh operating systems developed by Apple Inc. includes the graphical user interface-based operating systems it has designed for use with its Macintosh series of personal computers since 1984, as well as the related system software it once created for compatible third-party systems. <InlineLink href="https://en.wikipedia.org/wiki/Macintosh_operating_systems" text="Read more on Wikipedia" />.</p>
+  </Tab>
+  <Tab heading="Linux">
+    <p>Linux is a family of open-source Unix-like operating systems based on the Linux kernel,[12] an operating system kernel first released on September 17, 1991, by Linus Torvalds. <InlineLink href="https://en.wikipedia.org/wiki/Linux" text="Read more on Wikipedia" />.</p>
+  </Tab>
+</Tabs>
+`}</LiveComponent>
+
+## Examples
```
zchsh, line 40: Really dig seeing these examples, nicely done 💯

src/components/tabs/index.tsx
```diff
@@ -0,0 +1,204 @@
...
+  ariaLabel,
+  ariaLabelledBy,
+  children,
+  initialActiveIndex = 0,
+  showAnchorLine = true,
+}: TabsProps): ReactElement => {
+  /**
+   * Disallow rendering a Tabs component with only one Tab.
+   */
+  if (!Array.isArray(children)) {
```
zchsh, line 26: Totally non-blocking enhancement, but if we want we could validate that children of `Tabs` are all `Tab` components. We do [something like this in CodeTabs](https://github.com/hashicorp/react-components/blob/963d39ac51d63ff45a8a85ab078c7130502c972d/packages/code-block/partials/code-tabs/index.js#L47), which might be useful as a reference.

src/components/tabs/tabs.module.css
```diff
@@ -0,0 +1,55 @@
+/*
+NOTE: This file mainly uses selectors to apply styles because it ensures that
```
zchsh, line 2: I really like this approach 💖 we don't do it a ton in other components, I for one will definitely try to adopt this approach elsewhere 💯

src/pages/style.css
```diff
@@ -92,15 +92,10 @@ and negative margin.
...
-  /*
-  Not styling ::before with position: absolute and -5px positions because that
-  does not work on links that span more than one line.
-  */
   &:focus-visible {
     box-shadow: var(--token-focus-ring-box-shadow);
-    margin: -5px;
     outline: none;
-    padding: 5px;
+    z-index: 1;
```
zchsh, line 98: Minor nit, but will `z-index` here also need a non-default / non-`static` `position` in order to do anything? eg should we add `position: relative` to `.focus-ring-from-box-shadow` to ensure this applies? (Or should we apply `z-index` in the specific instances where it felt needed rather than for all `.focus-ring-from-box-shadow`?)

zchsh (approved):
💯 🚀  Looks great to me! Super impressed by the thoroughness 💖 

Had a couple of very minor, non-blocking nit things. Can definitely skip them & merge unless any of them really speak to ya 👍

src/components/tabs/components/tab/index.tsx
ashleemboyer: This is a good idea! I think adding a `console.warn` inside the component might be good as well, I'm going to test that real quick before adding. But will definitely add the comment above the component!

src/components/tabs/index.tsx
ashleemboyer: I think this is a great idea, I'll take a look at that example and try it out!

src/pages/style.css
ashleemboyer: Good call, I think I'll move `z-index`. We have some `absolute` positioned stuff that will eventually use this class (or some version of it) so setting `position: relative` will probably be too heavy of an assumption to make. I'll make this change as well!

src/components/tabs/components/tab/index.tsx
ashleemboyer: Updated in: https://github.com/hashicorp/dev-portal/pull/104/commits/7f0bcd5d2d1c67a0b6a34b803b55bcda798f3d8b

src/components/tabs/index.tsx
ashleemboyer: Updated in: https://github.com/hashicorp/dev-portal/pull/104/commits/5477198bd086bfc3789aa44464bd98481e90eded

src/pages/style.css
ashleemboyer: Updated in: https://github.com/hashicorp/dev-portal/pull/104/commits/89a70fb34e5c240c23182986fc78f570fc047457

zchsh (approved):
🚀

PR #109, opened 2022-02-11
feat: add basic config, limit static paths

BRKalow:

---

## Relevant links

- [Preview link](https://vercel.com/hashicorp/dev-portal/8LKTABJFSW342T7oKXNJBbNx3nH4) 🔎
- [Asana task](https://app.asana.com/0/1100423001970639/1201745229608930/f) 🎟️

## What

Introducing a lightweight config pattern, which makes config values available via a global `__config` object. Building off of that, I added an `__config.io_sites.max_static_paths` value, which I used in a wrapper around `getStaticGenerationFunctions` to ensure that we only render 10 static paths per dynamic docs route.

Mini-RFC for the config approach: https://www.notion.so/hashicorp/dev-portal-application-config-d39ba8dff9924edf84e8d67d7229c3c2

## Why

Our deploy preview build times were a bit slow (5m), with this change they are less than 2m. This will speed up iterations and reduce unnecessary load on our content API.

Speed up build times.

## Testing

- [ ] Observe build here: https://vercel.com/hashicorp/dev-portal/8LKTABJFSW342T7oKXNJBbNx3nH4
- [ ] See that only 160 static paths were generated

### DIFF
--- a/config/development.json
+++ b/config/development.json
@@ -1,0 +1,6 @@
+{
+  "io_sites": {
+    "max_static_paths": 0,
+    "revalidate": 360
+  }
+}
--- a/next.config.js
+++ b/next.config.js
@@ -1,3 +1,5 @@
+const fs = require('fs')
+const path = require('path')
 const withHashicorp = require('@hashicorp/platform-nextjs-plugin')
 const withSwingset = require('swingset')
 const redirectsConfig = require('./config/redirects')
@@ -14,6 +16,19 @@ const temporary_hideDocsPaths = {
     },
   ],
   has: [{ type: 'host', value: '(^(?!.*(waypointproject)).*$)' }],
+}
+
+function getHashiEnvironmentConfig() {
+  const env = process.env.HASHI_ENV || 'development'
+  try {
+    const envConfig = JSON.parse(
+      fs.readFileSync(path.join(process.cwd(), 'config', `${env}.json`))
+    )
+    return envConfig
+  } catch (err) {
+    console.log('Error loading environment config:', err)
+    return {}
+  }
 }
 
 module.exports = withSwingset({ componentsRoot: 'src/components/*' })(
@@ -47,6 +62,7 @@ module.exports = withSwingset({ componentsRoot: 'src/components/*' })(
       ENABLE_VERSIONED_DOCS: process.env.ENABLE_VERSIONED_DOCS || false,
       IS_CONTENT_PREVIEW: process.env.IS_CONTENT_PREVIEW,
       DEV_IO: process.env.DEV_IO,
+      ...getHashiEnvironmentConfig(),
     },
     svgo: {
       plugins: [
--- a/next.config.js
+++ b/next.config.js
@@ -1,5 +1,7 @@
 const fs = require('fs')
 const path = require('path')
+const flat = require('flat')
+const { webpack } = require('next/dist/compiled/webpack/webpack')
 const withHashicorp = require('@hashicorp/platform-nextjs-plugin')
 const withSwingset = require('swingset')
 const redirectsConfig = require('./config/redirects')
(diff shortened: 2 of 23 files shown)


next.config.js
```diff
@@ -47,6 +62,7 @@ module.exports = withSwingset({ componentsRoot: 'src/components/*' })(
       ENABLE_VERSIONED_DOCS: process.env.ENABLE_VERSIONED_DOCS || false,
       IS_CONTENT_PREVIEW: process.env.IS_CONTENT_PREVIEW,
       DEV_IO: process.env.DEV_IO,
+      ...getHashiEnvironmentConfig(),
```
dstaley, line 65: I have a bit of trepidation about exposing a configuration as a JSON file that _could_ contain deeply-nested values when it's going to get spread into the `env`. What do you think about using `dotenv.parse()` to parse a ``${env}.env`` file instead (or ``.env.${env}`` if you wanted to be consistent with the naming convention used by Next)? That way environment variable configuration is consistent between data in `config/` and files like `.env.local`.

dstaley: We could add support for nested keys right off the bat by using something like [flat](https://replit.com/@dstaley/BuoyantExternalLicenses#index.js). If we come up with an alternative in the future that allows for traversing the object in user code, we could implement these side-by-side as a transition period.

src/lib/_proxied-dot-io/get-static-generation-functions.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  return {
+    async getStaticPaths(ctx) {
+      const result = await getStaticPathsBase(ctx)
+
+      return {
+        ...result,
+        paths: result.paths.slice(
+          0,
+          Number.parseInt(process.env.IO_SITES_MAX_STATIC_PATHS, 10)
```
zchsh, line 19: What happens here if `IO_SITES_MAX_STATIC_PATHS` isn't defined? (eg since `config/development.json` is empty at the moment, that might happen if `HASHI_ENV` isn't set, I think?). At a glance it looks like we might end up with `NaN` as the second arg to `result.paths.slice`, but not sure 🤔

next.config.js
zchsh: 👍 on both fronts - I dig the possibility of having nested values in config, and dig `flat` as a solution to enable that. I like the idea of the nice safe feeling we'd have if we account for nesting off the bat - either by using `flat`, or by doing some validation after `JSON.parse` in `getHashiEnvironmentConfig` to ensure the config object isn't deeply nested.

zchsh (commented):
This seems great for local deploy previews of `dev-portal` work 💯 

Only potentially blocking thing, and might just be my lack of understanding: would there be a separate `config` for deploy previews in content repos, eg if I were to run a deploy preview while working in `vault`? I might just be out of the loop, but wanted to double-check that context is accounted for.

next.config.js
```diff
@@ -16,6 +20,44 @@ const temporary_hideDocsPaths = {
...
+    ...Object.fromEntries(
+      Object.entries(getHashiConfig(envConfigPath)).map(([key]) => {
+        return [
+          `config.${key}`,
+          webpack.DefinePlugin.runtimeValue(
+            () => {
+              return JSON.stringify(getHashiConfig(envConfigPath)[key])
+            },
+            { fileDependencies: [envConfigPath] }
+          ),
```
BRKalow, line 54: This allows the values to update during development. It's a little thing, but a nice QOL improvement for iterating locally.

(note: new values won't get added, but existing values will update when the config file changes)

next.config.js
```diff
@@ -16,6 +20,44 @@ const temporary_hideDocsPaths = {
...
+        safe: true,
+      })
+      return ret
+    } catch (err) {
+      console.log('Error loading environment config:', err)
+      return {}
+    }
+  }
+
+  return new webpack.DefinePlugin({
```
BRKalow, line 44: I'm opting to use `DefinePlugin` for the first iteration because this injects the literal value into the code, which enables dead code elimination.

So if we use this for flags:
```js
if (config.flags.enable_new_downloads_page) {
  return <NewDownloadsPage />
} else {
  return <OldDownloadsPage />
}
```

the value would get replaced and the dead code path would get removed:
```js
if (true) {
  return <NewDownloadsPage />
} else {
  // this code path is removed from the bundle
  return <OldDownloadsPage />
}
```

config/development.json
```diff
@@ -0,0 +1,6 @@
+{
+  "io_sites": {
+    "max_static_paths": 0,
+    "revalidate": 360
+  }
+}
```
BRKalow, line 6: I'm not sure why I prefer snake_case for configs, but it's a strong opinion loosely held so happy to convert to camelCase 😄

zchsh: +1, also loosely held, I think I just dig the `CONST` energy 💯

dstaley (approved):

PR #117, opened 2022-02-18
Set up Waypoint downloads page with data

ashleemboyer:

---

## Relevant links

- [Preview link](https://dev-portal-git-ambset-up-waypoint-downloads-with-data-hashicorp.vercel.app/waypoint/downloads) 🔎

## What/Why

- Adds `@types/semver`
- Moves Sidebar interfaces to a new `types.ts` file local to the component. This is a practice we decided on after the component was made so this was just cleanup.
- Adds a `showFilterInput` boolean prop to `Sidebar` to conditionally render the `SidebarFilterInput`. This was done because the new install view doesn't need the filter input in the sidebar.
  - `SidebarSidecarLayout` also now has the same boolean prop, it passes it down to `Sidebar`.
- Removes the `version` prop from `src/data/waypoint.json` since the latest version can be determined automatically now with the releases helper.
- Checks out `src/lib/fetch-release-data.ts` and `src/lib/fetch-with-retry.ts` from the `main` branch and updates them to work with with a `Product` object instead of just a product slug
- Moves placeholder data from `src/pages/waypoint/downloads/index.tsx` and moves it to `src/views/product-downloads-view/index.tsx`
- Adds a skateboard version of the version select to the content of the new `ProductDownloadsView` component

## How

Nothing in particular to add here that's not a typical React strategy (like `useState` for the version select).

## Testing

- [ ] Go to [the Waypoint downloads page](https://dev-portal-git-ambset-up-waypoint-downloads-with-data-hashicorp.vercel.app/waypoint/downloads)
- [ ] The `ProductSwitcher` in the navigation header should still show the correct information
- [ ] There should be a version select like the following, showing the latest version:
  image
- [ ] Change the version of the select
- [ ] The page title and the select should update and show the correct version

## Anything else?

Not much, the real version select coming soon after this!


src/lib/fetch-release-data.ts
```diff
@@ -0,0 +1,96 @@
...
+    })
+    .catch(() => {
+      throw new Error(
+        `--------------------------------------------------------
+        Unable to resolve release data on releases.hashicorp.com from link
+        <https://releases.hashicorp.com/${productSlug}/index.json>.
+        ----------------------------------------------------------`
+      )
+    })
+}
```
BRKalow, line 96: The change to passing the full product object here diverges the implementation and is something we'll need to address when merging the assembly branch back into main, either updating all of the usages or supporting either a slug or a product.

What do you think about supporting either a slug OR a product object?

BRKalow (commented):
Looks good!

src/lib/fetch-release-data.ts
ashleemboyer: We actually _need_ a full `product` object returned from this function so that the `CurrentProductContext` will be initialized correctly in `_app.tsx`, so I think this will only work if an object is passed as `product`. Do we add a TODO somewhere or an Asana task for now to make sure this divergence is addressed correctly when merging the UI assembly branch?

BRKalow: The io sites don't need/use the `CurrentProductContext` at the moment, so supporting the full product object or a slug here would let us have both implementations.

ashleemboyer: Will we have both cases running at the same time? I don't think we've decided when we're going to merge the assembly branch yet, so I'm not sure on that.

BRKalow: The io code is structured so that it can live alongside the UI work on the same branch, but yeah I guess we haven't made a definitive call on it yet. It depends on how we're going to be rolling out the refreshed UI I suppose. It's a possibility.

It seems like if we can support both a string or the product object we can avoid headache later. Or do you see downsides to supporting both?

ashleemboyer: I think long term we shouldn't support both because it makes the behavior more ambiguous, but to avoid headaches in the short term we can do both. I'll make the update and also add a TODO comment so we don't lose track of the long term goal.

ashleemboyer: Updated in https://github.com/hashicorp/dev-portal/pull/117/commits/d47455687d53e00c30b8a6ec621372c72512c45e!

BRKalow (approved):
:shipit:

PR #119, opened 2022-02-18
Refactor SidebarSidecarLayout props

ashleemboyer:

---

## Relevant links

- [Preview link](https://dev-portal-git-ambrefactor-sidebar-sidecar-lay-c2c408-hashicorp.vercel.app/) 🔎

## What

- Refactors some of `SidebarSidecarLayoutProps` into a new `sidebarProps` prop
- Makes a few TypeScript improvements on components for clarity, those have been annotated as PR comments
- Adds `types.ts` file for `SidebarBackToLink`, bringing up to our latest practice for components
- Renames the `backToLink` property in `SidebarProps` to `backToLinkProps` for clarity and alignment with the propose `sidebarProps` property in `SidebarSidecarLayoutProps`

## Why

Parent component props that are objects representing props sent to a child component are sometimes contentious, but I see two main benefits to using them:

- It's clearer how the properties in the object are being used and what child component they're rendered in if the object is clearly named. This can potentially help speed up debugging.
- It's less maintenance on the API of the parent component. Instead of having to add a prop to the parent component every time a prop is added to the child component, all we have to do is add another line in the props object being passed to the child.

## How

Changes here are fairly straightforward. Added a new `types.ts` file as we have done for other components, and other changes were mainly moving existing code around.

## Testing

- [ ] product landing pages should no longer show a filter input in the sidebar (past design request)
- [ ] product landing pages should should still work as expected
- [ ] docs pages should still work as expected
- [ ] install views should still work as expected

## Anything else?

Nope!


src/components/sidebar/components/sidebar-back-to-link/index.tsx
```diff
@@ -26,7 +33,7 @@ const SidebarBackToLink: React.FC<{
   )
 }
 
-const PlaceholderBackToLink: React.FC = () => {
+const PlaceholderBackToLink = (): ReactElement => {
```
ashleemboyer, line 36: Made this update to more accurately reflect what's happening. `React.FC` means that the `children` prop is passed to the component, but that's not happening here.

src/components/sidebar/components/sidebar-back-to-link/index.tsx
```diff
@@ -10,14 +12,19 @@ const BackToLinkText: React.FC = ({ children }) => (
...
-  if (!text || !url) return <PlaceholderBackToLink />
+/**
+ * TODO: Design is still planning this functionality & states. Since not all
+ * cases are covered, we return a placeholder link if necessary props aren't
+ * provided.
+ */
+const SidebarBackToLink = ({
+  text,
+  url,
+}: SidebarBackToLinkProps): ReactElement => {
```
ashleemboyer, line 23: This update of the addition of `SidebarBackToLinkProps` brings the component up to speed with how we've been building components lately: adding a `types.ts` file next to the component that contains all the types for the component, including the props interface for the component.

src/components/sidebar/index.tsx
```diff
@@ -88,12 +88,12 @@ const getFilteredMenuItems = (items: MenuItem[], filterValue: string) => {
...
 
-const Sidebar: React.FC<SidebarProps> = ({
+const Sidebar = ({
+  backToLinkProps = {},
   menuItems,
-  backToLink = {},
   showFilterInput = true,
   title,
-}) => {
+}: SidebarProps): ReactElement => {
```
ashleemboyer, line 96: Same case here, using `ReactElement` instead of `React.FC` since this component does not accept a `children` prop.

src/components/sidebar/types.ts
```diff
@@ -22,10 +24,7 @@ export interface MenuItem {
 
 export interface SidebarProps {
   /** Optional { text, url } to use for the "← Back to..." link at the top of the sidebar */
-  backToLink?: {
-    text: string
-    url: string
-  }
+  backToLinkProps?: SidebarBackToLinkProps
```
ashleemboyer, line 27: Referencing newly created `SidebarBackToLinkProps` as the source of truth for this property.

ashleemboyer:
Note to self: check that this file merges correctly after #167 is in `assembly-ui-v1`:

- `src/layouts/sidebar-sidecar/server.ts`

zchsh (approved):
✅ Looks great! Apologies for those merge conflicts, I just tried to resolve those, and assuming that doesn't break anything, this looks good to go to me 🚀 So much cleaner! ✨ 💯

ashleemboyer:
Thanks for the review and thanks for handling those merge conflicts @zchsh! 💖

PR #123, opened 2022-02-22
feat: introduce app config for the UI branch

BRKalow:

---

## Relevant links

- [Preview link](url) 🔎
- [Asana task](url) 🎟️

## What

Introduces the new app configuration implementation to the UI branch. Access values defined in the environment configs in `config/` via the global `__config`.

## Why

This will allow us to define global config values which we can use throughout the app! They will get replaced at build time with the associated value, so it will also unlock some new patterns leveraging dead code elimination to exclude certain code paths from the production build.

## How

Using webpack's [DefinePlugin](https://webpack.js.org/plugins/define-plugin/), we can statically replace each reference to a config value with its literal value. The `HASHI_ENV` environment variable controls which environment file is loaded at build time.

## Testing

See [the original PR](https://github.com/hashicorp/dev-portal/pull/109) for some usages. Not a huge need to test here as it's already been validated.

## Anything else?

We can also use this for feature flags! Example:

In `config/production.json`:
```json
{
  "flags": {
    "enable_new_downloads_view": false
  }
}
```

In `config/preview.json`:
```json
{
  "flags": {
    "enable_new_downloads_view": true
  }
}
```

In `config/development.json`:
```json
{
  "flags": {
    "enable_new_downloads_view": true
  }
}
```

Usage:
```js
export default function DownloadsPage() {
  if (__config.flags.enable_new_downloads_view) {
    return <NewDownloadsView />
  } else {
    return <OldDownloadsView />
  }
}

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,31 @@ SomePage.layout = SomeLayout
 export default SomePage
 ```
 
+### Configuration
+
+Per-environment configuration values are defined in JSON files in the `config/` folder. Each environment has its own config file, currently:
+
+```
+config/
+  development.json
+  preview.json
+  production.json
+```
+
+The configuration values are available globally within the application. They can be accessed from a global `__config` object:
+
+```js
+// config file:
+{
+   "my_config_value": "foo"
+}
+
+// in code:
+console.log(__config.my_config_value)
+```
+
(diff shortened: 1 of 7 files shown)


ashleemboyer (commented):
This looks great! Will do a PR in the morning for setting up a feature flag on the new downloads view. 😊

ashleemboyer (approved):

PR #124, opened 2022-02-22
Add DownloadStandaloneLink component

ashleemboyer:

---

## Relevant links

- [Preview link](https://dev-portal-git-ambadd-download-link-component-hashicorp.vercel.app/swingset/components/downloadstandalonelink) 🔎
- [Asana task](https://app.asana.com/0/1201010428539925/1201654639085730/f) 🎟️

## What

- Adds a new component with a Swingset docs page called `DownloadStandaloneLink`
- Adds two new props to `StandaloneLink`: `ariaLabel` and `download`

## Why

- This will be a highly repeated element on the new install page. Having it packaged up in a single reusable component will save some time and repeated code.
- The new component is built on top of the `StandaloneLink` component because there is a lot of unique styling handled for us by the main component and because the designs specify this relationship.
- The `ariaLabel` prop was added to `StandaloneLink` so that an accessible label could be applied to `DownloadStandaloneLink`s, which are otherwise not presented with their full context when brought into focus with just the Tab key and a screen reader.
- The `download` prop was added to `StandaloneLink` so that there is a way to set the `download` property of the final `<a>` element that gets rendered. This property is an added indication that a link is meant to be downloaded and not viewed. [As far as I currently know, this property does not affect screen reader output.](https://www.maxability.co.in/2018/06/24/html5-download-attribute/)

## How

Nothing extra to note here. Just built on top of another component and added some props where needed.

### Mouse testing

- [ ] Go to [the new Swingset page for `DownloadStandaloneLink`](https://dev-portal-git-ambadd-download-link-component-hashicorp.vercel.app/swingset/components/downloadstandalonelink)
- [ ] Click the Download link that's rendered in the Playground section
- [ ] The correct file should download

### Screen reader testing

- [ ] Start VoiceOver
- [ ] Go to [the new Swingset page for `DownloadStandaloneLink`](https://dev-portal-git-ambadd-download-link-component-hashicorp.vercel.app/swingset/components/downloadstandalonelink)
- [ ] Bring the Download link into focus with the keyboard
- [ ] The link should be presented as "link, Download waypoint 0.7.1 darwin amd64"

## Anything else?

It's possible we'll want to add an `ariaLabelledBy` prop down the road so that we can add a visual label and associate it with the download link in the future. If that becomes the case, we should add it and require either `ariaLabel` or `ariaLabelledBy`, but not allow both. I have not added that here because we don't need it at the moment.


src/components/standalone-link/types.ts
```diff
@@ -13,6 +21,14 @@ export interface StandaloneLinkProps {
    */
   color?: 'primary' | 'secondary'
 
+  /**
+   * Same as the <a> element's download prop. Passed directly to the internal
+   * link element.
+   *
+   * ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-download
+   */
+  download?: boolean | string
```
BRKalow, line 30: til about the `download` attribute! 🔖

BRKalow (approved):
LGTM! Nice work

PR #130, opened 2022-02-23
Add install view DownloadsSection

ashleemboyer:

---

## Relevant links

- [Preview link](https://dev-portal-git-ambadd-waypoint-downloads-tabbed-card-hashicorp.vercel.app/waypoint/downloads) 🔎

## What/Why

- Adds new `DownloadsSection` portion of the page that handles the whole downloads card where links and information can be changed based on the version selected from the switcher.
- Adds `textSize` prop to `DownloadStandaloneLink` with a default value of `200`. Can undo these changes if we feel it's unneeded, but since the component is in the main `components` directory (readily available for reuse), I figured it would be a helpful prop for future-proofing.
- Adds `openInNewTab` prop to `StandaloneLink`. There are a few instances of pairing an external link icon with a `StandaloneLink` in the new install view, so this seemed like a useful addition in those cases.
- Moved `packageManagers` property from `src/data/waypoint.json` to `src/data/waypoint-install.json` since the latter file is new and intended to house data that's only used in the new install view. This property also has the latest data according to 0.7.2 at the time of writing: https://github.com/hashicorp/waypoint/blob/v0.7.2/website/data/version.js.
- Adds `sortPlatforms` and `prettyOs` helpers to `src/views/product-downloads-view/helpers.ts`

## How

A few things, such as helpers, are directly copied from `react-components/product-download-page`. Otherwise, this work was accomplished by using some foundational components worked on over the past several weeks. I also tried to split things up into separate files to prevent:

- a single, overwhelmingly long file
- long and difficult to parse markup

## Testing

- [ ] Go to [/waypoint/downloads on the preview link](https://dev-portal-git-ambadd-waypoint-downloads-tabbed-card-hashicorp.vercel.app/waypoint/downloads)
- [ ] The downloads section will look something like this with the macOS tab highlighted and "Package manager for macOS", "Binary download for macOS", "Release information", and "Notes" sections within:
  image
- [ ] Ensure all version labels and links have the same version as selected in the version switcher
- [ ] Pick a different version in the switcher
- [ ] Ensure all version labels and links update to the correct version
- [ ] Ensure the package manager section never shows if the version selected is not the latest
- [ ] Select the latest version from the version switcher
- [ ] Select the Windows tab
- [ ] Ensure the information displayed is correct
- [ ] Select the Linux tab
- [ ] Interact with the package manager tabs
- [ ] Ensure information displayed is correct

## Anything else?

We want to use `CodeTabs` where there are multiple package managers for an OS, but there is a bug that is preventing it. For now, we're using the regular `Tabs` component and have a `TODO` in the code noting the technical debt and the Asana task for the bug we've identififed.

The version switcher belongs inside of this card, but that's a more complex change to make so that the page subtitle under "Install Waypoint" reflects the correct version. I want to spend a little more time thinking on this and whether or not it's worth adding something like a local context for any area of the page to have access to the currently selected version. Happy to hear more thoughts on this idea in comments!

### DIFF
--- a/src/pages/waypoint/downloads/index.tsx
+++ b/src/pages/waypoint/downloads/index.tsx
@@ -1,6 +1,7 @@
 import { ReactElement } from 'react'
 import { GetStaticProps } from 'next'
 import waypointData from 'data/waypoint.json'
+import installData from 'data/waypoint-install.json'
 import { Product } from 'types/products'
 import { generateStaticProps, GeneratedProps } from 'lib/fetch-release-data'
 import EmptyLayout from 'layouts/empty'
@@ -11,7 +12,11 @@ const WaypointDownloadsPage = (props: GeneratedProps): ReactElement => {
   if (__config.flags.enable_new_downloads_view) {
     const { latestVersion, releases } = props
     return (
-      <ProductDownloadsView latestVersion={latestVersion} releases={releases} />
+      <ProductDownloadsView
+        latestVersion={latestVersion}
+        pageContent={installData}
+        releases={releases}
+      />
     )
   } else {
     return <PlaceholderDownloadsView />
--- a/src/views/product-downloads-view/components/downloads-section/index.tsx
+++ b/src/views/product-downloads-view/components/downloads-section/index.tsx
@@ -1,19 +1,13 @@
 import { ReactElement, useMemo } from 'react'
 import CodeBlock from '@hashicorp/react-code-block'
-import { ReleaseVersion } from 'lib/fetch-release-data'
 import Card from 'components/card'
 import Heading from 'components/heading'
-import { PackageManager } from 'views/product-downloads-view/types'
 import { sortPlatforms, prettyOs } from 'views/product-downloads-view/helpers'
 import Tabs, { Tab } from 'components/tabs'
 import Text from 'components/text'
 import s from './downloads-section.module.css'
 import DownloadStandaloneLink from 'components/download-standalone-link'
-
-interface DownloadsSectionProps {
-  packageManagers: PackageManager[]
-  selectedRelease: ReleaseVersion
-}
+import { DownloadsSectionProps } from './types'
 
 const DownloadsSection = ({
   packageManagers,
--- a/src/views/product-downloads-view/components/downloads-section/index.tsx
+++ b/src/views/product-downloads-view/components/downloads-section/index.tsx
@@ -1,4 +1,5 @@
 import { ReactElement, useMemo } from 'react'
+import classNames from 'classnames'
 import { IconExternalLink16 } from '@hashicorp/flight-icons/svg-react/external-link-16'
 import CodeBlock from '@hashicorp/react-code-block'
 import { PackageManager } from 'views/product-downloads-view/types'
@@ -159,8 +160,11 @@ const NotesSection = ({ selectedRelease }) => {
     <>
       <Heading
         {...SHARED_HEADING_LEVEL_3_PROPS}
+        className={classNames(
+          SHARED_HEADING_LEVEL_3_PROPS.className,
+          s.specialSubHeading
+        )}
         slug="notes"
-        style={{ marginBottom: 8 }}
       >
         Notes
       </Heading>
--- a/src/views/product-downloads-view/components/downloads-section/index.tsx
+++ b/src/views/product-downloads-view/components/downloads-section/index.tsx
@@ -8,6 +8,7 @@ import { sortPlatforms, prettyOs } from 'views/product-downloads-view/helpers'
 import Card from 'components/card'
 import DownloadStandaloneLink from 'components/download-standalone-link'
 import Heading from 'components/heading'
+import InlineLink from 'components/inline-link'
 import Tabs, { Tab } from 'components/tabs'
 import Text from 'components/text'
 import StandaloneLink from 'components/standalone-link'
@@ -95,7 +96,6 @@ const DownloadsSection = ({
                       tabs={packageManagers.map(({ label }) => label)}
                     >
                       {packageManagers.map((packageManager) => {
-                        console.log(packageManager)
                         return (
                           <CodeBlock
                             key={packageManager.label}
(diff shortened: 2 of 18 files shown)


src/pages/waypoint/downloads/index.tsx
```diff
@@ -13,8 +13,8 @@ const WaypointDownloadsPage = (props: GeneratedProps): ReactElement => {
     const { latestVersion, releases } = props
     return (
       <ProductDownloadsView
-        pageContent={installData}
         latestVersion={latestVersion}
+        pageContent={installData}
         releases={releases}
       />
```
ashleemboyer, line 19: This change is just alphabetizing the props.

src/views/product-downloads-view/index.tsx
```diff
@@ -14,12 +14,13 @@ import {
...
 } from './helpers'
 import { ProductDownloadsViewProps } from './types'
+import DownloadsSection from './components/downloads-section'
 import s from './product-downloads-view.module.css'
 
 const ProductDownloadsView = ({
   latestVersion,
-  releases,
   pageContent,
+  releases,
```
ashleemboyer, line 23: This is an alphabetization change.

PS alphabetization is apparently a word?

src/views/product-downloads-view/index.tsx
```diff
@@ -83,25 +84,23 @@ const ProductDownloadsView = ({
...
+            <option key={option.value} value={option.value}>
+              {option.label}
+            </option>
+          ))}
+        </select>
+      </div>
+      <DownloadsSection
+        packageManagers={pageContent.packageManagers}
+        selectedRelease={releases.versions[selectedVersion]}
+      />
```
ashleemboyer, line 103: This is the main change in this file other than adding a `<div>` around the temp switcher for styling purposes.

src/views/product-downloads-view/product-downloads-view.module.css
```diff
@@ -9,6 +9,7 @@
 }
 
 .pageHeaderTitle {
+  color: var(--token-color-neutral-foreground-primary);
```
ashleemboyer, line 12: Something I missed before.

src/views/product-downloads-view/components/downloads-section/index.tsx
```diff
@@ -0,0 +1,259 @@
...
+  groupPackageManagersByOS,
+} from './helpers'
+import s from './downloads-section.module.css'
+
+const SHARED_HEADING_LEVEL_3_PROPS = {
+  className: s.subHeading,
+  level: 3 as HeadingProps['level'],
+  size: 200 as HeadingProps['size'],
+  weight: 'semibold' as HeadingProps['weight'],
+}
```
BRKalow, line 27: I wonder if this would work 💭 
```suggestion
const SHARED_HEADING_LEVEL_3_PROPS: Partial<HeadingProps> = {
  className: s.subHeading,
  level: 3,
  size: 200,
  weight: 'semibold',
}
```

src/views/product-downloads-view/components/downloads-section/index.tsx
```diff
@@ -0,0 +1,259 @@
...
+        )}
+        slug="notes"
+      >
+        Notes
+      </Heading>
+      <Text size={200}>
+        You can find the{' '}
+        <InlineLink
+          href={`https://releases.hashicorp.com/${name}/${version}/${shasums}`}
+          text={`SHA256 checksums for Waypoint ${version}`}
```
BRKalow, line 175: Is the plan to address some of the hardcoded product name usages later on?

src/views/product-downloads-view/components/downloads-section/index.tsx
```diff
@@ -0,0 +1,259 @@
...
+    selectedRelease,
+  ])
+  const packageManagersByOS = useMemo(
+    () => groupPackageManagersByOS(packageManagers),
+    [packageManagers]
+  )
+
+  return (
+    <article>
+      <Card elevation="base">
```
BRKalow, line 211: I wonder if it'd be worth it to allow `Card` to render `article` under the hood 💭

src/views/product-downloads-view/components/downloads-section/index.tsx
```diff
@@ -0,0 +1,259 @@
...
+          {Object.keys(downloadsByOS).map((os) => {
+            const packageManagers: PackageManager[] = packageManagersByOS[os]
+            const prettyOSName = prettyOs(os)
+
+            /**
+             * TODO: it might be nice to introduce a local Context here with all
+             * the information needed so that these helper components don't have
+             * APIs that could potentially require changes with every visual
+             * change.
+             */
```
BRKalow, line 231: 👍

BRKalow (commented):
Nothing block here, looking good

BRKalow (approved):
Oops I meant to approve 😄

src/views/product-downloads-view/components/downloads-section/index.tsx
ashleemboyer: Oops! Great catch. This should already be using `name` instead of `Waypoint`. Will fix it!

src/views/product-downloads-view/components/downloads-section/index.tsx
ashleemboyer: I've not seen this before! I'll give it a try and see what happens

src/views/product-downloads-view/components/downloads-section/index.tsx
ashleemboyer: Personally I think that's too heavy of an assumption for `Card` to make, especially since components like `CardLink` are built on top of it. If we made `Card` render an `article`, then `CardLink` would end up rendering an `article` element within an `a` element and I think we probably don't always want that as the default.

src/views/product-downloads-view/components/downloads-section/index.tsx
ashleemboyer: No luck on that one. Still getting the error I was resolving with all the type casting.

image

src/views/product-downloads-view/components/downloads-section/index.tsx
BRKalow: Yeah, sorry I didn't mean by default, but with something like `asElement`

src/views/product-downloads-view/components/downloads-section/index.tsx
ashleemboyer: Fixed in https://github.com/hashicorp/dev-portal/pull/130/commits/0567f6247f916b98c25af7c7bb81e863cdd7214c

src/views/product-downloads-view/components/downloads-section/index.tsx
ashleemboyer: Ah, I see! I think that could potentially be really useful!

PR #131, opened 2022-02-24
feat: add consul docs routes

zchsh:

---

## Relevant links

- [Preview link][docs] 🔎
- [Asana task](https://app.asana.com/0/1100423001970639/1201647459964504/f) 🎟️

## What

This PR adds `docs-page` routes for `/consul`.

## Why

To iterate towards supporting all product docs content within the `dev-portal` project.

## How

- Re-uses existing templates, as used for `vault` and `waypoint`
- Builds on work done for `vault` to rewrite `/consul/api/*` links to `/consul/api-docs/*` with a `remark` plugin
    - We have [a related task for Vault](https://app.asana.com/0/1100423001970639/1201655704127120/f) that details a longer-term fix - for example, doing this transform as part of our content pipeline.

## Testing

- [ ] Navigate to the docs views, click around and confirm behaviour is expected
    - [ ] [/consul/docs][docs]
    - [ ] [/consul/api-docs][api-docs]
    - [ ] [/consul/commands][commands]

## Anything else?

One known issue is that on the [/consul/docs][docs] view, the links in the sidebar to `API` and `Commands (CLI)`, with the "external" icons, did not initially function as expected, as they reference absolute URLs as organized on `consul.io`.

To address this issue, I've modified the `prepareNavNodeForClient` function. Previously this function did not touch the `href` of `NavDirectLink` nodes. This PR changes this - now, `NavDirectLink` nodes will be tested to see if they're absolute, and:
- If absolute, eg `https://www.example.com`, will be unchanged
- If not absolute, eg `/api-docs`, will be prefixed with the `product.slug`. In this case specifically, `/api-docs` becomes `/consul/api-docs`.

[docs]: https://dev-portal-git-zsadd-consul-docs-views-hashicorp.vercel.app/consul/docs
[api-docs]: https://dev-portal-git-zsadd-consul-docs-views-hashicorp.vercel.app/consul/api-docs
[commands]: https://dev-portal-git-zsadd-consul-docs-views-hashicorp.vercel.app/consul/commands

### DIFF
--- a/src/data/consul-landing.json
+++ b/src/data/consul-landing.json
@@ -15,7 +15,7 @@
       "text": "In this guide we cover what Consul is, what problems it can solve, how it compares to existing software, and how you can get started using it.",
       "link": {
         "text": "Learn More",
-        "url": "https://www.consul.io/docs/intro"
+        "url": "/consul/docs/intro"
       }
     },
     {
@@ -26,25 +26,25 @@
           "icon": "IconDocs",
           "heading": "Consul Reference Documentation",
           "text": "The documentation is reference material for all available features and options of Consul",
-          "url": "https://www.consul.io/docs"
+          "url": "/consul/docs"
         },
         {
           "icon": "IconTerminal",
           "heading": "Consul CLI",
           "text": "Consul is controlled via a very easy to use command-line interface (CLI).",
-          "url": "https://www.consul.io/commands"
+          "url": "/consul/commands"
         },
         {
           "icon": "IconServer",
           "heading": "Consul API",
           "text": "The main interface to Consul is a RESTful HTTP API. The API can perform basic CRUD operations on nodes, services, checks, configuration, and more.",
-          "url": "https://www.consul.io/api-docs"
+          "url": "/consul/api-docs"
         },
         {
           "icon": "IconDownload",
           "heading": "Consul Install",
           "text": "Please download the proper package for your operating system and architecture.",
-          "url": "https://www.consul.io/downloads"
+          "url": "/consul/downloads"
         }
       ]
     },
@@ -61,17 +61,17 @@
         {
           "heading": "Connect",
           "text": "Consul Connect provides service-to-service connection authorization and encryption...",
-          "url": "https://www.consul.io/docs/connect"
+          "url": "/consul/docs/connect"
         },
         {
           "heading": "Services",
           "text": "One of the main goals of service discovery is to provide a catalog of available services. ",
-          "url": "https://www.consul.io/docs/discovery/services"
+          "url": "/consul/docs/discovery/services"
         },
         {
           "heading": "Configuration",
           "text": "The agent has various configuration options that can be specified via the...",
-          "url": "https://www.consul.io/docs/agent/options"
+          "url": "/consul/docs/agent/options"
         }
       ]
     },
@@ -88,17 +88,17 @@
         {
           "heading": "Etiam quis in ac non quam.",
           "text": "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Ultricies nisi interdum vitae id.",
-          "url": "https://www.consul.io"
+          "url": "/consul/docs"
         },
         {
           "heading": "Aliquet eu risus dis sed.",
           "text": "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sit neque mattis phasellus quam.",
-          "url": "https://www.consul.io"
+          "url": "/consul/docs"
         },
         {
           "heading": "Ornare faucibus blandit id.",
           "text": "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Urna donec gravida viverra.",
-          "url": "https://www.consul.io"
+          "url": "/consul/docs"
         }
       ]
     },
@@ -115,17 +115,17 @@
         {
           "heading": "Interdum sollicitudin.",
           "text": "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed elementum lectus posuere.",
-          "url": "https://www.consul.io"
+          "url": "/consul/docs"
         },
         {
           "heading": "Ornare nibh semper.",
           "text": "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Ut commodo facilisis diam.",
-          "url": "https://www.consul.io"
+          "url": "/consul/docs"
         },
         {
           "heading": "Fames nibh gravida.",
           "text": "Lorem ipsum dolor sit amet, consectetur adipiscing elit.",
-          "url": "https://www.consul.io"
+          "url": "/consul/docs"
         }
       ]
     },
--- a/src/data/consul.json
+++ b/src/data/consul.json
@@ -1,6 +1,7 @@
 {
   "name": "Consul",
   "slug": "consul",
+  "basePaths": ["docs", "api-docs", "commands"],
   "sidebar": {
     "landingPageNavData": [
       { "title": "Introduction", "fullPath": "/consul" },
@@ -28,5 +29,32 @@
         "href": "https://support.hashicorp.com"
       }
     ]
-  }
+  },
+  "navigationHeaderItems": [
+    {
+      "label": "Overview",
+      "path": "/consul",
+      "id": "overview"
+    },
+    {
+      "label": "Docs",
+      "id": "docs",
+      "path": "/consul/docs"
+    },
+    {
+      "label": "API",
+      "path": "/consul/api-docs",
+      "id": "api-docs"
+    },
+    {
+      "label": "CLI",
+      "path": "/consul/commands",
+      "id": "commands"
+    },
+    {
+      "label": "Install",
+      "path": "/consul/downloads",
+      "id": "downloads"
+    }
+  ]
 }
--- a/src/data/consul.json
+++ b/src/data/consul.json
@@ -4,10 +4,10 @@
   "basePaths": ["docs", "api-docs", "commands"],
   "sidebar": {
     "landingPageNavData": [
-      { "title": "Introduction", "fullPath": "/consul" },
+      { "title": "Introduction", "fullPath": "/consul/docs/intro" },
       {
         "title": "Getting Started",
-        "fullPath": "/consul"
+        "fullPath": "/consul/docs/install"
       }
     ],
     "resourcesNavData": [
--- a/src/layouts/sidebar-sidecar/utils/prepare-nav-data-for-client.ts
+++ b/src/layouts/sidebar-sidecar/utils/prepare-nav-data-for-client.ts
@@ -70,17 +70,32 @@ function prepareNavNodeForClient(node: NavNode, basePaths: string[]): MenuItem {
       id: slugify(`menu-item-${node.path}`, { lower: true }),
     }
   } else if (isNavDirectLink(node)) {
-    // For direct links, if they're absolute URLs,
-    // such as https://www.example.com, then leave them untouched.
-    // Otherwise, prefix them with the product slug (basePath[0])
-    const adjustedHref = isAbsoluteUrl(node.href)
-      ? node.href
-      : `/${basePaths[0]}${node.href}`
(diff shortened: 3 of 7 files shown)


src/data/consul.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "Consul",
   "slug": "consul",
+  "basePaths": ["docs", "api-docs", "commands"],
```
zchsh, line 4: Added these `docs` routes 👍

src/data/consul.json
```diff
@@ -28,5 +29,32 @@
         "href": "https://support.hashicorp.com"
       }
     ]
-  }
+  },
+  "navigationHeaderItems": [
```
zchsh, line 33: This is likely not final, fully correct `navigationHeaderItems` - I based this off the current `.io` nav. The intent here is to have something to navigate to the new `docs` routes with.

src/layouts/sidebar-sidecar/utils/product-url-adjusters.ts
```diff
@@ -14,12 +14,16 @@ import remarkPluginAdjustLinkUrls from 'lib/remark-plugin-adjust-link-urls'
...
-      const isBadApiUrl = url == '/api' || url.startsWith('/api/')
-      if (isBadApiUrl) url = url.replace(/^\/api/, '/api-docs')
-      return url
-    },
-  },
+  { urlAdjustFn: rewriteApiToApiDocs },
 ]
+export const consulUrlAdjuster: Pluggable = [
+  remarkPluginAdjustLinkUrls,
+  { urlAdjustFn: rewriteApiToApiDocs },
```
zchsh, line 21: Both `vault` and `consul` share the need to ensure `/api` links get rewritten to `/api-docs` (otherwise we'd have to set up more complex redirects within `dev-portal`). I've re-used the approach from `vault`, and shifted things a bit so it's re-usable for other products as well 👍

src/data/consul-landing.json
```diff
@@ -15,7 +15,7 @@
       "text": "In this guide we cover what Consul is, what problems it can solve, how it compares to existing software, and how you can get started using it.",
       "link": {
         "text": "Learn More",
-        "url": "https://www.consul.io/docs/intro"
+        "url": "/consul/docs/intro"
```
zchsh, line 18: Have updated all URLs in `consul-landing`, these routes are now accessible within `dev-portal` 🎉

src/layouts/sidebar-sidecar/utils/prepare-nav-data-for-client.ts
```diff
@@ -69,9 +70,16 @@ function prepareNavNodeForClient(node: NavNode, basePaths: string[]): MenuItem {
...
+    // For direct links, if they're absolute URLs,
+    // such as https://www.example.com, then leave them untouched.
+    // Otherwise, prefix them with the product slug (basePath[0])
+    const adjustedHref = isAbsoluteUrl(node.href)
+      ? node.href
+      : `/${basePaths[0]}${node.href}`
+    // Also add `id` for direct links
     return {
       ...node,
+      href: adjustedHref,
```
zchsh, line 82: At present, folks sometimes use `NavDirectLink` nodes for _non-external_ URLs, such as `/api-docs`. Unlike absolute, external URLs, such as `https://www.example.com`, these URLs make assumptions about the URL structure of the `.io` sites, and these assumptions are not true in `dev-portal`, so we end up with broken links.

To fix this, if a `NavDirectLink` node has a non-absolute URL, we prefix it with the product.slug (which in this context is `basePath[0]`). For example, the [Consul docs sidebar](https://dev-portal-git-zsadd-consul-docs-views-hashicorp.vercel.app/consul/docs) has an API link with the URL `/api-docs`, and we prefix this to end up with the valid URL `/consul/api-docs`.

ashleemboyer: I took a deeper look at this because I saw that external link icons were being rendered for the "API" and "Commands (CLI)" sidebar items on `/consul/docs` when that's not the desired behavior. To word what you already said in another way, sometimes `href` is used for specifying internal link paths instead of `path`, and this presents an issue in the `SidebarNavMenuItem` component that conditionally renders an external link icon based on the `fullPath` property that is only added to nav data if the `path` property is defined.

So here's how I approached handling that, I think there needs to be replacement of `href` with `path` and `fullPath` instead of always setting `href`, because the latter will tell the sidebar to render the external link:

```typescript
function prepareNavNodeForClient(node: NavNode, basePaths: string[]): MenuItem {
  if (isNavBranch(node)) {
    ...
  } else if (isNavLeaf(node)) {
    ...
  } else if (isNavDirectLink(node)) {
    const id = slugify(`external-url-${node.title}`, { lower: true })

    // check if there is data that disagrees with DevDot's assumptions/needs
    const hrefIsNotAbsolute = !isAbsoluteUrl(node.href)

    // if there is data that needs fixed, correct it by:
    // - removing the `href` prop
    // - adding the `path` prop
    // - adding the `fullPath` that is usually added in the previous block of this if-statement
    if (hrefIsNotAbsolute) {
      return {
        ...node,
        href: null,
        id,
        path: node.href,
        fullPath: `/${basePaths[0]}${node.href}`,
      }
    } else {
      // otherwise, just add the id prop as we did before
      return { ...node, id }
    }
  } else {
    ...
  }
}
```

ashleemboyer (changes requested):

src/pages/consul/api-docs/[[...page]].tsx
```diff
@@ -0,0 +1,28 @@
...
+  product,
+  basePath,
+  baseName,
+  additionalRemarkPlugins: [consulUrlAdjuster],
+})
+
+consulDocsPage.layout = SidebarSidecarLayout
+
+export { getStaticPaths, getStaticProps }
+export default consulDocsPage
```
Comment (ashleemboyer) [inline src/pages/consul/api-docs/[[...page]].tsx:28]: This component name should be in PascalCase and needs `Api` added in too.

```suggestion
const ConsulApiDocsPage = ({ mdxSource }): ReactElement => {
  return <DocsView {...mdxSource} />
}

const { getStaticPaths, getStaticProps } = getStaticGenerationFunctions({
  product,
  basePath,
  baseName,
  additionalRemarkPlugins: [consulUrlAdjuster],
})

ConsulApiDocsPage.layout = SidebarSidecarLayout

export { getStaticPaths, getStaticProps }
export default ConsulApiDocsPage
```

src/pages/consul/commands/[[...page]].tsx
```diff
@@ -0,0 +1,28 @@
...
+  product,
+  basePath,
+  baseName,
+  additionalRemarkPlugins: [consulUrlAdjuster],
+})
+
+consulDocsPage.layout = SidebarSidecarLayout
+
+export { getStaticPaths, getStaticProps }
+export default consulDocsPage
```
Comment (ashleemboyer) [inline src/pages/consul/commands/[[...page]].tsx:28]: This component name should be in PascalCase and should have `Commands` instead of `Docs` in the name.

```suggestion
const ConsulCommandsPage = ({ mdxSource }): ReactElement => {
  return <DocsView {...mdxSource} />
}

const { getStaticPaths, getStaticProps } = getStaticGenerationFunctions({
  product,
  basePath,
  baseName,
  additionalRemarkPlugins: [consulUrlAdjuster],
})

ConsulCommandsPage.layout = SidebarSidecarLayout

export { getStaticPaths, getStaticProps }
export default ConsulCommandsPage
```

src/pages/consul/docs/[[...page]].tsx
```diff
@@ -0,0 +1,30 @@
...
+  product,
+  basePath,
+  baseName,
+  additionalRemarkPlugins: [consulUrlAdjuster],
+})
+
+consulDocsPage.layout = SidebarSidecarLayout
+
+export { getStaticPaths, getStaticProps }
+export default consulDocsPage
```
Comment (ashleemboyer) [inline src/pages/consul/docs/[[...page]].tsx:30]: This component name should be in PascalCase.

```suggestion
const ConsulDocsPage = ({ mdxSource }): ReactElement => {
  return <DocsView {...mdxSource} additionalComponents={additionalComponents} />
}

const { getStaticPaths, getStaticProps } = getStaticGenerationFunctions({
  product,
  basePath,
  baseName,
  additionalRemarkPlugins: [consulUrlAdjuster],
})

ConsulDocsPage.layout = SidebarSidecarLayout

export { getStaticPaths, getStaticProps }
export default ConsulDocsPage
```

ashleemboyer (changes requested):
Second pass, just found a few minor things that needed updated in addition to things from my previous review!

src/layouts/sidebar-sidecar/utils/prepare-nav-data-for-client.ts
zchsh: Ah thanks for this, this makes sense! Will take this approach 👍

One UX concern I had was that when these links are rendered without an "external" icon, it implies to me that'll they'll behave like any other sidebar link (ie the sidebar context stays consistent), so without the "external" icon it can feel jarring for the sidebar to shift. However, and I think this echoes your suggested approach, this is probably more a "tweak the sidebar content" issue, the "switch to `path` when internal" implementation seems more correct 👍💯

ashleemboyer: I'd argue that the Consul nav data isn't currently being used correctly. The root nav items that link to API and Commands for example are redundant with the main navigation header, and I've noticed that some links actually link to other docs that live in a different section in the same sidebar. Both of these are misuses of the sidebar IMO, so I think we'll have to fix these in the future. But overall and what we can address in the short term: if they link to a page that lives in DevDot (or will live in DevDot very soon), then the external link icon shouldn't be rendered.

zchsh: +1 great points, fully agreed on this being a nav-data content issue. Have updated this with the suggested approach ☺️

zchsh:
@ashleemboyer Thanks for the excellent review & quick-to-apply suggestions 💖 💯 Have updated with those, let me know what you think 🙇‍♂️

src/layouts/sidebar-sidecar/utils/prepare-nav-data-for-client.ts
```diff
@@ -69,10 +70,32 @@ function prepareNavNodeForClient(node: NavNode, basePaths: string[]): MenuItem {
...
+      // convert it to a NavLeaf node, and treat it similarly.
+      // Note that the `fullPath` added here differs from typical
+      // NavLeaf treatment, as we only use the first part of the `basePath`.
+      // (We expect this to be the product slug, eg "consul").
+      return {
+        title: node.title,
+        path: node.href,
+        id,
+        fullPath: `/${basePaths[0]}${node.href}`,
+      }
```
ashleemboyer, line 94: I think instead of setting every property (like `title`) we want to spread `node`, "delete" `href` (by setting it to `null`), and only add the properties we're using to correct the node. That way if we ever want to add more properties to nodes we won't have tricky bugs coming from not spreading in this block. Other proposed change is to alphabetize the properties we're setting.

```suggestion
      return {
        ...node,
        fullPath: `/${basePaths[0]}${node.href}`,
        href: null,
        id,
        path: node.href,
      }
```

ashleemboyer (changes requested):

src/layouts/sidebar-sidecar/utils/prepare-nav-data-for-client.ts
zchsh: Sounds good, applied!

For context, had switched this because it felt like we're trying to switch the node structure here from `NavDirectLink` to `NavLeaf`. It made me think we might even want to do this transform as a separate, pre-cursor step to the transformations we have set up in `prepare-nav-data-for-client`. Otherwise, eg if we do add properties to `NavDirectLink`, and spread all of them, we could end up with a weird hybrid between the two node types (which I think could also cause unexpected bugs).

ashleemboyer (approved):
🥇

PR #133, opened 2022-02-25
feat: multi-arch docker images

thiskevinwang:

---

## Relevant links

- [Preview link](url) 🔎
- [Asana task](url) 🎟️

### DIFF
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -6,6 +6,7 @@ name: Docker Publish
 # documentation.
 
 on:
+  pull_request:
   push:
     branches: [main]
 
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -6,7 +6,6 @@ name: Docker Publish
 # documentation.
 
 on:
-  pull_request:
   push:
     branches: [main]
 
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -21,12 +21,14 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
-
+      # Emulate ARM Processor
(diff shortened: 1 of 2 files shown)


.github/workflows/docker_publish.yml
```diff
@@ -6,6 +6,7 @@ name: Docker Publish
 # documentation.
 
 on:
+  pull_request:
```
thiskevinwang, line 9: 💭 Do we want to verify that Docker builds properly on a PR? (no push occurs)

BRKalow: I don't think it's necessary, as all it really does is run npm install, but I don't feel too strongly about it. If we do want to keep it, we'll need to update something with the build and push step as it looks like the latest run pushed to the latest tag.

thiskevinwang: Will remove this!

thiskevinwang, line 9:
```suggestion
```

BRKalow:
@thiskevinwang Do we need to remove [this](https://github.com/hashicorp/dev-portal/blob/main/Dockerfile#L1) flag from the Dockerfile, or do you know if it'll get overwritten by the platform specified from our config here?

thiskevinwang:
> @thiskevinwang Do we need to remove [this](https://github.com/hashicorp/dev-portal/blob/main/Dockerfile#L1) flag from the Dockerfile, or do you know if it'll get overwritten by the platform specified from our config here?

@BRKalow ah good catch, also TIL 👇! ... Yes we should (I will) remove that

> All FROM commands inside the Dockerfile without their own --platform flag will pull base images for this platform and this value will also be the platform of the resulting image.
> - [docs](https://docs.docker.com/engine/reference/commandline/buildx_build/#platform)

BRKalow (approved):

PR #138, opened 2022-02-28
feat: add nomad docs views

zchsh:

---

## Relevant links

- [Preview link](https://dev-portal-git-zsadd-nomad-docs-views-hashicorp.vercel.app) 🔎
- [Asana task](https://app.asana.com/0/1100423001970639/1201647459964518/f) 🎟️

## What

This PR adds `docs-page` routes for `/nomad`.

## Why

To iterate towards supporting all product docs content within the `dev-portal` project.

## How

- Re-uses existing templates, as used for `vault`, `waypoint`, and [`consul`](https://github.com/hashicorp/dev-portal/pull/131)
- Fixes an issue in breadcrumb generation, where an unusual edge case in `/nomad` content broke typical generation
    - Previously, if there were no matches for a given nav node to be displayed in a breadcrumb bar, we would throw an error
    - In what looks like a content mishap, Nomad has a [`/docs/autoscaling/agent/source`](https://www.nomadproject.io/docs/autoscaling/agent/source) page, but does not have nav nodes or docs pages for the "parent" `/docs/autoscaling/agent` or `/docs/autoscaling` pages. This seems to be because content has moved to [`/tools/autoscaling`](https://www.nomadproject.io/tools/autoscaling).
    - The broader issue here is a content one, I think. However there is no immediate solution to prevent the build error other than changing from an error to a warning. We can still render unlinked breadcrumb bar items for `/docs/autoscaling` and `/docs/autoscaling/agent`. You can see these on [the preview's `/nomad/docs/autoscaling/agent/source`]() page.

## Testing

- [ ] Navigate to the docs views, click around, and confirm behaviour is as expected
    - [ ] [`/nomad/api-docs`](https://dev-portal-git-zsadd-nomad-docs-views-hashicorp.vercel.app/nomad/api-docs)
    - [ ] [`/nomad/docs`](https://dev-portal-git-zsadd-nomad-docs-views-hashicorp.vercel.app/nomad/docs)
    - [ ] [`/nomad/intro`](https://dev-portal-git-zsadd-nomad-docs-views-hashicorp.vercel.app/nomad/intro)
    - [ ] [`/nomad/plugins`](https://dev-portal-git-zsadd-nomad-docs-views-hashicorp.vercel.app/nomad/plugins)
    - [ ] [`/nomad/tools`](https://dev-portal-git-zsadd-nomad-docs-views-hashicorp.vercel.app/nomad/tools)

## Anything else?

Nope, do want to call out the change to `get-docs-breadcrumbs` now logging a warning and rendering fallback unlinked breadcrumb bar items (instead of throwing an error as was done previously).

### DIFF
--- a/src/components/breadcrumb-bar/utils/get-docs-breadcrumbs.ts
+++ b/src/components/breadcrumb-bar/utils/get-docs-breadcrumbs.ts
@@ -89,8 +89,8 @@ function findAllPathMatchedNodes(navNodes, pathString, depth = 0) {
 }
 
 function findPathMatchedNodes(navNode, pathString, depth) {
-  // If it's a node with child routes, return true
-  // if any of the child routes are active
+  // If it's a node with child routes, look for matches
+  // within the child routes
   if (navNode.routes) {
     // Check for an index (aka "Overview") node in the child routes.
     // These nodes will have paths with a number of parts
--- a/src/components/breadcrumb-bar/utils/get-docs-breadcrumbs.ts
+++ b/src/components/breadcrumb-bar/utils/get-docs-breadcrumbs.ts
@@ -37,7 +37,7 @@ function getPathBreadcrumbs({
   // associated navData node. This gets slightly
   // complex due to index (aka "Overview") nodes.
   const breadcrumbNodes = breadcrumbPaths.map((p) =>
-    getPathMatchedNode(navData, p)
+    getPathMatchedNode(navData, p, basePath)
   )
   // Map the matched navData nodes into { title, url }
   // objects as needed for breadcrumb link rendering.
@@ -49,7 +49,7 @@ function getPathBreadcrumbs({
   })
 }
 
-function getPathMatchedNode(navNodes, pathString) {
+function getPathMatchedNode(navNodes, pathString, basePath) {
   const matches = findAllPathMatchedNodes(navNodes, pathString)
   // If we have exactly one match, this is great, and expected
   if (matches.length == 1) return matches[0]
@@ -63,17 +63,22 @@ function getPathMatchedNode(navNodes, pathString) {
   if (matches.length > 1) {
     if (IS_DEV) {
       console.warn(
-        `Ambiguous breadcrumb path: Found ${matches.length} matches for "${pathString}". Please ensure there is exactly one node or index-less category with the path "${pathString}" in the provided navData.`
+        `Ambiguous breadcrumb path under "${basePath}": Found ${matches.length} matches for "${pathString}". Please ensure there is exactly one node or index-less category with the path "${pathString}" in the provided navData.`
       )
     }
     return matches[0]
   }
   // Otherwise, we have zero matches, which would mean a breadcrumb with missing parts.
-  // This would feel broken to a site visitor, so should be a blocking
-  // issue we need to fix, so we throw an error.
-  throw new Error(
-    `Missing breadcrumb path: Found zero matches for "${pathString}". Please ensure there is a node (or index-less category) with the path "${pathString}" in the provided navData.`
-  )
+  // We have no nav-data to render a nice "title" for the breadcrumb bar, and
+  // there isn't a matched path to link to, but we can still render an
+  // unlinked breadcrumb item using the "pathString" as the title
+  if (IS_DEV) {
+    console.warn(
+      `Missing breadcrumb path under "${basePath}": Found zero matches for "${pathString}". Please ensure there is a node (or index-less category) with the path "${pathString}" in the provided navData.`
+    )
+  }
+  const lastPathPart = pathString.split('/').pop()
+  return { title: lastPathPart }
 }
 
 function findAllPathMatchedNodes(navNodes, pathString, depth = 0) {
--- a/src/data/nomad-landing.json
+++ b/src/data/nomad-landing.json
@@ -26,25 +26,25 @@
           "icon": "IconDocs",
           "heading": "Nomad Reference Documentation",
           "text": "This documentation is a reference for all available features and options of Nomad.",
-          "url": "https://www.nomadproject.io/docs"
+          "url": "/nomad/docs"
         },
         {
           "icon": "IconTerminal",
           "heading": "Nomad CLI",
           "text": "Nomad is controlled via a very easy to use command-line interface (CLI).",
-          "url": "https://www.nomadproject.io/docs/commands"
+          "url": "/nomad/docs/commands"
         },
         {
           "icon": "IconServer",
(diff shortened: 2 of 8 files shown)


src/components/breadcrumb-bar/utils/get-docs-breadcrumbs.ts
```diff
@@ -63,17 +63,22 @@ function getPathMatchedNode(navNodes, pathString) {
...
-        `Ambiguous breadcrumb path: Found ${matches.length} matches for "${pathString}". Please ensure there is exactly one node or index-less category with the path "${pathString}" in the provided navData.`
+        `Ambiguous breadcrumb path under "${basePath}": Found ${matches.length} matches for "${pathString}". Please ensure there is exactly one node or index-less category with the path "${pathString}" in the provided navData.`
       )
     }
     return matches[0]
   }
   // Otherwise, we have zero matches, which would mean a breadcrumb with missing parts.
-  // This would feel broken to a site visitor, so should be a blocking
-  // issue we need to fix, so we throw an error.
-  throw new Error(
```
zchsh, line 74: Previously we threw an error here. But, this broke the build when rendering [`/nomad/docs/autoscaler/agent/source`](https://dev-portal-git-zsadd-nomad-docs-views-hashicorp.vercel.app/nomad/docs/autoscaling/agent/source) due to some quirks in Nomad's nav-data (described in more detail in the PR description).

To fix this, we now render a kind of "fallback" breadcrumb bar item if we can't match up a nav node. This item is:
1. Unlinked, because there's nothing to link to
2. Titled with the raw `path` part to the page, because there's no nav-node with a pretty title.

We definitely want to avoid these situations, so I've retained warnings here. We may also be able to more strictly validate `nav-data` in order to prevent these types of nav-data structures at the source.

src/data/nomad-landing.json
```diff
@@ -26,25 +26,25 @@
           "icon": "IconDocs",
           "heading": "Nomad Reference Documentation",
           "text": "This documentation is a reference for all available features and options of Nomad.",
-          "url": "https://www.nomadproject.io/docs"
+          "url": "/nomad/docs"
```
zchsh, line 29: Updated these links, these URLs are available locally now 🎉

src/components/breadcrumb-bar/utils/get-docs-breadcrumbs.ts
```diff
@@ -84,8 +89,8 @@ function findAllPathMatchedNodes(navNodes, pathString, depth = 0) {
 }
 
 function findPathMatchedNodes(navNode, pathString, depth) {
-  // If it's a node with child routes, return true
-  // if any of the child routes are active
+  // If it's a node with child routes, look for matches
+  // within the child routes
```
zchsh, line 93: This comment was a lil inaccurate, I think I'd used the recursion we do for highlighting active nodes as a template.

zchsh:
@ashleemboyer This is https://github.com/hashicorp/dev-portal/pull/131, but make it `nomad`!

Of note: I ran into an annoying issue with the sidebar data on [`/nomad/docs/autoscaling/agent/source`](https://dev-portal-git-zsadd-nomad-docs-views-hashicorp.vercel.app/nomad/docs/autoscaling/agent/source). I have a fix for it, but one that continues to warn us in Dev, cause I think the root here might ultimately be more of a content issue.

src/components/breadcrumb-bar/utils/get-docs-breadcrumbs.ts
```diff
@@ -63,17 +63,22 @@ function getPathMatchedNode(navNodes, pathString) {
   if (matches.length > 1) {
     if (IS_DEV) {
       console.warn(
-        `Ambiguous breadcrumb path: Found ${matches.length} matches for "${pathString}". Please ensure there is exactly one node or index-less category with the path "${pathString}" in the provided navData.`
+        `Ambiguous breadcrumb path under "${basePath}": Found ${matches.length} matches for "${pathString}". Please ensure there is exactly one node or index-less category with the path "${pathString}" in the provided navData.`
```
ashleemboyer, line 66: Love verbose errors, I think this is a nice improvement! ✨

src/components/breadcrumb-bar/utils/get-docs-breadcrumbs.ts
```diff
@@ -63,17 +63,22 @@ function getPathMatchedNode(navNodes, pathString) {
...
+  // there isn't a matched path to link to, but we can still render an
+  // unlinked breadcrumb item using the "pathString" as the title
+  if (IS_DEV) {
+    console.warn(
+      `Missing breadcrumb path under "${basePath}": Found zero matches for "${pathString}". Please ensure there is a node (or index-less category) with the path "${pathString}" in the provided navData.`
+    )
+  }
+  const lastPathPart = pathString.split('/').pop()
+  return { title: lastPathPart }
 }
```
ashleemboyer, line 82: Just leaving a comment for future reference, since we talked about it off this PR:

Sometime after we've onboarded content, and identified issues that need fixing, and fixed those issues, it would be nice to convert checks like this to:

1. Throw errors instead of warnings (so we can start enforcing content rules)
2. Make sure errors are thrown in all environments (is a more thorough check, warnings can probably still be dev only!)
3. Only throw errors when the latest version is being viewed, otherwise log a warning to the console (means we don't have to update every single version of content that conflicts with new/updated content rules)

src/data/nomad.json
```diff
@@ -28,5 +29,37 @@
...
+    {
+      "label": "Plugins",
+      "path": "/nomad/plugins",
+      "id": "commands"
+    },
+    {
+      "label": "Tools",
+      "path": "/nomad/tools",
+      "id": "commands"
+    },
```
ashleemboyer, line 58: These two have the same `id` specified, and it's making React unhappy because they're used as `key` props in a map.

```suggestion
      "label": "Plugins",
      "path": "/nomad/plugins",
      "id": "plugins"
    },
    {
      "label": "Tools",
      "path": "/nomad/tools",
      "id": "tools"
    },
```

Side note: I think these IDs probably aren't needed anymore, but let's address that outside of this PR. 😊

ashleemboyer (changes requested):

src/data/nomad.json
zchsh: ah dang thanks for catching this, totally missed it 😅 Applied! ✅

ashleemboyer: All good! You and Bryce catch my copy-paste errors all the time 😁

ashleemboyer (approved):
All good to :shipit: when those checks pass! 👏🏻

PR #159, opened 2022-03-04
Add new VersionContextSwitcher component

ashleemboyer:

---

## Relevant links

- [Preview link](https://dev-portal-git-ambadd-version-context-switcher-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1201010428539925/1201342966970641/f) 🎟️

## What

- Adds `VersionContextSwitcher` (`index.tsx`, `types.ts`, `version-context-switcher.module.css`)
- Adds `Version` type in `src/lib/fetch-release-data.ts`
- Moves the version switcher to be inside of the `DownloadsSection` card and located at the top right of the card
- Adds `src/views/product-downloads-view/components/index.ts` barrel file to help clean up some imports
- Adds `PageHeader` component local to `product-downloads-view` to help clean up some markup
- Adds a current version context and a hook to consume the value
- Adds `initializeVersionSwitcherOptions` helper to `src/views/product-downloads-view/helpers.ts`
- Improves some TypeScript in `src/views/product-downloads-view/helpers.ts`

## Why

- Adding the `VersionContextSwitcher` now, especially before we have a foundational `ContextSwitcher`, has two benefits I can think of at the moment:
  - Gets us thinking about what the `ContextSwitcher` API will look like since it will eventually be used in `VersionContextSwitcher`
  - Helps keep the code clean and easy to update later since it's in its own component folder 
- The new current version context local to the view is helpful because there are many places in the view where the current version is displayed, many of them being at a higher level in the component tree than the version switcher itself.

## How

- When I'm creating a Context I usually double-check my work against Kent C. Dodds's [How to use React Context effectively](https://kentcdodds.com/blog/how-to-use-react-context-effectively) article.
- The [ARIA live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) MDN page helped me decide what semantic elements to use in `VersionContextSwitcher`. I also checked with the Design Systems Team on whether or not an ARIA live region (`aria-live`) was needed for this component, and was given guidance to use an informative `aria-label` instead.
- Otherwise, this PR was mostly moving around a lot of code.

### General interaction

- [ ] Go to an install view on the preview URL. [Waypoint's is fine](https://dev-portal-git-ambadd-version-context-switcher-hashicorp.vercel.app/waypoint/downloads).
- [ ] The new `VersionContextSwitcher` should be at the top-right of the first card in the page:
  image
- [ ] Choose a different version from the dropdown
- [ ] The breadcrumb, page subtitle, version switcher, and all links in the downloads card should update to reflect the selected version
- [ ] The Package managers section of the downloads card should no longer render

### VoiceOver testing

- [ ] [Turn on VoiceOver](https://support.apple.com/guide/voiceover/turn-voiceover-on-or-off-vo2682/mac)
- [ ] Bring the version switcher into focus with the keyboard
- [ ] VoiceOver should present the switcher as "Waypoint 0.7.2 (latest), Choose a different version to install, menu pop up collapsed, button"
  - Depending on where your focus is coming from, such as tabbing from the breadcrumb to the switcher, there may be more content appended to the end of this presentation

## Anything else?

Nope!


src/views/product-downloads-view/components/downloads-section/index.tsx
```diff
@@ -208,17 +211,31 @@ const DownloadsSection = ({
   )
 
   return (
-    <article className={s.root}>
+    <div className={s.root}>
```
ashleemboyer, line 214: Changed this to `div` because the other sections on the page don't use `article`. Rather than update all the sections to use `article` now, I'd prefer to double-check whether `section` or `article` should be used for these section components later and update them then.

src/views/product-downloads-view/helpers.ts
```diff
@@ -11,6 +13,9 @@ const PLATFORM_MAP = {
   Linux: 'linux',
 }
 
+// exclude pre-releases and such
+const VALID_SEMVER_REGEX = /^\d+\.\d+\.\d+$/
```
zchsh, line 17: Super minor & non-blocking nit, but I wonder whether it'd be worth the reassurance of using [semver](https://www.npmjs.com/package/semver). (not gonna lie i find the docs to be quite hard to parse, but I think it'd be something like `semver.valid(maybeSemverString)` to check it's valid and then `semver.satisfies(maybeSemverString, '*')` to check that it's not a pre-release).

zchsh (approved):
✅ Looks great! I poked around and tested different versions, and also confirmed the presentation with VoiceOver.

Super minor non-blocking nit around using [semver](https://www.npmjs.com/package/semver) for version string validation, def feel free to ignore or defer to a later PR (may not even be worth the extra dependency, I'm not sure).

src/views/product-downloads-view/helpers.ts
ashleemboyer: Woah this is a great idea to look into! We're using the semver package for sorting already, so I'm guessing one of these validation functions wouldn't be hard to add. I'll work on adding this in one of my follow-up PRs!

PR #170, opened 2022-03-09
Add new install view for Boundary

ashleemboyer:
## "Ready for Review" checklist

- [x] The Vercel preview link has been added to this PR's description
- [x] The Asana task has been added to this PR's description
- [x] This PR's link has been added to the "📐 Pull Request" field on the Asana task
- [x] This Vercel preview link has been added to the "🔍 Preview" field on the Asana task
- [x] The description template has been filled in below

---

## Relevant links

- [Preview link](https://dev-portal-git-ambadd-boundary-install-view-hashicorp.vercel.app/boundary/downloads) 🔎
- [Asana task](https://app.asana.com/0/0/1201893669632603/f) 🎟️

## What/Why

- Adds `src/data/boundary-install.json`, which contains the data needed for Boundary's `ProductDownloadsView`
- Adds `src/pages/boundary/downloads/index.tsx`, which conditionally renders the new downloads view based on the `enable_new_downloads_view` flag

## How

Using the same process as:

- #140
- #143
- #148
- #150
- #151

## Testing

- [ ] Go to https://dev-portal-git-ambadd-boundary-install-view-hashicorp.vercel.app/boundary/downloads
- [ ] Make sure the page renders how you expect

## Anything else?

Nope!

PR #151, opened 2022-03-03
Add new install view for Vagrant

ashleemboyer:
## "Ready for Review" checklist

- [x] The Vercel preview link has been added to this PR's description
- [x] The Asana task has been added to this PR's description
- [x] This PR's link has been added to the "📐 Pull Request" field on the Asana task
- [x] This Vercel preview link has been added to the "🔍 Preview" field on the Asana task
- [x] The description template has been filled in below

---

## Relevant links

- [Preview link](https://dev-portal-git-ambadd-vagrant-install-view-hashicorp.vercel.app/vagrant/downloads) 🔎
- [Asana task](https://app.asana.com/0/0/1201893669632605/f) 🎟️

## What/Why

- Adds `src/data/vagrant-install.json`, which contains the data needed for Vagrant's `ProductDownloadsView`
- Adds `src/pages/vagrant/downloads/index.tsx`, which conditionally renders the new downloads view based on the `enable_new_downloads_view` flag
- Adds support for `packageManagerOverrides`
  - Adds `generatePackageManagers` helper

## How

- Using the same process as #140, #143, #148, and #150 😊
- Referenced how [`ProductDownloadsPage` handles `packageManagerOverrides`](https://github.com/hashicorp/react-components/blob/main/packages/product-download-page/index.tsx#L70-L103)

## Testing

- [ ] Go to https://dev-portal-git-ambadd-vagrant-install-view-hashicorp.vercel.app/vagrant/downloads
- [ ] Make sure the macOS package manager is correctly overridden
- [ ] Make sure the Linux -> Homebrew package manager is correctly overridden
- [ ] Makes sure the rest of the page still behaves as you expect

## Anything else?

- The current Vagrant downloads page shows a "Download Vagrant vmware Utility" link in the merchandising slot. This link will eventually live in the sidebar of the main Vagrant downloads page
- A relevant Asana ticket: [Dynamically pull titles & descriptions for cards on product landings](https://app.asana.com/0/1201010428539925/1201646299837754/f)

PR #143, opened 2022-03-01
Add new install view for Consul

ashleemboyer:
## "Ready for Review" checklist

- [x] The Vercel preview link has been added to this PR's description
- [x] The Asana task has been added to this PR's description
- [x] This PR's link has been added to the "📐 Pull Request" field on the Asana task
- [x] This Vercel preview link has been added to the "🔍 Preview" field on the Asana task
- [x] The description template has been filled in below

---

## Relevant links

- [Preview link](https://dev-portal-git-ambadd-consul-install-view-hashicorp.vercel.app/consul/downloads) 🔎
- [Asana task](https://app.asana.com/0/0/1201893669632604/f) 🎟️

## What/Why

- Adds `src/data/consul-install.json`, it contains all of the data needed for Consul's `ProductDownloadsView`, plus some placeholder data
- Adds `src/pages/consul/downloads/index.tsx`, which conditionally renders the new downloads view based on the `enable_new_downloads_view` flag
- Adds the `generateDefaultPackageManagers`, which is something I mostly copied from [react-components/product-download-page](https://github.com/hashicorp/react-components/blob/HEAD/packages/product-download-page/package-managers.ts).
  - The only change I made was for it to accept a whole product object instead of just a slug. I see this is a maintenance improvement in case we ever need to change the behavior of the helper to use more parts of a `Product` object.
- Updated `ProductDownloadsView` to use the new `generateDefaultPackageManagers` helper
- Updated the `packageManagers` prop in the `ProductDownloadsViewProps['pageContent']` shape to be optional since we now have a default array we can use

## How

Nothing extra to add here. [react-components/product-download-page](https://github.com/hashicorp/react-components/blob/HEAD/packages/product-download-page) was very helpful as a reference!

## Testing

- [ ] Go to [/consul/downloads on the preview](https://dev-portal-git-ambadd-consul-install-view-hashicorp.vercel.app/consul/downloads)
- [ ] Make sure the page renders how you expect

## Anything else?

Nope!

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1201893669632604

PR #171, opened 2022-03-09
Add new install view for Terraform

ashleemboyer:

---

## Relevant links

- [Preview link](https://dev-portal-git-ambadd-terraform-install-view-hashicorp.vercel.app/terraform/downloads) 🔎
- [Asana task](https://app.asana.com/0/0/1201893669632607/f) 🎟️

## What/Why

- Adds `src/data/terraform-install.json`, which contains the data needed for Terraform's `ProductDownloadsView`
- Adds `src/pages/terraform/downloads/index.tsx`, which conditionally renders the new downloads view based on the `enable_new_downloads_view` flag

## How

Using the same process as:

- #140
- #143
- #148
- #150
- #151
- #170

## Testing

- [ ] Go to https://dev-portal-git-ambadd-terraform-install-view-hashicorp.vercel.app/terraform/downloads
- [ ] Make sure the version selected has limited versions as options
- [ ] Make sure the rest of the page renders how you expect

## Anything else?

Nope!

### DIFF
--- a/src/pages/terraform/downloads/index.tsx
+++ b/src/pages/terraform/downloads/index.tsx
@@ -1,0 +1,92 @@
+import { ReactElement } from 'react'
+import { GetStaticProps } from 'next'
+import semverGte from 'semver/functions/gte'
+import semverMajor from 'semver/functions/major'
+import semverMinor from 'semver/functions/minor'
+import semverPatch from 'semver/functions/patch'
+import terraformData from 'data/terraform.json'
+import installData from 'data/terraform-install.json'
+import { Product } from 'types/products'
+import { generateStaticProps, GeneratedProps } from 'lib/fetch-release-data'
+import EmptyLayout from 'layouts/empty'
+import ProductDownloadsView from 'views/product-downloads-view'
+import PlaceholderDownloadsView from 'views/placeholder-product-downloads-view'
+
+const VERSION_DOWNLOAD_CUTOFF = '1.0.11'
+
+const TerraformDownloadsPage = (props: GeneratedProps): ReactElement => {
+  if (__config.flags.enable_new_downloads_view) {
+    const { latestVersion, releases } = props
+    return (
+      <ProductDownloadsView
+        latestVersion={latestVersion}
+        pageContent={installData}
+        releases={releases}
+      />
+    )
+  } else {
+    return <PlaceholderDownloadsView />
+  }
+}
+
+/**
+ * Pulled from terraform-website/pages/downloads/index.jsx on 3/9/2022:
+ * https://github.com/hashicorp/terraform-website/blob/master/pages/downloads/index.jsx#L55-L98
+ */
+function filterOldVersions(props) {
+  if (!props?.props?.releases?.versions) return props
+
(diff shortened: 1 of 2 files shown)


src/pages/terraform/downloads/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+  })
+
+  props.props.releases.versions = newVersions
+}
+
+export const getStaticProps: GetStaticProps = async () => {
+  const product = terraformData as Product
+  const generatedProps = await generateStaticProps(product)
+
+  filterOldVersions(generatedProps)
```
zchsh, line 85: Non-blocking nit, and almost certainly outside the scope of this PR, but the way `filterOldVersions` takes in the whole `{ props }` object and modifies it with a kind of side-effect-y approach, also using the global `VERSION_DOWNLOAD_CUTOFF`, feels a little opaque to me... It feels hard to tell exactly what's happening to generatedProps, and how much of it is affected, without reading the function source of `filterOldVersions`.

A kind of stop-gap-y way to mitigate this might be:

```ts
const rawVersions = generatedProps.props?.releases?.versions
if (rawVersions) {
  const newVersions = filterOldVersions(rawVersions, VERSION_DOWNLOAD_CUTOFF)
  generatedProps.props.releases.versions = newVersions
}

// & also.. modify `filterOldVersions` so it's something like:
// (rawVersions: ???, versionCutoff: string): ??? return `newVersions` 
// ... sorry for the `???`, just not sure on the actual type for incoming versions and don't wanna get it wrong!
```

And maybe another solution, potentially a little clearer, could be to have `fetch-release-data`'s `generateStaticProps` accept a second, optional `options` arg:

```ts
const generatedProps = await generateStaticProps(product, { versionCutoff: VERSION_DOWNLOAD_CUTOFF })
// use something like `filterVersions(rawVersions, versionCutoff)` within `generateStaticProps`
```

Again, def not blocking at all, just wanted to get the idea out there and see if it resonated 👍

zchsh (approved):
✅ Preview looks great, version cutoff seems to work perfectly!

Had a minor, non-blocking nit around how `filterOldVersions` works, but since it's being moved over from an existing codebase, could also see how we might want to keep them consistent for now - so def feel free to defer or even ignore 👍

src/pages/terraform/downloads/index.tsx
ashleemboyer: I had the same thoughts about it being side-effect-y and it definitely took me a minute to figure out what was going on! Updating `filterOldVersions` in a way like you mention where it returns a new array would make it clearer sooner what's happening. I think this had been an urgent request so there wasn't much time for refinement.

I'm less sure about extending `generateStaticProps` since I don't think there are any other products that do this. I think we would want to have two new options:
- `versionCutoff`, like you've proposed
- a boolean variable, maybe `onlyLatestMinorVersions`, that signals to exclude additional versions like what was requested when `filterOldVersions` was created

I've created an Asana ticket for the first part of this, refactoring `filterOldVersions`!
https://app.asana.com/0/1201010428539925/1201944491379825/f

PR #176, opened 2022-03-11
feat: adjust pull req template utl

kendallstrautman:
Adjusting the url so we can use the perma git branch pattern for the PR template!

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -14,7 +14,7 @@ Check these items off in the GitHub PR UI as you complete them.
 
 ## Relevant links
 
-- [Preview link](url) 🔎
+- [Preview link](https://dev-portal-git-BRANCH_NAME-hashicorp.vercel.app/PATH_TO_VIEW) 🔎
 - [Asana task](url) 🎟️
 
 ## What


zchsh (approved):
I dig this! 💯 Since the `amb.update-assembly-with-main` branch will hopefully land today, might makes sense to wait to merge this? But seems like it could also be a non-issue, I'm not sure 👍

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -14,7 +14,7 @@ Check these items off in the GitHub PR UI as you complete them.
 
 ## Relevant links
 
-- [Preview link](url) 🔎
+- [Preview link](https://dev-portal-git-BRANCH_NAME-hashicorp.vercel.app/PATH_TO_VIEW) 🔎
```
ashleemboyer, line 17: @kendallstrautman Could you add a note via MD comment about how Vercel removes `.` characters from branch names? This was something I had no idea about before I joined, I think it'd be valuable to mention.

```suggestion
<!--
Comment about the `.` character in branch names
-->

- [Preview link](https://dev-portal-git-BRANCH_NAME-hashicorp.vercel.app/PATH_TO_VIEW) 🔎
```

kendallstrautman: Good call!

PR #178, opened 2022-03-14
feat: custom vault dot io docs landing

zchsh:

---

## Relevant links

- [Preview link][docs] 🔎
- [Asana task](https://app.asana.com/0/1201010428539925/1201965330165954/f) 🎟️

## What

This PR adds a new `vaultproject.io/docs` landing page.

## Why

To improve the user experience of the vaultproject.io/docs landing page.

## How

This new page replaces the existing `/docs` landing page authored in MDX with a customized version authored in JSX. It makes this replacement for the latest version only. It leaves the `/docs` landing page for previous versions untouched.

## Testing

👉 Note - ensure you're in "Vault preview" mode by using the dropdown selector in the lower left of the screen

- [ ] Visit the [Vault docs view][docs], and confirm that it renders our new custom page
    - [ ] Check that the page renders as expected in your default browser, and in other browsers
    - [ ] View the page on mobile viewports, to confirm that responsive behaviour is as expected
    - [ ] Turn on VoiceOver and navigate through the page with your keyboard, to ensure the page is not narrowly targeting pointing and clicking
    - Additional Visual QA on this page would be much appreciated! 🙇‍♂️
- [ ] Navigate to [an older version of the Vault docs view][docs/v1.6.x], and confirm the non-custom docs page is being rendered as expected
    - Toggle back to the "latest version" and confirm that the custom page is rendered
- [ ] Navigate to the `/docs` page using the header nav bar, and confirm the page renders as expected

## Anything else?

⚠️ Note that this does NOT replace the need for the [docs index.mdx file in vault/website/content](https://github.com/hashicorp/vault/blob/main/website/content/docs/index.mdx). This content `.mdx` file still needs to be maintained, as it will still be used (each current version eventually becomes a previous version, and each previous version needs to have the `docs/index.mdx` file present).

---
  - https://app.asana.com/0/0/1201965330165954

[docs]: https://dev-portal-git-zsvault-shiny-docs-landing-hashicorp.vercel.app/docs
[docs/v1.6.x]: https://dev-portal-git-zsvault-shiny-docs-landing-hashicorp.vercel.app/docs/v1.6.x


src/views/_proxied-dot-io/product-docs-landing/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+import CardLink from 'components/card-link'
+import FeaturedCard from './partials/featured-card'
+import UseCaseCard from './partials/use-case-card'
+import s from './style.module.css'
+import VAULT_CONTENT from './vault-content'
+
+function ProductDocsLanding(): ReactElement {
+  // Note: later we could use this view with products other than Vault,
+  // and put this content somewhere more author-friendly (eg DatoCMS).
+  const productSlug = 'vault'
```
zchsh, line 11: Very "hard-coded" for now, but with an eye on the possibility that this view will be re-used by other products, and with other content. (Note: we'd probably move to a "block"-based approach rather than have hard-coded "Use Cases" and "Developer" sections as well).

src/views/_proxied-dot-io/product-docs-landing/vault-content.js
```diff
@@ -0,0 +1,103 @@
+const vaultContent = {
+  pageTitle: 'Documentation',
+  pageSubtitle:
+    'Welcome to Vault documentation! Vault is an identity-based secret and encryption management system. This documentation covers the main concepts of Vault, what problems it can solve, and contains a quick start for using Vault.',
+  featuredCard: {
+    heading: 'Get Started',
+    image: {
+      src: require('./vault-get-started-diagram.png'),
```
zchsh, line 8: This is only reason this is a JS file and not a JSON file. Alternate here could be to upload this asset to Dato. (Incidentally, we need a high-res asset here, hence the `FPO` text!).

zchsh:
@ashleemboyer @BRKalow This felt ready for code review, though we'll need to wait on the FPO asset, and not sure if we want to merge this to main before doing the Vault → `dev-portal` serving switch over.

One thing to call out is the tweak to `CardLink` - curious to get your thoughts on this, definitely happy to take a different approach if there's one we prefer 🙇‍♂️

src/views/_proxied-dot-io/product-docs-landing/partials/featured-card/index.tsx
```diff
@@ -0,0 +1,51 @@
+import { ReactElement } from 'react'
```
BRKalow, line 1: nit: what do you think about renaming this directory to `components` from `partials`?

src/views/_proxied-dot-io/product-docs-landing/index.tsx
BRKalow: Just an observation that this is very vault focused, and maybe it's not worth trying to posture it as re-usable until that time comes? We're going to have to make a bunch of changes anyway, so it might be clearer to call it `VaultDocsLanding` for now....not a huge deal though.

src/components/card/card.module.css
```diff
@@ -1,3 +1,4 @@
+/* Note: also used by components/card-link */
```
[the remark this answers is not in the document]
zchsh: Update: went  with a different approach, no longer relevant.

src/components/card-link/card-link.module.css
```diff
@@ -3,6 +3,8 @@ NOTE: `color: black` is needed because the color is otherwise set by a global
 style applied to anchor elements.
 */
 .root {
+  composes: root from '../card/card.module.css';
```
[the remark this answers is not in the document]
zchsh: Edit: went with a different approach, setting `100%` on the `div` without `CardLink` via a passed `className` also works, and keeps things more the way they are:


CleanShot 2022-03-16 at 16 26 05@2x

src/views/_proxied-dot-io/product-docs-landing/index.tsx
ashleemboyer: I'd probably not opt for making a view yet even, and just putting this markup in `src/pages/_proxied-dot-io/vault/docs/index.tsx`. Not a very strong opinion though. 👍🏻

src/views/_proxied-dot-io/product-docs-landing/partials/use-case-card/types.ts
```diff
@@ -0,0 +1,10 @@
+import { Products } from '@hashicorp/platform-product-meta'
+export interface UseCaseCardProps {
```
ashleemboyer, line 2: nit suggestion

```suggestion
import { Products } from '@hashicorp/platform-product-meta'

export interface UseCaseCardProps {
```

src/views/_proxied-dot-io/product-docs-landing/vault-content.js
ashleemboyer: Random curiosity, is it possible to to set the `img.src` property to the path of the image and then do an inline require where this property is used?

src/views/_proxied-dot-io/product-docs-landing/partials/featured-card/index.tsx
ashleemboyer: Would be on board with that, didn't make a comment on it since this is in `_prodixed-dot-io` and we have different standards in each part of the app.

src/views/_proxied-dot-io/product-docs-landing/index.tsx
zchsh: Great feedback here - have renamed to `vault-docs-landing` 👍 Could try to remove the view entirely, but I think then we'd have to put `components` somewhere NOT in the `pages` folder (so as to not have each component render as a page) 🤔

src/views/_proxied-dot-io/product-docs-landing/partials/use-case-card/types.ts
zchsh: applied this! (well, renamed the file first - should have thought through that and committed the suggestion first 😅 )

src/views/_proxied-dot-io/product-docs-landing/partials/flex-float/style.module.css
```diff
@@ -0,0 +1,22 @@
...
+  /*
+  We want items to have space between them when on the same line,
+  but we also want them to wrap to multiple lines when space is limited.
+  We use a flex layout with some interesting margin settings to achieve this.
+  Each item creates a buffer around it with margin, and the container offsets
+  the "edges" of this buffer with negative margin.
+  
+  Note as well that margin does NOT collapse in flex layouts,
+  which is why the item buffer is half the desired spacing.
+  */
```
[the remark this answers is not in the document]
zchsh: 🤦 Yep, absolutely, thanks for catching this 💖 Apparently my brain is still in "support IE mode" 😭 😎

kylemac:
@zchsh from slack DM:
- [ ] Developer Quick Start button link should point to the [developer quickstart](https://dev-portal-git-zsvault-shiny-docs-landing-hashicorp.vercel.app/docs/get-started/developer-qs) (rather than https://learn.hashicorp.com/collections/vault/cloud)
- [ ] Rename link (in the Developers section) from GitHub Samples to Code Samples

zchsh:
@kylemac I've rebased this to pull in fixes so that the build builds. Also made content updates to match the lastest feedback:

- [x] Developer Quick Start button now links to the [developer quickstart](https://dev-portal-git-zsvault-shiny-docs-landing-hashicorp.vercel.app/docs/get-started/developer-qs)
- [x] Link (in the Developers section) now reads `Code Samples` (was previously `GitHub Samples`
- [x] The CLI Quick Start button has been changed a black theme (aka "neutral primary"... was previously the yellow "vault primary" theme)
- [x] Replaces FPO asset with SVG from Dato

zchsh:
@BRKalow @ashleemboyer I think this PR will be ready to merge shortly - would love code review from y'all 🙇‍♂️

The changes since I think you last reviewed are pretty minimal:
- Content has been tweaked (see above messages)
- Since we're now loading an image from a Dato URL, rather than using `require`, content has been changed from `.js` to `.json` format.

BRKalow:
> ⚠️ Note that this does NOT replace the need for the [docs index.mdx file in vault/website/content](https://github.com/hashicorp/vault/blob/main/website/content/docs/index.mdx?rgh-link-date=2022-03-14T16%3A47%3A58Z). This content .mdx file still needs to be maintained, as it will still be used (each current version eventually becomes a previous version, and each previous version needs to have the docs/index.mdx file present).

This feels odd. Is there any way we can avoid this? 🤔


---

[opinion] The links feel a little squished together on these cards, on my first look it took me a few seconds to deduce each individual link. Maybe more spacing could help?
Screen Shot 2022-03-30 at 3 02 55 PM

zchsh:
@BRKalow I'm sure if there's a feasible way to avoid it with the current "ejected from versioned docs" setup... but I'll try to rubber duck it for a sec...

> If we remove the docs landing `.mdx` from versioned docs, then we'll need some way for versioned docs URLs to the docs landing (eg `/docs/v1.8.x`) to show the "custom" landing page...
> ... but the "custom" landing page is outside versioned docs... maybe we could do a rewrite or something? 🤔
> ... though we'll also want to show the "older version" warning I think, rewrite wouldn't help much with that... or maybe it would, we could detect the version number in the URL and show it based on that?...
> ... and maybe we'd also want links on the "custom" landing page to link to the expected "older version" of content... though even if we can detect the current version, because the "custom" landing page is NOT versioned, kind of "stuck at latest", there'd be no guarantee the links work anyways...

... based on that I think if we want to remove the need to maintain a [docs index file](https://github.com/hashicorp/vault/blob/main/website/content/docs/index.mdx?rgh-link-date=2022-03-14T16%3A47%3A58Z) as a "versioned" fallback to the custom landing page, I think the most sound approach is probably to register the "custom" landing page components in MDX and switch to having the "custom" be a less custom, normal part of our docs content.

Re: spacing - great feedback. Will double the current horizontal spacing 👍

BRKalow:
> ... based on that I think if we want to remove the need to maintain a [docs index file](https://github.com/hashicorp/vault/blob/main/website/content/docs/index.mdx?rgh-link-date=2022-03-14T16%3A47%3A58Z) as a "versioned" fallback to the custom landing page, I think the most sound approach is probably to register the "custom" landing page components in MDX and switch to having the "custom" be a less custom, normal part of our docs content.

Yeah, clearly you've given this some thought already haha. I'd say we can definitely roll forward as-is and maybe talk about how we want to handle this w.r.t versioning more holistically as a follow-up!

src/views/_proxied-dot-io/vault-docs-landing/components/use-case-card/style.module.css
```diff
@@ -0,0 +1,16 @@
+.heading {
+  composes: g-type-display-5 from global;
+  margin: 0 !important;
+}
+
+.body {
+  margin: 8px 0 0 0 !important;
```
BRKalow, line 7: Are the `!important` necessary here / is there a way we can make them not necessary?

src/views/_proxied-dot-io/vault-docs-landing/components/featured-card/style.module.css
```diff
@@ -0,0 +1,48 @@
...
+  composes: g-type-display-5 from global;
+  /* !important is necessary here to stop leaks from enclosing
+      @hashicorp/react-content component styles */
+  margin: 0 !important;
+}
+
+.body {
+  /* !important is necessary here to stop leaks from enclosing
+      @hashicorp/react-content component styles */
+  margin: 8px 0 0 0 !important;
```
BRKalow, line 28: Okay I think I answered my own question re the `!important` usage haha

BRKalow (approved):
Looks good! I think if we make this available to other products in its current form we could do with a hardening pass to add docs comments to the interfaces and make it a bit more re-usable, but for this deliverable I'd say we're good to go!

👏

zchsh:
☝️ @BRKalow i just had another thought, more future-y though and almost certainly outside the scope of this PR... maybe Ashlee's work in dynamically importing MDX components per-product in `DocsView` (in the new `dev-portal` UI) would be relevant?

To be more specific, I think the reason we shied away from going straight to having this "custom" page be part of the normal content workflows was mainly concern around performance, maintenance, and iteration on custom components... _assuming_ they'd have to be shared across all docs views. But, if we can conditionally expose custom components to specific pages (such as the ones in this PR to only the Vault `/docs` landing page, for example), then maybe that alleviates those concerns a bit. (I think there'd still be a concern around supporting custom landing page components on all past versions of landing pages... but that seems like a much more manageable scope than supporting custom landing pages components on all past versions of any docs page).

BRKalow:
@zchsh Yeah I think we'd need to agree on a set of eligible landing page primitives. We could definitely expose those, instead of the full page component, like you said!

PR #183, opened 2022-03-15
Upgrade nextjs-bundle-analysis and add `minimumChangeThreshold`

dstaley:

---

## Relevant links

 ksmy-branch
-->
- [Asana task](url) 🎟️

## What

This PR opts us into a canary release of `nextjs-bundle-analysis` that supports a new option `minimumChangeThreshold`.

## Why

Previous versions of `nextjs-bundle-analysis` would report any page that changed in size, leading to noisy PRs and a low signal-to-noise ratio, thus reducing the likelihood that a PR author would notice a significant change in page size. This PR adds a 1,000 byte minimum change threshold, which should at least make the changed pages list more accurate to the changes that are being made in the PR.

## How

Upgrades the package and defines `minimumChangeThreshold` in `package.json`

## Testing

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "flat": "^5.0.2",
     "identity-obj-proxy": "^3.0.0",
     "jest": "^27.4.7",
+    "nextjs-bundle-analysis": "^0.5.0-canary.1",
     "postcss": "^8.4.5",
     "postcss-flexbugs-fixes": "^5.0.2",
     "postcss-normalize": "^10.0.1",
@@ -148,6 +149,7 @@
   "nextBundleAnalysis": {
     "budget": null,
     "budgetPercentIncreaseRed": 20,
-    "showDetails": true
+    "showDetails": true,
+    "minimumChangeThreshold": 300
   }
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -148,6 +149,7 @@
   "nextBundleAnalysis": {
     "budget": null,
     "budgetPercentIncreaseRed": 20,
-    "showDetails": true
+    "showDetails": true,
+    "minimumChangeThreshold": 1000
```
BRKalow, line 153: So many bytes!

BRKalow (approved):

package.json
dstaley: How about we go to 300 and see how that feels? Anything under that and we'll be reporting pages as changed when it's really just Next adding small bits of metadata to pages (for example, [this PR](https://github.com/hashicorp/dev-portal/pull/115)).

BRKalow: Yeah, I mean I'm totally fine with 1k here too. We have the risk of death by a thousand cuts in that a bunch of tiny changes over time might not be visible until we're in a bad spot, but the noise reduction is definitely nice, and we should really be storing the analysis data and graphing it over time to track trends and such too.

PR #201, opened 2022-03-18
Adding <product>.slug DevDot docs in src/data/README

ashleemboyer:

---

## Relevant links

- [Preview link](https://github.com/hashicorp/dev-portal/blob/amb.add-dev-dot-json-data-docs/src/data/README.md) 🔎
- [Asana task](https://app.asana.com/0/1201010428539925/1201959792840210/f) 🎟️

## What

- Adds DevDot details for `{product-slug}.json` files in `src/data/README.md`
- Adds some `<details>/<summary>` elements to existing section of docs
- Adds `ProductName` type
- Adds `productSlugsToNames` map in `config/products.ts`
- Adds `typecheck` npm script in `package.json`

## Why

See the annotations via PR comment below.

## Testing

Nothing special to test, just make sure [the README file](https://github.com/hashicorp/dev-portal/blob/amb.add-dev-dot-json-data-docs/src/data/README.md) reads well and the PR checks pass. 😊

## Anything else?

Nope!

### DIFF
--- a/config/products.ts
+++ b/config/products.ts
@@ -1,5 +1,20 @@
-import capitalize from '@hashicorp/platform-util/text/capitalize'
-import { ProductGroup, ProductSlug } from 'types/products'
+import { ProductGroup, ProductName, ProductSlug } from 'types/products'
+
+/**
+ * A map of product slugs to their proper noun names.
+ */
+const productSlugsToNames: { [slug in ProductSlug]: ProductName } = {
+  boundary: 'Boundary',
+  consul: 'Consul',
+  hcp: 'HashiCorp Cloud Platform',
+  nomad: 'Nomad',
+  packer: 'Packer',
+  sentinel: 'Sentinel',
+  terraform: 'Terraform',
+  vagrant: 'Vagrant',
+  vault: 'Vault',
+  waypoint: 'Waypoint',
+}
 
 /**
  * Holds the product slugs in the order ProductSwitcher will render them.
@@ -18,7 +33,7 @@ const productSwitcherSlugs = [
  */
 const products: ProductGroup[] = productSwitcherSlugs.map((slugGroup) => {
   const productGroup = slugGroup.map((slug: ProductSlug) => {
-    const name = slug === 'hcp' ? 'HashiCorp Cloud Platform' : capitalize(slug)
+    const name = productSlugsToNames[slug]
     const url = `/${slug}`
     return {
       name,
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
     "start": "next",
     "test:e2e": "playwright test",
     "test:watch": "jest --watch",
-    "test": "jest"
+    "test": "jest",
+    "typecheck": "tsc -p ."
   },
   "engines": {
     "npm": ">=7.0.0",
--- a/src/data/README.md
+++ b/src/data/README.md
@@ -16,16 +16,12 @@ TODO - describe the properties used
 
 ## `<product slug>-install.json`
 
-This file is only used in DevDot. It contains the data and content needed for each product's downloads page (`/boundary/downloads`, `/consul/downloads`, etc.).
+This file is only used in DevDot. It contains the data and content needed for each product's downloads page (`/boundary/downloads`, `/consul/downloads`, etc.). There are four possible data properties that can be provided at the highest level: `doesNotHavePackageManagers`, `featuredTutorials`, `packageManagerOverrides`, and `sidebarMarketingCard`.
 
-**Data Properties**
+<!-- doesNotHavePackageManagers -->
 
-- [`doesNotHavePackageManagers`](#doesnothavepackagemanagers)
-- [`featuredTutorials`](#featuredtutorials)
-- [`packageManagerOverrides`](#packagemanageroverrides)
-- [`sidecarMarketingCard`](#sidecarmarketingcard)
-
-### `doesNotHavePackageManagers`
+<details>
+<summary><code>doesNotHavePackageManagers</code></summary>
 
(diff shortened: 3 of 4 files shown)


config/products.ts
```diff
@@ -1,5 +1,20 @@
...
+  consul: 'Consul',
+  hcp: 'HashiCorp Cloud Platform',
+  nomad: 'Nomad',
+  packer: 'Packer',
+  sentinel: 'Sentinel',
+  terraform: 'Terraform',
+  vagrant: 'Vagrant',
+  vault: 'Vault',
+  waypoint: 'Waypoint',
+}
```
ashleemboyer, line 17: Adding this removes the need for `@hashicorp/platform-util/text/capitalize` and also fixes a TS error that appeared after I added the `ProductName` type in `src/types/products.ts`.

package.json
```diff
@@ -20,7 +20,8 @@
     "start": "next",
     "test:e2e": "playwright test",
     "test:watch": "jest --watch",
-    "test": "jest"
+    "test": "jest",
+    "typecheck": "tsc -p ."
```
ashleemboyer, line 24: Added this for local use currently. We could probably add a PR check for this though! I noticed some other projects do it.

src/data/README.md
```diff
@@ -16,16 +108,12 @@ TODO - describe the properties used
 
 ## `<product slug>-install.json`
 
-This file is only used in DevDot. It contains the data and content needed for each product's downloads page (`/boundary/downloads`, `/consul/downloads`, etc.).
+This file is only used in DevDot. It contains the data and content needed for each product's downloads page (`/boundary/downloads`, `/consul/downloads`, etc.). There are four possible data properties that can be provided at the highest level: `doesNotHavePackageManagers`, `featuredTutorials`, `packageManagerOverrides`, and `sidebarMarketingCard`.
```
ashleemboyer, line 111: Added some `<details>/<summary>` elements below to make this file a little more manageable and also added Markdown comments before each to make the raw Markdown a little more manageable too.

src/types/products.ts
```diff
@@ -2,14 +2,26 @@ import { ProductMeta, Products } from '@hashicorp/platform-product-meta'
...
+  | 'Boundary'
+  | 'Consul'
+  | 'HashiCorp Cloud Platform'
+  | 'Nomad'
+  | 'Packer'
+  | 'Sentinel'
+  | 'Terraform'
+  | 'Vagrant'
+  | 'Vault'
+  | 'Waypoint'
```
ashleemboyer, line 15: Added this type because we know what these values are and we also have all of these products in DevDot (at varying levels).

src/data/README.md
```diff
@@ -6,8 +6,100 @@ The `src/data/` directory contains all of the product data needed for various vi
 
 The `<product slug>.json` files (e.g. `boundary.json`, `consul.json`) contain metadata for each product that is used across multiple views, including their .io site and Dev Portal pages.
 
+### .io properties
+
 The canonical list of options (including a brief description of their usage) can be found in the [type definition for `ProductData`](./types.d.ts).
 
+### DevDot properties
```
dstaley, line 13: Since these props live in files described by `src/data/types.d.ts` would you mind adding these properties there as well (feel free to use simplified descriptions since they're really well documented here)? That way we can enforce type safety when accessing these DevDot properties.

ashleemboyer: I'm not sure what you mean. Are you asking for duplicate types added to `src/data/types.d.ts`?

Also not sure what you mean by "That way we can enforce type safety when accessing these DevDot properties" because we do enforce type safety (just in a manual way) where we use `{product}.json` files in DevDot. One example here, where we cast the data imported as `Product`: https://github.com/hashicorp/dev-portal/blob/amb.add-dev-dot-json-data-docs/src/pages/waypoint/index.tsx.

For context there are plans to improve the `Product` type and break up the interface so it's clearer where the different properties are used, and hopefully that can happen next week in the frontend refinement cycle I'm in. 🤞🏻

dstaley: Ah okay! It looks like we have two duplicate definitions for the type of `src/data/<productSlug>.json`. I'm happy to follow up this PR adding the properties and docs from `src/data/types.d.ts` to `src/types/product.ts` and removing `src/data/types.d.ts` so we have a single source of truth for what those files look like.

ashleemboyer: Question on that, does having `types.d.ts` in `src/data` get us type checking automatically? I think we need a mechanism that always checks that the required data is provided in these JSON files since they're easily changeable by anyone with write access.

Also, I like the idea of calling a single merged type `ProductData`. The struggle with what we currently call `Product` is that it had properties that don't really make sense with the name of `Product`. The type is used in a _ton_ of files though, so maybe the merge you're proposing is a good first step before renaming anything. `ProductData` has a much smaller footprint in the codebase. 😅

dstaley: It does not, however we _can_ add a JSON Schema file that'll get us fancy stuff in supported editors (like VS Code), so that could help with all of our various `.json` files we'll be allowing authors to edit. That's a separate file though, so even in that case we can get rid of `types.d.ts` (which, as you pointed out, is only being referenced in `lib/cms`, so it should be a really easy fix to point that to `types/product.ts` instead). Happy to explore these in a follow up PR, so this one has a 👍 from me!

package.json
zchsh: I dig this, nice and convenient and +1 on adding a PR check or pre-commit verification hook or something 💯

zchsh (approved):

src/data/README.md
ashleemboyer: Sounds good! We can chat about this more next week. 😊

PR #207, opened 2022-03-21
chore: update DevDot to use updated "focus-ring" design tokens

didoo:

- [ ] ~~This PR's link has been added to the "📐 Pull Request" field on the Asana task~~

---

## Relevant links

- [Preview link](https://dev-portal-ocryjc6cf-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1201687774781544/1202000336018759) 🎟️

## Why

In https://github.com/hashicorp/design-system/pull/98 the HDS team has updated the design tokens for the "focus-ring" to support both the `action` and the `critical` colors.

## How

In this PR I have:
- updated the `@hashicorp/design-system-tokens` dependency to version `0.7.0`
- updated all the references to `focus-ring` design tokens and colors in the codebase
   - notice: I have also replaced a couple of `hex` color values with their corresponding design tokens

## Testing

I have tested where I could the focus (eg. in a `standalone link` instance). Not sure if there is a more exact way to test the changes, I'm open to suggestions.

## Anything else?

I have added a comment in the "files changed" about one possible error in a color that could be fixed using the right design token (unless is an intentional design decision).

One last thing: when looking at the design tokens names, remember that `token-color-focus-***` refers to the actual color values, while `--token-focus-ring-***` refers to the "focus-ring" entity that it's built using the "focus" colors.

---
  - https://app.asana.com/0/0/1202000336018759


src/pages/style.css
```diff
@@ -162,7 +162,7 @@ and negative margin.
 
   &:focus-visible {
     border: 1px solid var(--token-color-palette-blue-200);
```
didoo, line 164: @ashleemboyer can you double check if `blue-200` is the right color (maybe is a specific design decision)? for us in HDS this is `blue-300` which means this can be declared using the specific token `var(--token-color-focus-action-internal)`

ashleemboyer: @didoo I assume this is something that just hasn't been updated in a while, or since the focus tokens were added. I say we leave this one for now because the `.g-focus-border-and-box-shadow-light` class has a wide area of impact and there are plans to refactor how focus styles are handled for many components.

ashleemboyer (commented):
Code looks good currently after a quick pass, will finish reviewing in the morning after some testing. 😊

src/pages/style.css
didoo: OK for me. Maybe we can add a comment, in this case? Like a TODO? What do you think?

ashleemboyer: I think the TODO above `.g-focus-ring-from-box-shadow` will suffice. 😊

ashleemboyer (approved):
Just tested against the following, looks great! Thank you for the updates @didoo! ✨

- `g-focus-ring-from-box-shadow`
  - `InlineLink`: https://developer.hashi-mktg.com/swingset/components/inlinelink
  - `StandaloneLink`: https://developer.hashi-mktg.com/swingset/components/standalonelink
  - `Tabs`: https://developer.hashi-mktg.com/swingset/components/tabs
  - `VersionContextSwitcher`: https://developer.hashi-mktg.com/waypoint/downloads
- `g-focus-border-and-box-shadow-light`
  - `SidebarFilterInput`: https://developer.hashi-mktg.com/waypoint/docs
  - `Sidebar` “Skip to main content” bypass block (Tab from `SidebarFilterInput`): https://developer.hashi-mktg.com/waypoint/docs
  - `SidebarNavMenuItem`: https://developer.hashi-mktg.com/waypoint/docs
  - `TableOfContents` links: https://developer.hashi-mktg.com/waypoint/docs
- `Disclosure`: https://developer.hashi-mktg.com/swingset/components/disclosure

PR #209, opened 2022-03-23
fix: use npx for build:deploy-preview as it runs before npm install

BRKalow:

---

## Relevant links

- [Asana task](url) 🎟️

## What / Why

Use `npx` for our `build:deploy-preview` script as it runs before `npm install`. This is necessary because we do some file moving related to `node_modules` based on the state of the build cache during a Vercel build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "dev-portal",
   "private": "true",
   "scripts": {
-    "build:deploy-preview": "hc-tools ./scripts/content-repo-preview/build.ts",
+    "build:deploy-preview": "npx --package @hashicorp/platform-tools hc-tools ./scripts/content-repo-preview/build.ts",
     "build:webpack-only": "hc-tools next-build-webpack-only",
     "build": "next build",
     "lint": "next-hashicorp lint",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "dev-portal",
   "private": "true",
   "scripts": {
-    "build:deploy-preview": "hc-tools ./scripts/content-repo-preview/build.ts",
+    "build:deploy-preview": "npx --package @hashicorp/platform-tools hc-tools ./scripts/content-repo-preview/build.ts",
```
dstaley, line 5: Since this is running before `npm install`, does it make sense to pin the version of `@hashicorp/platform-tools` so that we can safely release breaking changes of the package without also breaking the build?

BRKalow: Ah good call, this is probably the right posture to take in this scenario 👍

BRKalow, line 5:
```suggestion
    "build:deploy-preview": "npx --package @hashicorp/platform-tools@0.2.0 hc-tools ./scripts/content-repo-preview/build.ts",
```

PR #213, opened 2022-03-23
Disable @typescript-eslint/explicit-module-boundary-types rule; Add @typescript-eslint/typedef rule

ashleemboyer:

---

## Relevant links

- [Preview link](https://dev-portal-git-ambdisable-explicit-module-boun-f2d996-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1201979821989355/1201661195283538/f) 🎟️

## What

- Removes the `@typescript-eslint/explicit-module-boundary-types` rule
- Removes all `eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types` comments
- Adds the `@typescript-eslint/typedef` rule

## Why

- `@typescript-eslint/explicit-module-boundary-types` requires adding return types on function components, which really isn't necessary in this project.
- Disabling `@typescript-eslint/explicit-module-boundary-types` removes the requirement for typing arguments to function components, so the `@typescript-eslint/typedef` rule is needed to turn that back on. It's currently set at `warn` since over 150 new warnings appear with that rule. Both of the `arrowParameter` and `parameter` options are enabled since both kinds of declarations are used for declaring function components.

## How

Discussed with the team and used the following resources:

- https://typescript-eslint.io/rules/explicit-module-boundary-types
- https://typescript-eslint.io/rules/typedef/#variabledeclarationignorefunction

## Testing

There isn't anything to test for this one. Just need to make sure all PR checks should pass.

## Anything else?

Nope!

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,5 +3,6 @@ module.exports = {
   extends: './node_modules/@hashicorp/platform-cli/config/.eslintrc.js',
   rules: {
     curly: 'error',
+    '@typescript-eslint/explicit-module-boundary-types': 'off',
   },
 }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,5 +4,12 @@ module.exports = {
   rules: {
     curly: 'error',
     '@typescript-eslint/explicit-module-boundary-types': 'off',
+    '@typescript-eslint/typedef': [
+      'warn',
+      {
+        arrowParameter: true,
+        parameter: true,
+      },
+    ],
   },
 }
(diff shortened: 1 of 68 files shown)


dstaley (approved):
👍 I wonder if we can write our own rule that only enforces types for arrow functions assigned to variables, rather than also applying it to arrow functions that are used as arguments (like `.map()`). But for the time being, I think setting this to `warn` instead of error was the right call!

ashleemboyer:
> 👍 I wonder if we can write our own rule that only enforces types for arrow functions assigned to variables, rather than also applying it to arrow functions that are used as arguments (like `.map()`). But for the time being, I think setting this to `warn` instead of error was the right call!

@dstaley that would be fun! @BRKalow taught us how to create rules recently. 🥳

PR #215, opened 2022-03-23
Add Button component

ashleemboyer:

---

## Relevant links

- [Preview link](https://dev-portal-git-ambadd-button-component-hashicorp.vercel.app/swingset/components/button) 🔎
- [Asana task](https://app.asana.com/0/1201979821989355/1201654639085740/f) 🎟️

## What/Why

- Adds new `Button` component along with a swingset page for testing
- Adds `copy` icon option for `SwingsetTestIcon` so `Button` tests are more interesting

## How

Resources used:

- [HDS Figma designs](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/HDS-Product---Components-%5BWIP%5D?node-id=2340%3A21001)
- [HDS Button previews](https://design-system-components-hashicorp.vercel.app/components/button)

## Testing

Explore [the Button swingset page](https://dev-portal-git-ambadd-button-component-hashicorp.vercel.app/swingset/components/button) and see if everything behaves as you expect.

## Anything else?

Nope!

### DIFF
--- a/src/components/button/index.tsx
+++ b/src/components/button/index.tsx
@@ -1,0 +1,13 @@
+import s from './button.module.css'
+import { ButtonProps } from './types'
+
+const Button = ({ onClick, text, type = 'button' }: ButtonProps) => {
+  return (
+    <button className={s.root} onClick={onClick} type={type}>
+      {text}
+    </button>
+  )
+}
+
+export type { ButtonProps }
+export default Button
--- a/src/components/button/index.tsx
+++ b/src/components/button/index.tsx
@@ -27,7 +27,7 @@ const Button = ({
       type={type}
     >
       {hasLeadingIcon && icon}
-      <span className={s.text}>{text}</span>
+      {text && <span className={s.text}>{text}</span>}
       {hasTrailingIcon && icon}
     </button>
   )
--- a/src/components/button/index.tsx
+++ b/src/components/button/index.tsx
@@ -19,14 +19,27 @@ const Button = ({
   const className = classNames(s.root, s[size], s[color], {
     [s.fullWidth]: isFullWidth,
   })
-  const hasLeadingIcon = icon && iconPosition === 'leading'
-  const hasTrailingIcon = icon && iconPosition === 'trailing'
+  const hasIcon = !!icon
+  const hasLeadingIcon = hasIcon && iconPosition === 'leading'
+  const hasTrailingIcon = hasIcon && iconPosition === 'trailing'
   const hasText = !!text
   const hasLabel = !!ariaLabel || !!ariaLabelledBy || !!ariaDescribedBy
 
-  if (!hasText && !hasLabel) {
+  if (!hasIcon && !hasText) {
     throw new Error(
-      'Icon-only `Button`s require an accessible label. Either provide the `text` prop or one of: `ariaLabel`, `ariaLabelledBy`, `ariaDescribedBy`'
+      '`Button` must have either `text` or an `icon` with accessible labels.'
+    )
+  }
+
+  if (hasIcon && !hasLabel) {
+    throw new Error(
+      'Icon-only `Button`s require an accessible label. Either provide the `text` prop or one of: `ariaLabel`, `ariaLabelledBy`, `ariaDescribedBy`.'
+    )
+  }
+
+  if (ariaLabel && ariaLabelledBy) {
+    throw new Error(
+      '`Button` does not accept both `ariaLabel` and `ariaLabelledBy`. Only provide one or the other.'
     )
   }
 
--- a/src/components/button/index.tsx
+++ b/src/components/button/index.tsx
@@ -1,10 +1,24 @@
-import s from './button.module.css'
 import { ButtonProps } from './types'
+import s from './button.module.css'
+import classNames from 'classnames'
 
-const Button = ({ onClick, text, type = 'button' }: ButtonProps) => {
+const Button = ({
+  icon,
+  iconPosition = 'leading',
+  onClick,
+  size = 'medium',
+  text,
+  type = 'button',
+}: ButtonProps) => {
+  const className = classNames(s.root, s[size])
+  const hasLeadingIcon = icon && iconPosition === 'leading'
+  const hasTrailingIcon = icon && iconPosition === 'trailing'
+
   return (
-    <button className={s.root} onClick={onClick} type={type}>
-      {text}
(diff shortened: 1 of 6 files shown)


ashleemboyer:
FYI I still need to fill in the props table on the swingset page, but this is ready for review otherwise. 😊

src/components/button/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const hasText = !!text
+  const hasLabel = !!ariaLabel || !!ariaLabelledBy || !!ariaDescribedBy
+
+  if (!hasIcon && !hasText) {
+    throw new Error(
+      '`Button` must have either `text` or an `icon` with accessible labels.'
+    )
+  }
+
+  if (hasIcon && !hasLabel) {
```
zchsh, line 34: Noticed the Swingset examples for `Icon and Text` are throwing this error. If we have text in the button, then I think we don't need the label, right?

```suggestion
  if (hasIcon && !hasLabel && !hasText) {
```

src/layouts/core-dev-dot-layout/core-dev-dot-layout.module.css
```diff
@@ -21,8 +21,12 @@
   & button {
     font-family: var(--font-body);
 
-    &:hover {
+    &:not(:disabled) {
```
zchsh, line 24: TLDR; very non-blocking random thought, feel free to ignore!

Re: the TODO comment above... would we be open to starting with these styles within the button component?

I'm definitely on the fence here; I can't really imagine a `button` where we wouldn't want these styles... however I think it shouldn't be too much of a tradeoff to repeat these styles in every button-ish thing we build.

Alternately... maybe we split this out to a kind of "normalize"-ish stylesheet? I suspect this might be something the design systems team has an opinion on 🤔

zchsh (approved):
✅ Noticed an issue with icon + text button examples in Swingset, but however that gets resolved this looks good to me ☺️

src/components/button/index.tsx
ashleemboyer: Uh oh! I didn't notice the error message pop up below these, will work on fixing it 😊

ashleemboyer: Fixed in https://github.com/hashicorp/dev-portal/pull/215/commits/c9818e981f0c32e705dceaeb7eb6a142702fd844! Added an `isIconOnly` boolean and use it in this check now. Great catch! ✨

src/layouts/core-dev-dot-layout/core-dev-dot-layout.module.css
zchsh: I kinda like the idea of "normalize-on-the-fly" - eg if there's some other `<button>`-using components that need to be normalized as well (eg the ones in the sidebar), we'd repeat these styles within that component (and helpers seem like a great way to abstract rather than repeat! 👍).

If we go the more traditional normalize route, I'm really not sure where that stylesheet would live 🤔 i guess in most projects it'd be a global stylesheet; in our case since we have the co-located dot-ios I suppose we'd need something scoped to `core-dev-dot-layout` 🤷‍♂️

kendallstrautman (approved):
LGTM! 👏

src/components/button/types.ts
```diff
@@ -0,0 +1,19 @@
+import { ReactElement } from 'react'
+
+type ButtonElementProps = JSX.IntrinsicElements['button']
+
+// TODO: add jsdoc to each prop
+export type ButtonProps = Pick<
```
kendallstrautman, line 6: I think this should be an `interface`! Especially since its extending the button element props.

src/components/button/types.ts
```diff
@@ -0,0 +1,19 @@
...
+  ButtonElementProps,
+  'disabled' | 'onClick' | 'type'
+> & {
```
kendallstrautman, line 9:
```suggestion
export interface ButtonProps extends Pick<
  ButtonElementProps,
  'disabled' | 'onClick' | 'type'
> {
```

src/components/button/types.ts
kendallstrautman: Interesting notes in[ the comments](https://www.typescriptlang.org/play#example/types-vs-interfaces) here re: interfaces v. types

kendallstrautman: It's up to you though 👐

ashleemboyer: Just pushed up some commits that have the outcome of making this an `interface`. 😊

BRKalow (approved):
👏

src/layouts/core-dev-dot-layout/core-dev-dot-layout.module.css
ashleemboyer: I really like the idea of using a helper for this! For this cursor thing specifically, we could do something like this in `src/pages/style.css`:

```
.g-button-cursor-pointer {
  &:not(:disabled) {
    cursor: pointer;
  }

  &:disabled {
    cursor: not-allowed;
  }
}
```

I assume we'd want a separate one for handling `font-family`, or maybe that one actually stays here since font probably won't change between buttons? 🤔

ashleemboyer:
Have now added the `id` and `name` props (since those are probably very commonly used, easy to get them out of the way now), and added details to the `types.ts` file and the swingset page.

src/layouts/core-dev-dot-layout/core-dev-dot-layout.module.css
zchsh: ☝️ yeah i dig that! another approach could be to tie it to the HTML element to make it a bit more generic, and to signal the intent (i think?) to apply this to all `button` elements, eg:

```
/* src/styles/normalize.module.css */
.normalize-button {
  font-family: var(--font-body);
  font-size: 100%;
  line-height: inherit;
  
  &:not(:disabled) {
    cursor: pointer;
  }
 
  &:disabled {
    cursor: not-allowed;
  }
}

/* src/components/some-button/some-button.module.css */
.root {
  composes: normalize-button from "src/styles/normalize.module.css";
  /* ... other implementation-specific styles */
}
```

personally kind of on the fence... a "standard" global `normalize.css` might be better than the above in some ways (though arguably a little less "encapsulated").

in any case, global styles definitely seem more than fine for now 👍 especially since this stuff for `button` feels really genuinely global (rather than being convenient short-term but hard to maintain long-term, which I think can sometimes happen when abstracting to global styles too early).

PR #220, opened 2022-03-24
fix: use faster URL to test glob redirects

dstaley:

---

## Relevant links

- [Asana task](url) 🎟️

## What

This PR switches our E2E test for glob-based redirects to use `/security/index.html` instead of a docs page.

## Why

In E2E tests, the site is running in preview mode, which generates pages on the fly. Due to a combination of factors, this leads to sometimes exceeding the default timeout for page load when we try to hit a `/docs` route (since it needs to fetch the page from the content API). This test failure leads to a flaky test, which isn't a good thing since we want PR authors to have confidence that their changes haven't broken something.

## How

By switching to `/security/index.html`, we can load a page that's generated from local data as opposed to remote data.

## Testing

A passing E2E test without a warning that the glob-based redirects test was flaky will be considered a passing test.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -1,11 +1,8 @@
 name: E2E Tests
-on:
-  pull_request:
-  push:
-    branches:
-      - main
+on: [deployment_status]
 jobs:
   test:
+    if: github.event.deployment_status.state == 'success' && github.event.sender.id == 35613825
     timeout-minutes: 60
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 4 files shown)


BRKalow:
@dstaley This is a good adjustment! Had this thought the other day...I wonder if we could key off of deployment status and run the checks against the vercel deploy preview, instead of running the dev server?

BRKalow:
Success! https://github.com/hashicorp/dev-portal/runs/5681534093?check_suite_focus=true

BRKalow (approved):
👏

PR #221, opened 2022-03-24
feat: stub in custom img component

zchsh:

---

## Relevant links

- [Preview link - modified caption example][caption example] 🔎
- [Preview link - unmodified examples][unmodified examples] 🔎
- [Asana task](https://app.asana.com/0/1201987349274776/1201999966887531/f) 🎟️

## What

This PR adds `MdxImg` and `ImageConfig` components, and uses them in `views/tutorial-view`.

## Why

To implement the image components in tutorial content to match design specifications. This includes giving authors the ability to add captions to images.

## How

- For basic styling, uses an MDX custom component in place of the default `img` component
- For captions, leverages the existing `ImageConfig` component to allow captions to be added to images by wrapping them in `<ImageConfig caption="Some image caption text">`.
    - Note that for now, captions must be plain text. If we want, we can augment this by using the `caption` prop as a slot, allowing it to accept either a string or JSX.

## Testing

- [ ] Open [the design spec in Figma](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Dev-Portal?node-id=6298%3A73017) for reference
- [ ] Visit examples on any other page, for example [/vault/tutorials/getting-started-ui/getting-started-ui][unmodified examples], and confirm that the implementation is to spec
- [ ] Visit [the manually modified caption example][caption example], to verify that the component is to spec when used with a caption

## Anything else?

- The `!important` styles are necessary due to cascading from `:global` scoped content styles
    - I'd love to follow this PR up with a PR to encapsulate those `:global` styles into custom MDX components. I think that if this is something we want to do (I would advocate for it strongly for the sake of custom component flexibility, happy to dive into details as to why), we should do it sooner rather than later. 
  
- I'm unsure if we really need `caption` in this initial implementation.
    - I chose to take a shot at implementing it because captions, and additional image configuration more broadly such as `hideBorder`, seem like like a jagged edge we need to solve for.;
 
- I have _not_ implemented the "open image in separate tab" feature 
    - For context, the design spec mentions functionality to open the image source in a separate window, using a button that appears on hover on desktop. On mobile, where hovering to show the button is not possible, the intent is for a tap of the image to open it in a separate window. 
    - This is a very newly proposed feature (within the last few days), and is not something we have on Learn currently, so I think it makes sense to handle it in a follow-up PR (perhaps ideas around this feature will change in the meantime).

[caption example]: https://dev-portal-git-zsadd-tutorial-img-component-hashicorp.vercel.app/vault/tutorials/getting-started-ui/getting-started-intro-ui
[unmodified examples]: https://dev-portal-git-zsadd-tutorial-img-component-hashicorp.vercel.app/vault/tutorials/getting-started-ui/getting-started-ui


src/views/tutorial-view/components/image-config/README.md
```diff
@@ -0,0 +1,26 @@
+# ImageConfig
```
zchsh, line 1: I'd love to get this into Swingset, but it seems like we're out of luck unless we take all components out of the `views` folder and have them flat in the `components` folder. This felt like too big of a compromise for the moment.

src/views/tutorial-view/components/image-config/index.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * this is an approximation of this component's markup when used within an MDX document, eg:
+ * <ImageConfig>
+ *
+ *   ![hashicorp logo](/img/favicons/favicon-192x192.png)
+ *
+ * </ImageConfig>
+ */
+const renderBasicMock = () =>
```
zchsh, line 18: These tests were ported over from learn - [see original component](https://github.com/hashicorp/learn/tree/master/components/image-config). I've made modifications to ensure tests are still working as expected with the revised component.

src/views/tutorial-view/components/image-config/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+  children: ReactNode
+}): ReactElement {
+  //  emit a warning if no meaningful props are present
+  if (!hideBorder && !caption) {
+    console.warn(
+      `Warning: <ImageConfig /> was initialized with default props. Please remove <ImageConfig /> if configuration is not needed.`
+    )
+  }
+
+  // TODO: fully deprecate the `hideBorder` prop, then remove this warning
```
zchsh, line 30: I feel like this TODO has been here for a while. Is this deprecation something we can feasibly do? And if so, is it being actively worked on? My assumption is the answer to both is "maybe" rather than a clear "yes", so for now I've retained the `hideBorder` functionality (despite being very on board with deprecating it).

src/views/tutorial-view/components/image-config/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+    throw new Error(
+      `Images in MDX must have an source defined. Use the format ![some alt text](path/to/img.jpg). Image details: ${JSON.stringify(
+        { src, alt, title }
+      )} `
+    )
+  }
+
+  // TODO: remove this once we're done reviewing
+  // (here for now because otherwise, previewing this seemed difficult)
+  const TEMP_DEV_CAPTION =
```
zchsh, line 52: Going to remove this before merge; I needed some way to test the caption prop so hooked into a specific image source (on [this page](https://dev-portal-git-zsadd-tutorial-img-component-hashicorp.vercel.app/vault/tutorials/getting-started-ui/getting-started-intro-ui)).

Related: we could, I think, _not_ implement the caption prop for now, since authors can't really use it yet. Or we could implement it now, and also backport these changes to the `learn` repo so that authors can start using `caption` (and the new component look & feel, apart from the system fonts) right away 👍

src/views/tutorial-view/components/image-config/get-validated-image-child.tsx
```diff
@@ -0,0 +1,75 @@
+import { Children, isValidElement, ReactElement, ReactNode } from 'react'
+
+/**
+ * Given the incoming children passed to `<ImageConfig />`,
+ * validate that the children are in the expected structure.
+ *
+ * Return the child `<img />` element if the structure is expected,
+ * throw an error otherwise.
+ */
+function getValidatedImgChild(children: ReactNode): ReactElement {
```
zchsh, line 10: I've tried to make this validation function robust to avoid misuse. The idea here is to ensure we have a single `` child being used with `ImageConfig`, and return that child element. This is not as straightforward as it could be since in MDX we expect the structure `<p></p>`, but we could also see a single `` child if authors use HTML rather than the usual `Alt text` markdown syntax.

I spent a bit more time on this since I think it'll come in handy for MDX custom components generally. Other components that have similar validation needs: `CodeBlockConfig`, `CodeTabs`, `Tabs`.

src/views/tutorial-view/components/image-config/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+import getValidatedImgChild from './get-validated-image-child'
+import MdxImg from '../mdx-img'
+// import classNames from 'classnames'
+import s from './image-config.module.css'
+
+/**
+ * Configure extra settings for images via MDX.
+ * Designed to wrap a single markdown image.
+ */
+export default function ImageConfig({
```
zchsh, line 11: This component was largely ported over from Learn - [original component](https://github.com/hashicorp/learn/tree/master/components/image-config).

src/views/tutorial-view/components/mdx-img/index.tsx
```diff
@@ -0,0 +1,26 @@
+import { ReactElement } from 'react'
+import { MdxImgProps } from './types'
+import classNames from 'classnames'
+import s from './mdx-img.module.css'
+
+function MdxImg({
+  src,
+  alt,
+  title,
+  noMargin,
```
zchsh, line 10: `noMargin` and `noBorder` are here for use with `ImageConfig`
- `noMargin` - when used with a caption, we want the `img` element to not have any margin (so that we can set margin once on the enclosing `figure`)
- `noBorder` - we want to deprecate this, but it's something in `learn` already so figured we need to support it until we can make the necessary content lift

src/views/tutorial-view/components/mdx-img/mdx-img.module.css
```diff
@@ -0,0 +1,30 @@
...
+    margin: 0;
+  }
+
+  &.noBorder {
+    box-shadow: none;
+  }
+}
+
+.img {
+  /* TODO: remove content styles which force !important here */
```
zchsh, line 22: Added exhaustive properties here, these overlap with `:global` CSS.

The intent here is that we remove [the `:global` css for `img`](https://github.com/hashicorp/dev-portal/blob/78d0143803bd5826bf933f507546111a5bafce11/src/components/dev-dot-content/dev-dot-content.module.css#L85). I'd be happy to handle that in a follow-up PR if everyone's on board with the approach 🙇‍♂️

src/views/tutorial-view/components/mdx-img/types.ts
```diff
@@ -0,0 +1,12 @@
+export interface MdxImgProps {
+  /** Source attribute for the image tag. */
+  src: JSX.IntrinsicElements['img']['src']
+  /** Alternate text for the image tag. Should be a text equivalent to the information conveyed in the image. As an example, the alternate text "Diagram of Humans or Machines Clients" is likely not sufficient because it does not express all of the same information as the diagram does for sighted users. It's important to provide the same information experience for all users regardless of ability. */
```
zchsh, line 4: Tried to add some descriptive comments here to match the work done in the Figma file 👍

src/views/tutorial-view/index.tsx
```diff
@@ -123,9 +123,7 @@ export default function TutorialView({
           />
           {isInteractive ? <InteractiveLabToggle /> : null}
         </header>
-        <Content
-          content={<MDXRemote {...content} components={MDX_COMPONENTS} />}
-        />
+        <MDXRemote {...content} components={MDX_COMPONENTS} />
```
zchsh, line 126: I believe we no longer want to use our `@hashicorp/react-content` component. For now, I think it's replaced by `src/components/dev-dot-content` (which I think we want to convert to individual MDX components).

src/views/tutorial-view/components/image-config/docs.mdx
```diff
@@ -0,0 +1,39 @@
+---
+componentName: ImageConfig
+---
+
+# ImageConfig
```
zchsh, line 5: I'd love to get this into Swingset, but couldn't get the `glob` matching for Swingset's `componentsRoot` option (in `next.config.js`) to work. Probably due to a lack of experience with `glob`!

I would love any input here, for now I just toggled between `componentsRoot` options (loading from `src/views/tutorials-view/components` in order to see this Swingset page).

zchsh:
@kendallstrautman @ashleemboyer This implements the MDX image component for `/tutorials` views. TLDR of notes:
- There's a temporary content change to demo the `caption` prop, I need to remove this before merging (but figured it'd be helpful for review, since I wasn't able to get the MdxImg component into Swingset without file re-organization)
- Would love to do some clean-up in follow-up PRs - namely, using this component in `/docs` land too (and removing some related `:global` CSS to clean up some `!importants`), and perhaps also implementing the "Open image in separate tab" feature added by design last week (though this feels a little lower priority)

ashleemboyer:
@zchsh have you considered putting this component in `src/components` instead of localizing to the tutorial view? I don't see any blockers for making it a main component right now.

zchsh:
@ashleemboyer great point - i did consider it, figured i'd wait to elevate it until we're using it on `/docs` views as well (which felt like it would balloon the scope of this PR a little bit beyond what we'd want). happy to add it to `components` earlier if ya'll would prefer that 👍

ashleemboyer:
@zchsh I don't see any reason to wait! It'll lower the amount of files changed when we do decide to use this component in docs views. 😊

zchsh:
sounds good, i'll move it to `components` 👍

src/views/tutorial-view/components/image-config/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+  hideBorder = false,
+  /** A caption to display below the image. */
+  caption,
+  /** An MDX `img`, e.g. `![some alt text](path/to/img.jpg)` */
+  children,
+}: {
+  caption?: string
+  hideBorder?: boolean
+  children: ReactNode
+}): ReactElement {
```
ashleemboyer: Could you move the comments on these props and the props interface definition to a new `types.ts` file local to this component?

src/views/tutorial-view/components/image-config/index.tsx
ashleemboyer: I don't have context on the status of deprecation, but it does seem feasible to do IMO. Just did a search on Sourcegraph and it looks like `hideBorder` is still commonly used.

src/views/tutorial-view/components/image-config/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+  // Validate that the { src, alt, title } props are present
+  const { src, alt, title } = imgElement.props
+  // Validate that src is present, and is a non-empty string
+  if (typeof src !== 'string' || src == '') {
+    throw new Error(
+      `Images in MDX must have an source defined. Use the format ![some alt text](path/to/img.jpg). Image details: ${JSON.stringify(
+        { src, alt, title }
+      )} `
+    )
+  }
```
ashleemboyer: What do you think about adding a console warning for `alt` being `undefined` or `null`? Even if an image is decorative (doesn't require text to be written out), it's important for `alt` to be set to an empty string in those cases. Reference: https://www.w3.org/WAI/tutorials/images/decorative/

src/views/tutorial-view/components/image-config/index.tsx
ashleemboyer: I think the caption prop is a great addition and we should leave it in. Will save us time down the road. 😊

src/components/mdx-img/index.tsx
```diff
@@ -0,0 +1,26 @@
+import { ReactElement } from 'react'
+import { MdxImgProps } from './types'
+import classNames from 'classnames'
+import s from './mdx-img.module.css'
+
+function MdxImg({
```
ashleemboyer, line 6: Can you tell me more about your reasoning behind this name? I think `Image` would be an appropriate name, because we already have `Heading` and `Text` for example that are used in MDX content on docs views.

src/components/mdx-img/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+  src,
+  alt,
+  title,
+  noMargin,
+  noBorder,
```
ashleemboyer, line 11: What are some of the use cases for these two props?

src/components/mdx-img/types.ts
```diff
@@ -0,0 +1,12 @@
+export interface MdxImgProps {
```
ashleemboyer, line 1: Could you update this so the docs comments are always multi-line (like `Button`, as a recent example) and also put an empty line between each property for improved readability?

src/views/tutorial-view/index.tsx
ashleemboyer: I don't see `DevDotContent` going away totally since some of it's global handling is meant to be so for consistency. The following block for example is easier to keep in a global setting IMO:

```css
.h1,
.h2,
.h3,
.h4,
.h5,
.h6 {
  & + ul,
  & + ol {
    margin-top: 12px;
  }

  & + p {
    margin-top: 0;
  }

  /*
  TODO: these values are a special request from design for the purposes of the
  inital preview. They make <code> elements within heading elements a little
  less jarring.
  */
  & code {
    font-size: 0.6em;
    font-weight: 500;
  }
}
```

A side note for posterity, we are using quite a few custom components already here: https://github.com/hashicorp/dev-portal/blob/main/src/layouts/sidebar-sidecar/utils/_local_platform-docs-mdx/index.jsx

- `Tabs` and `Tab`
- `DocsAnchor` (uses `InlineLink` under the hood)
- `Heading` for all h1-h6 elements
- `Text` for p elements

ashleemboyer (changes requested):

src/views/tutorial-view/index.tsx
zchsh: ☝️ Great point, sibling selectors are hard to move out of the `:global` context.

The `code` though is to me an example of something we should move into the `Heading` component. So I think we should still try to convert `DevDotContent` to individual components where possible, so that styles are well-encapsulated and don't risk leaking into custom components (this is something we've already had a struggle contending with when styling MDX content in docs sites, and markdown content on `.com` as well).

src/views/tutorial-view/components/image-config/get-validated-image-child.tsx
```diff
@@ -0,0 +1,75 @@
...
+  // Validate that there is exactly one child element
+  const childCount = Children.count(children)
+  const validChildren = Children.toArray(children)
+  if (childCount !== 1 || validChildren.length !== 1) {
+    throw new Error(
+      `In ImageConfig, found ${childCount} total children and ${validChildren.length} valid children. Please ensure that ImageConfig has exactly one child element, and ensure it is a valid image element.`
+    )
+  }
```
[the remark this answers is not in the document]
zchsh: There are some examples here: https://blog.agney.dev/react-children-count/ - agreed that it's hard to picture the JSX. Intent here is to try to make the error message as helpful as possible.

src/views/tutorial-view/components/image-config/get-validated-image-child.tsx
```diff
@@ -0,0 +1,75 @@
...
+  // Validate that the child is a ReactElement
+  const onlyChild = validChildren[0]
+  if (!isValidElement(onlyChild)) {
+    throw new Error(
+      `In ImageConfig, found child that does not seem to be a valid React element. Please ensure that ImageConfig contains a valid image element.`
+    )
+  }
```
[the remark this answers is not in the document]
zchsh: Moved the file first so couldn't apply directly, but have made this change! ty 🙇

src/views/tutorial-view/components/image-config/get-validated-image-child.tsx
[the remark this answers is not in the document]
ashleemboyer: This makes sense, thank you for the link!

src/views/tutorial-view/index.tsx
ashleemboyer: Right, not in disagreement about using custom components for MDX content. We've been on the same page there all along. 😊

src/components/image-config/index.test.tsx
```diff
@@ -0,0 +1,96 @@
+import { render, screen } from '@testing-library/react'
+import '@testing-library/jest-dom/extend-expect'
+
+import ImageConfig from '.'
+
+const testImage = (
```
kendallstrautman, line 6: yay, ty for writing these tests 🥳

src/components/image-config/index.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+  beforeAll(() => mockConsoleWarn.mockImplementation(() => null))
+  afterAll(() => mockConsoleWarn.mockRestore())
+
+  it('should render children', () => {
+    render(<ImageConfig>{testImage}</ImageConfig>)
+
+    expect(screen.getByAltText('hashicorp logo')).toBeInTheDocument()
+  })
+
+  it('should apply `noBorder` styling if `hideBorder` is provided', () => {
```
kendallstrautman, line 37: `hideBorder` is a deprecated property for this image config in Learn. I'm not sure the history there but I see so many warnings about this in the build logs. I wonder if we should just do away with it altogether in dev portal.

kendallstrautman: Perhaps we need to carry it over though since its still used in the content

src/views/tutorial-view/components/image-config/index.tsx
kendallstrautman: Oh! just saw this note, I added another one in the 'files' view. This deprecation happened before my time. Actually deprecating it would require us to collaborate with content authors to remove the instances of it in the `tutorials` repo.

src/views/tutorial-view/components/image-config/index.tsx
kendallstrautman: Authors are planning on making dev dot content changes on a separate branch in tutorials so they could use this feature there? Or yes we could backport to Learn! Either way.

src/views/tutorial-view/components/image-config/get-validated-image-child.tsx
kendallstrautman: In looking at the tutorials repo, it seems like _all_ image references in tutorials use the plain markdown syntax. I still think this is a good thing to support though!

src/views/tutorial-view/components/mdx-img/mdx-img.module.css
kendallstrautman: I think removing global image styles for dev dot sounds good! The global styles will probably hurt more than they help imo.

src/views/tutorial-view/index.tsx
kendallstrautman: What would make the most sense to me is if we actually call `DevDotContent` (or whatever the name is) here, instead of at the sidebar layout level. Because we want these mdx-type content styles to be scoped to this component and not _everything_ in the layout. That would take additional refactoring though. So I'm fine with not doing it now, but any thoughts on this change? @ashleemboyer

kendallstrautman:
Screen Shot 2022-03-28 at 4 18 36 PM

The un-captioned image borders look very dark / slightly off. I'm not sure if this is just a stub or a fully styled state.

src/components/image-config/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+ * Designed to wrap a single markdown image.
+ */
+export default function ImageConfig({
+  /** Hide borders (temporarily configurable while we remove baked-in borders from existing images) */
+  hideBorder = false,
+  /** A caption to display below the image. */
+  caption,
+  /** An MDX `img`, e.g. `![some alt text](path/to/img.jpg)` */
+  children,
+}: {
```
kendallstrautman, line 17: I'm not sure if this was brought up before but could we extract this into an `interface` definition above the component? Just seems a bit easier to read imo :)

src/components/image-config/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Configure extra settings for images via MDX.
+ * Designed to wrap a single markdown image.
+ */
+export default function ImageConfig({
+  /** Hide borders (temporarily configurable while we remove baked-in borders from existing images) */
+  hideBorder = false,
+  /** A caption to display below the image. */
+  caption,
```
kendallstrautman, line 14: Should a default caption be set here if it's optional? Just looking at the default props warning below, curious how you see this working in practice. I would imagine that most images at the onset won't have captions.

src/components/mdx-img/index.tsx
kendallstrautman: I think it's interesting to have a way to define which components are used in MDX. Presumably there will be other views on DevDot where we could use a generic `Image` component that needs to be different from this one. Maybe just a comment in the file suffices for now.

src/components/image-config/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+import MdxImg from '../mdx-img'
+import s from './image-config.module.css'
+
+/**
+ * Configure extra settings for images via MDX.
+ * Designed to wrap a single markdown image.
+ */
+export default function ImageConfig({
+  /** Hide borders (temporarily configurable while we remove baked-in borders from existing images) */
+  hideBorder = false,
```
kendallstrautman, line 12: What's even more confusing to me about this prop RN (besides it being in a perma deprecated state) is that is basically has no effect on the actual rendering of the component — it always shows a border. I wonder if there is a way to just not define it as an option at all, but make it so if its passed the component doesn't break?

src/components/mdx-img/mdx-img.module.css
```diff
@@ -0,0 +1,28 @@
...
+}
+
+.img {
+  /* TODO: remove content styles which force !important here */
+  border-radius: 0 !important;
+  background-color: transparent !important;
+  display: block;
+  margin: 0 !important;
+  max-width: 100% !important;
+}
```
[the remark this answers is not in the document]
kendallstrautman: I think this is an example of why we need to be careful with global element styles. I _feel_ like this responsibility should live with the component.

src/views/tutorial-view/components/image-config/index.tsx
ashleemboyer: After all the convo we've had on this so far, honestly it seems like we can just totally remove the prop in DevDot and not add a deprecation warning, as long as we have an Asana task for removing all uses of it in the tutorials repo. This prop seems like an easy one to locate usage of with VS Code's search or Sourcegraph's.

src/views/tutorial-view/components/mdx-img/mdx-img.module.css
ashleemboyer: I'm in agreement that the styles can and should be moved from `DevDotContent`, we just need to make sure that the styles get applied to all content in the same way it currently is for consistency.

src/components/mdx-img/mdx-img.module.css
```diff
@@ -0,0 +1,28 @@
+.root {
+  background-color: var(--token-color-surface-primary);
+  box-shadow: var(--token-surface-base-box-shadow);
```
ashleemboyer, line 3: Can we clarify with design if borders via box-shadow are meant to be on all images on DevDot? I feel like shouldn't be applying borders to all images and we haven't been in docs content. It doesn't seem like something we can automate because we'd have issues with some images such as window screenshots from Mac:

image

Image location: https://developer.hashi-mktg.com/waypoint/docs/getting-started#viewing-the-web-ui

src/components/image-config/index.tsx
zchsh: I think the `TEMP_DEV_CAPTION` thing I did made the plan here opaque - if there's no `caption` provided, we'll render a plain `img` tag without a caption instead of a `figure` with a caption. Now that we have the `image-config` component in the main components folder and showing the example in Swingset, I'll remove the `TEMP_DEV_CAPTION` thing and that should make this more clear 🙇‍♂️

src/components/mdx-img/mdx-img.module.css
zchsh: ☝️ This is a great question. I feel like this relates to the `hideBorder` prop in Learn... Maybe this is a reason we need to retain that prop.

In Docs contexts, we can set the default to `true` to match existing behaviour (at present I believe putting a border around images in Docs never happens?). Will confirm this with design - thanks for flagging, this'll definitely come up in the follow-up PR to use `Image` for Docs MDX content as well 👍

src/components/mdx-img/mdx-img.module.css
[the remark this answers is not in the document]
zchsh: ☝️ @ashleemboyer ah yeah excellent clarification - I've been saying `:global` but really intended as "styles leaking in from other components", which don't have to be global. Targeting HTML elements rather than classNames in "wrapper" components has the same effect. And the specificity often ends up being an issue in either case.

Very much agreed on the point of consistency. In the context of this PR, my aim was to add exhaustive (even if duplicative) styles scoped only to Tutorials, based on the design specs I have from Brayden, and to follow-up with a PR that reconciles any differences between docs and learn. Let me know if that makes sense; happy to try to do the `/docs` reconciliation in this PR, but worry that it will balloon the scope quite a bit.

src/components/mdx-img/index.tsx
zchsh: The reasoning was that we likely don't want to use this component outside of MDX contexts. The component is intended solely for use as an MDX custom component (I think in JSX contexts, the use of `MdxImage` and `ImageConfig` would be an odd and opaque way to get the desired markup and styles). As well, I was a little worried about colliding with the `next/image` import name (often `Image`).

Can definitely see how a comment might be more useful though. I've renamed `MdxImage` to `Image`. I think another route we could go here, if capturing that intent of "some components should be used exclusively in MDX, not JSX" resonates, is to rename `ImageConfig` to `MdxImageConfig`.

src/views/tutorial-view/components/image-config/index.tsx
zchsh: 💖 Great callout - I've added this in!

Unfortunately led to a bit of a rabbit hole - the one case that looks difficult is for those cases where we want to set `alt=""`. In JSX this would be trivial. In MDX though, it looks like `` gives `alt` as undefined, and I'm not sure how to get `alt` to be an empty string (other than using `` directly). Have added [a task to discuss](https://app.asana.com/0/1201987349274776/1202042900781423/f), I feel like it's something we could address in more detail outside the scope of this PR.

Let me know what you think! :🙇‍♂️

src/views/tutorial-view/components/image-config/index.tsx
```diff
@@ -0,0 +1,72 @@
+import { ReactElement, ReactNode } from 'react'
+import getValidatedImgChild from './get-validated-image-child'
+import MdxImg from '../mdx-img'
+// import classNames from 'classnames'
```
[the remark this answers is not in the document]
zchsh: Yep, great call - 🔪 removed1

src/components/mdx-img/mdx-img.module.css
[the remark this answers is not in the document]
kendallstrautman: Good clarification 👐 I think its okay if we have some style duplicity now given the speed with which we need to move. However we can refactor so that these docs level sweeping styles don't interfere with the Learn components is probably the best path forward? Or we'll just need to be okay with the override debt we're taking on and have a plan to adjust it later.

src/views/tutorial-view/components/mdx-img/index.tsx
zchsh: 👆 I think solving the `noMargin` case with CSS would remove our model of having component styles encapsulated within each component. When we start writing styles that target HTML primitives in CSS modules, we complicate the guarantee in any nested components that their styles will not be affected by parent components.

In this specific case this is relatively trivial - at present, we don't expect anything but `` to be rendered within `figure`. Even so, there could be a relevant scenario on the horizon - it's likely we'll eventually support more arbitrary `caption` content rather than just plain text (for example, we could support links in captions). What if we have a case where we want to show an inline "external link" icon in the `caption` area? Then our `.figure img` CSS selector inadvertently targets that element.

I don't mean to suggest that this scenario is necessarily likely, or that we'll be able to come up with such examples in every possible case. But this aim for full encapsulation was the broader reasoning behind our adoption of CSS modules. There's a bit more detail on this specific case in Notion as well (search for `Issue: Generic selectors can be hard to override`).

Let me know if that context is helpful, happy to add anything else I can, or revisit the approach here if that rationale seems off 🙇‍♂️

src/views/tutorial-view/components/image-config/index.tsx
zchsh: ☝️ @ashleemboyer Base on your comment on the docs situation, I think we should hold off on removing `hideBorder` until we handle the `docs` use case. I believe in `docs` we'll need it, and we may even want to remove the deprecation warning 🤔 . I think we could conditionally render a warning for Learn content, or even ignore the `hideBorder` prop but only in Learn contexts, if the education team is taking a harder line on never using the prop.

src/views/tutorial-view/components/image-config/index.tsx
zchsh: ☝️ @kendallstrautman If we want to backport to `learn` - what's the best board on which to surface that as a task? I think it could be cool to do if authors want it, though I imagine it might be lower on the list than other priorities. (Random thought, but if we do want to backport, I think we could wait till we have the docs use case accounted for as well in our `dev-dot` `ImageConfig`, and _then_ do the backporting).

src/views/tutorial-view/index.tsx
zchsh: With this PR in mind, I think we're in a decent spot on this front, right? Plan being:

- This PR sets up custom MDX `img` → `Image` for the tutorials view (and also `ImageConfig`)
- Follow-up PR will use  `img` → `Image` on the docs view too (we'll need to reconcile some styles, this will involve moving any generic `img` styles in `DevDotContent` into the `Image` component)
- Future work will involve converting all non-sibling selector stuff in `DevDotContent` to MDX custom components, similar to the existing `Heading` and `Text`

 I've set up [an Asana task](https://app.asana.com/0/1201987349274776/1202042900781425/f) to discuss that "Future work" part. I've kicked it off with my opinions and perspective on what our acceptance criteria should be, based on past woes dealing with styling markdown and MDX across learn, docs, and dot-com via our `@hashicorp/react-content` component.

src/components/image-config/index.test.tsx
zchsh: No problem, if by "writing" you mean "copypasta-ing from `learn`" 😝

src/components/image-config/index.tsx
zchsh: 👍  Ah yep, great call - did for `Image`, now done for `ImageConfig` too ✅

zchsh:
@kendallstrautman @ashleemboyer Thanks for the reviews 🙇‍♂️ 💖  I've done my best to address feedback, hopefully didn't miss anything 🤞 

I think at a broad level, my intent so far has been to keep this scoped to "add `Image` to the `/tutorials` views" for now, and tackle the very related MDX custom components and consistency-with-docs concerns in separate follow-up PRS. Happy to adjust that if needed 👍

src/components/mdx-img/index.tsx
ashleemboyer: To clarify, my comment was only asking for context, not asking for a rename. If we want this component to only be used in MDX, I think that's fine and we don't have to use a more general name of `Image`. 💖

src/components/mdx-img/index.tsx
ashleemboyer: Didn't see the comment explaining these elsewhere because the `Files` view changed as I was reviewing, but resolving because I've since found my answer in another comment. 😊

src/components/mdx-img/mdx-img.module.css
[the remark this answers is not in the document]
ashleemboyer: > follow-up with a PR that reconciles any differences between docs and learn

This is totally fine, I just didn't know this was the plan until I asked about it. 😊

src/components/mdx-img/mdx-img.module.css
ashleemboyer: Sounds fine to address in a follow-up PR. 👍🏻

I think the necessity of `hideBorder` depends on the necessity of having borders set on images. I have no context about why borders are being added to images, so I'm not currently convinced they're necessary for any content. My current stance is that all content needs to be aligned, and I think it's easier to bring alignment sooner rather than take it out for half our content later (like several PRs from now, not in a PR or two from now).

src/components/mdx-img/index.tsx
zchsh: Ah cool, yeah I don't feel strongly so figured I might as well go with `Image` as you suggested - I dig the consistency with `Heading` and `Text`, and like that it now sits alongside `ImageConfig`. The `next/image` thing is pretty minor; i don't think there's much of a genuine risk for confusion, the name overlap just jumped out to me (and we can always adjust later, eg if we were to try to use `next/image` in MDX). And I guess another way to make the MDX intent clear could be a `components/mdx` folder or something (similar to `author-primitives` i guess? and pending swingset support for nested folders).

Happy to go with either `MdxImg` or `Image`, in the absence of anyone having a strong preference I'll leave as is with `ImageConfig` & `Image` ☺️

zchsh:
@kendallstrautman I forgot to address the weird border on that image:

> Screen Shot 2022-03-28 at 4 18 36 PM
> 
> The un-captioned image borders look very dark / slightly off. I'm not sure if this is just a stub or a fully styled state.

I believe the darker border is baked into the image. Previously the conflict between the CSS border and pixels-in-the-image border wouldn't have been as noticeable since we had 90˚ rather than rounded corners. I suspect we'll still want to proceed with the designs as spec'd, with the rounded borders, but CC @braydenlove for visibility on this.

src/components/image/types.ts
```diff
@@ -0,0 +1,30 @@
...
+   * Avoids adding margin to the component.
+   * For use in other elements, such as ImageConfig
+   */
+  noMargin?: boolean
+  /**
+   * Avoids adding border to the component.
+   * For use in other elements, such as ImageConfig
+   */
+  noBorder?: boolean
+}
```
ashleemboyer, line 30: Appreciate the comment formatting changes, empty lines between each will also be appreciated 🙏🏻

```suggestion
export interface ImageProps {
  /**
   * Source attribute for the image tag.
   */
  src: JSX.IntrinsicElements['img']['src']

  /**
   * Alternate text for the image tag. Should be a text equivalent to the
   * information conveyed in the image. As an example, the alternate text
   * "Diagram of Humans or Machines Clients" is likely not sufficient because
   * it does not express all of the same information as the diagram does
   * for sighted users. It's important to provide the same information
   * experience for all users regardless of ability.
   */
  alt: JSX.IntrinsicElements['img']['src']

  /**
   * Title attribute for the image. Typically only used if the image
   * is nested in an <a /> tag and needs to serve as text for the link.
   */
  title?: JSX.IntrinsicElements['img']['title']

  /**
   * Avoids adding margin to the component.
   * For use in other elements, such as ImageConfig
   */
  noMargin?: boolean

  /**
   * Avoids adding border to the component.
   * For use in other elements, such as ImageConfig
   */
  noBorder?: boolean
}
```

ashleemboyer (approved):
2 comments, but approving so you don't have to re-request review after these last 2 requested changes! 💖

braydenlove:
> 

Gosh some of these images are going to be a nightmare :(. Let me dig in and see if I can find a solution...

src/components/mdx-img/mdx-img.module.css
braydenlove: I think I can live with this sort of style (dropshadow UI element inside a container with a border). It honestly looks very strange either way so at least we would be treating it consistently. The image with the border baked in is another story though, we might just need to find something that looks mostly good for all images and try to update older ones that do not work.

src/views/tutorial-view/components/mdx-img/index.tsx
ashleemboyer: Sounds good! No firm opinion on this specific thing, just looking for context. I do have opinions on what we render for `caption`, so I'm glad we're forcing that to be a string for now, but that's for a later discussion. 😊

src/views/tutorial-view/index.tsx
ashleemboyer: I added a comment to the new discussion task that links to our existing task for MDX components: https://app.asana.com/0/1201987349274776/1201653123138729/f

src/components/mdx-img/mdx-img.module.css
ashleemboyer: @zchsh @kendallstrautman @braydenlove Okay so to summarize there are two separate issues here: border via `box-shadow` and rounded corners via `border-radius`.

I think we can apply consistent values across all images added via MDX and then come up with solutions (such as with the `noMargin` and `noBorder` props) for edge cases (such as images with built-in borders or drop shadows) as we go. Any objections?

braydenlove: I chatted with @zchsh about it and we have a plan. I will have Zach walk you through what we discussed as it will be easier to follow :)

zchsh:
An update here since we've done a couple syncs outside of this PR: broadly, we want to commit to the new design spec, cause consistent borders with rounded corners are nice.

- Our plan is to remove borders built into image pixels at the source. We'll try to get content folks to help us. We also have some automation tools up our sleeves (`imagemagick`) to help with cropping off borders. It'll still be probably half a day's work to get together a list of images that need to have their borders removed, and remove them
- A fallback for the above is to ask authors to use `<ImageConfig hideBorder />` for cases where there's no clear path to making the image look nice with a border around it.

A couple of related design notes:
- The latest design spec has 6 px border radius, which differs from the existing 5px radius on `/docs`. This change is cause we may have a "open image in new tab" button within that radius, and from the design system side these nested border radii need to work nicely together.
- We'll retain `background: white` for images, since in eventual-dark-mode we want to make sure image backgrounds are white even if the "surface primary" colour changes based on `prefers-color-scheme` stuff. Thanks Ashlee for catching this 🙇‍♂️

cc @kendallstrautman @braydenlove @ashleemboyer for visibility.

src/components/mdx-img/mdx-img.module.css
zchsh: ☝️ @ashleemboyer yeah, this sounds good!

- Re: `box-shadow` for "border", this is coming from the design system. I think it's mostly just an annoying semantics thing, but is generally a non-issue - an alternate way of achieving a border-ish visual effect. Personally I'm going to continue to refer to the shadow on the `Image` component as a "border" cause that's what it looks like to me, even if it's not using the `border` CSS property. Happy to adjust this if anyone finds it confusing (I certainly do, but less confusing than talking about image's "box-shadow"s).
- Re: `border-radius`, with `<ImageConfig hideBorder />` authors can remove both the `box-shadow` and `border-radius`, giving them a very "plain" representation of their image. This will act as a fallback if we're not able to make all the content changes we want (content changes being: crop built-in borders out of images for Vault and Waypoint learn content before June 20th)

I'd propose we start by rolling out this change as-is. The issues we're seeing, being the black borders on some images now looking wonky due to duplicate bordering, were already technically "broken": they already had duplicate borders, the duplicate borders just didn't look quite as bad.

I don't think we need an additional props solution - if authors can't find a way to make an image work with a border around it, they can use `hideBorder` as an escape hatch.

kendallstrautman (approved):

src/components/image/image.module.css
```diff
@@ -0,0 +1,29 @@
...
+  justify-content: center;
+  margin: 24px auto;
+
+  &.noMargin {
+    margin: 0;
+  }
+
+  &.noBorder {
+    box-shadow: none;
+    border-radius: 0;
```
kendallstrautman, line 18: 👍

PR #222, opened 2022-03-25
Throwing error in App if needed product data is missing

ashleemboyer:

---

## Relevant links

- [Preview link](https://dev-portal-git-ambadd-check-for-product-hydration-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1201979821989355/1201980530846413/f) 🎟️

## What

- Adds a check in `_app.tsx` on DevDot pages that need product data hydrated.

## Why

- I was going to write this out in a README or something, but decided an error might be more effective on the fly. This will help remind us to include a `product` property in every DevDot page component's props. Currently, it's required for `ProductSwitcher`, but some changes will come soon to the main header navigation and more properties in that object will be required.

## How

- Mainly needed to figure out which pages to throw this error on, so I started by just checking if the current page was `/`, then worked through main page paths one by one. That led to getting rid of `/_proxied-dot-io` and `/swingset` pages from the check.

## Testing

There's nothing to manually test here. If there were pages that were failing the check, we would see the build checks below failing.

## Anything else?

Nope!


src/pages/_app.tsx
```diff
@@ -33,6 +34,23 @@ const { ConsentManager, openConsentManager } = createConsentManager({
...
+  const isDevDotPage =
+    !currentPath.startsWith(`/_proxied-dot-io`) &&
+    !currentPath.startsWith(`/swingset`)
+  const isNotHomePage = currentPath !== '/'
+  const isMissingProductData = !pageProps.product
+  if (isDevDotPage && isNotHomePage && isMissingProductData) {
+    throw new Error(
+      `\`product\` property is missing from \`pageProps\` prop on path: \`${currentPath}\`. Product data is needed to hydrate \`CurrentProductContext\`.`
+    )
+  }
```
BRKalow, line 52: what do you think about gating this behind `if (process.env.HASHI_ENV !== 'production')`? Realistically we shouldn't need to do this check in production once the app has been built and the pages have been rendered.

src/pages/_app.tsx
```diff
@@ -33,6 +34,23 @@ const { ConsentManager, openConsentManager } = createConsentManager({
...
+  const currentPath = useCurrentPath({ excludeHash: true, excludeSearch: true })
+  const isDevDotPage =
+    !currentPath.startsWith(`/_proxied-dot-io`) &&
+    !currentPath.startsWith(`/swingset`)
+  const isNotHomePage = currentPath !== '/'
+  const isMissingProductData = !pageProps.product
+  if (isDevDotPage && isNotHomePage && isMissingProductData) {
+    throw new Error(
+      `\`product\` property is missing from \`pageProps\` prop on path: \`${currentPath}\`. Product data is needed to hydrate \`CurrentProductContext\`.`
+    )
```
BRKalow, line 51: Suggestion: Where can we get the data if it's missing? Maybe mention that the data can be read from `src/data/{product}.json`

BRKalow (changes requested):
Good idea!

src/pages/_app.tsx
ashleemboyer: @BRKalow do you know what `HASHI_ENV` is during the Vercel deploy check? I think it's `'preview'`, but I'm not 100% sure. If it is `'preview'`, I think this'll be a good add!

src/pages/_app.tsx
ashleemboyer: Great idea! I'll put that in the message.

src/pages/_app.tsx
ashleemboyer: Testing it out with: https://github.com/hashicorp/dev-portal/pull/222/commits/b1ea71ac69ec0f33c4092cb4dc2d751ac72348c6. Will fix the line afterwards. 😊

src/pages/_app.tsx
ashleemboyer: Updated in: https://github.com/hashicorp/dev-portal/pull/222/commits/346090bf5a9349344b7bb6428d79231d31cd51ba

src/pages/_app.tsx
BRKalow: Yep, it is `preview` for deploy previews, production builds on main have `production` 👍

BRKalow (approved):

src/pages/_app.tsx
```diff
@@ -27,6 +28,25 @@ if (typeof window !== 'undefined' && process.env.AXE_ENABLED) {
...
+
+  /**
+   * TODO: this will be temporary and last for as long as we have to manually
+   * hydrate product data from page components. Ideally, we won't have to do
+   * this we use more dynamic routes.
+   */
+  if (process.env.HASHI_ENV !== 'production') {
+    const isDevDotPage =
+      !currentPath.startsWith(`/_proxied-dot-io`) &&
+      !currentPath.startsWith(`/swingset`)
```
BRKalow, line 41: @ashleemboyer I don't think `currentPath` will include `/_proxied-dot-io` when we apply the rewrites, so this check fires when viewing the preview in the context of an io site.

PR #227, opened 2022-03-25
Fixing hydration check in App

ashleemboyer:
## Relevant links

- [Preview link](https://dev-portal-git-ambfix-hydration-check-hashicorp.vercel.app/) 🔎

## What

- Removes product hydration check, fingers crossed that some in-progress TS refactors will get us the same affect as the error

## Testing

- [ ] All PR checks should pass

### DIFF
--- a/src/pages/_app.tsx
+++ b/src/pages/_app.tsx
@@ -1,13 +1,11 @@
 import React from 'react'
 import dynamic from 'next/dynamic'
-import Cookies from 'js-cookie'
 import { SSRProvider } from '@react-aria/ssr'
 import '@hashicorp/platform-util/nprogress/style.css'
 import { ErrorBoundary } from '@hashicorp/platform-runtime-error-monitoring'
 import useAnchorLinkAnalytics from '@hashicorp/platform-util/anchor-link-analytics'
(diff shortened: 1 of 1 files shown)


src/pages/_app.tsx
```diff
@@ -35,7 +36,7 @@ export default function App({ Component, pageProps, layoutProps }) {
    * hydrate product data from page components. Ideally, we won't have to do
    * this we use more dynamic routes.
    */
-  if (process.env.HASHI_ENV !== 'production') {
+  if (process.env.HASHI_ENV !== 'production' && !Cookies.get('io_preview')) {
```
ashleemboyer, line 39: @BRKalow does this look right?

```suggestion
  const hasPreviewCookie = typeof window !== 'undefined' && Cookies.get('io_preview')
  if (process.env.HASHI_ENV !== 'production' && !hasPreviewCookie) {
```

BRKalow: Yes, but this introduces an unnecessary cookie read on each render of _app on the client, even in production. Ideally we'd move it into the `HASHI_ENV` check.

BRKalow (approved):

PR #228, opened 2022-03-25
Improving Product TypeScript

ashleemboyer:

---

## Relevant links

- [Preview link](https://dev-portal-git-ambimprove-product-ts-hashicorp.vercel.app/) 🔎

## What/Why

This PR mainly adds a new `ProductData` interface separate, but extending from the `Product` interface. This allows TypeScript to help us ensure we're hydrating the correct data for other parts of the app (`ProductSwitcher` and `NavigationHeader` for example).

- We don't get this benefit on tutorial pages yet since `getTutorialPageProps` ([in TutorialView server](https://github.com/hashicorp/dev-portal/blob/main/src/views/tutorial-view/server.ts)) defines its own interface for the `product` object it accepts and returns as a page prop. We might be able to address this in the long term with a `ComponentPageProps` interface since we require `product` to be a page prop on all DevDot pages other than the home page.

There are some other changes in this PR for readability or fixing typedef warnings, but I've annotated all of those with PR comments to hopefully ease the review process. 🙏🏻

## How

The commit history walks through the steps I took, but this was mainly an effort of replacing most of the `Product` interface uses with `ProductData`.

## Testing

There should be no change in functionality with these refactors. A clickthrough of the main pages for a product (Terrform might be good since it has the most pages) should suffice.

## Anything else?

Nope!


src/components/docs-anchor/index.tsx
```diff
@@ -4,11 +4,11 @@ import { useCurrentProduct } from 'contexts'
 import InlineLink from 'components/inline-link'
 import s from './docs-anchor.module.css'
 
-const DocsAnchor: React.FC<JSX.IntrinsicElements['a']> = ({
+const DocsAnchor = ({
   href,
   children,
   ...rest
-}) => {
+}: JSX.IntrinsicElements['a']) => {
```
ashleemboyer, line 11: These changes fix a typedef warning.

config/products.ts
```diff
@@ -29,19 +29,19 @@ const productSwitcherSlugs = [
...
+    const productGroup = slugGroup.map((slug: ProductSlug) => {
+      const name = productSlugsToNames[slug]
+      return {
+        name,
+        slug,
+      }
+    })
+    return productGroup
+  }
+)
```
ashleemboyer, line 45: Couple of changes here:

- The outer `map` call had a typedef warning on `productSlug`
- The `url` property became irrelevant with some other changes, so I removed it
- The comment above this was incomplete, so I updated it
- I felt it was beneficial to add a type of `productSwitcherSlugs` to remove ambiguity

src/components/navigation-header/index.tsx
```diff
@@ -43,10 +46,10 @@ const NavigationHeader: React.FC = () => {
           </Link>
           <ProductSwitcher />
         </div>
-        {currentProduct?.navigationHeaderItems && (
+        {hasNavigationItems && (
           <div className={s.headerRight}>
             <ul className={s.navLinks}>
-              {currentProduct?.navigationHeaderItems.map((navLink) => {
+              {navItems.map((navLink: NavigationHeaderItem) => {
```
ashleemboyer, line 52: There was a typedef warning on this `map` call, so that's why `NavigationHeaderItem` is now imported for use here. It made sense to also move the export of the interface to this file as well since we do that in most places.

ashleemboyer, line 52: The introduction of the `navItems` and `hasNavigationItems` variables helped make this code a little more readable.

src/components/navigation-header/types.ts
```diff
@@ -1,7 +1,5 @@
-interface NavigationHeaderItem {
+export interface NavigationHeaderItem {
```
ashleemboyer, line 1: See comment in `src/components/navigation-header/index.tsx` that explains the changes here.

src/components/product-icon/index.tsx
```diff
@@ -9,7 +9,7 @@ import { IconVaultColor16 } from '@hashicorp/flight-icons/svg-react/vault-color-
 import { IconWaypointColor16 } from '@hashicorp/flight-icons/svg-react/waypoint-color-16'
 import { ProductIconProps } from './types'
 
-const productNamesToIcons = {
+const productSlugsToIcons = {
```
ashleemboyer, line 12: I meant to put this refactor in #229 but forgot. This is just a name change to something clearer.

src/components/product-switcher/index.tsx
```diff
@@ -157,7 +157,7 @@ const ProductSwitcher: React.FC = () => {
             s.switcherOptionAnchor,
             s.switcherOptionContainer
           )}
-          href={product.url}
+          href={`/${product.slug}`}
```
ashleemboyer, line 160: This change made the `url` property no longer needed on a product object. This was the only place the property was used.

src/components/product-switcher/index.tsx
```diff
@@ -182,7 +182,9 @@ const ProductSwitcher: React.FC = () => {
         )}
         <li>
           <ul role="group">
-            {productGroup.map((product) => renderProductListItem(product))}
+            {productGroup.map((product: Product) =>
+              renderProductListItem(product)
+            )}
```
ashleemboyer, line 187: This fixed a typedef warning on the `map` call.

src/contexts/__tests__/current-product.test.tsx
```diff
@@ -1,3 +1,4 @@
+import { ReactNode } from 'react'
```
ashleemboyer, line 1: This file needed to be updated since `CurrentProductProvider` now expects a `ProductData` object. I filled in empty values for the properties associated with that interface since other those aren't the focus of the tests here.

src/contexts/current-product.tsx
```diff
@@ -1,17 +1,22 @@
...
 export type RouteChangeStartHandler = (url: string) => void
 
-type CurrentProduct = Product | undefined
+type CurrentProduct = ProductData | undefined
 
 const CurrentProductContext = createContext<CurrentProduct>(undefined)
 
-const CurrentProductProvider: FC<{ currentProduct: Product }> = ({
+interface CurrentProductProviderProps {
+  children: ReactNode
```
ashleemboyer, line 12: Went with this approach instead of the previously used `React.FC` approach because there was a typedef warning about the props not being typed.

src/contexts/current-product.tsx
```diff
@@ -1,17 +1,22 @@
...
+  currentProduct: ProductData
```
ashleemboyer, line 13: This is the main update of interest. This Context now expects a `ProductData` object.

src/lib/cms.ts
```diff
@@ -69,3 +68,254 @@ function proxiedRivetClient(productSlug: ProductSlug) {
 
 export default instance
 export { client, proxiedRivetClient }
+
+/**
+ *
+ * NOTE: types defined below are only used here. Refer to ProductSlug and
+ * ProductData from /src/data/types/products.ts if those are needed.
+ *
+ */
```
ashleemboyer, line 77: @dstaley we chatted about this briefly on another PR, what do you think about these changes?

src/types/products.ts
```diff
@@ -20,17 +20,19 @@ type ProductSlug = Exclude<Products, 'hashicorp'> | 'hcp' | 'sentinel'
...
  */
 interface Product extends ProductMeta {
-  basePaths?: string[]
   name: ProductName
-  navigationHeaderItems?: NavigationHeaderItem[]
-  sidebar?: {
+  slug: ProductSlug
+}
+
+interface ProductData extends Product {
```
ashleemboyer, line 27: New interface! 🎉

src/views/product-downloads-view/index.tsx
```diff
@@ -46,16 +46,18 @@ const ProductDownloadsViewContent = ({
...
     () => initializeBreadcrumbLinks(currentProduct, currentVersion),
     [currentProduct, currentVersion]
   )
-  const navData = useMemo(() => initializeNavData(currentProduct), [
-    currentProduct,
-  ])
+  const navData = useMemo(
+    () => initializeNavData(currentProduct),
+    [currentProduct]
+  )
```
ashleemboyer, line 60: These are formatting changes. Not sure why these were suddenly needing formatting? 🤷🏻‍♀️

src/components/docs-anchor/index.tsx
BRKalow: FWIW, the type was correct. I just don't think the lint rule has enough type information like the TS compiler does to deduce the correct type here. We also technically lose the return type from this change.

ashleemboyer: Does the inferred return type of `JSX.Element` still suffice?

src/lib/cms.ts
dstaley: I think I'd prefer these types to live in `src/types/product.ts` since it'd be a bit weird to import from `lib/cms` when you're trying to access a prop like `algoliaConfig` or `alertBanner`. I know that currently imports from `config/<slug>.json` aren't typed, but being able to do something like the following would be nice:

```ts
import type { ProductData } from 'types/product'
import _config from 'data/nomad.json'

const config = _config as ProductData

// ...access properties like config.algoliaConfig with type safety
```

You're totally right that currently the only file that used this `ProductData` type was `lib/cms.ts`, so the hassle right now might not be worth it, but it'd be nice to have the ability to use it in the future.

Totally fine punting this down the line and worrying about it later though!

ashleemboyer: @dstaley We do actually already do what's in your code snippet in many files in DevDot. 😊

I think where this gets really tricky is that there are properties in the `data/<slug>.json` that .io sites only use, and there are some in DevDot pages that aren't used in .io sites. I don't think we can have one type that they both share because their shapes differ too much. I think we can move the definitions to the existing `src/types/products.ts` as long as we define two different `ProductData` interfaces. I think that probably would best done in a separate PR like you mention so we can have a more focused discussion on it if needed.

dstaley: Yeah, let's explore this a bit more in a future PR then! I think your idea of two separate interfaces is a great idea. That'd even give us some signals when we're using a .io property on a devdot page and vice versa.

src/components/docs-anchor/index.tsx
BRKalow: Should be fine!

BRKalow (approved):
👍

dstaley (approved):
love me some TypeScript 😍

PR #231, opened 2022-03-28
fix: filterVersions for terraform page

zchsh:

---

## Relevant links

- [Preview link](https://dev-portal-git-zsfix-semver-filter-versions-hashicorp.vercel.app/terraform/downloads) 🔎
- [Asana task](https://app.asana.com/0/1201987349274776/1202035288193775/f) 🎟️

## What

Fixes an issue where the `filterVersions` function for `terraform/downloads` let through pre-release versions.

## Why

With pre-releases included after `filterVersions`, we'd end up with `undefined` values in the staticProps result, which causes build errors.

## How

- Switches from `semverGte` to `semverSatisfies`. This seems to fix the issue, though requires us to specify a range (eg `>=1.0.11`) rather than a version (eg `1.0.11`) as we did previously.
    - If desired we could retain the prop as a version eg `1.0.11` and prefix it with `>=` within `filterVersions`. However I felt passing a range instead felt a bit more transparent.
- Also did some light refactoring of `filterVersions`, adding types, and making it less side-effect-y
   
## Testing

Same as #171:

- [ ] Go to https://dev-portal-git-zsfix-semver-filter-versions-hashicorp.vercel.app/terraform/downloads
- [ ] Make sure the version selected has limited versions as options
    - Specifically, we expect to see `1.0.11` and `1.1.7 (latest)`
- [ ] Make sure the rest of the page renders how you expect

## Anything else?

Not that I can think of!

### DIFF
--- a/src/lib/fetch-release-data.ts
+++ b/src/lib/fetch-release-data.ts
@@ -73,7 +73,7 @@ export function getLatestVersionFromVersions(versions: string[]): string {
  */
 export function generateStaticProps(
   product: Product | string
-): Promise<{ props: GeneratedProps; revalidate?: number | boolean }> {
+): Promise<{ props: GeneratedProps; revalidate: number }> {
   let productSlug: string
   if (typeof product === 'string') {
     productSlug = product
--- a/src/lib/fetch-release-data.ts
+++ b/src/lib/fetch-release-data.ts
@@ -1,4 +1,3 @@
-import { GetStaticPropsResult } from 'next'
 import semverSort from 'semver/functions/rsort'
 import { Products as HashiCorpProduct } from '@hashicorp/platform-product-meta'
 import { Product } from 'types/products'
@@ -74,7 +73,7 @@ export function getLatestVersionFromVersions(versions: string[]): string {
  */
 export function generateStaticProps(
   product: Product | string
-): Promise<GetStaticPropsResult<GeneratedProps>> {
+): Promise<{ props: GeneratedProps; revalidate?: number | boolean }> {
   let productSlug: string
   if (typeof product === 'string') {
     productSlug = product
--- a/src/pages/terraform/downloads/index.tsx
+++ b/src/pages/terraform/downloads/index.tsx
@@ -1,18 +1,22 @@
 import { ReactElement } from 'react'
 import { GetStaticProps } from 'next'
-import semverGte from 'semver/functions/gte'
+import semverSatisfies from 'semver/functions/satisfies'
 import semverMajor from 'semver/functions/major'
 import semverMinor from 'semver/functions/minor'
 import semverPatch from 'semver/functions/patch'
 import terraformData from 'data/terraform.json'
 import installData from 'data/terraform-install.json'
 import { Product } from 'types/products'
-import { generateStaticProps, GeneratedProps } from 'lib/fetch-release-data'
+import {
+  generateStaticProps,
+  GeneratedProps,
+  ReleasesAPIResponse,
+} from 'lib/fetch-release-data'
 import CoreDevDotLayout from 'layouts/core-dev-dot-layout'
 import ProductDownloadsView from 'views/product-downloads-view'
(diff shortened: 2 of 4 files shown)


src/pages/terraform/downloads/index.tsx
```diff
@@ -30,33 +34,34 @@ const TerraformDownloadsPage = (props: GeneratedProps): ReactElement => {
...
  */
-function filterOldVersions(props) {
-  if (!props?.props?.releases?.versions) {
-    return props
-  }
-
-  const versions = props.props.releases.versions
-
-  // versions is in the form of { [version]: { ...metadata } }
+function filterVersions(
```
zchsh, line 43: Refactored to avoid modifying `props.props.releases.versions` as a kind of side-effect. Now accepts versions, and returns filteredVersions.

src/pages/terraform/downloads/index.tsx
```diff
@@ -30,33 +34,34 @@ const TerraformDownloadsPage = (props: GeneratedProps): ReactElement => {
...
-
-  /** @type {{[x: string]:{ [y: string]: any}}} */
-  const tree = {}
+  )
 
   /**
    * Computes the latest patch versions for each major/minor
-   * e.g. given [1.1.2, 1.1.1, 1.1.0, 1.0.9, 1.0.8] -> return [1.1.2, 1.0.9]
+   * e.g. given [1.1.2, 1.1.1, 1.1.0, 1.0.9, 1.0.8]
+   * return [1.1.2, 1.0.9]
```
zchsh, line 61: Added some comments on this part since I had a bit of trouble parsing what was going on initially.

zchsh:
@ashleemboyer @BRKalow this is a fix for the `filterVersion` issue discussed in Slack. Thanks for your input there 🙇‍♂️💖

BRKalow (approved):
Fix looks good, one comment about the type change. I think it'd be better long-term if we could use the type provided by next. Will approve so we can fix preview deploys, we can always follow-up with a type change. Thanks!

src/pages/terraform/downloads/index.tsx
```diff
@@ -30,33 +34,34 @@ const TerraformDownloadsPage = (props: GeneratedProps): ReactElement => {
...
   // Filter by arbitrary & reasonable version cutoff
-  const filteredVersions = Object.keys(versions).filter((version) => {
-    if (!semverGte(version, VERSION_DOWNLOAD_CUTOFF)) {
-      return false
+  const filteredVersions = Object.keys(versions).filter(
+    (versionNumber: string) => {
+      console.log({
+        versionNumber,
+        satisfies: semverSatisfies(versionNumber, versionRange),
+      })
```
ashleemboyer, line 53: Do you want this log in here still?

zchsh: ah no i do not, great catch ty 💖

src/lib/fetch-release-data.ts
```diff
@@ -74,7 +73,7 @@ export function getLatestVersionFromVersions(versions: string[]): string {
  */
 export function generateStaticProps(
   product: Product | string
-): Promise<GetStaticPropsResult<GeneratedProps>> {
+): Promise<{ props: GeneratedProps; revalidate: number }> {
```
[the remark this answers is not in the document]
zchsh: Here's the error, basically it gets mad when I try to access `props`:

CleanShot 2022-03-28 at 12 05 08@2x

Definitely happy to try to debug this and fix - can definitely see how it'd be ideal to use the type provided by next (in case that changes in the future or whatnot)

src/pages/terraform/downloads/index.tsx
zchsh: ✅ 🔪 Fixed!

src/lib/fetch-release-data.ts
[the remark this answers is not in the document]
BRKalow: Huh, weird! Well, let's roll with your fix for now 🙇  Thanks for the insight

PR #240, opened 2022-03-30
Fix typo in Helm example

bendrucker:
Fixes a typo in the Docker + Helm example, where `artifact` is printed as "artfact"


BRKalow (approved):
Good catch, thanks for the contribution! ❤️

PR #241, opened 2022-03-31
[NavigationHeader] Add keyboard interaction

ashleemboyer:

---

## Relevant links

- [Preview link](https://dev-portal-git-ambprimary-nav-keyboard-interaction-hashicorp.vercel.app/swingset/components/navigationheader) 🔎
- Asana Tasks 🎟️
  - [[NavigationHeader] Update dropdown component to close on "blur"](https://app.asana.com/0/1201987349274776/1202037558549088/f)
  - [[NavigationHeader] Update dropdown component to allow keyboard interaction via Tab, Shift+Tab, and Escape](https://app.asana.com/0/1201987349274776/1202037558549094/f)

## What

- Added keyboard interaction to `NavigationHeaderDropdownMenu`
- Turned off `@typescript-eslint/no-inferrable-types` (see PR comment for explanation)
- Added `shouldListen` argument to `useOnFocusOutside` (see PR comment for explanation)
- Added `deriveKeyEventState`
- Added `g-focus-ring-from-box-shadow-dark` CSS helper class

## Why

- `deriveKeyEventState` is helpful because it'll save having to type out boolean conditions against keyboard events.
  - I will add a follow-up PR for using in other places. 😊
- `g-focus-ring-from-box-shadow-dark` is needed because the focus rings look different when the focusable element lives on a darker background.
  - I've added a TODO above this new class for asking design about the hardcoded colors associated with this focus ring to see if those have been updated in Figma yet to use DS tokens.

## How

- Closing the menus on "blur" (when focus leaves the menu or when there's a click outside of it) is handled by the `useOnClickOutside` and `useOnFocusOutside` hooks. These hooks made it possible to leverage default keyboard functionality for button and anchor elements and required no extra `onKeyDown` or `onBlur` event handlers. 🎉 
- Only the ESCAPE key functionality needed to be added to the button and anchor elements.

## Testing

- [ ] Each header element is one TAB stop
- [ ] Menus do not open when the dropdown menu button receives focus
- [ ] If focus is on a link in the main header:
  - [ ] ENTER will activate the link and a new page will load
- [ ] If focus is on a dropdown menu’s button:
  - [ ] And the menu is closed:
    - [ ] SPACE and ENTER will open the dropdown and focus will remain on the button
  - [ ] And the menu is open:
    - [ ] ESCAPE will close the menu and focus will remain on the button
    - [ ] SPACE/ENTER will close the dropdown (focus will remain on the button)
    - [ ] TAB will move focus to the first menu item in the dropdown
    - [ ] SHIFT+TAB will close the menu and move focus to the previous focusable element in the page (may or may not be an element in the header)
- [ ] If focus is on a menu item within a dropdown menu:
  - [ ] ESCAPE will close the menu and focus will move to the dropdown menu's button
  - [ ] And it is the first menu item within a dropdown menu:
    - [ ] SHIFT+TAB will move focus to the dropdown menu’s button
    - [ ] TAB will move focus to the next menu item in the dropdown (if there are are least 2 menu items)
  - [ ] And it is the last menu item within a dropdown menu:
    - [ ] SHIFT+TAB will move focus to the previous menu item (if there are at least 2 menu items)
    - [ ] TAB will close the menu and move focus to the next focusable element in the page (may or may not be an element in the header)

## Anything else?

- Some follow-up PRs I have planned:
  - Migrating the `useOnClickOutside` and `useOnFocusOutside` hooks to TypeScript
  - Cleaning up the markup in the body of `NavigationHeaderDropdownMenu`
  - Adding tests and either a README or details to the Swingset page that document keyboard interaction
- Remaining primary nav tasks are in the [[DevDot] Primary Header Nav Implementation](https://app.asana.com/0/1202022787106804/1202037558549094) Asana board
- I found one quirk about `useOnFocusOutside` that I'm not ultimately concerned with in this specific case, so I don't think any changes are needed.
  - On the linked Swingset page, when in the very last dropdown menu and the very last menu item is focused, when TAB is pressed, focus does in fact leave the menu but it doesn't close.
  - This is happening because a `focusin` event does not occur, because focus moves to the browser's tabs (because the menu item in this demo is the last focusable element in the page).
  - I believe this can be fixed by listening to the `focusout` event (and also supplying different `refs` to the hook).
  - However, I am not concerned about this behavior because there will be 100% be focusable elements in the header after the last dropdown menu. They just aren't in this demo so that's why this behavior is happening.


.eslintrc.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
   rules: {
     curly: 'error',
     '@typescript-eslint/explicit-module-boundary-types': 'off',
+    '@typescript-eslint/no-inferrable-types': 'off',
```
ashleemboyer, line 7: The `typedef` rule was upset that the `shouldListen` prop did not have a definition, and when I added one, this rule was upset about that. I feel like it's OK if we define types that are inferable? It doesn't seem like it hurts anything IMO, it seems like it's extra clarity.

src/hooks/use-on-focus-outside.js
```diff
@@ -3,15 +3,18 @@ import { useEffect } from 'react'
...
       const isOutside = ref.current && !ref.current.contains(event.target)
       if (isOutside) handler(event)
     }
 
-    // Bind the event listener
-    document.addEventListener('focusin', handleFocusChange)
+    // Bind the event listener if the listener should be applied via `shouldListen`
+    if (shouldListen) {
+      document.addEventListener('focusin', handleFocusChange)
+    }
```
ashleemboyer, line 16: Added `shouldListen` that works the same as `useOnClickOutside`'s `shouldListen`, that way it's possible to conditionally add this listeners instead of adding it all the time.

src/lib/derive-key-event-state.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  return {
+    isArrowDown,
+    isArrowLeft,
+    isArrowRight,
+    isArrowUp,
+    isEscape,
+    isShiftTab,
+    isSpace,
+    isTab,
```
ashleemboyer, line 34: TODO: update these names to include `Key` in the name. Still getting feedback on whether or not variable names for multiple keys should have `Keys` instead of `Key` (e.g. `isShiftTabKey` vs. `isShiftTabKeys`).

src/lib/derive-key-event-state.ts
```diff
@@ -0,0 +1,38 @@
...
+  // Arrow keys
+  const isArrowDown = key === 'ArrowDown'
+  const isArrowLeft = key === 'ArrowLeft'
+  const isArrowRight = key === 'ArrowRight'
+  const isArrowUp = key === 'ArrowUp'
+
+  // Escape, Space, and Enter
+  const isEscape = key === 'Escape'
+  const isSpace = key === ' '
+
```
ashleemboyer, line 21: TODO: add section here for Home, End, Page Up, and Page Down

src/components/navigation-header/components/dropdown-menu/index.tsx
```diff
@@ -144,39 +99,38 @@ const NavigationHeaderDropdownMenu = ({
...
-          weight="medium"
-        >
-          {label}
-        </Text>
-        <IconChevronDown16 className={s.activatorIcon} />
-      </button>
-    )
-  }
-
   return (
```
ashleemboyer, line 102: TODO: see if I can clean up this markup at all without this file becoming overwhelmingly long

ashleemboyer: Will actually do this in a follow-up PR

src/lib/derive-key-event-state.ts
ashleemboyer: Done in https://github.com/hashicorp/dev-portal/pull/241/commits/eeac712df1c217d6479b4f6deabdd0862041f886

src/lib/derive-key-event-state.ts
ashleemboyer: Done in https://github.com/hashicorp/dev-portal/pull/241/commits/0abbdbc67e534f3e00618990272ff842fcf8e480

src/components/navigation-header/components/dropdown-menu/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+   * button or the dropdown menu list, then the menu will be closed.
+   */
+  const handleMouseLeave = () => {
+    if (isOpen) {
+      setIsOpen(false)
+    }
+  }
+
+  return (
+    <div className={s.root} onMouseLeave={handleMouseLeave} ref={menuRef}>
```
alexcarpenter, line 122: Have we considered using a tool like hover intent to determine when to show menu on hover? This makes it so that the hover only shows after the mouse has stopped above the trigger to prevent menus opening unintentionally. 

Hook: https://github.com/natelindev/react-use-hoverintent
Demo: https://codesandbox.io/s/lucid-phoebe-5ut4ey?file=/src/App.tsx

src/components/navigation-header/components/dropdown-menu/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+        id={menuId}
+        style={{ display: isOpen ? 'block' : 'none' }}
+      >
+        {itemGroups.map((items: NavigationHeaderItem[], groupIndex: number) => {
+          const groupId = getItemGroupId(groupIndex)
+          const isLastItemGroup = groupIndex === numberOfItemGroups - 1
+          const showDivider = numberOfItemGroups > 1 && !isLastItemGroup
+          return (
+            <>
+              <ul className={s.itemGroup}>
```
alexcarpenter, line 155: More of just a question, but is it misleading to keyboard users to open the menu when there are groups and only get announced that there is 1 item? Should we flatten this to be a single ul, and use `<li role="separator" />` to render the hr? Or is this a use case to use role group on the unordered list?

In the video following, you can see when I enter the first menu it reads "Developer home list 1 item" but there is actually 3 items within the menu.


https://user-images.githubusercontent.com/825855/161989956-c734c189-80b1-4609-848e-becd2301c6ac.mp4

src/components/navigation-header/components/dropdown-menu/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+  const handleMouseLeave = () => {
+    if (isOpen) {
+      setIsOpen(false)
+    }
+  }
+
+  return (
+    <div className={s.root} onMouseLeave={handleMouseLeave} ref={menuRef}>
+      <div className={s.activatorWrapper}>
+        <button
```
alexcarpenter, line 124: Looks to be some unexpected behavior with regards to keyboard navigation happening in Firefox. Unable to tab into the menu items.

https://user-images.githubusercontent.com/825855/161990495-301a7869-921d-4f03-ac5c-b833df7b3c9d.mp4

image

ashleemboyer: I am able to tab to menu items in Firefox. Can you see if you have `accessibility.tabfocus` in your config as described here: https://www.a11yproject.com/posts/macos-browser-keyboard-navigation/#highlight-each-item-in-firefox.

When I have that configured and set to 7 as that section describes, I cannot TAB to links in the dropdown. When I delete the setting, I am then able to TAB to the links. It's possible that this is no longer a needed setting? [Scott O'Hara has a post with a 11/13/2019 update](https://www.scottohara.me/blog/2014/10/03/link-tabbing-firefox-osx.html#latest_update) that says: "However, apparently in Firefox 71+, an update will be put in place where this may no longer be necessary"

src/components/navigation-header/components/dropdown-menu/index.tsx
ashleemboyer: I don't think it's misleading, because we're expressing to sighted users that these are two separate groups of links. When there are more products, sighted users will be able to infer that the top group of links are "general" and the second group of links are "products". Since non-sighted users aren't able to infer the same thing, I think adding a brief label/name/description for each group would definitely be more informative!

src/components/navigation-header/components/dropdown-menu/index.tsx
alexcarpenter: I do have it enabled 
CleanShot 2022-04-06 at 11 19 13@2x

ashleemboyer: Could you delete it and see if that fixes your issue?

src/components/navigation-header/components/dropdown-menu/index.tsx
ashleemboyer: I've not seen this before! I forked your sandbox to try it out, and when I increase the `interval` option, I see that it works for the mouseover event but not for the mouseout event. Can you tell if I'm using it correctly? https://codesandbox.io/s/beautiful-kowalevski-u0orb3?file=/src/App.tsx

src/components/navigation-header/components/dropdown-menu/index.tsx
alexcarpenter: Still wondering why i'd be able to tab to the top level buttons but not the menu items without the setting correctly in place. I'd assume I wouldn't have been able to tab through the trigger elements. 🤷🏼

src/components/navigation-header/components/dropdown-menu/index.tsx
alexcarpenter: Looks like `timeout` effects the `mouseout` event https://github.com/natelindev/react-use-hoverintent#options

ashleemboyer: Ah, that does what I was looking for!

I have some general hesitancy about using either [`react-use-hoverintent`](https://www.npmjs.com/package/react-use-hoverintent) or [`react-hoverintent`](https://www.npmjs.com/package/react-hoverintent) because it's not clear how well supported and maintained they are. Some additional context: I'm not sure if we should keep these hover interactions, but we can't evaluate that at the moment so they're sticking around for now. With all of that in mind, do you think it's fine to make a ticket for evaluating this package outside of this PR before using it? We may also want to consider using react-aria's `useHover` hook, but I'm [still reading](https://react-spectrum.adobe.com/blog/building-a-button-part-2.html) about that one. 😊

alexcarpenter: Yeah, no problem, sounds good.

ashleemboyer: Ticket added! https://app.asana.com/0/1201987349274776/1202074017129387/f

BRKalow:
@ashleemboyer I noticed that these menus behave differently than the product switcher from a keyboard navigation perspective, which cycles focus within the menu and uses one list element. Do you think we should align on one pattern going forward, or are these two different types of navigation menus?

ashleemboyer:
> @ashleemboyer I noticed that these menus behave differently than the product switcher from a keyboard navigation perspective, which cycles focus within the menu and uses one list element. Do you think we should align on one pattern going forward, or are these two different types of navigation menus?

@BRKalow The DevDot `ProductSwitcher` (not the one that shows on preview URLs) is going away with the new primary nav refactor, so they don't need aligned. 😊 The arrow key functionality that `ProductSwitcher` supports is actually over-engineered based on feedback last week from Melanie on the Design Systems team ([see this Slack message](https://hashicorp.slack.com/archives/C7KTUHNUS/p1648742386580669?thread_ts=1648741500.802189&cid=C7KTUHNUS)), so that's why I went back on supporting that here. Arrow key functionality will be more important for replicating `<select>` element functionality in future `ContextSwitcher` components.

ashleemboyer:
Just pushed an update that adds some custom tokens that don't live in the design system. There should be no visual changes though!

PR #244, opened 2022-03-31
feat: add inline, height, and width props to ImageConfig and Image

BRKalow:

---

## Relevant links

- [Preview link](https://dev-portal-git-brkfeat-image-config-io-sites-hashicorp.vercel.app/swingset/components/imageconfig) 🔎
- [Asana task](url) 🎟️

## What

Add additional props to Image/ImageConfig:

* `inline` forces the wrapping element of `Image` to be `inline-flex` instead of `flex`
* `height` and `width` set the height and width on the underlying `img` element using the `style` prop.

## Why

To facilitate setting explicit dimensions on images which are rendering in markdown syntax. I chose to set the dimensions using `style` instead of the same-named attributes to ensure that any existing base styles are overridden. This ensures it can work in both the dev-portal context as well as the current .io site context.

We want to enforce usage of the markdown image syntax, but we also want to address the use-case of explicit dimensions being set on images ([sourcegraph](https://hashicorp-mktg.sourcegraph.com/search?q=context:global+file:website/.*.mdx+%3Cimg&patternType=literal)).

## Testing

- [ ] Navigate to https://dev-portal-git-brkfeat-image-config-io-sites-hashicorp.vercel.app/swingset/components/imageconfig
- [ ] See that the example has explicit dimensions

### DIFF
--- a/src/components/_proxied-dot-io/common/docs-page/index.tsx
+++ b/src/components/_proxied-dot-io/common/docs-page/index.tsx
@@ -1,0 +1,24 @@
+import ReactDocsPage, { DocsPageProps } from '@hashicorp/react-docs-page'
+import ImageConfigBase from 'components/image-config'
+import { ImageConfigProps } from 'components/image-config/types'
+
+const ioComponents = {
+  ImageConfig: (props: ImageConfigProps) => (
+    <ImageConfigBase hideBorder {...props} />
+  ),
+}
+
+/**
+ * A shared wrapper around react-docs-page to inject common authoring components for our .io docs pages.
+ */
+export default function DocsPage({
+  additionalComponents,
+  ...props
+}: DocsPageProps) {
+  return (
+    <ReactDocsPage
+      additionalComponents={{ ...ioComponents, ...additionalComponents }}
+      {...props}
+    />
+  )
+}
--- a/src/components/image-config/docs.mdx
+++ b/src/components/image-config/docs.mdx
@@ -16,7 +16,7 @@ componentName: ImageConfig
 
 ## Props
 
-- `caption` `string` - a plain text caption to show below the image.
+- `caption: string` - a plain text caption to show below the image.
 
   <details><summary>Example of `caption` in use</summary>
 
@@ -34,6 +34,20 @@ componentName: ImageConfig
 
   </details>
 
+- `inline: boolean` - Indicate whether or not the rendered image element should be inline
+- `height: number` - An explicit height value (in px) to be passed to the rendered image element
+- `width: number` - An explicit width value (in px) to be passed to the rendered image element
+
 ### Deprecated props
 
-- 🚫 `hideBorder` `boolean` - was a boolean prop. This prop is now deprecated. Please remove borders from all images.
+- 🚫 `hideBorder: boolean` - was a boolean prop. This prop is now deprecated. Please remove borders from all images.
+
+## Explicit dimensions
+
+Pass `height` and `width` to apply explicit dimensions to the `img` element.
+
+<LiveComponent>{`
+<ImageConfig height={200} width={200} inline>
+  <img src="https://place-hold.it/500x300" />
+</ImageConfig>
+`}</LiveComponent>
--- a/src/components/image/image.module.css
+++ b/src/components/image/image.module.css
@@ -22,6 +22,13 @@
     display: inline-flex;
   }
 
+  /*
+  This is nested inside .root, instead of delcared as its own class, to ensure that the selector
+  is specfic enough to take precedence over the styles declared in @hashicorp/react-content.
+
+  TODO: we should be able to switch this back to a standalone class after all products are migrated to
+  dev-portal.
+  */
   & img {
     display: block;
     max-width: 100%;
--- a/src/components/image/image.module.css
+++ b/src/components/image/image.module.css
@@ -25,5 +25,6 @@
   & img {
     display: block;
     max-width: 100%;
+    margin: 0;
   }
 }
--- a/src/components/image/image.module.css
+++ b/src/components/image/image.module.css
@@ -17,9 +17,13 @@
     box-shadow: none;
     border-radius: 0;
   }
-}
 
-.img {
-  display: block;
-  max-width: 100%;
+  &.inline {
+    display: inline-flex;
+  }
+
+  & img {
+    display: block;
+    max-width: 100%;
+  }
 }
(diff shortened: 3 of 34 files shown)


src/components/image/index.tsx
```diff
@@ -1,8 +1,28 @@
...
+ * Create an object to be passed as a style prop to the underlying img element
+ */
+function makeStyleProp(
+  width: ImageProps['width'],
+  height: ImageProps['height']
+): Pick<CSSProperties, 'width' | 'height'> {
+  const style: Pick<CSSProperties, 'width' | 'height'> = {}
+
+  if (width) {
+    style.width = `${width}px`
```
dstaley, line 16: q: do we want to support specifying dimensions in non `px` units? We could get that for free by reusing the `HTMLImageElement`'s types.

dstaley (approved):

src/components/image/index.tsx
BRKalow: We certainly could, I was initially basing it off of the fact that the height/width attributes only support integer values: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-height

ashleemboyer: Personally, I think `style` gives far too much flexibility. We have a general plan to create image guidelines (for example: don't add padding within an image when creating it, because the site will already add margin/padding via CSS), and the `style` prop would enable people to break the guidelines easily.

I think we should add in props one at a time like we have in this PR, and wait to see what kinds of requests are made for customizing image styling. If we end up seeing a lot of requests for the same thing, then that seems like the time to reassess what props are available.

src/components/image/index.tsx
```diff
@@ -27,15 +50,22 @@ function Image({
     )
   }
 
+  /**
+   * We're passing width and height as style here to ensure they override
+   * any base styles.
+   */
+  const style = makeStyleProp(width, height)
```
ashleemboyer, line 57: 💭 Is `make` a common naming pattern we use in other places? I'm used to `generate`, but I like `make` quite a bit here. Or maybe, are `generate` and `make` actually different?? It's probably not be that deep, just a curious question. 😊

src/lib/_proxied-dot-io/get-static-generation-functions.ts
```diff
@@ -1,26 +1,23 @@
 import { getStaticGenerationFunctions as getStaticGenerationFunctionsBase } from '@hashicorp/react-docs-page/server'
```
ashleemboyer, line 1: Double-checking, are the changes here just formatting changes?

src/pages/_proxied-dot-io/waypoint/docs/[[...page]].tsx
```diff
@@ -4,16 +4,22 @@ import NestedNode from 'components/_proxied-dot-io/waypoint/nested-node'
 import DocsPage from '@hashicorp/react-docs-page'
 import productData from 'data/waypoint.json'
 import { isVersionedDocsEnabled } from 'lib/env-checks'
+import ImageConfigBase from 'components/image-config'
+import { ImageConfigProps } from 'components/image-config/types'
 // Imports below are used in getStatic functions only
 import { getStaticGenerationFunctions } from 'lib/_proxied-dot-io/get-static-generation-functions'
 
+const ImageConfig = (props: ImageConfigProps) => (
+  <ImageConfigBase hideBorder {...props} />
```
Comment (ashleemboyer) [inline src/pages/_proxied-dot-io/waypoint/docs/[[...page]].tsx:13]: It's not clear to me yet, why does this one have `hideBorder` but the Vault one doesn't?

ashleemboyer (changes requested):
Really cool to see your idea in action so quickly after you proposed it! ✨

Have one request for a comment in here, and a few other comments sprinkled in. One more general question: should we go ahead and make `ImageConfig` available to [all proxied .io sites](https://github.com/hashicorp/dev-portal/tree/main/src/pages/_proxied-dot-io)?

src/lib/_proxied-dot-io/get-static-generation-functions.ts
BRKalow: Yep!

src/components/image/index.tsx
ashleemboyer: Just did a quick search, and it looks like `generate` is used to name functions that return strings, objects, arrays, etc. that are built by the helper function (e.g. `generateStaticProps`, `generateStaticPaths`, `generatePackageManagers`, etc.). Seems like `make` is used to name functions that return a function/function component (e.g. `makeHeadingElement`, `makeFetchWithRetry`).

What do you think about using `make` and `generate` separately in this way? I formed my thoughts on this while writing this comment, so not deeply invested in the idea. If we come to a consensus, I think it'd be valuable to write it down (maybe in the `docs` folder). 😊

src/pages/_proxied-dot-io/waypoint/docs/[[...page]].tsx
Comment (ashleemboyer) [reply src/pages/_proxied-dot-io/waypoint/docs/[[...page]].tsx:13]: Awesome, that sounds good!

src/components/image/image.module.css
```diff
@@ -17,9 +17,14 @@
...
-  max-width: 100%;
+  &.inline {
+    display: inline-flex;
+  }
+
+  & img {
+    display: block;
+    max-width: 100%;
+    margin: 0;
+  }
```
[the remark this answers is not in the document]
ashleemboyer: The comment you've added is great! 🙌🏻 Thank you for getting that in there. 💖

src/components/image/index.tsx
BRKalow: Works for me! Thanks for doing some research here 🙏

src/components/image-config/docs.mdx
```diff
@@ -34,6 +34,20 @@ componentName: ImageConfig
 
   </details>
 
+- `inline: boolean` - Indicate whether or not the rendered image element should be inline
```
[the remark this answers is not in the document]
BRKalow: Just for posterity, we had chatted about this on Friday synchronously, and we're going to roll with the `inline` prop for now. I don't necessarily want to rework the images here, and it seems like the `inline` prop _could_ have some uses elsewhere.

src/components/_proxied-dot-io/common/docs-page/index.tsx
```diff
@@ -0,0 +1,24 @@
+import ReactDocsPage, { DocsPageProps } from '@hashicorp/react-docs-page'
```
ashleemboyer, line 1: Oh nice!! This is a great idea ✨

ashleemboyer (approved):

PR #226, opened 2022-03-25
feat: tutorial next prev stub

kendallstrautman:
## "Ready for Review" checklist

- [x] The Vercel preview link has been added to this PR's description
- [x] The Asana task has been added to this PR's description
- [x] This PR's link has been added to the "📐 Pull Request" field on the Asana task
- [x] This Vercel preview link has been added to the "🔍 Preview" field on the Asana task
- [x] The description template has been filled in below

---

## Relevant links

- [Preview link](https://dev-portal-git-kstutorial-next-prev-hashicorp.vercel.app/waypoint/tutorials/get-started-kubernetes/get-started-kubernetes-destroy) 🔎
- [Asana task](https://app.asana.com/0/1201987349274776/1201999966887544) 🎟️

## What

This PR adds a stub for the next / previous tutorial component. This copies over the basic component, logic and some stub styles from Learn. The final styling the component to spec will happen in another PR. 

## Why

To facilitate a good end of page UX. 

## How

Copied over and adjusting logic slightly now that we have a new API endpoint to use that will return the 'next collection' in the sorted sidebar order. This is only called if the tutorial is the last in the collection list, which reduces a call in `getStaticProps`.

## Testing

- Each tutorial page should show "next / previous" links at the bottom of the page, which lead to adjacent tutorials in the collection
- On the first tutorial in a collection, the "previous" button should go "back to collection"
- On the last tutorial in a collection, the "next" button should go to "next collection", which should reflect the product sidebar collection order. See image below or [refer to Learn](https://learn.hashicorp.com/waypoint) for correct order (currently the product tutorial pages in dev portal don't have the proper sidebar sort logic) 

Screen Shot 2022-03-25 at 3 45 12 PM

PR #251, opened 2022-04-04
feat: next prev tutorial chrome

zchsh:

---

## Relevant links

- [Preview link][/vault/tutorials/tokens/token-management] 🔎
- [Asana task](https://app.asana.com/0/1201987349274776/1202016460396717/f) 🎟️

## What

Refines styling and implementation of "Next / Previous" links in the `tutorials` end-of-page UX.

## Why

To match design specs as we build out the chromed UI for the `/tutorials` view.

## How

- Adds support for `ariaLabel` and `next/link` `as` to `CardLink`
- Refactors `DirectionalLinkBox` to use `CardLink` under the hood
- Refactors `DirectionalLinkBox` to use React icons from Flight rather than `InlineSvg`

## Testing

> Note: next & previous functionality has been validated in #226. Additional validation is never a bad thing, but for the purpose of this PR, feel free to focus on the design spec only.

Visit pages with each of the "Next / Previous" end-of-page variations, and ensure they match [the design spec](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Dev-Portal?node-id=6298%3A73041).

- [ ] starting tutorial: "back to collection / next": [/vault/tutorials/adp/transform][/vault/tutorials/adp/transform]
- [ ] middle tutorial: "previous / next": [/vault/tutorials/getting-started-ui/getting-started-auth-ui][/vault/tutorials/getting-started-ui/getting-started-auth-ui]
- [ ] end tutorial: "previous / next collection": [/vault/tutorials/encryption-as-a-service/eaas-spring-demo][/vault/tutorials/encryption-as-a-service/eaas-spring-demo]
- [ ] final collection end tutorial: "previous / browse tutorials": [/vault/tutorials/tokens/token-management][/vault/tutorials/tokens/token-management]

## Anything else?

The mobile implementation matches the design spec, but may feel off due to the long “next collection” and “back to collection” text (which I couldn't find in the design spec; they may not have been accounted for).

We could refine the mobile styles further in future iterations where we have more bandwidth to focus on the mobile experience. For example, we could make the longer text buttons wrap to a new line. However, this would likely require additional design input.

Previewing content on mobile is difficult, so it's difficult to see the exact behaviour of next / previous on small viewports:
- Swingset does not support responsive design mode
- `SidebarSidecarLayout` does not support small screens, so custom style overrides are needed to give an approximate preview of mobile behaviour

To see what this difference looks like, expand the `Design spec` & `Implementation` detail sections below:

<details>
<summary>Design spec</summary>

design-spec

</details>

<details>
<summary>Implementation</summary>

In most cases, `Previous` & `Next` appear, and look fine:

CleanShot 2022-04-04 at 12 03 11@2x

At the start or end of collections, longer text appears, which causes awkward wrapping: 

CleanShot 2022-04-04 at 12 02 22@2x

</details>

[/vault/tutorials/adp/transform]: https://dev-portal-git-zstutorial-next-prev-chrome-hashicorp.vercel.app/vault/tutorials/adp/transform
[/vault/tutorials/getting-started-ui/getting-started-auth-ui]: https://dev-portal-git-zstutorial-next-prev-chrome-hashicorp.vercel.app/vault/tutorials/getting-started-ui/getting-started-auth-ui
[/vault/tutorials/encryption-as-a-service/eaas-spring-demo]: https://dev-portal-git-zstutorial-next-prev-chrome-hashicorp.vercel.app/vault/tutorials/encryption-as-a-service/eaas-spring-demo
[/vault/tutorials/tokens/token-management]: https://dev-portal-git-zstutorial-next-prev-chrome-hashicorp.vercel.app/vault/tutorials/tokens/token-management

### DIFF
--- a/src/components/card-link/index.tsx
+++ b/src/components/card-link/index.tsx
@@ -9,7 +9,7 @@ const CardLink = ({
   children,
   className,
   href,
-  'aria-label': ariaLabel,
+  ariaLabel,
 }: CardLinkProps): ReactElement => {
   const classes = classNames(s.root, className)
 
--- a/src/components/card-link/index.tsx
+++ b/src/components/card-link/index.tsx
@@ -9,13 +9,12 @@ const CardLink = ({
   children,
   className,
   href,
-  as,
   'aria-label': ariaLabel,
 }: CardLinkProps): ReactElement => {
   const classes = classNames(s.root, className)
 
   return (
-    <MaybeInternalLink href={href} as={as} aria-label={ariaLabel}>
+    <MaybeInternalLink href={href} aria-label={ariaLabel}>
       <Card className={classes}>{children}</Card>
     </MaybeInternalLink>
   )
--- a/src/components/card-link/index.tsx
+++ b/src/components/card-link/index.tsx
@@ -9,12 +9,13 @@ const CardLink = ({
   children,
   className,
   href,
+  as,
   'aria-label': ariaLabel,
 }: CardLinkProps): ReactElement => {
   const classes = classNames(s.root, className)
 
   return (
-    <MaybeInternalLink href={href} aria-label={ariaLabel}>
+    <MaybeInternalLink href={href} as={as} aria-label={ariaLabel}>
       <Card className={classes}>{children}</Card>
     </MaybeInternalLink>
   )
--- a/src/components/card-link/index.tsx
+++ b/src/components/card-link/index.tsx
@@ -9,11 +9,12 @@ const CardLink = ({
   children,
   className,
   href,
+  'aria-label': ariaLabel,
 }: CardLinkProps): ReactElement => {
   const classes = classNames(s.root, className)
 
   return (
-    <MaybeInternalLink href={href}>
+    <MaybeInternalLink href={href} aria-label={ariaLabel}>
       <Card className={classes}>{children}</Card>
     </MaybeInternalLink>
   )
--- a/src/components/card-link/types.ts
+++ b/src/components/card-link/types.ts
@@ -20,7 +20,7 @@ export interface CardLinkProps {
   /**
    * Optional decorator for the path that will be shown in the browser URL bar.
    */
-  as?: string
+  as?: LinkProps['as']
 
   /**
    * An optional string value that labels the link. For use when content visible in the DOM meant to give the link meaning is either missing, or does not accurately describe the link.
--- a/src/components/card-link/types.ts
+++ b/src/components/card-link/types.ts
@@ -1,3 +1,4 @@
+import { LinkProps } from 'next/link'
 import { ReactNode } from 'react'
 
 export interface CardLinkProps {
@@ -15,6 +16,11 @@ export interface CardLinkProps {
    * The destination of the link.
    */
   href: string
+
+  /**
+   * Optional decorator for the path that will be shown in the browser URL bar.
+   */
+  as?: string
 
   /**
    * An optional string value that labels the link. For use when content visible in the DOM meant to give the link meaning is either missing, or does not accurately describe the link.
--- a/src/components/card-link/types.ts
+++ b/src/components/card-link/types.ts
@@ -1,5 +1,5 @@
-import { MaybeInternalLinkProps } from 'components/maybe-internal-link/types'
 import { ReactNode } from 'react'
+import { MaybeInternalLinkProps } from 'components/maybe-internal-link'
 
 export interface CardLinkProps {
   /**
--- a/src/components/card-link/types.ts
+++ b/src/components/card-link/types.ts
@@ -20,5 +20,5 @@ export interface CardLinkProps {
   /**
    * An optional string value that labels the link. For use when content visible in the DOM meant to give the link meaning is either missing, or does not accurately describe the link.
    */
-  'aria-label'?: string
+  ariaLabel?: string
 }
--- a/src/components/card-link/types.ts
+++ b/src/components/card-link/types.ts
@@ -15,4 +15,9 @@ export interface CardLinkProps {
    * The destination of the link.
    */
   href: string
+
+  /**
+   * An optional string value that labels the link. For use when content visible in the DOM meant to give the link meaning is either missing, or does not accurately describe the link.
+   */
+  'aria-label'?: string
 }
--- a/src/components/card-link/types.ts
+++ b/src/components/card-link/types.ts
@@ -18,7 +18,9 @@ export interface CardLinkProps {
   href: string
 
   /**
-   * An optional string value that labels the link. For use when content visible in the DOM meant to give the link meaning is either missing, or does not accurately describe the link.
+   * An optional string value that labels the link. For use when content visible
+   * in the DOM meant to give the link meaning is either missing, or does not
+   * accurately describe the link.
    */
   ariaLabel?: MaybeInternalLinkProps['aria-label']
 }
--- a/src/components/card-link/types.ts
+++ b/src/components/card-link/types.ts
@@ -18,11 +18,6 @@ export interface CardLinkProps {
   href: string
 
   /**
-   * Optional decorator for the path that will be shown in the browser URL bar.
-   */
-  as?: LinkProps['as']
-
-  /**
    * An optional string value that labels the link. For use when content visible in the DOM meant to give the link meaning is either missing, or does not accurately describe the link.
    */
   'aria-label'?: string
--- a/src/components/card-link/types.ts
+++ b/src/components/card-link/types.ts
@@ -1,4 +1,4 @@
-import { LinkProps } from 'next/link'
+import { MaybeInternalLinkProps } from 'components/maybe-internal-link/types'
 import { ReactNode } from 'react'
 
 export interface CardLinkProps {
@@ -20,5 +20,5 @@ export interface CardLinkProps {
   /**
    * An optional string value that labels the link. For use when content visible in the DOM meant to give the link meaning is either missing, or does not accurately describe the link.
    */
-  ariaLabel?: string
+  ariaLabel?: MaybeInternalLinkProps['aria-label']
 }
--- a/src/components/maybe-internal-link/index.tsx
+++ b/src/components/maybe-internal-link/index.tsx
@@ -3,10 +3,7 @@ import Link from 'next/link'
 import isAbsoluteUrl from 'lib/is-absolute-url'
 import { MaybeInternalLinkProps } from './types'
 
-function MaybeInternalLink({
-  href,
-  ...rest
-}: MaybeInternalLinkProps): React.ReactElement {
+function MaybeInternalLink({ href, ...rest }: MaybeInternalLinkProps) {
   const Elem = isAbsoluteUrl(href) ? InternalLink : 'a'
   return <Elem href={href} {...rest} />
 }
--- a/src/components/maybe-internal-link/index.tsx
+++ b/src/components/maybe-internal-link/index.tsx
@@ -5,16 +5,15 @@ import { MaybeInternalLinkProps } from './types'
 
 function MaybeInternalLink({
   href,
-  as,
   ...rest
 }: MaybeInternalLinkProps): React.ReactElement {
   const Elem = isAbsoluteUrl(href) ? InternalLink : 'a'
   return <Elem href={href} {...rest} />
 }
 
-function InternalLink({ href, as, ...rest }: MaybeInternalLinkProps) {
+function InternalLink({ href, ...rest }: MaybeInternalLinkProps) {
   return (
-    <Link href={href} as={as}>
+    <Link href={href}>
       {/* Disabling anchor-has-content, children is in ...rest  */}
       {/* eslint-disable-next-line jsx-a11y/anchor-has-content */}
       <a {...rest} />
--- a/src/components/maybe-internal-link/index.tsx
+++ b/src/components/maybe-internal-link/index.tsx
@@ -1,20 +1,20 @@
+import React from 'react'
 import Link from 'next/link'
 import isAbsoluteUrl from 'lib/is-absolute-url'
-import React from 'react'
+import { MaybeInternalLinkProps } from './types'
 
 function MaybeInternalLink({
   href,
+  as,
   ...rest
-}: {
-  href: string
-} & React.HTMLProps<HTMLAnchorElement>): React.ReactElement {
+}: MaybeInternalLinkProps): React.ReactElement {
   const Elem = isAbsoluteUrl(href) ? InternalLink : 'a'
   return <Elem href={href} {...rest} />
 }
 
-function InternalLink({ href, ...rest }) {
+function InternalLink({ href, as, ...rest }: MaybeInternalLinkProps) {
   return (
-    <Link href={href}>
+    <Link href={href} as={as}>
       {/* Disabling anchor-has-content, children is in ...rest  */}
       {/* eslint-disable-next-line jsx-a11y/anchor-has-content */}
       <a {...rest} />
--- a/src/components/maybe-internal-link/index.tsx
+++ b/src/components/maybe-internal-link/index.tsx
@@ -18,4 +18,5 @@ function InternalLink({ href, ...rest }: MaybeInternalLinkProps) {
   )
 }
 
+export type { MaybeInternalLinkProps }
 export default MaybeInternalLink
--- a/src/components/maybe-internal-link/types.ts
+++ b/src/components/maybe-internal-link/types.ts
@@ -1,3 +1,5 @@
-export type MaybeInternalLinkProps = {
-  href: string
-} & React.HTMLProps<HTMLAnchorElement>
+type AnchorElementProps = JSX.IntrinsicElements['a']
+
+export interface MaybeInternalLinkProps extends AnchorElementProps {
+  href: AnchorElementProps['href']
+}
--- a/src/components/maybe-internal-link/types.ts
+++ b/src/components/maybe-internal-link/types.ts
@@ -1,6 +1,3 @@
-import { LinkProps } from 'next/link'
-
 export type MaybeInternalLinkProps = {
   href: string
-  as?: LinkProps['as']
-} & Omit<React.HTMLProps<HTMLAnchorElement>, 'as'>
+} & React.HTMLProps<HTMLAnchorElement>
--- a/src/components/maybe-internal-link/types.ts
+++ b/src/components/maybe-internal-link/types.ts
@@ -1,0 +1,6 @@
+import { LinkProps } from 'next/link'
+
+export type MaybeInternalLinkProps = {
+  href: string
+  as?: LinkProps['as']
+} & React.HTMLProps<HTMLAnchorElement>
--- a/src/components/maybe-internal-link/types.ts
+++ b/src/components/maybe-internal-link/types.ts
@@ -3,4 +3,4 @@ import { LinkProps } from 'next/link'
 export type MaybeInternalLinkProps = {
   href: string
   as?: LinkProps['as']
-} & React.HTMLProps<HTMLAnchorElement>
+} & Omit<React.HTMLProps<HTMLAnchorElement>, 'as'>
--- a/src/components/text/index.tsx
+++ b/src/components/text/index.tsx
@@ -7,7 +7,7 @@ const Text: React.FC<TextProps> = ({
   size = 300,
   weight = 'regular',
   ...rest
-}) => {
+}: TextProps) => {
   const className = classNames(
     s.root,
     `hds-typography-body-${size}`,
--- a/src/views/tutorial-view/components/next-previous/components/directional-link-box/directional-link-box.module.css
+++ b/src/views/tutorial-view/components/next-previous/components/directional-link-box/directional-link-box.module.css
@@ -6,7 +6,7 @@
   justify-content: center;
   height: 100%;
 
-  @media (min-width: 729px) {
+  @media (--dev-dot-tablet-up) {
     &.direction-previous {
       text-align: left;
       justify-content: flex-start;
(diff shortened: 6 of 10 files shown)


src/components/card-link/index.tsx
```diff
@@ -9,11 +9,12 @@ const CardLink = ({
   children,
   className,
   href,
+  'aria-label': ariaLabel,
 }: CardLinkProps): ReactElement => {
   const classes = classNames(s.root, className)
 
   return (
-    <MaybeInternalLink href={href}>
+    <MaybeInternalLink href={href} aria-label={ariaLabel}>
```
zchsh, line 17: `CardLink` now supports `aria-label`, passing it to `MaybeInternalLink`.

I opted for snake casing here to match use on `<a />` tags (and therefore in `MaybeInternalLink`). Loosely held opinion is that the minor annoyance of having to use an alias when de-structuring props within `CardLink` is worth the consistency.

src/components/maybe-internal-link/index.tsx
```diff
@@ -1,18 +1,17 @@
+import React from 'react'
 import Link from 'next/link'
 import isAbsoluteUrl from 'lib/is-absolute-url'
-import React from 'react'
+import { MaybeInternalLinkProps } from './types'
```
zchsh, line 4: Split this out into a `types.ts` file for consistency (and to remove the "arg should be typed" warning in `InternalLink` within this file).

src/components/text/index.tsx
```diff
@@ -7,7 +7,7 @@ const Text: React.FC<TextProps> = ({
   size = 300,
   weight = 'regular',
   ...rest
-}) => {
+}: TextProps) => {
```
zchsh, line 10: Super minor nit, noticed this was missing, so added it 👍

src/views/tutorial-view/components/next-previous/components/directional-link-box/directional-link-box.module.css
```diff
@@ -1,101 +1,41 @@
-.card {
-  border: 1px solid magenta; /** @TODO style these to spec, remove any prev imp styles necessary - https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Dev-Portal?node-id=6298%3A73041 */
```
zchsh, line 2: Styles are now mostly from `CardLink`, so much of this can be removed.

src/views/tutorial-view/components/next-previous/components/directional-link-box/directional-link-box.module.css
```diff
@@ -1,101 +1,41 @@
...
-  position: relative;
-  display: block;
-  background: var(--white);
-  border-radius: 5px;
-  margin: 0;
-  padding: 20px;
-}
-
-/* visually hidden */
```
zchsh, line 11: Now using `aria-label` rather than visually hiding a label element.

src/views/tutorial-view/components/next-previous/components/directional-link-box/directional-link-box.module.css
```diff
@@ -1,101 +1,41 @@
...
-  box-shadow: var(--token-surface-mid-box-shadow);
-  transition: box-shadow 0.3s;
+  justify-content: center;
 
-  &[data-direction='previous'] {
-    text-align: left;
-    justify-content: flex-start;
-    &:hover {
-      box-shadow: var(--token-surface-high-box-shadow);
+  @media (--medium-up) {
```
zchsh, line 6: I've refactored the media-queries here to be mobile-first. I find this much easier to reason about if it's consistent, and it also happens to DRY things up a bit.

I've also switched from the old-habit-of-mine `data-` attributes for styling to classNames. Details on the history here are in Notion, under `Code Style & Practices` > `CSS Modules`.

> Note: not to say Notion guidelines are ideal; they're just what we have. I think we're very open to changing them in any & all ways.

src/views/tutorial-view/components/next-previous/index.tsx
```diff
@@ -58,7 +58,7 @@ export function NextPrevious({
       return (
         <DirectionalLinkBox
           label="Next"
-          link={{ href: tutorial.next.path }}
+          href={tutorial.next.path}
```
zchsh, line 61: I've switched the props interface here, to avoid `as` as it's no longer needed.

In [previous commits I tried supporting `as` in both `CardLink` and `MaybeInternalLink`](https://github.com/hashicorp/dev-portal/commit/83f4a08799abbb785a5bd7f7a8c49d97bd2b0ffa). This worked okay, but felt overly complex given that we no longer need to use the `as` prop. Definitely open to re-adding it in future iterations if we see a need for it.

zchsh:
@kendallstrautman @ashleemboyer Uses `CardLink` for `next/prev`.

As usual this ended up being a lil more complex than expected, at first I tried to support `as` as it was present in `DirectionalLinkBox`, then decided to refactor `DirectionalLinkBox` so it accepts `href` rather than the full set of `link` props (previously passed to `next/link`).Happy to adjust the approach here, let me know what y'all think 🙇‍♂️

src/views/tutorial-view/components/next-previous/components/directional-link-box/directional-link-box.module.css
```diff
@@ -1,101 +1,36 @@
...
-    text-align: right;
-    justify-content: flex-end;
-
-    &:hover {
-      box-shadow: var(--token-surface-high-box-shadow);
+  text-align: center;
+  justify-content: center;
+  height: 100%;
+
+  @media (--medium-up) {
```
ashleemboyer, line 9: Can you update this to use one of the `--dev-dot-` prefxied `@custom-media` in [`src/pages/style.css`](https://github.com/hashicorp/dev-portal/blob/main/src/pages/style.css)? If you need a new one all good to create it there! 😊

src/views/tutorial-view/components/next-previous/components/directional-link-box/directional-link-box.module.css
```diff
@@ -1,101 +1,36 @@
...
+    &.direction-previous {
```
ashleemboyer, line 10: Could you update these classnames to be `camelCase` instead of `kebab-case`?

src/views/tutorial-view/components/next-previous/components/directional-link-box/directional-link-box.module.css
```diff
@@ -1,101 +1,36 @@
...
-  color: var(--black);
-
-  & > div {
-    display: flex;
-  }
+  display: block;
 
-  &[data-direction='previous'] {
+  &.direction-previous {
     order: 0;
```
ashleemboyer, line 27: Just noticed these, are the `order` properties actually needed?

src/views/tutorial-view/components/next-previous/components/directional-link-box/index.tsx
```diff
@@ -1,46 +1,38 @@
...
-/*
- * @TODO style to spec
- * look at the `CardLink` component to use for base styles!
- */
-
 function DirectionalLinkBox({
-  link,
+  href,
   label,
   title,
```
ashleemboyer, line 20: Can this `title` prop be renamed to `ariaLabel` so it's clearer how it's used?

src/views/tutorial-view/components/next-previous/components/directional-link-box/types.ts
```diff
@@ -0,0 +1,8 @@
+export type DirectionOption = 'next' | 'previous' | 'final'
+
+export interface DirectionalLinkBoxProps {
+  href: string
+  label: string
+  title: string
```
ashleemboyer, line 6: Thoughts on referencing `CardLinkProps['href']` for `href` and `CardLinkProps['ariaLabel']` for `title` (potentially renaming to `ariaLabel`)?

ashleemboyer (changes requested):

src/components/maybe-internal-link/types.ts
```diff
@@ -0,0 +1,3 @@
+export type MaybeInternalLinkProps = {
+  href: string
+} & React.HTMLProps<HTMLAnchorElement>
```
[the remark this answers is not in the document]
BRKalow: `IntrinsicElements` is effectively a superset of `React.HTMLProps`, and I would say it's preferable in cases like this as it's more clear what tag we're "extending".

From the type definitions:
```ts
        interface IntrinsicElements {
            // HTML
            a: React.DetailedHTMLProps<React.AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>;
```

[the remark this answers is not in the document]
zchsh: ☝️ Sweet, thanks for the feedback here. I've updated with @ashleemboyer 's suggested block:

```ts
type AnchorElementProps = JSX.IntrinsicElements['a']

export interface MaybeInternalLinkProps extends AnchorElementProps {
  href: AnchorElementProps['href']
}
```

Out of curiosity, is the above equivalent to...

```ts
export type MaybeInternalLinkProps = JSX.IntrinsicElements['a']
```

... but we just want to be more explicit about the presence of `href`?

src/views/tutorial-view/components/next-previous/components/directional-link-box/directional-link-box.module.css
zchsh: I can try, but the incoming props are in lower case (`next`, `previous`, `final`). I opted for kebab-case since otherwise I think I'd have to mess around with capitalization. Let me know if that seems alright! ☺️

src/views/tutorial-view/components/next-previous/components/directional-link-box/directional-link-box.module.css
zchsh: Great call, the `order: 0` isn't, so I've removed it 🔪 I think it was there just to signal that `.direction-next` does need the `order` prop - it moves the icon after the text in the `flex` flow (removing the need for different markup if the icon is meant to be shown before or after the text).

src/views/tutorial-view/components/next-previous/components/directional-link-box/types.ts
zchsh: I'm on board with it! Have made the change since I don't have a strong opinion 👍

My weak opinion: Personally I don't see a huge difference either way - I _think_ TypeScript would warn us if the `string` type we declare here was incompatible with `CardLinkProps`. I slightly prefer the transparency of saying `string` rather than the relative opacity of `CardLinkProps['href']`. Super minor and weak opinion though, seeing that's it's `string` is just a hover away so 🤷‍♂️

src/components/maybe-internal-link/index.tsx
```diff
@@ -1,18 +1,17 @@
...
 
 function MaybeInternalLink({
   href,
   ...rest
-}: {
-  href: string
-} & React.HTMLProps<HTMLAnchorElement>): React.ReactElement {
+}: MaybeInternalLinkProps): React.ReactElement {
```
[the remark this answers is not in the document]
zchsh: 🔪 Removed! Much cleaner, ty 💖

src/components/card-link/types.ts
```diff
@@ -15,4 +16,9 @@ export interface CardLinkProps {
    * The destination of the link.
    */
   href: string
+
+  /**
+   * An optional string value that labels the link. For use when content visible in the DOM meant to give the link meaning is either missing, or does not accurately describe the link.
+   */
+  'aria-label'?: string
```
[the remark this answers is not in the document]
zchsh: ✅ Done, great call 👍

src/components/card-link/types.ts
```diff
@@ -1,3 +1,4 @@
+import { LinkProps } from 'next/link'
```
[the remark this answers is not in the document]
zchsh: 🔪 Removed, thanks for catching this 👍

src/views/tutorial-view/components/next-previous/components/directional-link-box/directional-link-box.module.css
zchsh: ✅ Great call, updated!

src/components/maybe-internal-link/types.ts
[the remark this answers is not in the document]
ashleemboyer: Yep those are the differences, if what you mean by "more explicit about the presence of `href`" is that `href` is required. I have a preference to export `interface`s over `types`, but I don't know why that is... I think `type` and `interface` are mostly equivalent? That's another gap in my TS knowledge maybe @BRKalow can answer. 😆 Or maybe @dstaley!

src/views/tutorial-view/components/next-previous/components/directional-link-box/directional-link-box.module.css
zchsh: ☝️ I suppose I could change the class names to be `.next` , `.previous`, & `.final` which would technically be `camelCase`, with no need to capitalization mess-arounds. I have a slight preference for the explicitness of prefixing with `direction` (matches the prop nicely, makes it a little more transparent), but very slight, so happy to adjust to whatever feels best for everyone else.

zchsh:
@ashleemboyer Thanks for the thorough review 💯  I think I got everything, let me know if there's anything I missed 🙇‍♂️

src/views/tutorial-view/components/next-previous/components/directional-link-box/directional-link-box.module.css
ashleemboyer: Ohhh I didn't notice that! Then `kebab-case` makes total sense here.

src/components/maybe-internal-link/types.ts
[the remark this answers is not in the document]
BRKalow: I think @kendallstrautman might have posted a resource recently, but interfaces are generally preferable when possible. I believe they provide more helpful type errors when used. 👍 

Here's a section from TS docs: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces

src/views/tutorial-view/components/next-previous/components/directional-link-box/directional-link-box.module.css
ashleemboyer: Okay great! Thank you 😊

src/components/card-link/types.ts
```diff
@@ -1,3 +1,4 @@
+import { MaybeInternalLinkProps } from 'components/maybe-internal-link/types'
 import { ReactNode } from 'react'
```
ashleemboyer, line 2: Two things here:

- Could you move the `'react'` import to be first?
- Could you update `MaybeInternalLink` to export the type from the component file to shorten this import statement?

src/views/tutorial-view/components/next-previous/components/directional-link-box/types.ts
```diff
@@ -0,0 +1,9 @@
+import { CardLinkProps } from 'components/card-link/types'
+export type DirectionOption = 'next' | 'previous' | 'final'
```
ashleemboyer, line 2: Two things here:

- Need a space between these lines
- `CardLinkProps` is exported from the component file, so the `/types` isn't needed in the import statement

```suggestion
import { CardLinkProps } from 'components/card-link'

export type DirectionOption = 'next' | 'previous' | 'final'
```

ashleemboyer (changes requested):
Couple of requests, and a quick question. Is the preview link behaving weirdly for you? This is what I see when I go to: https://dev-portal-git-zstutorial-next-prev-chrome-hashicorp.vercel.app/vault/tutorials/adp/transform

image

The text alignment is off in the prev/next cards but when I run this locally I don't see the issue. 🤔

src/components/maybe-internal-link/types.ts
```diff
@@ -0,0 +1,5 @@
+type AnchorElementProps = JSX.IntrinsicElements['a']
+
+export interface MaybeInternalLinkProps extends AnchorElementProps {
+  href: AnchorElementProps['href']
```
kendallstrautman, line 4: Interesting...curious why we need to redefine href here if it's already a part of the `AnchorElementProps` extension. Not a blocker, just curious.

src/views/tutorial-view/components/next-previous/components/directional-link-box/index.tsx
```diff
@@ -1,46 +1,38 @@
...
+import Text from 'components/text'
+import { DirectionalLinkBoxProps, DirectionOption } from './types'
 import s from './directional-link-box.module.css'
 
-interface DirectionalLinkBoxProps {
-  link: LinkProps
-  label: string
-  title: string
-  direction: DirectionOption
+const IconDict: { [k in DirectionOption]: typeof IconArrowRight16 } = {
```
kendallstrautman, line 11: Oh nice, that's a good way of working around the svg icon typing.

src/views/tutorial-view/components/next-previous/components/directional-link-box/index.tsx
```diff
@@ -1,46 +1,38 @@
...
-        </span>
-      </a>
-    </Link>
+    <CardLink
+      className={classNames(s.linkbox, s[`direction-${direction}`])}
+      href={href}
+      ariaLabel={ariaLabel}
+    >
+      <Icon className={classNames(s.icon, s[`direction-${direction}`])} />
+      <Text asElement="span" size={200} weight="medium">
```
kendallstrautman, line 32: Maybe check on the text color and weight, it seems a bit different than the design?

src/views/tutorial-view/components/next-previous/index.tsx
kendallstrautman: I think we maybe used to need the `as` because of query param issues in Learn, but now we should be ok 👍

kendallstrautman (changes requested):
Code looks good! I think we just need to double check the VQA and make style adjustments to match the spec.

zchsh:
@kendallstrautman @ashleemboyer I was confused about the responsive behaviour stuff...

I think it came out of switching to the custom media query `@media (--dev-dot-tablet-up)`. Which is weird cause it seems to be working for `TutorialMeta`, for example? Going to try to dive deeper to debug a bit. Where are we loading in custom-media queries? I know we do it [here](https://github.com/hashicorp/web-platform-packages/blob/f7ff67154f3c2de5bd3c0ae58f13c26f9df9c7c0/packages/postcss-config/index.js#L13) for the old media queries, but couldn't track down where we're doing it for the new ones.

src/views/tutorial-view/components/next-previous/components/directional-link-box/index.tsx
zchsh: ✅ Great call - the color was a off, needed to set it to `var(--token-color-foreground-primary)`.

Re: font-weight, I think it's correct should be `500` in both Figma and the browser. I think possibly Figma's rendering is a little different? I do notice that if we disable `-webkit-font-smoothing: antialiased;` on body, then we get a look that seems more closely matched to Figma.

zchsh:
@kendallstrautman @ashleemboyer ☝️ I've spiked out a fix for the `custom-media` issue in #259 . I think we'll likely want to take this to the `platform-postcss-config` package. But I think that tracks down the issue with `custom-media`.

For now, would we be comfortable sticking with the hard-coded `@media (min-width: 729px)` (equivalent to the doesn't-seem-to-be-working `--dev-dot-tablet-up`, and similar to the approach in `featured-tutorials-section.module.css`?). If so, I think this PR is ready for re-review - with that hard-coded value, responsive styles should now be working as expected.

Or should #259 block this PR, and we should resolve the CSS issue first rather than follow this up? 🤔

src/components/card-link/index.tsx
zchsh: ✅ makes sense - i've renamed to `ariaLabel` for `CardLink`.

src/components/maybe-internal-link/types.ts
zchsh: ☝️ I had the same question. Good to know! 💯

src/views/tutorial-view/components/next-previous/components/directional-link-box/index.tsx
zchsh: Yeah i Kinda threw up my arms and put `$TSFixMe` there for a sec... and then was like wait I can throw up my arms in a more "correct" way 🤣

zchsh:
@ashleemboyer @kendallstrautman I think I've got the `@media` issue nailed down - fixed in #259 , and rebased this PR, so styles should be as expected now 👍

I _just_ noticed we're using a `24px` icon for the "final tutorial in final collection" `Browse Tutorials`, going to switch that to `16px` for consistency (in absence of design spec, definitely open to iterating on that further if design wants something different, but figure consistent size is probably a good base assumption).

Let me know what ya'll think 🙇‍♂️

ashleemboyer (approved):
Looks great! 👏🏻

Re-reviewed the tablet-up media query, and that's working right in the preview link now. I wonder why the query worked locally but didn't work in the preview link. 🤔 Regardless, I think that PR you spiked to change how custom media are loaded is good! And it should prevent us seeing that issue again.

src/views/tutorial-view/components/next-previous/components/directional-link-box/directional-link-box.module.css
```diff
@@ -1,101 +1,39 @@
...
-  height: 1px;
-  overflow: hidden;
-  position: absolute;
-  white-space: nowrap;
-  width: 1px;
-}
-
 .linkbox {
-  composes: card;
   display: flex;
```
kendallstrautman, line 2: You may want to add `color` here so that the icon picks up the color as well. I think in the spec they are the same dark grey. 
```suggestion
  color: var(--token-color-foreground-primary);
  display: flex;
```

kendallstrautman (approved):
Looks good! Thanks for fixing that media query thing 😄 I'd say just look into that icon color thing as they look very contrasted as all-black. After that, feel free to 🚢 🥳

PR #256, opened 2022-04-05
Rename Disclosure to AccordionDisclosure

ashleemboyer:

---

## Relevant links

- [Preview link](https://dev-portal-git-ambrename-disclosure-hashicorp.vercel.app/swingset/components/accordiondisclosure) 🔎

## What

- Renames `Accordion` to `AccordionDisclosure`
- Resolves a couple of `typedef` issues in the component
- Adds block comments in the `types.ts` file for each component prop

## Why

- The design system has an [internal `Disclosure` component](https://design-system-components-hashicorp.vercel.app/utilities/disclosure) that is very generic for showing and hiding content. It has what we call "slots" for the trigger/activator/toggle element that shows/hides content, and a second slot for the content that should be disclosed. That component will be really useful for us to have in the future to ensure all disclosure components show/hide content consistently and accessibly. This rename opens up a space for adding the `Disclosure` utility component when we're ready to.
- The `typedef` issues were easy to resolve while I was looking at the file, and the block comments in `types.ts` were also easy to add while I was here.

## How

Good ol' find and replace. 😊

## Testing

- [ ] Go to the `AccordionDisclosure` swingset page
- [ ] The page should behave the same, only the component should be renamed

## Anything else?

Nope!


BRKalow (approved):
Radical!

PR #257, opened 2022-04-05
feat: implement FeaturedInCollections layout

zchsh:

---

## Relevant links

- [Preview link][two-cards] 🔎
- [Asana task](https://app.asana.com/0/1201987349274776/1201999966887523/f) 🎟️

## What

Implements `FeaturedInCollections`, which is an end-of-page UX component that lays out some `CollectionCards`.

## Why

To finish chroming the `/tutorials` view.

## How

Uses a `flex` layout for less "empty slot"-ish look in edge cases. Previously I'd used a `grid` layout. This worked great with one or two cards, but as soon as there are three cards, we can end up with a layout that looks like it has an "empty slot":

grid

I thought this looked less than ideal. I couldn't find a design spec to cover this situation, but figured switching to a `flex` layout might be preferable. This layout has identical behaviour in the most common cases, and in the "odd-number" type cases, we get a slightly cleaner (IMO) layout with the last item occupying the full width of the container:

flex

## Testing

- [ ] Visit [a page with one "featured in" collection][one-card], and ensure the cards match the design spec layout
- [ ] Visit [a page with two "featured in" collections][two-cards], and ensure the cards match the design spec layout
- [ ] Visit a page with three "featured in" collections, and ensure the cards match the design spec layout
    - I couldn't find such a page, so I set up [a story in Storybook][three-cards-storybook] so that I could test the use case, while also seeing responsive behaviour
    - Storybook may also be useful for reviewing responsive behaviour for one-up and two-up layouts.

## Anything else?

Note on collection cards: collection cards themselves are not yet implemented. We're waiting on a finalized design spec.

[one-card]: https://dev-portal-git-zstutorial-chrome-featured-in-c-2c74d9-hashicorp.vercel.app/vault/tutorials/getting-started-ui/getting-started-install
[two-cards]: https://dev-portal-git-zstutorial-chrome-featured-in-c-2c74d9-hashicorp.vercel.app/vault/tutorials/secrets-management/vault-pki-consul-connect-ca
[three-cards-storybook]: https://dev-portal-git-zsspike-featured-in-collections-zchsh.vercel.app/?path=/story/components-featuredincollections--three-up


src/views/tutorial-view/components/featured-in-collections/index.tsx
```diff
@@ -1,46 +1,43 @@
-import { getCollectionSlug } from 'views/collection-view/helpers'
-import CardLink from 'components/card-link'
-import { Collection as ClientCollection } from 'lib/learn-client/types'
-
-interface FeaturedInCollectionsProps {
-  collections: CollectionCardProps[]
-}
-
-// @TODO move this to the collection card component file when we make it
```
zchsh, line 9: Moved `CollectionCard` out to a separate file, even though we're not _quite_ ready to make it yet (design file still says WIP).

src/views/tutorial-view/components/featured-in-collections/index.tsx
```diff
@@ -1,46 +1,43 @@
...
-  numTutorials: number
-}
+import CollectionCard from 'views/tutorial-view/components/collection-card'
+import { FeaturedInCollectionsProps } from './types'
+import s from './featured-in-collections.module.css'
 
 // This should render the eventual `CollectionCard` component (doesn't exist yet)
 // which will be used on many other views
 export function FeaturedInCollections({
+  className,
```
zchsh, line 8: Passing `className` to the root element allows us to set spacing styles on the component, to space it relatively to adjacent components at the page level.

src/views/tutorial-view/components/featured-in-collections/featured-in-collections.module.css
```diff
@@ -0,0 +1,46 @@
+.heading {
+  margin: 0;
+  padding: 0;
+  font-weight: var(--token-typography-font-weight-semibold);
+  font-family: var(--token-typography-display-200-font-family);
```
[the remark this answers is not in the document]
zchsh: ✅ figured it out

src/views/tutorial-view/components/featured-in-collections/featured-in-collections.module.css
```diff
@@ -0,0 +1,44 @@
...
+  flex-wrap: wrap;
+  gap: 24px;
+  list-style: none;
+
+  /**
+   * Note: !important is necessary here as we're within a content container
+   * that uses nested generic selectors to style <ul />.
+   * TODO: convert the nested generic selectors to MDX components.
+   * Asana task:
+   * https://app.asana.com/0/1201987349274776/1201653123138729/f
```
zchsh, line 21: Going to try to resolve this issue next because it keeps coming up. I think there's two independent yet related things I'd like to tackle:
1. Move the content container out of `SidebarSidecarLayout`.
    - Proposed approach: Views like `tutorial-view` & `docs-view` will instead consume the content container, so that it can wrap the MDXContent part of things only (rather than the entire content area). I think Kendall already proposed something similar (or identical!) to this
    - No matter how well-encapsulated our content styles are, I think we want to do this. And seems relatively low-effort now that docs-view is better abstracted thanks to Ashlee's work on consolidating product-specific `additionalComponents`. So I'm gonna try to tackle this first
2. Do the MDX custom component thing for `<ul />` & `<ol />`.
    - Once we do the above container refactor, this will be less "necessary", but I'm hoping to have bandwidth to try it out anyways (eg if we add a similar "list of linked cards" or really "list of anything" custom component for use in the content area in the future, then it will come up again).

src/views/tutorial-view/components/collection-card/collection-card.module.css
```diff
@@ -0,0 +1,8 @@
+/**
+ * TODO: styles and functionality dependent on design spec being finalized.
+ * (currently "WIP" in Figma)
+ * Figma: https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Dev-Portal?node-id=6860%3A79233
```
kendallstrautman, line 4: For context, these card designs will be finished with the _collection_ view.

src/views/tutorial-view/components/collection-card/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+  slug,
+  numTutorials,
+  name,
+  description,
+  theme,
+}: CollectionCardProps) {
+  return (
+    <CardLink
+      className={classNames(s.root, className)}
+      href={getCollectionSlug(slug)}
```
kendallstrautman, line 23: Another note here is that some links will not work for product collections that haven't been added to dev dot yet. [Asana task here](https://app.asana.com/0/1201987349274776/1201991101510148) to address that

src/views/tutorial-view/components/featured-in-collections/featured-in-collections.module.css
kendallstrautman: I vote #1! Seems like the most straightforward way to address this problem and it also doesn't prevent us from pursuing #2 as we need it.

kendallstrautman: I don't think it should block this PR, we can follow up and remove the `!important`s when we make that change.

src/views/tutorial-view/tutorial-view.module.css
```diff
@@ -0,0 +1,3 @@
+.featuredInCollections {
+  margin-top: 48px;
```
kendallstrautman, line 2: Thoughts on adding the margin bottom as well? Or do you think that would be handled elsewhere? Perhaps its not needed, but its difficult to tell without the scaled layout. 

Seems like the designs have it as 48 - desktop, 50 - tablet, 32 - mobile. It's up to you. We can always add it later 👌

kendallstrautman (approved):
Looks great! 🚀  i'll defer to your discretion as to the order of making the adjustment with content to avoid the `!important`s

src/views/tutorial-view/components/featured-in-collections/featured-in-collections.module.css
ashleemboyer: If you do 2 first, will 1 still be needed?

src/views/tutorial-view/components/collection-card/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+import CardLink from 'components/card-link'
+import classNames from 'classnames'
+import { CollectionCardProps } from './types'
+import s from './collection-card.module.css'
+
+/**
+ * TODO: styles and functionality dependent on design spec being finalized.
+ * (currently "WIP" in Figma)
+ * Figma: https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Dev-Portal?node-id=6860%3A79233
+ */
```
ashleemboyer, line 11: 👍🏻

src/views/tutorial-view/components/collection-card/types.ts
```diff
@@ -0,0 +1,11 @@
+import { Collection as ClientCollection } from 'lib/learn-client/types'
+
+export interface CollectionCardProps
+  extends Pick<
+    ClientCollection,
+    'id' | 'name' | 'slug' | 'theme' | 'description'
```
ashleemboyer, line 6: Have you considered making all of these props one single prop (maybe named `collection`) that's object? One benefit would be less maintenance on the `CollectionCardProps` API, because the maintenance would move to the prop object if more display/logic attributes are needed from `ClientCollection`.

src/views/tutorial-view/components/collection-card/types.ts
```diff
@@ -0,0 +1,11 @@
...
+  > {
+  /** Optional class name to pass to the root element, which is expected to be the underlying CardLink component. */
+  className?: string
+  numTutorials: number
```
ashleemboyer, line 10: Formatting request - feel free to not apply this suggestion via GitHub PR UI and do it manually if that's preferable! 💖 

```suggestion
  /**
   * Optional class name to pass to the root element, which is expected to be
   * the underlying CardLink component.
   */
  className?: string

  /**
   * TODO: add description
   */
  numTutorials: number
```

src/views/tutorial-view/components/collection-card/types.ts
```diff
@@ -0,0 +1,11 @@
+import { Collection as ClientCollection } from 'lib/learn-client/types'
```
ashleemboyer, line 1: Should this component live in the main `src/components` directory? Or do you have plans for that already? I can see it being super reusable!

src/views/tutorial-view/components/featured-in-collections/index.tsx
```diff
@@ -1,46 +1,46 @@
...
-                <p>{c.theme} Logo</p>
-              </CardLink>
+            <li key={id} className={s.listItem}>
+              <CollectionCard
+                /**
+                 * Note: className here necessary because CardLink
+                 * doesn't automatically behave as one might expect.
+                 * It requires height 100% to be set on the inner <a /> element,
+                 * which we do through this className.
+                 */
```
ashleemboyer, line 30: This comment breaks up the markup a quite a bit (makes it a little harder to read), could it be moved to live right above the `.listItemCard` class block in `src/views/tutorial-view/components/featured-in-collections/featured-in-collections.module.css`?

src/views/tutorial-view/tutorial-view.module.css
ashleemboyer: Is the "edit on github" `StandaloneLink` what will be placed after this section? I poked through the designs for a minute but couldn't find a frame that specifies what'll come after these.

ashleemboyer (changes requested):
Looking great! ✨  I like the idea of using `flex` instead of `grid` for that awkward space that happens with 3 items.

> Storybook may also be useful for reviewing responsive behaviour for one-up and two-up layouts

It should be easier to test in the app next week or the week after, I'll be making the sidebar collapse on tablet/mobile as part of my upcoming work in that area! 🙏🏻

src/views/tutorial-view/components/collection-card/index.tsx
zchsh: Great callout - added a comment that links to that task 👍

src/views/tutorial-view/tutorial-view.module.css
zchsh: ☝️ both great callouts - i didn't noticed spacing was responsive, looks like there's other stuff I missed too 😅 I missed that in `NextPrevious` as well I think 😞 Will open a follow-up PR for that 👍

`Edit on GitHub` does appear to come after - [Figma link](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Dev-Portal?node-id=6298%3A73041) (should hopefully link directly to the artboard with the stuff on it; if not it's the `Learn components` page in the `Dev Portal` file).

zchsh: re: `margin-bottom`: It might be a weird MO, but my plan is to generally stick to `margin-top` for elements. I find it less confusing, I dunno 🤷 And then if an element is omitted from the flow, it "takes its spacing with it" (doesn't leave an accidental margin below itself). Let me know if it's cool if I take that approach, happy to adjust if we have more prescriptive guidance.

ashleemboyer: I don't think that's weird at all! "takes its spacing with it" is a super sound mentality. I'm a fan ✨

src/views/tutorial-view/components/featured-in-collections/featured-in-collections.module.css
zchsh: ☝️ great Q - i think either 2 or 1 will make the other much less necessary.

But I also think it's probably ideal if we do both? I guess if our styles are really well encapsulated then there's technically no harm in not doing `1` - I guess sibling selectors could still apply unexpectedly though? 🤔

src/views/tutorial-view/components/collection-card/types.ts
zchsh: I dig that! 💯 I like that it'll also prevent repetitious `thing={thing}` passing of props (without the dreaded `{...spread}`. Would it be alright if we handle that in the PR where we implement `CollectionCard` rather than this PR?

src/views/tutorial-view/components/collection-card/types.ts
zchsh: Ditto, can also see it being re-usable! For now, it' just used in one place, so I figure better not to abstract it earlier than necessary.

src/views/tutorial-view/components/featured-in-collections/index.tsx
zchsh: ✅ Moved

zchsh:
@ashleemboyer Thanks for the review, I think I've addressed everything!

One thing I'm hoping to defer is refactoring the `CollectionCard` props API - I'm thinking we can do this in a subsequent PR when we implement `CollectionCard` ➡️ 

@kendallstrautman I made those responsive spacing changes - thanks for calling those out, apologies for being kinda sloppy with following the design specs 😅 Will work on that for next PR 👍

src/views/tutorial-view/components/collection-card/types.ts
ashleemboyer: That's totally good with me! 👍🏻

ashleemboyer (approved):
💅🏻✨

PR #266, opened 2022-04-06
Migrate useOnClickOutside and useOnFocusOutside to TypeScript

ashleemboyer:

---

## Relevant links

- [Preview link](https://dev-portal-git-ambmigrate-click-focus-outside-to-ts-hashicorp.vercel.app/) 🔎

## What

- Renames `src/hooks/use-on-click-outside.js` and `src/hooks/use-on-focus-outside.js` to have `.ts` extensions
- Fixes all typedef issues that arose from changing the file extensions
- Updates both hooks to have the same argument API
- Adds `shouldListen` argument to `useOnFocusOutside`
- Uses `Array.some` rather than `Array.reduce` when checking the array of `refs` given for click/focus
- Updates `Popover` to use the `shouldListen` arguments
- Moves some Typescript interfaces defined in `Popover` to a new `types.ts` file in the component's folder (aligns with what we do for other components)
  - I updated some of the `PopoverProps` to reference `DialogArrowProps` since some are passed straight into that inner component.

## Why

- Don't feel like I need to rehash the benefits of TypeScript on this one. 😊
- The hooks are nearly the same, just listening to different types of events, so making their APIs and logic nearly the same improves how intuitive they are to use. If learn how to use one of them, it's now easier to learn how to use the other one.
- Two reasons for replacing `Array.reduce`:
  - It can be difficult to read and understand as quickly as other `Array` methods
  - `Array.some` will stop iterating as soon as it finds an array item that satisfies the condition, whereas `Array.reduce` iterates over every item. We only need to find a single ref object that satisfies the condition in the code, so this is a performance optimization.
- Updated `Popover` to leverage the `shouldListen` arguments so that the mouse, touch, and focus event listeners are only added to the `document` when the popover is open. That's the only time we care about those listening to those events.

## How

For the `typedef` fixing changes, I worked on those one at a time. For the other changes, I mainly moved around code, compared and contrasted the two hooks to make sure they matched, and tested that the usage of the hooks was still functional in `Popover`.

## Testing

- [ ] Go to a page where the `Popover` is used ([such as the search input on Waypoint's landing page](https://dev-portal-git-ambmigrate-click-focus-outside-to-ts-hashicorp.vercel.app/waypoint))
- [ ] Click the search input in the page header
- [ ] A popover should appear below it
- [ ] Click outside of the popover
- [ ] The popover should go away
- [ ] Click the search input again
- [ ] The popover should open
- [ ] Hit the TAB key until your browser focus leaves the popover
- [ ] The popover should close

## Anything else?

Nope!


src/components/popover/types.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+   * Arrow coloration.
+   */
+  themeClass: string
+
+  /**
+   * An object of style properties.
+   */
+  themeProps: $TSFixMe
```
dstaley, line 32: I think this can be updated to `CSSProperties`? Looks like it's being spread into an object being cast as that type.

dstaley (approved):

src/components/popover/types.ts
ashleemboyer: Great catch! Updated in https://github.com/hashicorp/dev-portal/pull/266/commits/582226575965eddebd117c0bf08894bf20cb9fd0

ashleemboyer: Do you have an opinion on module augmentation like described here? If I understand it correctly, it'll prevent us from having to cast so many objects in this component `as CSSProperties`.

https://github.com/frenic/csstype#what-should-i-do-when-i-get-type-errors

dstaley: I think that's an awesome idea! I think the only augmentation I'd like to see would be adding `[index: `--${string}`]: any;` to support custom properties. Adding `[index: string]: any;` would be a bit too permissive.

ashleemboyer: Nice! That's a great way to limit it to CSS variables, our whole goal! I'll write up a ticket for it. 😊

ashleemboyer: New ticket here: https://app.asana.com/0/1201987349274776/1202090858665463/f

PR #267, opened 2022-04-06
chore: address broken featured collection links

kendallstrautman:

---

## Relevant links

- [Preview link](https://dev-portal-git-kshandle-non-beta-collection-links-hashicorp.vercel.app/vault/tutorials/cross-products/kubernetes-consul-vault-pipeline#next-steps) 🔎
- [Asana task](https://app.asana.com/0/1199634971449915/1201991101510148) 🎟️

## What

This PR makes a small workaround so that featured collections in tutorial views don't link to 404 pages. If the product referenced in the collection is not a part of the beta rollout, the featured collection card links externally to Learn. 

This isn't ideal UX, but I think it's _ok_ for the sake of beta. These cards linked to a mix of external and internal routes is a bit odd, but since its not a long term solution I don't think we necessarily need to adjust the cards to support true external linking. Open to feedback on that though. 

Another option woulda be to not render cards for collections that aren't included in dev dot yet 🤷 

## Testing

- [Visit this tutorial,](https://dev-portal-git-kshandle-non-beta-collection-links-hashicorp.vercel.app/vault/tutorials/cross-products/kubernetes-consul-vault-pipeline#next-steps) go to the bottom of the page. Click on the first collection card to validate that internal links still work. Then click on a collection card that says 'terraform logo' and it should link to the proper collection in production Learn.

### DIFF
--- a/src/views/collection-view/helpers.ts
+++ b/src/views/collection-view/helpers.ts
@@ -9,6 +9,8 @@ import { splitProductFromFilename } from 'views/tutorial-view/utils'
  * reference a different product context
  */
 
+const BETA_PRODUCTS = ['vault', 'waypoint']
+
 export function getTutorialSlug(
   tutorialDbSlug: string,
   collectionDbSlug: string
@@ -20,5 +22,13 @@ export function getTutorialSlug(
 
 export function getCollectionSlug(collectionDbSlug: string): string {
   const [product, collectionFilename] = collectionDbSlug.split('/')
+  const isBetaProduct = BETA_PRODUCTS.indexOf(product) === -1
+
+  // if not a 'sanctioned product', link externally to Learn
+  // interim solution for BETA where not all products are onboarded
+  if (isBetaProduct) {
+    return `https://learn.hashicorp.com/collections/${collectionDbSlug}`
+  }
+
   return `/${product}/tutorials/${collectionFilename}`
 }
--- a/src/views/collection-view/helpers.ts
+++ b/src/views/collection-view/helpers.ts
@@ -9,8 +9,6 @@ import { splitProductFromFilename } from 'views/tutorial-view/utils'
  * reference a different product context
  */
 
-const BETA_PRODUCTS = ['vault', 'waypoint']
-
 export function getTutorialSlug(
   tutorialDbSlug: string,
   collectionDbSlug: string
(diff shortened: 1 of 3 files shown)


src/views/collection-view/helpers.ts
```diff
@@ -9,6 +9,8 @@ import { splitProductFromFilename } from 'views/tutorial-view/utils'
  * reference a different product context
  */
 
+const BETA_PRODUCTS = ['vault', 'waypoint']
```
kendallstrautman, line 12: Perhaps this is something we lift up, presuming other areas of the codebase may need this conditional treatment of products. Or does this already exist somewhere??

ashleemboyer: Agree that this should probably be higher up and I don't believe it exists yet, I'd like to try handling these with feature flags and _potentially_ adding something like `isBetaProduct` to the `CurrentProduct` context. I plan on toying around with this a little in primary nav work! What do you think of these ideas? At the very least, we can do the latter without feature flags and just hardcode the array of beta products in `CurrentProduct` context.

kendallstrautman: I think that's a good idea! Although in this case the collection links wouldn't be referencing the `CurrentProduct`....like they could be called within the vault product context (if that were current) and be referencing _other_ product collections, if that makes sense.

ashleemboyer: Ah that's a great point! I keep forgetting how that's possible in Learn content. 😅

We can probably create a new hook/util instead then that accepts a product slug as an argument, and checks against an array of beta products (like you have here) or another option is that we use a beta feature flag for each product. I think either would work! The latter would require a config change as more products are added, the former would require code changes. Both require rebuilds I believe, so I think they have the same tradeoffs!

src/views/collection-view/helpers.ts
```diff
@@ -20,5 +20,14 @@ export function getTutorialSlug(
 
 export function getCollectionSlug(collectionDbSlug: string): string {
   const [product, collectionFilename] = collectionDbSlug.split('/')
+  const isBetaProduct =
+    __config.dev_dot.products_with_content_preview_branch.includes(product)
+
+  // if not a 'sanctioned product', link externally to Learn
+  // interim solution for BETA where not all products are onboarded
+  if (isBetaProduct) {
```
ashleemboyer, line 28: I think this needs flopped!

```suggestion
  if (!isBetaProduct) {
```

ashleemboyer: Wait... I think I mean flipped? Words are hard. Flooped? 🤣

kendallstrautman: 🤣

ashleemboyer (approved):
Works beautifully! ✨

kendallstrautman:
@ashleemboyer I followed up and removed that duplicate `getCollectionSlug` function as well 😄

PR #271, opened 2022-04-07
Add AllProductData context, useIsBetaProduct hook, and getIsBetaProduct utility

ashleemboyer:

---

## Relevant links

- [Preview link](https://dev-portal-git-ambadd-use-is-beta-hook-hashicorp.vercel.app/waypoint) 🔎

## What

- Renames `products_with_content_preview_branch` config variable to `beta_product_slugs`
- Exports existing `productSlugsToNames` from existing `config/products.ts` file
- Adds `AllProductData` context, which exposes a `useIsBetaProduct` hook
- Updates `src/pages/_app.tsx` to wrap much of the app in `AllProductDataProvider`
- Adds `getIsBetaProduct` utility under `src/lib`
- Uses `getIsBetaProduct` in `src/layouts/sidebar-sidecar/server.ts`

## Why

- In a Slack conversation, it was decided that `beta_product_slugs` is a clearer name than  `products_with_content_preview_branch` in config
- `productSlugsToNames` is needed in the new `AllProductData` Context
- We need a global, DRY, optimized way to determine if a product is a beta product in multiple parts of the application. The existing need in `src/layouts/sidebar-sidecar/server.ts` shows that we need a utility that works in server-side code, and an upcoming need in #267 shows that we need a utility that works in client-side code. There will also be a client-side need in the in-progress primary header navigation refactor so that only beta product links are shown in the menus.
  - `AllProductData` Context solves the client-side need, and handles deriving its state once, which is why I opted for using a Context and handling the deriving there instead of in the `useIsBetaProduct` hook. The hook just handles looking up data and doesn't need to calculate it.
  - `getIsBetaProduct` solves the server-side need. It's less optimized since it calculates the boolean it returns every time it runs, but I think that's probably okay since the server-side code runs much less often than the client-side code.

## How

Not really sure how to explain this one in a way that doesn't repeat too much of the Why section above. Please ask questions though if there are any! 💖

## Testing

There isn't much to test here since the usage is minimal.

- [ ] The build check should pass
- [ ] Pages that use `SidebarSidecarLayout` should still work ([/waypoint](https://dev-portal-git-ambadd-use-is-beta-hook-hashicorp.vercel.app/waypoint) for example)

## Anything else?

- Would like to wait to add tests until after these are used more and we have determined if we want to keep them.
- Can add a follow-up PR for using the new hook in `getCollectionSlug`?
  - BTW @zchsh and @kendallstrautman, I just noticed there are two duplicate helpers by this name. They live in:
    - `src/views/collection-view/helpers.ts`
    - `src/views/product-tutorials-view/helpers.ts`
- Will have a PR Thursday or Friday that uses the new hook in the primary navigation header.

### DIFF
--- a/config/development.json
+++ b/config/development.json
@@ -7,7 +7,7 @@
     "max_static_paths": 0,
     "revalidate": 360,
     "content_preview_branch": "dev-portal",
-    "products_with_content_preview_branch": ["waypoint", "vault"]
+    "beta_product_slugs": ["waypoint", "vault"]
   },
   "learn": {
     "max_static_paths": 0
--- a/config/preview.json
+++ b/config/preview.json
@@ -7,7 +7,7 @@
     "max_static_paths": 10,
     "revalidate": 360,
     "content_preview_branch": "dev-portal",
-    "products_with_content_preview_branch": ["waypoint", "vault"]
+    "beta_product_slugs": ["waypoint", "vault"]
   },
   "learn": {
     "max_static_paths": 10
--- a/config/production.json
+++ b/config/production.json
@@ -7,7 +7,7 @@
     "max_static_paths": 10,
     "revalidate": 360,
     "content_preview_branch": "dev-portal",
-    "products_with_content_preview_branch": ["waypoint", "vault"]
+    "beta_product_slugs": ["waypoint", "vault"]
   },
   "learn": {
     "max_static_paths": 10
--- a/config/products.ts
+++ b/config/products.ts
@@ -44,4 +44,4 @@ const products: ProductGroup[] = productSwitcherSlugs.map(
   }
 )
 
-export { products }
+export { products, productSlugsToNames }
--- a/src/contexts/all-product-data.tsx
+++ b/src/contexts/all-product-data.tsx
@@ -1,0 +1,97 @@
+import { createContext, ReactNode, useContext, useMemo } from 'react'
+import { ProductSlug } from 'types/products'
+import { productSlugsToNames } from '../../config/products'
+
+/**
+ * Describes the shape of each product's state manged by this Context.
+ */
+interface ProductState {
+  isBetaProduct: boolean
+}
+
+/**
+ * Describes the shape of the state object this Context manages.
+ *
+ * NOTE this cannot be an interface or an error occurs: "A mapped type may not
+ * declare properties or methods"
+ *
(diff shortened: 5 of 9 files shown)


BRKalow:
@ashleemboyer what are your long-term thoughts on what else will go into `AllProductData`? Initially I had expected it to contain the data from `src/data` based on its name, but I see that is not the case.

ashleemboyer:
> @ashleemboyer what are your long-term thoughts on what else will go into `AllProductData`? Initially I had expected it to contain the data from `src/data` based on its name, but I see that is not the case.

@BRKalow I think that's exactly what we should put in it! Since that wasn't the main goal of this PR, I didn't attempt to include that data yet because I'm not sure what the most performant way to do that is and wanted to discuss more as a group. I _think_ the way we'd hydrate the data would be to import all of those data files in `_app.tsx` (or via a helper so that file doesn't get too long?) and then pass it in as a prop to the Provider. Not sure what the implications are yet though.

I see `AllProductData` being a context that stores everything and exposes multiple hooks for accessing specific data. I think it could mostly replace the `CurrentProduct` context, and expose it's own `useCurrentProduct` hook. If we can pull it off, a huge benefit of this would be not having to import from src/data to hydrate product data in every single page component anymore.

BRKalow:
Makes sense, thanks. So would the context have all of the product data all the time? I'm not sure if there's a super elegant way to get at the product data outside of going through `getStaticProps` without impacting bundle size, but that can be a separate conversation haha.

Just offering this up, but the `getIsBetaProduct` utility seems reasonable for use in all scenarios. That computation is so light I'm not sure it makes sense to worry about optimizing it yet.

ashleemboyer:
@BRKalow `AllProductData` and `useIsBetaProduct` were both low effort to create and enable us to experiment with storing all product data in one place to see how it affects the app. (Also I misspoke earlier and meant to say in my other comment that we'd import just the `src/data/{productSlug}.json` files, not _all_ of them.) Are there concerns you have with adding them?

BRKalow:
No, if you've got plans to experiment with the context further then it seems reasonable to me! 😃

 If we don't have immediate ideas on further usage of the context though, then I would say this adds unnecessary complexity for the needs of `getIsBetaProduct`.

ashleemboyer:
@BRKalow yep, I'd like to do some experimenting with the context further! 😊

BRKalow (approved):

PR #273, opened 2022-04-07
feat: implement InteractiveLabCallout component

zchsh:

---

## Relevant links

- [Preview link][preview] 🔎
- [Asana task](https://app.asana.com/0/1201987349274776/1201999966887538/f) 🎟️

## What

Implements the `InteractiveLabCallout` component, and exposes it for use in the `/tutorials` view.

## Why

To implement the tutorials view UI.

## How

- Migrates the existing component from `learn`
- Styles to match [the design spec](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Dev-Portal?node-id=6298%3A73033)

## Testing

- [ ] Visit a view with `InteractiveLabCallout`, such as [`/vault/tutorials/app-integration/agent-read-secrets`][preview]. Ensure the `InteractiveLabCallout` is rendering as expected. Ensure the button opens the interactive lab.

## Anything else?

Debated renaming the component folder so it's `kebab-case` of the component name (this would also make Swingset happy). Have left without this rename for now, so that `lab-callout` and `lab-embed` are adjacent in the components folder (this folder is getting large; both those components are specific to `tutorials-view` so maybe we'd want to move them into `tutorials-view/components`?)

[preview]: https://dev-portal-git-zstutorial-interactive-lab-callout-hashicorp.vercel.app/vault/tutorials/app-integration/agent-read-secrets


src/components/button/button.module.css
```diff
@@ -15,6 +15,7 @@
   font-size: var(--font-size);
   justify-content: center;
   line-height: var(--line-height);
+  min-height: var(--min-height);
```
zchsh, line 18: I've added this to match the [HDS implementation](). Without it, it seems that `small` size buttons are smaller than the expected `28px` height.

Happy to drop these changes from this PR and handle later - aim here was to match the design spec exactly, but I think it could make sense to get very close (does not require these changes) and iron out the `button` details more intentionally (if they don't seem quite right).

src/components/lab-callout/interactive-lab-callout.module.css
```diff
@@ -0,0 +1,61 @@
...
+.description {
+  composes: hds-typography-body-100 from global;
+
+  /**
+   * TODO: raise line-height rounding issue with design-systems
+   * TLDR; line-height defined in relative-to-font-size units often
+   * ends up with weird rounding. Re-defined in rems here to fix that.
+   * (line-height is already indirectly rems, since font-size is rems).
+   * Details in Asana task:
+   * https://app.asana.com/0/1201987349274776/1202093639417968/f
```
zchsh, line 34: This is _really_ nitpicky. Intent here was to try to match the design spec exactly, and this `line-height` issue was preventing me from doing so.

src/pages/swingset/[[...swingset]].tsx
```diff
@@ -4,13 +4,15 @@ import CoreDevDotLayout from 'layouts/core-dev-dot-layout'
...
 import { SearchProvider } from '@hashicorp/react-search'
+import InstruqtProvider from 'contexts/instruqt-lab'
 import SwingsetColorToken from '__swingset-components/swingset-color-token'
 import SwingsetTestIcon from '__swingset-components/swingset-test-icon'
 
 // TODO: temporary for building new navigation menus
 import { NavigationHeaderDropdownMenu } from 'components/navigation-header/components'
 
 const components = {
+  InstruqtProvider,
```
Comment (zchsh) [inline src/pages/swingset/[[...swingset]].tsx:15]: Necessary to get `InteractiveLabCallout` to render in Swingset.

src/views/tutorial-view/utils/mdx-components.tsx
```diff
@@ -2,15 +2,13 @@ import codeBlockPrimitives from '@hashicorp/react-code-block/mdx'
...
+import InteractiveLabCallout from 'components/lab-callout'
 import VideoEmbed from 'components/video-embed'
 
 /**
  * @TODO move over these components from learn, update to new spec
  *
  * import Accordion from 'components/accordion'
- * import VideoEmbed from 'components/video-embed'
- * import ImageConfig from 'components/image-config'
- * import InteractiveLabCallout from 'components/lab-callout'
```
zchsh, line 13: These are all done now! 🎉

src/components/button/button.module.css
ashleemboyer: Which buttons are you seeing this behavior for? I checked a couple at https://developer.hashi-mktg.com/swingset/components/button in Chrome, Safari, and Firefox and haven't found the same thing you have yet.

ashleemboyer: Did some investigation, and turns out the `line-height` for the `small` variant was changed 10 days ago! 😄
https://github.com/hashicorp/design-system/commit/822154e4b84447d93a574c5845754d5beb856788

The new value is `0.875rem` (as you have it below), so you can just add that and remove the `min-height` stuff since it isn't needed.

ashleemboyer (changes requested):
Requesting this change while I work on reviewing the rest of the PR. 😊

src/components/lab-callout/interactive-lab-callout.module.css
```diff
@@ -0,0 +1,61 @@
...
+   */
+  line-height: 1.125rem;
+  color: var(--token-color-surface-faint);
+
+  /**
+   * Note: !important needed, as generic <p> styles leak from DevDotContent.
+   * Task to fix:https://app.asana.com/0/1201987349274776/1201653123138729/f
+   */
+  margin: 4px 0 0 0 !important;
```
ashleemboyer, line 43: Same comment here as the other `!important`, I'm not seeing any styles leak in from `DevDotContent`. They're coming from the user agent, then the `hds-typography-body-100` class, then this local `description` class. Since the `hds-typography-body-100` clears out the `margin`, you should be able to just set `margin-top: 4px` here.

```suggestion
  margin-top: 4px;
```

src/components/lab-callout/docs.mdx
```diff
@@ -0,0 +1,40 @@
...
+# Interactive Lab Callout
+
+This component showcases a callout to draw attention to the interactive lab feature. It is a non-configurable component that is defined inline along content. This means that this component accepts no props or customization options whatsoever. It offers a button that when clicked, will open the interactive lab tab. If this component is defined inline on a tutorial that does not have an interactive lab, it will not render.
+
+## Playground
+
+<!--
+Note: component is actually called InteractiveLabCallout.
+But the name mismatch makes Swingset angry.
+I think we'd need to rename the component folder in order
```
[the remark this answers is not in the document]
kendallstrautman: That makes sense! Feel free to rename it.

src/components/lab-callout/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+  return (
+    <div className={s.interactiveCallout}>
+      <div>
+        <h2 className={s.title}>Launch Terminal</h2>
+        <p className={s.description}>
+          This tutorial includes a free interactive command-line lab that lets
+          you follow along on actual cloud infrastructure.
+        </p>
+        <div className={s.ctaButton}>
+          <Button
```
ashleemboyer, line 22: I think this needs the `type="secondary"` prop.

src/components/lab-callout/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+            // TODO: <Button /> within dev-dot does not take "data-heap-track"
```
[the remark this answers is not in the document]
kendallstrautman: Ya maybe we should wait to handle analytics until we have a more unified plan. I'd say remove it and just add a note that we may want to track interaction here 👍

ashleemboyer (changes requested):

ashleemboyer:
> Debated renaming the component folder so it's kebab-case of the component name (this would also make Swingset happy). Have left without this rename for now, so that lab-callout and lab-embed are adjacent in the components folder (this folder is getting large; both those components are specific to tutorials-view so maybe we'd want to move them into tutorials-view/components?)

I think moving these to `tutorials-view/components` is totally appropriate!

kendallstrautman:
Just a little VQA note, is the button supposed to be white? As in the designs? Not sure if another decision was made to make it blue 
Screen Shot 2022-04-07 at 9 07 35 AM

kendallstrautman:
I'm wondering if want it to break sooner than the 'normal' breakpoints to remove the image under tablet 

Screen Shot 2022-04-07 at 9 10 58 AM

src/components/lab-callout/docs.mdx
```diff
@@ -0,0 +1,40 @@
...
+
+```md
+## This is the first heading within the mdx content body
+
+Consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
+
+<InteractiveLabCallout />
+
+Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
+```
```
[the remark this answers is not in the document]
zchsh: I think the unstyled `pre` is expected (though very much agreed, not ideal). I don't believe we've set up `CodeBlock` as a custom component within Swingset.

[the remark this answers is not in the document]
ashleemboyer: Ah, okay! That makes sense. I wasn't sure if that was setup by default or not.

src/components/button/button.module.css
zchsh: I think the `min-height` might still be needed - I'm seeing a height of `26px` when I remove it:

CleanShot 2022-04-07 at 17 27 02@2x

vs the expected `28px` when I have it in place:

CleanShot 2022-04-07 at 17 27 19@2x

Let me know if that’s acceptable, happy to adjust if there’s another approach we want to take to meet the design spec 🙇‍♂️ Or also happy to discard the change if we don't mind the `2px` difference.

src/components/lab-callout/index.tsx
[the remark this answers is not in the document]
zchsh: ✅ I've added a comment on Heap, and set up a [follow-up task to support data-heap-track in Button](https://app.asana.com/0/1201987349274776/1202096857327389/f).

src/components/lab-callout/interactive-lab-callout.module.css
zchsh: As in `.title` - thanks for this, have cleaned this up! ✅

src/components/lab-callout/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const ctx = useInstruqtEmbed()
+
+  if (!ctx.labId) {
+    return null
+  }
+
+  return (
+    <div className={s.interactiveCallout}>
+      <div>
+        <h2 className={s.title}>Launch Terminal</h2>
```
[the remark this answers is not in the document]
zchsh: Great catch - have switched to `<p />` ✅

src/components/lab-callout/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export default function InteractiveLabCallout() {
+  const ctx = useInstruqtEmbed()
+
+  if (!ctx.labId) {
+    return null
+  }
+
+  return (
+    <div className={s.interactiveCallout}>
```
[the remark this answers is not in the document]
zchsh: Ah nice, thanks for calling this out - have used `Card` now, passing `className` 👍 also passing `elevation="low"` (and added support for that in `Card`) to match the design spec ✅

src/components/lab-callout/docs.mdx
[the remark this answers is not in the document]
zchsh: ✅ renamed. Swingset is happy now 🌞

src/components/lab-callout/interactive-lab-callout.module.css
```diff
@@ -0,0 +1,61 @@
+.interactiveCallout {
+  align-items: center;
+  background-color: var(--token-color-foreground-strong);
+  border-radius: 6px;
+  padding: 16px;
+
+  @media (--medium-up) {
```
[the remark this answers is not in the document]
zchsh: ✅ yep, great call, replaced

src/components/lab-callout/index.tsx
zchsh: ✅ made this `"secondary"` - i think this addresses @kendallstrautman 's question about the blue button below as well (talk about me missing basic stuff from the design spec 😅 sorry about that!)

kendallstrautman (approved):
LGTM! Great work 😄

src/components/button/button.module.css
ashleemboyer: Took another look at the padding in small buttons here: https://design-system-components-hashicorp.vercel.app/components/button, and found that our top/bottom padding is off by 1! That update was made in the same one as linked above for the `line-height` change and I didn't notice it until now.

So request is:

- Remove `--min-height` (For context: we shouldn't be setting any `height` properties because it makes buttons less responsive with font size. Buttons (or containers with text) should always grow with the font size.)
- Update the padding top/bottom for the `.small` variant to from `--padding-top-bottom: calc(6px - var(--border-width));` to `--padding-top-bottom: calc(7px - var(--border-width));`

ashleemboyer (changes requested):

src/components/button/button.module.css
zchsh: ☝️ sounds good, just made this change ✅

For context on why I added `min-height`: I was going off the HDS implementation, which seems to use `min-height` (I think unlike `height` this still allows buttons to grow with the font size).

ashleemboyer (approved):

src/components/button/button.module.css
ashleemboyer: That's true, but unfortunately `min-height` doesn't work in the other direction as far as I recall, like when users set their font size to something smaller than whatever `1rem` comes out to be. What'll happen is the text will shrink but the button will stop at whatever `min-height` is defined to be, and then the text will no longer be vertically aligned. That adds a need for `min-height` to be set in `rem` units, which is complex in our case because we're using `px` units for `padding` (that's another topic lol), so then we'd need to use `calc` to do `px` + `rem` math and IMO that's all more complex than ensuring height is correct in both font-size directions using just the `line-height` and `padding` properties.

PR #279, opened 2022-04-08
feat: build tutorial sidebar

zchsh:

---

## Relevant links

- [Preview link](https://dev-portal-git-zstutorial-sidebar-hashicorp.vercel.app/swingset/components/tutorialssidebar) 🔎
- [Asana task](https://app.asana.com/0/1201987349274776/1201999966887514/f) 🎟️

## What

Implements the tutorial view sidebar.

## Why

To continue with tutorial view chroming work.

## How

- Takes an approach that relies composition of smaller existing elements from the existing docs `Sidebar`, rather than adding complexity to the props API of the existing docs `Sidebar`

## Testing

- [ ] Visit the [Swingset page](), and ensure the examples look as expected
    - Looking at the markup in each `LiveComponent` may be helpful to get a sense of component use. Of course, in practice we'll be able to `map` over incoming sections and links within those sections rather than manually repeating them. But hopefully the general idea around composition is clear.

## Anything else?

- Focus in this PR is establishing implementation approach to `TutorialSidebar`. Intent is to fully match the design spec in a subsequent PR (as those changes will affect the docs `Sidebar` as well)
- Intent is to implement the _use_ of this component on the "Tutorial Product Landing", "Collection", and "Tutorial" views in subsequent PRs.

### DIFF
--- a/src/components/sidebar/components/sidebar-nav/sidebar-nav-menu-item/index.tsx
+++ b/src/components/sidebar/components/sidebar-nav/sidebar-nav-menu-item/index.tsx
@@ -1,4 +1,5 @@
 import { KeyboardEventHandler, useEffect, useRef, useState } from 'react'
+import Link from 'next/link'
 import classNames from 'classnames'
 import { IconChevronRight16 } from '@hashicorp/flight-icons/svg-react/chevron-right-16'
 import { IconExternalLink16 } from '@hashicorp/flight-icons/svg-react/external-link-16'
@@ -11,21 +12,35 @@ interface SidebarMenuItemProps {
   item: MenuItem
 }
 
-function SidebarNavLink({ item }: SidebarMenuItemProps) {
-  const { isActive, fullPath, href, title } = item
+const SidebarNavLink = ({ item }) => {
+  if (item.fullPath) {
+    return (
+      <li>
+        <Link href={item.fullPath}>
+          <a
+            aria-current={item.isActive ? 'page' : undefined}
+            className={s.sidebarNavMenuItem}
+            // TODO: this might break some accessible labels, probably need aria-label
+            dangerouslySetInnerHTML={{ __html: item.title }}
+          />
+        </Link>
+      </li>
+    )
+  }
+
   return (
-    <li className={s.sidebarNavListItem}>
+    <li>
       <MaybeInternalLink
-        aria-current={isActive ? 'page' : undefined}
+        aria-current={item.isActive ? 'page' : undefined}
         className={s.sidebarNavMenuItem}
-        href={fullPath || href}
+        href={item.href}
       >
         <span
           className={s.navMenuItemLabel}
           // TODO: this might break some accessible labels, probably need aria-label
-          dangerouslySetInnerHTML={{ __html: title }}
+          dangerouslySetInnerHTML={{ __html: item.title }}
         />
-        {href ? <IconExternalLink16 /> : null}
+        <IconExternalLink16 />
       </MaybeInternalLink>
     </li>
   )
--- a/src/components/sidebar/components/sidebar-nav/sidebar-nav-menu-item/index.tsx
+++ b/src/components/sidebar/components/sidebar-nav/sidebar-nav-menu-item/index.tsx
@@ -1,4 +1,5 @@
 import { KeyboardEventHandler, useEffect, useRef, useState } from 'react'
+import classNames from 'classnames'
 import { IconChevronRight16 } from '@hashicorp/flight-icons/svg-react/chevron-right-16'
 import { IconExternalLink16 } from '@hashicorp/flight-icons/svg-react/external-link-16'
 import MaybeInternalLink from 'components/maybe-internal-link'
@@ -92,9 +93,13 @@ const SidebarNavSubmenu: React.FC<SidebarMenuItemProps> = ({ item }) => {
   )
 }
 
+export function HorizontalRule({ className }: { className?: string }) {
+  return <hr className={classNames(s.divider, className)} />
+}
+
 const SidebarNavMenuItem: React.FC<SidebarMenuItemProps> = ({ item }) => {
   if (item.divider) {
-    return <hr className={s.divider} />
+    return <HorizontalRule />
   }
 
   // TODO: 2022-01-03: designs show a heading on the product home page,
--- a/src/components/sidebar/components/sidebar-nav/sidebar-nav-menu-item/index.tsx
+++ b/src/components/sidebar/components/sidebar-nav/sidebar-nav-menu-item/index.tsx
@@ -108,7 +108,7 @@ const SidebarNavSubmenu: React.FC<SidebarMenuItemProps> = ({ item }) => {
   )
 }
 
-export function HorizontalRule({ className }: { className?: string }) {
+function HorizontalRule({ className }: { className?: string }) {
   return <hr className={classNames(s.divider, className)} />
 }
 
@@ -134,5 +134,5 @@ const SidebarNavMenuItem: React.FC<SidebarMenuItemProps> = ({ item }) => {
   return <SidebarNavLink item={item} />
 }
 
-export { SidebarNavLink }
+export { HorizontalRule, SidebarNavLink }
 export default SidebarNavMenuItem
--- a/src/components/sidebar/components/sidebar-nav/sidebar-nav-menu-item/index.tsx
+++ b/src/components/sidebar/components/sidebar-nav/sidebar-nav-menu-item/index.tsx
@@ -1,5 +1,4 @@
 import { KeyboardEventHandler, useEffect, useRef, useState } from 'react'
-import Link from 'next/link'
 import { IconChevronRight16 } from '@hashicorp/flight-icons/svg-react/chevron-right-16'
 import { IconExternalLink16 } from '@hashicorp/flight-icons/svg-react/external-link-16'
 import MaybeInternalLink from 'components/maybe-internal-link'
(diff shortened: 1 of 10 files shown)


src/components/sidebar/components/sidebar-nav/sidebar-nav-menu-item/sidebar-nav-menu-item.module.css
```diff
@@ -7,7 +11,8 @@
   color: var(--token-color-palette-neutral-500);
   display: flex;
   font-size: 14px;
-  font-weight: 400;
+  font-weight: var(--token-typography-font-weight-regular);
+  gap: 8px;
```
zchsh, line 15: This `gap` property ensures there is space between the item's text and icon (if present), which can be an issue for long text on small viewports.

zchsh:
@ashleemboyer Re: thread in Slack, have done my best to make some changes here to get `TutorialSidebar` more in line with the existing docs `Sidebar` component.

Further details: I've tried to take an approach that composes the smaller elements we've already built for `Sidebar`, rather than inheriting a more complex props API and toggling off functionality such as sidebar filter. Thanks for your excellent notes in Slack 💖

Here's where I'm at on each of them, very much open to iterating further on them and making any changes necessary:

- Have moved `SkipToMainContent` link after the "back to" link, to match the existing docs sidebar
- Have ensured the "back to" link in `TutorialSidebar` uses `StandaloneLink`
- Have added support for a title for `TutorialSidebar`'s `<nav />` element, by adding support for `ariaLabelledBy` to TutorialSidebar. I've updated the examples in Swingset as a demo (definitely open to feedback here on whether this usage feels correct)
- Re: markup, have got the wrapping element as a `<nav />`, which I think defined its own content context similar to `<section />`. With this idea in mind, I then have a consistent `<h1>` element for the top-level title of the sidebar (`Tutorials` on the product landing and collection pages; and the current collection title on tutorial pages). Each section of links has a section title (an `<h2 />`) and then a `<ul />` holding the list of links under that section title. Very open to adjusting this, was just debating what markup to use for `SectionTitle` elements, and `<h2 />` seemed more correct than eg `<p>` or `<div>`.
- Have switched to using the same `<hr />` component as in the docs sidebar. To match design spec, due to the way `TutorialSidebar` uses a flex layout with `gap` rather than setting `margin` on individual elements, it seems we currently need to pass in a `className` to adjust the `margin-top` and `margin-bottom`. Definitely open to trying to reconcile that difference in whatever way we can (eg we could change `TutorialSidebar` to use individual `margin` properties in this PR, or in a follow-up PR we could change the docs `Sidebar` to use a flex layout with `gap`).
- Re: `SidebarNavLink`, I previously had some changes there that removed some repetitious markup, but weren't substantial otherwise. I've reverted those so that we can focus on other aspects of the PR 👍

zchsh:
@kendallstrautman Tagging you here as well for review in case you have any opinions on the markup structure of `TutorialSidebar`, and on the composition-y pattern used in this PR.

I still have this marked as draft cause I'm not sure if the markup will need to change substantially based on the discussion in [Ashlee's YTB Slack thread](https://hashicorp.slack.com/archives/C01KCU4HDPY/p1649425685653469). If we're okay with the markup and broader approach as-is, then I think this'll be ready for a more "final" review 👍

zchsh:
@ashleemboyer @kendallstrautman  I decided to just mark this ready for full review now that i've got the PR template filled out 👍

With the caveat: in the context of this PR, I think it could make sense to focus on the overall approach (rather than eg the design spec) since I think we have a decent amount of substance there already. I _think_ we're already pretty close to design spec... but I think it could make sense to focus for now on the structure, and follow up either in a subsequent PR, or subsequent commits, once we've finalized the approach.

src/components/sidebar/components/sidebar-nav/sidebar-nav-menu-item/sidebar-nav-menu-item.module.css
ashleemboyer: This should already be handled with the `margin-right` property inside of the `navMenuItemLabel` class defined lower in this file. 😊

src/components/sidebar/components/sidebar-nav/sidebar-nav-menu-item/index.tsx
```diff
@@ -107,9 +108,13 @@ const SidebarNavSubmenu: React.FC<SidebarMenuItemProps> = ({ item }) => {
   )
 }
 
+export function HorizontalRule({ className }: { className?: string }) {
+  return <hr className={classNames(s.divider, className)} />
+}
```
ashleemboyer, line 113: Can you remove the `export` from here and move `HorizontalRule` down to the `export { SidebarNavLink }` statement? It makes it easier to tell what all is exported from this file and what isn't.

src/components/tutorials-sidebar/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+    <nav className={s.root} aria-labelledby={ariaLabelledBy}>
+      <div className={s.backToLink}>
+        <StandaloneLink
+          href={backToLink.href}
+          text={backToLink.text}
+          icon={<IconChevronLeft16 />}
+          iconPosition="leading"
+          textSize={200}
+        />
+      </div>
```
ashleemboyer, line 32: `StandaloneLink` supports a `className` prop, did you try that already and find that a wrapper `div` worked out better?

src/components/tutorials-sidebar/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+
```
ashleemboyer, line 33: Not sure if this space between was intentional or not, feel free to ignore if intentional!

```suggestion
```

src/components/tutorials-sidebar/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+    </ul>
+  )
+}
+
+function ListItem({ href, isActive, text, isExternal }: ListItemProps) {
+  const parsedIsExternal =
+    typeof isExternal == 'boolean' ? isExternal : isAbsoluteUrl(href)
+  const hrefOrFullPath = parsedIsExternal ? { href } : { fullPath: href }
+  return <SidebarNavLink item={{ isActive, title: text, ...hrefOrFullPath }} />
+}
```
ashleemboyer, line 65: Curious on why `isExternal` is being added rather than using the two separate props (`href` & `fullPath`) that the docs sidebar currently relies on, and are in effect being used here anyways?

src/components/tutorials-sidebar/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const Wrapper = visuallyHidden ? VisuallyHidden : Fragment
+  const Elem = as
+  return (
+    <Wrapper>
+      <Elem className={s.sectionTitle} id={id}>
+        {text}
+      </Elem>
+    </Wrapper>
+  )
+}
```
ashleemboyer, line 82: Let's chat about this one more on Monday. It's almost there, but it gives too much flexibility for using the incorrect heading level for labeling sections. The levels also need to be updated, because we can't have multiple `h1` elements in a page, and I also expect the sidebar title will never be the most important heading element in any page.

ashleemboyer (changes requested):

src/components/tutorials-sidebar/index.tsx
zchsh: I tried that initially - the issue I ran into was that on `focus`, we modify `margin` as well, & that conflicted. The wrapper avoids that issue ☺️

src/components/tutorials-sidebar/index.tsx
zchsh: @ashleemboyer  Sounds good - I think I confused`<nav />` being classified as [sectioning content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#sectioning_content) with the concept of the "sectioning root" ... and also, TIL this seems to have been [_supposed_ to be part of HTML5](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#multiple_h1_elements_on_one_page) but seems to just have not been implemented yet? Fascinating stuff 😅 

Happy to update these however makes sense. I guess since there's really no "section" isolation, the`h1` could become an `h2`, and the `h2`s would become `h3`s, and that might have a stronger guarantee to fit in nicely with the outline of the page as a whole? 🤔

src/components/tutorials-sidebar/index.tsx
zchsh: Great question. Here's the rationale:
1. I wanted to modify `SidebarNavLink` as little as possible, so leave the `href` & `fullPath` in place...
2. This meant that to get an external icon, I'd have to pass `fullPath`; to omit it I'd have to pass `href`...
3. I wanted the props for `ListItem` to be as simple as possible - `href` for all links. We can handle this pretty nicely with `MaybeInternalLink` so figured this was worth it...
5. Initially I had `isExternal` not exposed as a prop, and directly determined it from `isAbsoluteUrl`. Worked fine, mostly...
6. But then realized there will be some non-internal links marked with the external link icon (namely `All Tutorials`) which need to be marked with the external icon...
7. So that led me to expose `isExternal` as an optional prop, with the default behaviour being show it only if `isAbsoluteUrl(href)`

I considered that we could, instead of handling this at the `ListItem` component level, lift up the responsibility to where `ListItem` is used, and ultimately to whatever data massaging we do in the tutorials view. However I think the data massaging would be more work than what we're doing here - we'd have to either author content with separate `href` and `fullPath` properties depending on link location, or traverse the incoming data or something.

Let me know if that makes sense 🙇‍♂️

src/components/sidebar/components/sidebar-nav/sidebar-nav-menu-item/sidebar-nav-menu-item.module.css
zchsh: Ah cool - I'd started to try to reduce some repetition in `SidebarNavLink` but then reverted my changes, forgot to fix this 👍 ty 💖

src/components/sidebar/components/sidebar-nav/sidebar-nav-menu-item/index.tsx
zchsh: ✅ Great call, moved!

src/components/tutorials-sidebar/index.tsx
zchsh: ☝️ I took a shot at the semantic markup changes, and at reducing the chance of incorrect heading levels:

- `TutorialSidebar` now requires a `title` string prop, which _always_ renders an `h2` into the `<nav />`, and _always_ references it with `aria-labelledby`
- `SectionTitle` now _only_ ever renders an `h3`

Definitely just making an attempt here - looking forward to ironing it out Monday 🙇‍♂️

src/components/tutorials-sidebar/index.tsx
ashleemboyer: Thank you for the thorough context! Makes sense. 🙏🏻

> It's possible that we could eliminate the need for the isExternal prop by being super consistent in design on only showing the "external" link for links that are outside dev-portal.

Yep, this is what we need to do. "All Tutorials" might be "external" to the current page, but it is not external to the whole site. The external link icon should only be used for links that are totally outside of the DevDot site.

You don't need to make any changes in this PR, can save it for a follow-up if you've got one coming. 😊

src/components/tutorials-sidebar/index.tsx
ashleemboyer: This is good! Here's the reference I was waiting to share today: https://www.w3.org/WAI/tutorials/page-structure/headings/. It has an example of using a single `h1` for the most important heading on the page, and then using `h2`s to give the sidebar a title, and `h3`s to title each section of the sidebar. Nicely done! 👏🏻

ashleemboyer (approved):
🥇

src/components/tutorials-sidebar/index.tsx
zchsh: 👍 Sounds good, I think it makes sense to follow up on this.

This feels like something we should possibly raise with design - it's definitely possible this was an unintentional slip in the sidebar mockup, though I think it's also possible the design intent here doesn't always match the model of external link icons being strictly for separate domains, and may align to a fuzzier concept of using external link icons for separate experiences.

PR #285, opened 2022-04-08
feat(config): support extending from base config

BRKalow:

---

## Relevant links

- [Asana task](url) 🎟️

## What
- Refactoring our config code to support extending from a base config file, which I have added. Going forward, all configs can either specify an "extends" property, or extend from `config/base.json` by default.
- Moved all non-config related code out of `config/` and into either `build-libs/` or somewhere in `src/`

New base config:
```
{
  "io_sites": {
    "max_static_paths": 9999,
    "revalidate": 360
  },
  "dev_dot": {
    "canonical_base_url": "https://developer.hashi-mktg.com",
    "max_static_paths": 10,
    "revalidate": 360,
    "content_preview_branch": "dev-portal",
    "beta_product_slugs": ["waypoint", "vault"]
  },
  "learn": {
    "max_static_paths": 10
  },
  "flags": {
    "enable_new_downloads_view": true
  }
}
```

New preview config:
```
{
  "extends": "base",
  "io_sites": {
    "max_static_paths": 10
  }
}
```

Final config for preview environments:
```
{
  "io_sites": {
    "max_static_paths": 10,
    "revalidate": 360
  },
  "dev_dot": {
    "canonical_base_url": "https://developer.hashi-mktg.com/",
    "max_static_paths": 10,
    "revalidate": 360,
    "content_preview_branch": "dev-portal",
    "beta_product_slugs": ["waypoint", "vault"]
  },
  "learn": {
    "max_static_paths": 10
  },
  "flags": {
    "enable_new_downloads_view": true
  }
}
```

## Why

- Reduce repetition and unnecessary feedback cycles where people might forget to populate a value across all environment configs.
- This should make it clearer exactly what is changed from environment to environment when looking at the config files
- Clean up some code structure to separate concerns
- Prepare for usage of config values outside of application code (during build time, for example, to stop rendering of non beta products)


config/index.js
```diff
@@ -0,0 +1,57 @@
...
+  try {
+    const baseConfigPath = path.join(process.cwd(), 'config', `base.json`)
+    const baseConfig = JSON.parse(fs.readFileSync(baseConfigPath))
+    const envConfig = JSON.parse(fs.readFileSync(configPath))
+
+    // Load the extended config, if no extends property is defined use the base config
+    let extendsConfig = baseConfig
+    if ('extends' in envConfig && envConfig.extends !== 'base') {
+      // Determine which config this environment is extending from
+      const extendsConfigPath = path.join(
```
dstaley, line 29: Would love to hear if you specifically wanted to use the format `"extends": "filename"` in lieu of `"extends": "./filename.json"`.

config/index.js
```diff
@@ -0,0 +1,57 @@
...
+ * Load an environment config from a specific path.
+ */
+function getHashiConfig(configPath) {
+  try {
+    const baseConfigPath = path.join(process.cwd(), 'config', `base.json`)
+    const baseConfig = JSON.parse(fs.readFileSync(baseConfigPath))
+    const envConfig = JSON.parse(fs.readFileSync(configPath))
+
+    // Load the extended config, if no extends property is defined use the base config
+    let extendsConfig = baseConfig
```
dstaley, line 26: If we're always going to extend from `base` when `extends` isn't set, should we omit the `extends` key from our configs?

config/index.js
```diff
@@ -0,0 +1,57 @@
...
+    if ('extends' in envConfig && envConfig.extends !== 'base') {
+      // Determine which config this environment is extending from
+      const extendsConfigPath = path.join(
+        process.cwd(),
+        'config',
+        `${envConfig.extends}.json`
+      )
+      extendsConfig = getHashiConfig(extendsConfigPath)
```
dstaley, line 34: love that we can recursively handle `extends` 😍

config/index.js
```diff
@@ -0,0 +1,57 @@
...
+
+    const envFlattened = flat(envConfig, {
+      safe: true,
+    })
+
+    // Because we are "flattening" the object, a simple spread should be sufficient here
+    const finalConfig = { ...extendsFlattened, ...envFlattened }
+
+    if (process.env.DEBUG_CONFIG) {
+      console.log(finalConfig)
```
dstaley, line 47: Can you tag this debug log so we know where it's coming from, like we do [for redirect debugging](https://github.com/hashicorp/dev-portal/blob/main/config/redirects.js#L363)?

config/index.js
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    return finalConfig
+  } catch (err) {
+    console.log('Error loading environment config:', err)
```
dstaley, line 52: I'm curious in what case we'd want to continue building the site if we were unable to load the config.

config/plugin.js
```diff
@@ -0,0 +1,55 @@
...
+const { webpack } = require('next/dist/compiled/webpack/webpack')
+const { getHashiConfig } = require('./index')
+
+/**
+ * Reads in config files from config/[env].json and replaces references in the
+ * code with the literal values using webpack.DefinePlugin.
+ *
+ * Supports extending configs via the `extends` property.
+ *
+ * Example:
```
dstaley, line 11: Can you illustrate some deeply nested properties to show that we don't overwrite the entire object at that path?

dstaley (changes requested):
Love it! Just two small changes I'd like to see.

config/index.js
BRKalow: Good call! :nod:

config/index.js
BRKalow: Haha I had the same thought, I don't think we want to actually. I will adjust!

config/index.js
BRKalow: We could, I kinda like that it is a little more self-documenting with the explicit `extends` in each file though.

config/index.js
BRKalow: It was deliberate, as I'm not sure I would want to enable folks to think about referencing a file from elsewhere. This way it's scoped to the `/config` directory without having to add another check (though I guess someone could nefariously put `"./filename"` and the path would resolve).

Can you think of a scenario where we might want the way you've proposed? I do like that it's a bit more explicit.

config/index.js
dstaley: I'm totally down to keep the `extends` in the config file. What was the reasoning behind always extending `base` even when `extends` wasn't specified? Was it more of a safety net against an author creating a new config but forgetting to extend `base`?

config/plugin.js
BRKalow: Added to the test case and referenced that in this docs comment, good call!

config/index.js
BRKalow: @dstaley Yep, correct. The app will fail to build if all of the referenced config values aren't defined, so it seemed reasonable to default to extending base to guard against that. 

An alternative would be to validate that all keys that exist in base also exist in the final config file. I had started down this route, but it seemed potentially overkill in lieu of just the default extends.

config/index.js
dstaley: One use case I can think of is if we, in some far-flung future, wanted to reference a config that was distributed through npm. That being said, I did a little snooping and found that while files like `tsconfig.json` use file paths, others like `eslintrc` use both paths and simple names. My vote is that if we need that functionality in the future, we worry about adding it then. I was just curious as to your reasoning!

dstaley (approved):
✨

config/index.js
BRKalow: Yeah that makes total sense! In general I try to limit functionality and only expose what's really needed, thanks for bringing this up though 🙏

PR #289, opened 2022-04-11
feat(Hompeage): `<ProductNav />`

alexcarpenter:

---

## Relevant links

- [Preview link](https://dev-portal-git-achomepageproduct-nav-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1200641486648086/1202112386641526) 🎟️

## What

Implement `<ProductNav />` component on the homepage.

CleanShot 2022-04-13 at 15 14 20@2x

## Why

New homepage design.

### DIFF
--- a/src/views/homepage/components/product-nav/index.tsx
+++ b/src/views/homepage/components/product-nav/index.tsx
@@ -95,6 +95,7 @@ export default function ProductNav({ notice, products }: ProductNavProps) {
                       {productIcons[product.slug].color}
                       <Text
                         weight="medium"
+                        size={200}
                         className={s.productName}
                         asElement="span"
                       >
@@ -107,6 +108,7 @@ export default function ProductNav({ notice, products }: ProductNavProps) {
                     {productIcons[product.slug].neutral}
                     <Text
                       weight="medium"
+                      size={200}
                       className={s.productName}
                       asElement="span"
                     >
--- a/src/views/homepage/components/product-nav/index.tsx
+++ b/src/views/homepage/components/product-nav/index.tsx
@@ -1,5 +1,6 @@
 import { CSSProperties, ReactElement } from 'react'
 import Link from 'next/link'
+import classNames from 'classnames'
 import type { ProductSlug } from 'types/products'
 import { productSlugsToNames } from 'lib/products'
 import getIsBetaProduct from 'lib/get-is-beta-product'
@@ -89,10 +90,14 @@ export default function ProductNav({ notice, products }: ProductNavProps) {
       ) : null}
       <nav className={s.nav}>
         <ul className={s.list}>
-          {products.map((product) => {
+          {products.map((product, index) => {
             const isBetaProduct = getIsBetaProduct(product)
             const productName =
               product === 'hcp' ? 'HCP' : productSlugsToNames[product]
+            const productClassName = classNames(s.product, {
+              [s.isFirstChild]: index == 0,
+              [s.isLastChild]: index == products.length - 1,
+            })
             return (
               <li
                 className={s.listItem}
@@ -105,7 +110,7 @@ export default function ProductNav({ notice, products }: ProductNavProps) {
               >
                 {isBetaProduct ? (
                   <Link href={`/${product}/`}>
-                    <a className={s.product}>
+                    <a className={productClassName}>
                       {productIcons[product].color}
                       <Text
                         weight="semibold"
@@ -118,7 +123,7 @@ export default function ProductNav({ notice, products }: ProductNavProps) {
                     </a>
                   </Link>
                 ) : (
-                  <span className={s.product}>
+                  <span className={productClassName}>
                     {productIcons[product].neutral}
                     <Text
                       weight="semibold"
--- a/src/views/homepage/components/product-nav/index.tsx
+++ b/src/views/homepage/components/product-nav/index.tsx
@@ -3,6 +3,8 @@ import Link from 'next/link'
 import type { ProductSlug } from 'types/products'
 import { productSlugsToNames } from 'lib/products'
 import getIsBetaProduct from 'lib/get-is-beta-product'
+import { IconHashicorp24 } from '@hashicorp/flight-icons/svg-react/hashicorp-24'
+import { IconHashicorpColor24 } from '@hashicorp/flight-icons/svg-react/hashicorp-color-24'
 import { IconTerraform24 } from '@hashicorp/flight-icons/svg-react/terraform-24'
 import { IconTerraformColor24 } from '@hashicorp/flight-icons/svg-react/terraform-color-24'
 import { IconPacker24 } from '@hashicorp/flight-icons/svg-react/packer-24'
@@ -23,14 +25,18 @@ import Text from 'components/text'
 import s from './product-nav.module.css'
 
 const productIcons: {
-  [key in Exclude<ProductSlug, 'sentinel'>]: {
+  [key in ProductSlug]: {
     neutral: ReactElement
     color: ReactElement
   }
 } = {
   hcp: {
-    neutral: <IconTerraform24 />,
-    color: <IconTerraformColor24 />,
+    neutral: <IconHashicorp24 />,
+    color: <IconHashicorpColor24 />,
+  },
+  sentinel: {
+    neutral: <IconHashicorp24 />,
+    color: <IconHashicorpColor24 />,
   },
   terraform: {
     neutral: <IconTerraform24 />,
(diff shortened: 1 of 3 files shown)


src/lib/products.ts
```diff
@@ -20,10 +20,10 @@ const productSlugsToNames: { [slug in ProductSlug]: ProductName } = {
  * Holds the product slugs in the order ProductSwitcher will render them.
  */
 const productSwitcherSlugs: ProductSlug[][] = [
-  ['terraform', 'packer', 'vagrant'],
-  ['vault', 'boundary'],
+  ['terraform', 'packer'],
   ['consul'],
-  ['nomad', 'waypoint'],
+  ['vault', 'boundary'],
+  ['nomad', 'waypoint', 'vagrant'],
```
alexcarpenter, line 26: Update ordering to reflect the current groupings and order for rendering.

- Moves vagrant to be alongside nomad and waypoint
- Moves consul to be the second array within the grouping

image

src/views/homepage/components/product-nav/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+            {notice}
+          </Text>
+        </div>
+      ) : null}
+      <nav className={s.nav}>
+        <ul className={s.list}>
+          {products.map((product) => {
+            const isBetaProduct = getIsBetaProduct(product)
+            const productName =
+              product === 'hcp' ? 'HCP' : productSlugsToNames[product]
```
alexcarpenter, line 95: Needed because our `productSlugsToNames` map returns HashiCorp Cloud Platform.

src/views/homepage/components/product-nav/product-nav.module.css
```diff
@@ -0,0 +1,124 @@
...
+      box-shadow: var(--token-surface-high-box-shadow);
+
+      @media (--dev-dot-desktop) {
+        box-shadow: none;
+
+        &::after {
+          box-shadow: var(--token-surface-high-box-shadow);
+          border-bottom-left-radius: 4px;
+          border-bottom-right-radius: 4px;
+          height: calc(100% + 4px);
```
alexcarpenter, line 110: animating height here because scale doesn't allow px values, I know height is not super performant for animations, but it felt light enough that its ok.

alexcarpenter: Also using after element as to not scale the icon, text, or box-shadow on hover.

src/views/homepage/components/product-nav/product-nav.module.css
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    @nest .listItem:last-child & {
+      border-top-right-radius: 14px;
+      border-bottom-right-radius: 14px;
+
+      &::after, &:hover::after {
+        border-bottom-right-radius: 14px;
+      }
+    }
```
zchsh, line 92: Oooh I think this is the first time I've seen `@nest`! 🤩 

Nitpick: I find `@nest` stuff to be a bit confusing - and it feels like it might break our convention of [grouping styles by selector](https://www.notion.so/hashicorp/CSS-Modules-753f2647f7aa4a598e28990d9f0d05a8).

Definitely optional, but in the interest of keeping styles grouped by selector, would it make sense here to use `index` in the `products` map to generate & pass `isFirstItem` & `isLastItem` classNames rather than only `s.product`? This way we could avoid the parent-dependent `@nest` selector:


```jsx
/* products.map((product, index: number) => { */

const productClassName = classNames(s.product, {
  [s.isFirstChild]: index == 0,
  [s.isLastChild]: index == products.length - 1,
})

/* ... and pass `productClassName` instead of `s.product` ... */
```

```suggestion
    &.isFirstChild {
      border-top-left-radius: 14px;
      border-bottom-left-radius: 14px;

      &::after, &:hover::after {
        border-bottom-left-radius: 14px;
      }
    }

    &.isLastChild {
      border-top-right-radius: 14px;
      border-bottom-right-radius: 14px;

      &::after, &:hover::after {
        border-bottom-right-radius: 14px;
      }
    }
```

zchsh (approved):
✅ 💯 Looks great!

Had a couple minor nitpicks:
- `@nest` feels new & unfamiliar to me, and like it might break our convention of grouping styles by selector. Made an optional suggestion for how we could avoid it, if we end up wanting to 👍
    - I should note this might be more unfamiliarity bias than anything else 😅 so feel free to ignore the suggestion! If we do start using `@nest` though, I think we start a Slack thread or internal RFC or something on when to use it vs when not to use it (especially since I think it's a new introduction to the `dev-dot` project 👯 )
- Found an annoying, Safari-specific (I think?) visual bug. Might just need a `background` on `.list` to fix 🤞 Felt small enough to be non-blocking.

src/views/homepage/components/product-nav/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+import { IconNomad24 } from '@hashicorp/flight-icons/svg-react/nomad-24'
+import { IconNomadColor24 } from '@hashicorp/flight-icons/svg-react/nomad-color-24'
+import { IconWaypoint24 } from '@hashicorp/flight-icons/svg-react/waypoint-24'
+import { IconWaypointColor24 } from '@hashicorp/flight-icons/svg-react/waypoint-color-24'
+import { IconVagrant24 } from '@hashicorp/flight-icons/svg-react/vagrant-24'
+import { IconVagrantColor24 } from '@hashicorp/flight-icons/svg-react/vagrant-color-24'
+import Text from 'components/text'
+import s from './product-nav.module.css'
+
+const productIcons: {
```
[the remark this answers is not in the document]
alexcarpenter: Thought about it, but I am assuming the neutral usage will not be common, and to keep bundle sizes from growing It seemed to make sense to keep it separate. That is an assumption though.

src/views/homepage/components/product-nav/product-nav.module.css
```diff
@@ -0,0 +1,132 @@
...
+
+.list {
+  list-style: none;
+  margin: 0 auto;
+  padding: 0;
+  max-width: var(--container-max-width);
+  display: grid;
+  grid-template-columns: repeat(2, 1fr);
+  gap: 24px;
+  border-radius: 14px;
```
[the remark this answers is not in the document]
alexcarpenter: Got it, not seeing that on my end but I went ahead and added it. I put it within the desktop media query since we don't want backgrounds once within the grid.

src/views/homepage/components/product-nav/index.tsx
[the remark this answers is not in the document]
zchsh: makes sense - +1 that the neutral use case won't be too common, that's a great point 💯

src/views/homepage/components/product-nav/product-nav.module.css
alexcarpenter: Got it, I am on board with this change.

I tend to prefer to use nest as to reduce the code needed within the component, since this can be done in CSS without the extra code. This is a bit trivial since its just two classes but on a component when more children need to change based on a state or position of index I'd prefer not to have to add conditional classes throughout the component when CSS can handle it. I see these two variations one of the same with regards to grouping, these two selectors are at the same level, we're offloading more work to the component code.

zchsh: ☝️ Great point about reducing the code & # of tacked-on classNames within the component - I can definitely see the rationale for that.

Also 🤦 on mentioning grouping, I had a more convoluted idea in mind (see `Random related thoughts`) but didn't go through with it - very much not applicable in this case

I should also be clear, the avoidance so far of `@nest` in `dev-dot` is I think more related to us not being aware of it that anything else -  we definitely haven't discussed it as a team! It's quite possibly, likely even, that we have a ton of great use cases for `@nest`, and it seems like this could be one of them. Just nervous about introducing it into the project without that context of how & when we want to use it - but super open to it 🙇‍♂️

<details>
<summary>Random related thoughts, why I mentioned grouping</summary>

Something that's been bouncing around my mind is balancing strictness with regards to reducing `className` & JSX complexity vs strictness in grouping CSS by element (& maybe even pseudo-element?). As you mentioned the grouping even with the extra classNames is still the same - what sticks out to me about this is that we have to jump into a few different places within the CSS to figure out what's going on with the `::after` pseudo element across viewports (at left in screenshot below).

With this example, and the `.isFirstChild` & `.isLastChild` classNames in place, I think we could move things around a bit for more strict grouping... but also more lines of CSS, and the `border-radius: 14px` aren't as closely grouped...



I'm very on the fence about this though - can definitely see how `@nest` is preferable is a lotta ways 👍 But wanted to get this thought out there in case it sparks anything! 

</details>

PR #294, opened 2022-04-12
Update SidebarBackToLink

ashleemboyer:

---

## Relevant links

- [Preview link](https://dev-portal-git-ambupdate-back-to-link-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202022787106807/1202105084146694/f) 🎟️
- [Designs](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/?node-id=7502%3A69101) 🎨

## What

- Updates `SidebarBackToLink` to
  - render a primary style `StandaloneLink`
  - no longer render a dev popover
  - require all of its props (since it no longer renders dev popover)
  - have the latest `padding-left` and `margin-bottom` specified by designs
- Updates `Sidebar` to conditionally render `SidebarBackToLink` based on `backToLinkProps`.
- Updates `Sidebar` to have `32px` on the top and `24px` otherwise, to match the latest values in designs.
- Replaces `StandaloneLink`'s `textSize` prop with `size` to more accurately apply `font-size`, `line-height`, and `font-weight` styles instead of inheriting from `Text`.
- Removes references to `StandaloneLink`'s old `textSize` prop
- Updates spacing between text and icon in `StandaloneLink` to match [the HDS implementation](https://design-system-components-hashicorp.vercel.app/components/link/standalone)
- Updates `TutorialsSidebar` to use `SidebarBackToLink`

## Why

These changes are made to update the existing sidebar with design updates that have happened since the learn content pages have been designed.

## Testing

- [ ] The back to link in sidebar should look correct on:
  - [ ] [Waypoint landing page](https://dev-portal-git-ambupdate-back-to-link-hashicorp.vercel.app/waypoint)
  - [ ] [Waypoint /docs page](https://dev-portal-git-ambupdate-back-to-link-hashicorp.vercel.app/waypoint/docs)
  - [ ] [Waypoint /commands page](https://dev-portal-git-ambupdate-back-to-link-hashicorp.vercel.app/waypoint/commands)
  - [ ] [Waypoint /plugins page](https://dev-portal-git-ambupdate-back-to-link-hashicorp.vercel.app/waypoint/plugins)
  - [ ] [Waypoint /downloads page](https://dev-portal-git-ambupdate-back-to-link-hashicorp.vercel.app/waypoint/downloads)
  - [ ] [Vault landing page](https://dev-portal-git-ambupdate-back-to-link-hashicorp.vercel.app/vault)
  - [ ] [Vault /docs page](https://dev-portal-git-ambupdate-back-to-link-hashicorp.vercel.app/vault/docs)
  - [ ] [Vault /api-docs page](https://dev-portal-git-ambupdate-back-to-link-hashicorp.vercel.app/vault/api-docs)
  - [ ] [Vault /downloads page](https://dev-portal-git-ambupdate-back-to-link-hashicorp.vercel.app/vault/downloads)
- [ ] On the product landing views, the "Get started" standalone link should look correct
- [ ] On the install views, the following `StandaloneLink`s should look correct:
  - [ ] The "Download" links that appear under the "Binary download for macOS" section
  - [ ] The "GitHub" link that appears under "Release information"
  - [ ] The "Learn more" link that appears in the sidecar marketing card
- [ ] The [`TutorialsSidebar` Swingset page](https://dev-portal-git-ambupdate-back-to-link-hashicorp.vercel.app/swingset/components/tutorialssidebar) should still behave as expected

## Anything else?

Nope!


src/components/sidebar/index.tsx
```diff
@@ -102,12 +102,15 @@ const Sidebar = ({
   const [filterValue, setFilterValue] = useState('')
   const filteredMenuItems = getFilteredMenuItems(itemsWithMetadata, filterValue)
 
+  let backToLink
+  if (backToLinkProps) {
+    const { text, url } = backToLinkProps
+    backToLink = <SidebarBackToLink text={text} url={url} />
+  }
```
ashleemboyer, line 109: Using this approach instead of `&&` or a ternary with `null` in the false-y branch.

src/components/sidebar/components/sidebar-back-to-link/index.tsx
```diff
@@ -1,17 +1,9 @@
...
 import s from './sidebar-back-to-link.module.css'
 
-const BackToLinkText: React.FC = ({ children }) => (
-  <Text asElement="span" size={200} weight="medium">
-    {children}
-  </Text>
-)
-
 /**
  * TODO: Design is still planning this functionality & states. Since not all
```
zchsh, line 8: Nit, non-blocking: I think we can remove this comment now that you've covered all the cases ☺️

src/components/sidebar/index.tsx
zchsh: Neat! 🤩 

Non-blocking & very optional question: I'm curious to get your take on when we should reach for this pattern. My understanding of the current approaches:

- the `&&` approach is for when we're conditionally showing or not showing a component, and the condition being checked is known not to be a falsy value such as `0` or `""`
- the `{condition ? X : Y}` approach is for situations where we're either switching between X and Y components, or for show / hide situations where `condition` might be a falsy value such as `0` or `""` 

It seems like the `let` & `if` block approach might be useful for situations where there's some more complex logic or destructing we want to do when we do want to render the component. Does that seem accurate? 🤔

src/components/sidebar/components/sidebar-back-to-link/index.tsx
```diff
@@ -21,42 +13,14 @@ const SidebarBackToLink = ({
...
-        </>
-      }
-    >
-      <div className={s.backToLink}>
-        <IconArrowLeft16 className={s.icon} />
-        <BackToLinkText>Back to [Page]</BackToLinkText>
-      </div>
-    </DevPopover>
+    <StandaloneLink
+      className={s.root}
```
zchsh, line 18: Blocking: I think we might need a wrapper element with `s.root` applied - otherwise, I think this is causing `margin` & `padding` issues on `:focus`:

CleanShot 2022-04-13 at 09 28 09

src/components/tutorials-sidebar/tutorials-sidebar.module.css
```diff
@@ -1,16 +1,3 @@
-.root {
-  display: flex;
-  flex-direction: column;
-  gap: 20px;
```
zchsh, line 4: 💖  I dig that adding `margin-bottom` on `BackToLink` removed the need for this layout stuff 🙌 Nicely done!

zchsh (changes requested):
💯 Looks great! 

Found one thing which I think will be a quick fix: there seems to be a `margin` & `padding` issue on `:focus` of `SidebarBackToLink`:

CleanShot 2022-04-13 at 09 28 09

I think `StandloneLink` ends up with `:focus`-specific styles that take precedence over `SidebarBackToLink`'s `margin-bottom` and `padding-left`. I think we might need a wrapping element of some kind to avoid that 🤔

src/components/sidebar/components/sidebar-back-to-link/index.tsx
ashleemboyer: I need to make some spacing adjustments after some convos yesterday anyways, I'm going to see if a different approach helps resolve this. 🤔  Great catch!

src/components/sidebar/components/sidebar-back-to-link/index.tsx
ashleemboyer: Oh, right! Yes I'll remove. 😊

src/components/sidebar/index.tsx
ashleemboyer: Hopefully not misinterpreting @kendallstrautman's take on this from a PR I had a couple of days ago, but as I understand it we might just want to avoid `&&` altogether? She [linked a post](https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx#conclusion) that said even if you're sure the condition is a boolean, that using `&&` is abusing the operator.

As for ternaries, I don't think we should have multiline ternaries inside of JSX. If it's one line, that's not as difficult to read. But ternaries that span over several lines are hard to read and understand.

With both of these things in mind, that leaves using `let` and `if` before the `return` with markup. It's a little extra work, but it really does keep the markup clean and easy to read!

ashleemboyer:
@zchsh resolved both of the things you noted in https://github.com/hashicorp/dev-portal/pull/294/commits/27497e504a88b77463cebf9033ac9b94f0d5b200! This is ready for re-review. 😊

zchsh (approved):

PR #297, opened 2022-04-12
feat(homepage): `<MerchandisingSlots />`

alexcarpenter:

---

## Relevant links

- [Preview link](https://dev-portal-git-achomepagemerchandising-slots-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202111817660946/1202112386641557) 🎟️

## Why

Implements `<MerchandisingSlots />` components we bespoke slot implementations.

CleanShot 2022-04-14 at 16 12 07@2x

## How

I went with nested slot components and there is a bit of duplication between each of them, but until we see how these will be used evolved in the future I didn't want to prematurely abstract a slot component.


src/views/homepage/components/merchandising-slots/slots/hashiconf-global-slot/hashiconf-global-slot.module.css
```diff
@@ -0,0 +1,48 @@
+.root {
+  --padding: 24px;
+  --min-height: 200px;
+
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  padding: var(--padding);
+  min-height: var(--min-height);
+  border-radius: 6px;
```
zchsh, line 10: Nit, non-blocking: if we're being strict, I think we should leverage `CardLink` here (& in other slots) for the border-radius and shadow properties. This would also give us the right `:hover` property ☺️ But... will require that `height: 100%` workaround, I think 🤔

zchsh (approved):
✅ This looks good to me!

Noticed the lack of `:hover` effect on the cards. IMO this isn't blocking, fine to follow up on whenever since this PR seems perfecto otherwise, but wanted to flag - I think the "correct" solution here would be to leverage `CardLink` in each of the `slot` components 👍

src/views/homepage/components/merchandising-slots/slots/hashiconf-global-slot/hashiconf-global-slot.module.css
alexcarpenter: I will give it a shot, initially didn't go that route because card includes padding. I will actually test it though!

zchsh: ahhhh interesting... dang it, another hidden cost of diverging from the design system 😬 i guess we'd have to override our default card padding to match the designs.... so yeah no stress here 👍 would be nice if we could have full consistency (IMO), but I think ejecting from `CardLink` rather than overriding it are kind of equivalently "meh" given the design needs here. I like the call you made 💯

alexcarpenter: I do have a hover state in place for the cards, its just hard to see since its on dark cards.

zchsh: 🤦 I see it now, oops 😬 do we need a `transition` there?

alexcarpenter: Added transition, still pretty faint. We could bump up to `box-shadow: var(--token-surface-higher-box-shadow);` to  make it a touch more noticeable. We ran into a similar thing on dotcom homepage. The darker merchandising slots have a darker box-shadow to make the hovers look consistent.

alexcarpenter: Updated to use CardLink.

zchsh: very much agreed on how subtle it is... i wonder if we need like a `theme="dark"` for `CardLink` or something (seems like this is something that comes up a lot, so worth encoding there). Made a subtask [on the polish task you set up](https://app.asana.com/0/1202111817660946/1202127264881981/f) (thanks for setting that up btw! 💖 )

PR #302, opened 2022-04-12
Update Sidebar title, section headings, and horizontal rules

ashleemboyer:

---

## Relevant links

- [Preview link](https://dev-portal-git-ambupdate-sidebar-labels-hashicorp.vercel.app/) 🔎
- Asana tasks 🎟️
  - [Update Sidebar title and section headings to have the same styles](https://app.asana.com/0/1202022787106807/1202105084146705/f)
  - [Update Sidebar horizontal rule styles](https://app.asana.com/0/1202022787106807/1202105084146707/f)

## What

- Adds a few components to Sidebar's `components` directory:
  - `SidebarHorizontalRule`
  - `SidebarSectionHeading`
  - `SidebarTitleHeading`
- Updates `SidebarNav` to use `SidebarTitleHeading`
- Updates `SidebarNavMenuItem` to use `SidebarHorizontalRule` and `SidebarSectionHeading`
- Places styles shared by `SidebarTitleHeading` and `SidebarSectionHeading` in `src/components/sidebar/sidebar.module.css`
- Uses `SidebarHorizontalRule`, `SidebarSectionHeading`, and `SidebarTitleHeading` in `TutorialsSidebar`
- Removes styles now handled by the new subcomponents of `Sidebar`

## Why

We need our Sidebar elements to look and behave the same, so exporting components from `Sidebar` that handles consistent style and function is a step towards ensuring all Sidebars look and feel the same.

## How

For each new sub component, I worked on adding it to the main `Sidebar` component first. Once that was solid, I made use of the component in `TutorialsSidebar` and then made any necessary CSS or JSX changes so the component would look and feel the same in both sidebars.

## Testing

Make sure the sidebars on the following pages are styled as expected with respect to the sidebar title, section headings, and horizontal rules:

- [ ] https://dev-portal-git-ambupdate-sidebar-labels-hashicorp.vercel.app/swingset/components/tutorialssidebar
- [ ] https://dev-portal-git-ambupdate-sidebar-labels-hashicorp.vercel.app/vault
- [ ] https://dev-portal-git-ambupdate-sidebar-labels-hashicorp.vercel.app/vault/api-docs
- [ ] https://dev-portal-git-ambupdate-sidebar-labels-hashicorp.vercel.app/vault/docs
- [ ] https://dev-portal-git-ambupdate-sidebar-labels-hashicorp.vercel.app/vault/downloads
- [ ] https://dev-portal-git-ambupdate-sidebar-labels-hashicorp.vercel.app/waypoint
- [ ] https://dev-portal-git-ambupdate-sidebar-labels-hashicorp.vercel.app/waypoint/commands
- [ ] https://dev-portal-git-ambupdate-sidebar-labels-hashicorp.vercel.app/waypoint/docs
- [ ] https://dev-portal-git-ambupdate-sidebar-labels-hashicorp.vercel.app/waypoint/plugins
- [ ] https://dev-portal-git-ambupdate-sidebar-labels-hashicorp.vercel.app/waypoint/downloads

## Anything else?

This PR only handles aligning the styles of the sidebar title, section headings, and horizontal rules. There will be more work done in follow up PRs for updating the various types of menu items.


src/components/sidebar/components/index.ts
```diff
@@ -0,0 +1,5 @@
+import SidebarHorizontalRule from './sidebar-horizontal-rule'
+import SidebarSectionHeading from './sidebar-section-heading'
+import SidebarTitleHeading from './sidebar-title-heading'
+
+export { SidebarHorizontalRule, SidebarSectionHeading, SidebarTitleHeading }
```
kendallstrautman, line 5: Love me a barrel export file ❤️

src/components/sidebar/components/sidebar-horizontal-rule/index.tsx
```diff
@@ -0,0 +1,7 @@
+import s from './sidebar-horizontal-rule.module.css'
+
+const SidebarHorizontalRule = () => {
```
kendallstrautman, line 3: Just thinking outloud, but I wonder if it's needed to have the explicit `SidebarComponentName` pattern with these. It may read easier in the sidebar component and working with it to have it be like: 

```
<Sidebar >
  <SectionHeading>Yo</SectionHeading>
  <HorizontalRule />
</Sidebar>
```

Given that these components will always be used within the context of the sidebar, as children or a composed component it seems. I could see it either way though! It's up to you.

src/components/sidebar/components/sidebar-title-heading/index.tsx
```diff
@@ -0,0 +1,16 @@
+import sSidebar from 'components/sidebar/sidebar.module.css'
+
+interface SidebarTitleHeadingProps {
```
kendallstrautman, line 3: I'm probably behind on the convo on this, but in my previous interaction with the `MenuItem` interface, the title and heading were different. Is this just the compounding of those two as they share styles? Or whats the story with the `TitleHeading` combo?

kendallstrautman (approved):
LGTM! I'll defer to @zchsh for the visual QA though since he's recently done sidebar work with tutorials.

src/components/sidebar/components/index.ts
ashleemboyer: Me too! ✨  Wish we could use one for our main `src/components` but I think it had negative impacts on the build time. 😢

src/components/sidebar/components/sidebar-horizontal-rule/index.tsx
ashleemboyer: So the reason I prefixed these with `Sidebar` has to do with codebase-wide searching. `SectionHeading` and `HorizontalRule` are both generic enough that we could eventually create those components again later for other parts of the app. That would leave us with two (or more, potentially) versions of components with the same name, which makes searching for usage of a component harder.

src/components/sidebar/components/sidebar-title-heading/index.tsx
ashleemboyer: I put `Heading` in the names of `SidebarTitleHeading` and `SidebarSectionHeading` in hopes of implying that a heading HTML element is what gets rendered by these components. I think it also implies a relationship between the two since their names are similar, which I think is important because they're directly related to one another in both style (they currently have the same styles despite rendering different heading element levels) and hierarchy (`SidebarTitleHeading` is like a parent heading to `SidebarSectionHeading`) in the markup. I could add some comments in each file that help clarify their usage, would that be valuable?

src/components/sidebar/components/index.ts
zchsh: +1 this is cool and also kinda new to me and I like it 💖

src/components/sidebar/components/sidebar-horizontal-rule/index.tsx
zchsh: Will +1 to both 😝 I can see this either way, and personally lean towards the `Sidebar` prefix for extra clarity, and for code-search & future-maybe-duplicate-name reasons.

src/components/sidebar/components/sidebar-title-heading/index.tsx
zchsh: This seems like a hard naming problem 😅  I like the idea of capturing both the HTML `heading` element part of things, and the hierarchy here... 

... it seems like the design-side name of `Title` is maybe making this more complex? I think in the absence of that, we might reach for something like `SidebarHeading` & `SidebarSectionHeading`?

In any case, I'm on board with any of the names suggested here 👍

src/components/sidebar/components/sidebar-title-heading/index.tsx
```diff
@@ -0,0 +1,16 @@
+import sSidebar from 'components/sidebar/sidebar.module.css'
```
zchsh, line 1: Nit, non-blocking: what if we co-located `SidebarTitleHeading` & `SidebarSectionHeading` in the same file? Potential benefits:

- hierarchical relationship between the two might be clearer
- we'd have `sidebar-title-heading.module.css` which could have one className rather than separate but identical `.title` & `.heading`; this might also make it more clear that the two components share identical styles

src/components/sidebar/sidebar.module.css
```diff
@@ -6,3 +6,15 @@
     margin-left: 1em;
   }
 }
+
+.title,
+.heading {
+  color: var(--token-color-foreground-primary);
+  font-size: var(--token-typography-display-100-font-size);
+  font-weight: var(--token-typography-font-weight-semibold);
+  line-height: var(--token-typography-display-100-line-height);
```
zchsh, line 15: Non-blocking: This is _super_ nitpicky, but in the Figma file it looks like `body-100` is being used.

I _cannot_ tell at a glance if there's a difference between `body-100` & `display-100` (there doesn't seem to be), and also `display-100` seems more "correct" in way here since these are heading elements, but wanted to flag just in case ☺️

zchsh (approved):
✅ Looks great!

Had only nit-picky hot takes on naming & use of `display-100` vs `body-100`. Definitely non-blocking 👍

src/components/sidebar/components/sidebar-title-heading/index.tsx
ashleemboyer: Maybe this is a case where we need some terminology alignment. Where is `Title` being used on the design-side? Want to make sure I'm understanding where the confusion is stemming from. 😊

src/components/sidebar/components/sidebar-title-heading/index.tsx
ashleemboyer: I really like this idea! Could also lower the risk for misusing one for the other. I'll work on this update. 😊

ashleemboyer:
@kendallstrautman @zchsh have made a couple of updates that need re-review! 😊

- Moved `SidebarTitleHeading` and `SidebarSectionHeading` to their own file, where they also have a shared classname and stylesheet
- Fixed `StandaloneLink` spacing for when there is an icon, I did that in my `SidebarBackToLink` PR but there was a merge conflict that I didn't resolve correctly that undid my change
- Replaced the arrow-left icon in `SidebarBackToLink` with chevron-left, I should have done this before but I didn't notice the change until now 😁

src/components/sidebar/components/sidebar-horizontal-rule/index.tsx
kendallstrautman: The search-ability aspect and potential name duplicates makes sense to me!

src/components/sidebar/sidebar.module.css
zchsh: ☝️ ahah i had the thought about `compose` as well... tbh sometimes the individual tokens feel more readable to me? Like, `compose` feels very opaque...

I wish we had like "go to definition" like we do in TS, but for `compose`'d classNames 🤔

ashleemboyer: That would be great! I wonder if it'd be easy to make a little vscode extension for it.

src/components/sidebar/components/sidebar-title-heading/index.tsx
zchsh: +1 on getting alignment on this. Here's where I've seen it in design specs:

CleanShot 2022-04-13 at 15 07 27@2x

What I find confusing is how the distinction between `title` and `heading` seems like it varies in dev contexts (`title` being the named HTML attribute, `heading` being HTML element) vs design contexts (I'm honestly not sure on exact definitions, I feel like `title` and `heading` are kind of interchangeable, but folks lean towards `title` sometimes to avoid confusion with `header` & `heading` sounding so similar).

zchsh (approved):
✅ Looks great to me!

Great catch on `arrow` vs `chevron` 💯

src/components/sidebar/components/sidebar-title-heading/index.tsx
ashleemboyer: Ahh I see now! With title/heading _probably_ being interchangeable on the design side, and less so on the development side, I'd vote for the following names:

- Sidebar heading (code name: `SidebarHeading`)
- Section heading (code name: `SidebarSectionHeading`)

These code names are exactly what you proposed! 😊 What are your thoughts on waiting to rename until we've confirmed terminology alignment with design? I can go ahead and start that convo and then update these names in a follow-up PR.

zchsh: ☝️ Sounds great! Very on board with keeping the names you have, & following up later once we have design alignment 💯

kendallstrautman (approved):

src/components/sidebar/components/sidebar-title-heading/index.tsx
ashleemboyer: Created an Asana ticket for discussing this with design! https://app.asana.com/0/1202110981600689/1202132404550212/f

PR #307, opened 2022-04-13
Remove unused ProductSwitcher

ashleemboyer:
## "Ready for Review" checklist

- [x] The Vercel preview link has been added to this PR's description
- [x] The description template has been filled in below

---

## Relevant links

- [Preview link](https://dev-portal-git-ambremove-unused-product-switcher-hashicorp.vercel.app/) 🔎

## What

- Removes unused `ProductSwitcher`
- Replaces `productSwitcherSlugs` with a flat array of all product slugs
- Updates the `products` array to be generated from the new `productSlugs`

## Why

`ProductSwitcher` is no longer needed since the latest primary nav doesn't use it. This left `productSwitcherSlugs` only being used indirectly by `PreviewProductSwitcher`, which was flattening the array anyways, so that informed the decision to use a flat array instead.

## Testing

- [ ] Make sure the `PreviewProductSwitcher` still works

## Anything else?

Nope!

PR #310, opened 2022-04-13
feat(homepage): `<TutorialCard />`

alexcarpenter:

---

## Relevant links

- [Preview link](https://dev-portal-git-achomepagetutorial-card-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/inbox/1200641486636491/1202118917268192/1202122065211612) 🎟️

## What

Adding rough implementation of `<TutorialCard />`

CleanShot 2022-04-13 at 12 29 50@2x

### Example usages

```tsx
<TutorialCard
  link="/"
  duration="10min"
  heading="Title Max 70 Characters Try and Keep it to a Max of Three Lines Loremi"
  description="Body maximum 130 characters. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Auctor vitae pharetra accumsan risu, eu v..."
  badges={["vault", "boundary", "interactive", "video"]}
/>;
```

## Why

Implementation a rough version for the homepage, knowing there is more needed of this component once the product version gets implemented.

### DIFF
--- a/src/views/homepage/components/tutorial-card/index.tsx
+++ b/src/views/homepage/components/tutorial-card/index.tsx
@@ -1,11 +1,11 @@
 import type { ProductSlug } from 'types/products'
 import { productSlugs } from 'lib/products'
-import ProductIcon from 'components/product-icon'
 import { IconPlay16 } from '@hashicorp/flight-icons/svg-react/play-16'
 import { IconTerminalScreen16 } from '@hashicorp/flight-icons/svg-react/terminal-screen-16'
 import Badge from 'components/badge'
 import CardLink from 'components/card-link'
 import Heading from 'components/heading'
+import ProductIcon from 'components/product-icon'
 import Text from 'components/text'
 import s from './tutorial-card.module.css'
 
--- a/src/views/homepage/components/tutorial-card/index.tsx
+++ b/src/views/homepage/components/tutorial-card/index.tsx
@@ -1,0 +1,74 @@
+import Link from 'next/link'
+import { VisuallyHidden } from '@reach/visually-hidden'
+import type { ProductSlug } from 'types/products'
+import { productSlugs } from 'lib/products'
+import ProductIcon from 'components/product-icon'
+import { IconPlay16 } from '@hashicorp/flight-icons/svg-react/play-16'
+import { IconTerminalScreen16 } from '@hashicorp/flight-icons/svg-react/terminal-screen-16'
+import Heading from 'components/heading'
+import Text from 'components/text'
+import s from './tutorial-card.module.css'
+
+const getProductIcons = () => {
+  const icons = {}
+  for (const slug of productSlugs) {
+    icons[slug] = <ProductIcon productSlug={slug} />
+  }
+  return icons
+}
+
+const BADGE_ICON_MAP = {
+  ...getProductIcons(),
+  video: <IconPlay16 />,
+  interactive: <IconTerminalScreen16 />,
+}
+
+interface TutorialCardProps {
+  link: string
+  duration?: string
+  heading: string
+  description: string
+  badges: Array<ProductSlug | 'video' | 'interactive'>
+}
+
+export default function TutorialCard({
+  link,
+  duration,
+  heading,
+  description,
+  badges,
+}: TutorialCardProps) {
+  return (
+    <Link href={link}>
+      <a className={s.tutorialCard}>
+        {duration ? (
+          <Text className={s.duration} size={100} weight="medium">
+            {duration}
+          </Text>
+        ) : null}
+        <Heading
+          className={s.heading}
+          level={3}
+          size={200}
+          weight="semibold"
+          slug={heading}
+        >
+          {heading}
+        </Heading>
+        <Text className={s.description} size={100}>
+          {description}
+        </Text>
+        <ul className={s.badgesList}>
+          {badges.map((badge, index) => {
+            return (
+              <li key={index} className={s.badgesListItem}>
+                <VisuallyHidden>{badge}</VisuallyHidden>
+                <span className={s.badge}>{BADGE_ICON_MAP[badge]}</span>
+              </li>
+            )
+          })}
+        </ul>
+      </a>
+    </Link>
+  )
+}
--- a/src/views/homepage/components/tutorial-card/index.tsx
+++ b/src/views/homepage/components/tutorial-card/index.tsx
@@ -41,7 +41,7 @@ export default function TutorialCard({
   badges,
 }: TutorialCardProps) {
   return (
-    <CardLink href="/" className={s.tutorialCard}>
+    <CardLink href={link} className={s.tutorialCard}>
       {duration ? (
         <Text className={s.duration} size={100} weight="medium">
           {duration}
--- a/src/views/homepage/components/tutorial-card/index.tsx
+++ b/src/views/homepage/components/tutorial-card/index.tsx
@@ -31,8 +31,6 @@ interface TutorialCardProps {
   badges: Array<Exclude<ProductSlug, 'sentinel'> | 'video' | 'interactive'>
 }
 
-console.log(productSlugs)
-
 export default function TutorialCard({
   link,
   duration,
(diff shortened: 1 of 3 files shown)


src/views/homepage/components/tutorial-card/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+        </Heading>
+        <Text className={s.description} size={100}>
+          {description}
+        </Text>
+        <ul className={s.badgesList}>
+          {badges.map((badge, index) => {
+            return (
+              <li key={index} className={s.badgesListItem}>
+                <VisuallyHidden>{badge}</VisuallyHidden>
+                <span className={s.badge}>{BADGE_ICON_MAP[badge]}</span>
```
ashleemboyer, line 66: Can you use [the Badge component](https://developer.hashi-mktg.com/swingset/components/badge) here?

src/lib/products.ts
```diff
@@ -44,4 +44,6 @@ const products: ProductGroup[] = productSwitcherSlugs.map(
   }
 )
 
-export { products, productSlugsToNames }
+const productSlugs = Object.keys(productSlugsToNames) as ProductSlug[]
+
+export { products, productSlugs, productSlugsToNames }
```
alexcarpenter, line 49: Update will be coming from https://github.com/hashicorp/dev-portal/pull/307

src/views/homepage/components/tutorial-card/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+export default function TutorialCard({
+  link,
+  duration,
+  heading,
+  description,
+  badges,
+}: TutorialCardProps) {
+  return (
+    <Link href={link}>
```
ashleemboyer, line 42: Can you use [the CardLink component](https://developer.hashi-mktg.com/swingset/components/cardlink) here?

src/lib/products.ts
ashleemboyer: Just merged! 🥳

src/views/homepage/components/tutorial-card/index.tsx
alexcarpenter: Is there a way to get the link to have a class? Looking to have equal height cards, but I don't have access to the wrapping link. 

CleanShot 2022-04-13 at 12 50 09@2x

src/views/homepage/components/tutorial-card/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+        ) : null}
+        <Heading
+          className={s.heading}
+          level={3}
+          size={200}
+          weight="semibold"
+          slug={heading}
+        >
+          {heading}
+        </Heading>
```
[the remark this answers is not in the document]
alexcarpenter: For this implementation yes.

CleanShot 2022-04-13 at 12 52 21@2x

src/views/homepage/components/tutorial-card/index.tsx
alexcarpenter: Updated

src/views/homepage/components/tutorial-card/index.tsx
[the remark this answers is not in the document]
ashleemboyer: Sweet! Thanks for the quick screenshot. Just wanted to double check on the heading element hierarchy. 😊

src/views/homepage/components/tutorial-card/index.tsx
ashleemboyer: We've done this using the `className` prop on `CardLink` before, `FeaturedTutorialsSection` is one place. They're located at the bottom of the downloads pages.

This is the class that gets applied to those cards, can you give something like this a try?

```css
.card {
  display: flex;
  flex-direction: column;
  justify-content: space-between;
  height: 100%;
}
```

src/views/homepage/components/tutorial-card/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+        size={200}
+        weight="semibold"
+        slug={heading}
+      >
+        {heading}
+      </Heading>
+      <Text className={s.description} size={100}>
+        {description}
+      </Text>
+      <ul className={s.badgesList}>
```
ashleemboyer, line 62: This is a note for the future when we expand this component some more, I think we'll need an `aria-label` or something on this list that describes what it's a list of. I'm not 100% sure yet what it would be, so no need to do anything in this PR for this!

src/views/homepage/components/tutorial-card/index.tsx
alexcarpenter: That works! Can this be a default as to not require consumers to add this? Feels like it'd be a common pattern to have a grouping of these cards that need to be of equal height?

ashleemboyer: We don't be able to add it by default because there are several existing use cases where this wouldn't work, but I think we could either add a prop to `Card` that applies this class or we could create an additional `Card` component that's a variant of this. I'm leaning towards the prop idea since CSS is the only difference, maybe call it `isGridCard`? What do you think? Btw you aren't the first one to notice this pain, so totally valid issue to raise. 😊

src/views/homepage/components/tutorial-card/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+  video: <IconPlay16 />,
+  interactive: <IconTerminalScreen16 />,
+}
+
+interface TutorialCardProps {
+  link: string
+  duration?: string
+  heading: string
+  description: string
+  badges: Array<Exclude<ProductSlug, 'sentinel'> | 'video' | 'interactive'>
```
alexcarpenter, line 31: Excluding sentinel here as we don't have a product icon for it.

src/views/homepage/components/tutorial-card/index.tsx
alexcarpenter: Do you have an example of where this would be problematic to be the default? Is it the 100% height declaration that makes it grow to large?

From my isolated testing adding a classname to the link with flex and adding flex to the root of card link grows the card as needed. Not extra CSS needed at the consumer level.

CleanShot 2022-04-13 at 13 34 16@2x

ashleemboyer: Here are some places where `Card` is used, you can go ahead and try adding some default stuff now and testing it if you have time!

- The `InteractiveLabCallout` in the "Launch the lab environment" towards the bottom of this page: https://developer.hashi-mktg.com/vault/tutorials/app-integration/agent-read-secrets
- The Prev/Next cards at the bottom of learn content (can use same link as previous)
- The "This tutorial also appears in" section at the bottom of some pages: https://developer.hashi-mktg.com/vault/tutorials/app-integration/agent-kubernetes#help-and-reference
- The cards used on /vault/docs and /consul/docs (which can be tested by running `npm run start:vault` and `npm run start:consul`)
- Several cards located in the install pages, and cards within cards! https://developer.hashi-mktg.com/vault/downloads
- Product landing pages: https://developer.hashi-mktg.com/vault

zchsh: We ran into a similar issue in #178  - here's [a comment with some context](https://github.com/hashicorp/dev-portal/pull/178#discussion_r828156913).

At the time I'd proposed getting `Card` & `CardLink` to share the same CSS class name, rather than `CardLink` using the `Card` component. I think this could achieve the same styling consistency goal, without the extra wrapping element.

Maybe that's something we could revisit? I feel like this issue of having to add `height: 100%` is going to come up in a _ton_ of `Card` layouts 🤔

zchsh (approved):
✅ This looks great to me!

Definitely noticed no `max-width` & inline styles on the `section`, but I'm assuming the intent is to demo the cards and handle the page-level layout etc later 👍

src/views/homepage/components/tutorial-card/index.tsx
ashleemboyer: I think `CardLink` should continue to use `Card`. I think it'd be great to find a CSS solution for this and I like Alex's proposal, I just want to make sure that the changes we make don't negatively impact other uses of `Card` and `CardLink` since they're not just used within grids.

alexcarpenter:
@zchsh yup, LearnSection will handle the layout of these https://github.com/hashicorp/dev-portal/pull/315

PR #328, opened 2022-04-15
feat: use middleware for .io rewrites

dstaley:

---

## Relevant links

- [Asana task](url) 🎟️

## What

Simplifies the rewrite logic by shifting it to `pages/_middleware.ts`

## Why

To support more robust e2e testing, and to eventually support the easier removal of .io rewrites.

## How

In `pages/_middleware.ts`, we examine every incoming request to determine if the request is product-specific. If so, we rewrite all URLs (minus static assets) to the respective `/_proxied-dot-io` version.

## Testing

- Go to preview URL: https://dev-portal-5xxqaf5l9-hashicorp.vercel.app/
- Change to a product
- Ensure homepage renders without issue
- Repeat for remaining products
- Switch to Nomad
- Ensure this URL returns a JSON file: https://dev-portal-5xxqaf5l9-hashicorp.vercel.app/data/vault/nomad-cluster-role.json
- 

## Anything else?

---
  - https://app.asana.com/0/0/1201980045498143


src/pages/_middleware.ts
```diff
@@ -77,6 +91,19 @@ export function middleware(req: NextRequest, ev: NextFetchEvent) {
     return NextResponse.redirect(url, permanent ? 308 : 307)
   }
 
+  // Rewrite .io pages
```
BRKalow, line 94: We might want to change the precedence in `determineProductSlug` with this change. I think the hostname check should take priority, before the cookie.

It's likely not a huge deal, as I would need to have an `io_preview` cookie set on packer.io that is NOT packer, for example, but seems like the correct ordering.

build-libs/rewrites.js
```diff
@@ -1,85 +0,0 @@
-//@ts-check
```
BRKalow, line 1: So excited to see this simplified!

BRKalow, line 1: So excited to see this simplified!

dstaley:
This has probably diverged enough from the base branch that it should probably be re-created from scratch if this is functionality that's still desirable. I'll defer to you and @BRKalow on whether it makes sense to implement this approach given how the middleware currently works in dev-portal.

BRKalow:
Let's close this out for now 👍

PR #336, opened 2022-04-19
style: collection page meta

kendallstrautman:

- [X] The Vercel preview link has been added to this PR's description
- [X] The Asana task has been added to this PR's description

---

## Relevant links

- [Preview link](https://dev-portal-git-kscollection-page-meta-hashicorp.vercel.app/vault/tutorials/cloud) 🔎
- [Asana task](https://app.asana.com/0/1202001387788447/1202149167439223) 🎟️

## What

This PR adds and styles out the collection meta section of the collection view. 

Screen Shot 2022-04-19 at 3 58 34 PM

## Testing

- Visit a few [collection pages](https://dev-portal-git-kscollection-page-meta-hashicorp.vercel.app/vault/tutorials/cloud) and compare with the [design spec](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Dev-Portal?node-id=7053%3A54950)


src/components/icon-tile/types.ts
```diff
@@ -9,7 +9,7 @@ export interface IconTileProps {
   /** Pass a single child, which should be a Flight icon. For 'small' and 'medium' size, pass the 16px icon size; for other sizes pass the 24px icon size. Note that non-"color" icons will be colored using the "brandColor". */
   children: React.ReactNode
   /** Note: the "extra-large" option is not documented in the design system. It's being used for the IconTileLogo component, as used on the /{product} view pages. */
-  size: 'small' | 'medium' | 'large' | 'extra-large'
+  size?: 'small' | 'medium' | 'large' | 'extra-large'
```
kendallstrautman, line 12: This was being treated as optional within the component, a default of 'medium' is set.

src/views/collection-view/components/collection-meta/collection-meta.module.css
```diff
@@ -0,0 +1,45 @@
...
+  @media (--dev-dot-tablet-up) {
+    margin-top: 32px;
+    margin-bottom: 24px;
+  }
+}
+
+.ctaButton {
+  composes: root from '/src/components/button/button.module.css';
+  composes: primary from '/src/components/button/button.module.css';
+  composes: medium from '/src/components/button/button.module.css';
```
kendallstrautman, line 29: Decided to just composes button styles from that component. Lmk if there's a better way to do this.

src/views/collection-view/components/collection-meta/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+        weight="bold"
+        slug={heading.id}
+        className={s.heading}
+      >
+        {heading.text}
+      </Heading>
+      <Text className={s.description}>{description}</Text>
+      <div className={s.cta}>
+        <span className={s.ctaButton}>
+          <Link href={cta.href}>
```
kendallstrautman, line 44: Had to make a unique link here that has primary button-ish styles.

src/views/collection-view/components/collection-meta/collection-meta.module.css
```diff
@@ -0,0 +1,45 @@
...
+  & a {
```
zchsh, line 30: Possibly blocking: I think the `a` here looks odd when focused:

CleanShot 2022-04-20 at 11 20 59@2x

☝️ This is in Firefox, so you can see the focus ring, but in Safari I find it _really_ hard to see (not even sure if it's there? but am sure I'm focusing it, since `Enter` navigates as expected):

CleanShot 2022-04-20 at 11 22 38@2x

Maybe a potential fix: `compose`-ing the above classNames directly onto the `a` element might work? (Though I suspect that might break in some other way, so maybe a terrible suggestion, I'm not sure 😬 ). Possibly related to the above, very speculative suggestion of setting up a `components/button-link` component, & copy-pasting styles into there to make a more isolated component focused on the "`<a />` that looks like a `button`" use case.

src/views/collection-view/components/collection-meta/collection-meta.module.css
zchsh: Thinking a little farther out... maybe we should sync with design on this? Eg do we need to solve for the same "things that look like `Button` and have all the same variations, but could be a `<a />` or a `<button />`" use case that `@hashicorp/react-button` solves for? 🤔 

Or maybe they'd be willing to switch this up to a more standard `StandaloneLink` in order to avoid this whole conundrum?

zchsh (approved):
✅ This looks great 🙌 

---

🤔 I'm (possibly needlessly) worried about the `composes` for the button primary styles, as well as the focus state there.

I tried to express some random thoughts on that in a comment while keeping it brief, but it's kinda long-winded 😅 Approving regardless cause I think we could follow this work up, and refine the `compose` approach to `<a />`-styled-like-button, as we work through the Collection page UI implementation.

src/views/collection-view/components/collection-meta/collection-meta.module.css
kendallstrautman: @zchsh okay I made a new component and copied over the styles from 'Button'. There's still some more work to do on that component to make it fully legit, so I added some comments. But I figure for the sake of this PR, that fine-grained work isn't necessary. Take a look and lmk what you think!

zchsh (approved):
✅✅ Sweet, `components/button-link` looks great to me - `:focus` styles as well 👍

Future thing: Maybe during our sync with Brayden we can check in on how often "links styled as buttons" are used... and eg whether we need to support the full range of possibilities that `button` offers, or whether we can support just a subset 💭  Intent being to help inform whether it makes sense to keep `button-link` separate (if it's a small subset), or whether we should somehow fold it back into the `components/button` folder or something, so that it shares the same style file somehow (if it's super one-to-one like `@hashicorp/react-button`, which from what I understand is _much_ more often used as an `<a />` tag)

PR #351, opened 2022-04-21
Add mobile sidebar functionality to ProductTutorialsView, CollectionView, and TutorialView pages

ashleemboyer:

- [x]  The description template has been filled in below

---

## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambadd-mobile-learn-sidebars-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202022787106807/1202210975997090/f) 🎟️

## 🗒️ What

1. Adds `SidebarMobileControlsProps` to `SidebarMobileControls`
2. `generateProductLandingSidebarNavData` and `generateTopLevelSidebarNavData` now return a whole level of sidebar nav data rather than just the `menuItems` prop for the level.
3. Updates `TutorialsSidebar` to leverage the new `overviewItemHref` `Sidebar` prop.
4. Adds `CollectionViewSidebarContent` and `TutorialViewSidebarContent` in the `TutorialsSidebar` component file.
5. Replaced `Sidebar`'s `sidebarSlot` prop with an `AlternateSidebar` prop.
6. Added `getCollectionViewSidebarSections` helper in `collection-view/server.ts`. It is used in both `CollectionView` and `TutorialView`.

## 🤷 Why

1. `SidebarMobileControls` used to guess the text for the level up/down buttons, but that didn't work as well once there were 4 levels of sidebar nav data in some learn pages. So rather than abstract the guessing too early, I decided to make these explicit props for now.
2. Refactoring `generateProductLandingSidebarNavData` and `generateTopLevelSidebarNavData` makes the code DRYer and clearer what information is repeated for the level. The only differences are `menuItems`, and the rest of the data for the level is repeated.
3. Using `overviewItemHref` delegates some work to `Sidebar`, such as determining if that item is the active page or not.
4. `CollectionViewSidebarContent` and `TutorialViewSidebarContent` are useful for sharing duplicate sidebar content between `CollectionView` and `ProductTutorialsView`.
5. Since we now need to be more restrictive with what's placed in the `Sidebar`, `sidebarSlot` was replaced because it allowed anything to be rendered in the `Sidebar` area. `AlternateSidebar` naming was chosen with a couple of goals in mind:
  - Hopefully `PascalCase` communicates that what should be passed for the prop is a React component.
  - Hopefully putting `Sidebar` in the name indicates that the component passed through the prop renders `Sidebar` inside of it. Haven't figured out the TypeScript for this yet.
6. Abstracting the `getCollectionViewSidebarSections` made the code reusable in multiple views.

## 🛠️ How

Not sure what to add here, but all questions welcome!

## 📸 Design Screenshots

Nothing extra to add here since `Sidebar` designs don't change in this PR.

## 🧪 Testing

For all testing, make sure your viewport width is `1000px` or smaller.

### General

- [ ] Navigating between mobile sidebar nav data using level up/down buttons should still work as expected for all views

### `TutorialView` pages

1. Go to a tutorial page, such as [/waypoint/tutorials/get-started-docker/get-started-intro](https://dev-portal-git-ambadd-mobile-learn-sidebars-hashicorp.vercel.app/waypoint/tutorials/get-started-docker/get-started-intro)
2. Open the `Sidebar` using the icon button at the right of the primary nav bar
3. There should not be an `Overview` menu item at the top of the sidebar content
4. The sidebar items should be correct for the collection the tutorial belongs to
5. There should not be a visible title for the sidebar
6. There should not be a level down button
7. There should be a level up button
8. The text of the level up button should match the name of the collection the tutorial belongs to
9. Click the level up button
10. There should be a level down button that reads `Previous`
11. There should be a level up button that reads `${current product name} Home`
12. The title of this sidebar level should be `Tutorials`
13. There should be an `Overview` item after the title that links to the all tutorials view landing page for the current product
14. The sidebar should show items for all collections, categorized as usual
15. Click the level up button
16. There should be a level down button that reads `Previous`
17. There should be a level up button that reads `Main Menu`
18. The title of this sidebar level should be the current product's name
19. The sidebar menu items should match what's on product landing pages
20. Click the level up button
21. There should be a level down button (but on the lefthand side of the Sidebar) reading `${current product name} Home`
22. The title of this sidebar level should be `Main Menu`
23. There should be links to the main home page and the beta product landing pages

### `CollectionView` pages

1. Go to a collection page, such as [/waypoint/tutorials/get-started-docker](https://dev-portal-git-ambadd-mobile-learn-sidebars-hashicorp.vercel.app/waypoint/tutorials/get-started-docker)
2. Open the `Sidebar` using the icon button at the right of the primary nav bar
3. Items 11-23 from in the `TutorialView pages` above should be true here as well

### `ProductTutorialsView` pages

1. Go to a product tutorials page, such as [/waypoint/tutorials](https://dev-portal-git-ambadd-mobile-learn-sidebars-hashicorp.vercel.app/waypoint/tutorials)
2. Open the `Sidebar` using the icon button at the right of the primary nav bar
3. Items 11-23 from in the `TutorialView pages` above should be true here as well

## 💭 Anything else?

Most items are noted in new `TODO`s added to the code. In addition:

- When viewing a `TutorialView` page, after opening the `Sidebar` on mobile and clicking the level up button one time, the menu item for the collection the tutorial belongs to is active. I don't think this is the expect behavior, at least in terms of how that active state also sets `aria-current="page"` for active menu items. `aria-current` is not correct in this situation and it should only be set for the lowest-level sidebar nav data.


src/views/collection-view/index.tsx
```diff
@@ -1,29 +1,108 @@
...
-  product,
   layoutProps,
 }: CollectionPageProps): React.ReactElement {
-  const { name, slug, description, shortName, tutorials } = collection
+  const currentProduct = useCurrentProduct()
+  const currentPath = useCurrentPath({ excludeHash: true, excludeSearch: true })
+  const { name, slug, description, tutorials } = collection
+
+  // TODO: refactor, very long
+  const sidebarNavDataLevels = [
```
kendallstrautman, line 29: Perhaps we could do some of this data generation in `getStaticProps` and some of the base boilerplate could be shared between views? It's hard for me to see how _similar_ these are between each view. 

Or what are your plans for refactoring this?

src/views/collection-view/index.tsx
```diff
@@ -1,29 +1,108 @@
...
+              return (
+                <>
+                  <HorizontalRule />
+                  <SectionTitle text={section.title} />
+                  <SectionList items={section.items} />
+                </>
+              )
+            }
+          )}
+          <HorizontalRule />
```
kendallstrautman, line 79: I liked your idea of separating this resources section into its own component. And ya perhaps we could separate this whole children piece into a component?

src/views/collection-view/index.tsx
```diff
@@ -1,29 +1,108 @@
...
     <SidebarSidecarLayout
       breadcrumbLinks={layoutProps.breadcrumbLinks}
       headings={layoutProps.headings}
-      sidebarSlot={
-        <ProductCollectionsSidebar
-          product={{ name: product.name, slug: product.slug }}
-          sections={layoutProps.sidebarSections}
-        />
-      }
+      AlternateSidebar={TutorialsSidebar}
```
kendallstrautman, line 104: Could you walk me through the function of `AlternateSidebar` now?

src/views/collection-view/server.ts
```diff
@@ -27,14 +27,29 @@ export interface CollectionPageProps {
...
+> & { collectionViewSidebarSections: CollectionCategorySidebarSection[] }
 
 export interface CollectionPagePath {
   params: {
     collectionSlug: string
   }
 }
 
+export async function getCollectionViewSidebarSections(
+  product: LearnProductData,
```
kendallstrautman, line 39: 💭  This could probably just be `productSlug`, to limit the data being passed there?

kendallstrautman: I think this would be the function that we would call client-side on menu open if we wanted to do that dynamic approach.... then the sidebar children could be generated based on that data.

src/views/collection-view/server.ts
```diff
@@ -48,14 +63,6 @@ export async function getCollectionPageProps(
...
   const collection = await getCollection(`${product.slug}/${slug}`)
-  // For sidebar data
-  const allProductCollections = await getAllCollections({
-    product: { slug: product.slug, sidebarSort: true },
-  })
-  const filteredCollections = filterCollections(
-    allProductCollections,
-    product.slug
-  )
   const layoutProps = {
```
kendallstrautman, line 66: 💭  We could add some of the sidebar layout props here? If we moved some of that level logic into `getStaticProps`

src/views/collection-view/index.tsx
```diff
@@ -1,29 +1,108 @@
...
 import { CollectionPageProps } from './server'
 import CollectionMeta from './components/collection-meta'
 
 function CollectionView({
   collection,
-  product,
   layoutProps,
 }: CollectionPageProps): React.ReactElement {
-  const { name, slug, description, shortName, tutorials } = collection
+  const currentProduct = useCurrentProduct()
```
kendallstrautman, line 24: Isn't `product` passed anyways for the other layout pieces? Any reason why we wouldn't just use the prop here instead of the hook?

ashleemboyer: Either way would work! I don't have a strong preference for either. I just used the hook here so the API for `CollectionView` didn't _have_ to receive the `product` object. It's mainly passed through  for `App` to hydrate the context. It's just coincidental that we can grab the prop here, not necessarily standard for every view component to be able to do it.

src/views/collection-view/index.tsx
ashleemboyer: Yeah... 🤔 I think the top two levels are the same for all of these views, so that could probably be shared between them with a helper function. And then each view could handle appending whatever extra levels it has. Thoughts on that?

Second thought would be to add something like a `TutorialsSidebarChildren` React component that lives in `TutorialsSidebar`'s file, and then replacing this React code under `children` with `<TutorialsSidebarChildren />` (and whatever props it needs). I think that would make it reusable in multiple views!

src/views/collection-view/index.tsx
ashleemboyer: So it's a more structured version of `sidebarSlot`, and doesn't let just anything to be rendered in the `Sidebar` portion of the layout. It's currently pretty manual (depends on the developer to know exactly what to do) and I'm not sure yet how/if the TypeScript can be updated to enforce the rules the `AlternateSidebar` expects the component passed to it to follow. But the gist is: the name of the component passed to `AlternateSidebar` needs to render a `Sidebar` inside of it.

src/views/collection-view/server.ts
ashleemboyer: Riffing off of my other comment proposing something like `TutorialsSidebarContent`, I wonder if we could call this function from there? Doing that should defer the call until the component is rendered. For this view, that'll be immediately since the level that needs the data is at the level of this view. For `TutorialView`, I think `TutorialsSidebarContent` wouldn't be rendered until the `Sidebar` is opened and the level up button is pressed. 🤔 I'll experiment with this a little bit when I make `TutorialsSidebarContent`!

src/views/collection-view/server.ts
ashleemboyer: I think we can construct some of the levels that'll get passed to the layout, but this view will have to be responsible for adding one more level before the array is passed, since it's setting a `children` prop that will be a React component. Same for `TutorialsView`, but that one will have to add 2 levels instead of just one. Side note, I think the same component can be used for both of those levels. I'll get that implemented in my next pass!

src/views/collection-view/index.tsx
kendallstrautman: Since the data is already being passed, imo it would make more sense to just use that! It's up to you though 👍

src/views/collection-view/index.tsx
kendallstrautman: I think that sounds great! Condensing these various children iterations into their own composed components, and having those live within the `TutorialSidebar` dir.

src/views/collection-view/index.tsx
kendallstrautman: Got it, and the idea that were retaining `TutorialSidebar` (as the alternate passed) even though it is effectively the same as `Sidebar` is that later on we may adjust how this alternate sidebar behaves?

src/views/collection-view/server.ts
kendallstrautman: If we're going to try out the clientside call, I'd say we should try to defer it until a certain user action, like the opening of the menu. This way we ensure that the data is only called when the user is in fact is a mobile-esque context. If we move it to the client, but then call it everytime anyways, I think it'd be better to just keep it in the server, if that makes sense.

src/views/collection-view/index.tsx
ashleemboyer: 🗒️ Decided to keep this here since `product` won't always be passed as a prop.

src/views/collection-view/index.tsx
ashleemboyer: 🗒️ Ended up refactoring both `generateTopLevelSidebarNavData` and `generateProductLandingSidebarNavData` to return more data.

src/views/collection-view/server.ts
ashleemboyer: 🗒️ Will be deferring the work for optimizing the API call to a follow-up PR.

zchsh (approved):
✅  I worked through the testing steps, and everything seems to check out! Was a little lighter on code review but happy to dive in more deeply if that'd be helpful.

I did notice a scrolling quirk, I think I might've seen Bryce mention this in another PR - definitely something we could handle outside this PR:

scrolls-when-open

src/components/tutorials-sidebar/index.tsx
```diff
@@ -30,6 +35,66 @@ function TutorialsSidebar({
...
+        return (
+          <>
+            <HorizontalRule />
+            <SectionTitle text={section.title} />
+            <SectionList items={section.items} />
+          </>
+        )
+      })}
+      <HorizontalRule />
+      <SectionTitle text="Resources" />
```
kendallstrautman, line 55: 💭 Seems like there's small opportunity to create a `Resources` component to share between these two views.

src/views/product-tutorials-view/index.tsx
```diff
@@ -12,14 +18,30 @@ import s from './product-tutorials-view.module.css'
 function ProductTutorialsView({
   layoutProps,
   data,
-  product,
 }: ProductTutorialsViewProps): React.ReactElement {
+  const currentProduct = useCurrentProduct()
```
kendallstrautman, line 22: I do think it makes more sense to use the `product` props that are passed to this view.

src/views/tutorial-view/index.tsx
```diff
@@ -60,13 +72,17 @@ export type CollectionContext = {
...
+    SidebarSidecarLayoutProps,
+    'children' | 'headings' | 'breadcrumbLinks'
+  > & { sidebarSections: CollectionCategorySidebarSection[] }
 >
 
 export default function TutorialView({
   layout,
   tutorial,
 }: TutorialViewProps): React.ReactElement {
+  const currentProduct = useCurrentProduct()
```
kendallstrautman, line 85: Again `product` is already passed as a prop to this view. I feel like we should just use that.

src/views/tutorial-view/server.ts
```diff
@@ -60,6 +61,10 @@ export async function getTutorialPageProps(
     collection.data,
     fullTutorialData.collectionCtx
   )
+  const sidebarSections = await getCollectionViewSidebarSections(
```
kendallstrautman, line 64: We should add a note here re: that this calls _all_ product collections and how we plan to potentially refactor to load that data clientside.

src/views/collection-view/index.tsx
ashleemboyer: Hopefully we can get rid of it in the process of refactors that allow us to not have to manually hydrate `CurrentProductContext` anymore. 😊

src/views/tutorial-view/server.ts
ashleemboyer: Added in https://github.com/hashicorp/dev-portal/pull/351/commits/cf4703efef38589eaf7c4cd7c2bb2718f25684f8!

src/views/product-tutorials-view/index.tsx
ashleemboyer: I took this into consideration before choosing the hook, and ultimately decided to go with the hook because I think it benefits us more in the long term.

kendallstrautman (changes requested):
Looking good! The only thing I'd like to see changed is the use of the product prop versus context on the tutorial views. 

For VQA, I noticed a few things, but not necessary to address in this PR. Is there a place I should make follow-up tickets? 

Here's a [loom vid](https://www.loom.com/share/35d882d64a07417f81e85d81aecb5c73) of me walking through things on mobile. I mentioned a _ton_ of stuff that of course doesn't need to be handled on this PR — probably more of a reflection of the mobile refinement waterfall. 

🐛 There is one bug that's worth noting: if you scroll far down a page, and then open the menu, you may not see the menu — related to the vertical scroll bug Zach mentioned.

kendallstrautman (approved):
🙏 🙇‍♀️ 🚀

PR #357, opened 2022-04-21
fix: warn not throw for image-config issue

zchsh:

---

## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Fixes an issue where `ImageConfig` is too strict; we have invalid content, so we'd rather show it kind-of-working than have the whole page error.

## 🤷 Why

To avoid showing an error page.

## 🛠️ How

**Context:**
- Previously we strictly required _exactly_ `1` child in `ImageConfig` (note: nested `p` tag is cause MDX is fun like that, mostly irrelevant here)
- Because we had this strictness, after validation we could assume `children[0]` was definitely an `img`
- The error we encountered was because there were two children - a text node first, and an `img` tag second

**Fix:**
- Now, we filter `children` to find the `img` tag. So we can gracefully handle if other children are included...
- ⚠️ BUT, IMPORTANT: this means that the text node we previously rendered is now ignored!
    - However, _this is the intent behind the `ImagConfig` component_. So, I think this is a reasonable solution.
    - We now _warn_ in dev if multiple children are encountered.
    - I've set up 🎟️  [an Asana task](https://app.asana.com/0/1202097197789424/1202163585099211/f) to eventually escalate the warning to an error (once we can make the content fixes necessary)

## 🧪 Testing

- [ ] Visit [the page that was erroring][preview], and ensure that it now renders correctly.

## 💭 Anything else?

Nope!

[preview]: https://dev-portal-git-zsfix-image-config-strictness-hashicorp.vercel.app/vault/tutorials/auth-methods/approle
[task]: https://app.asana.com/0/1202001387788447/1202156371151365/f


zchsh:
@kendallstrautman I think this should fix the issue you raised [in this task](https://app.asana.com/0/1202001387788447/1202156371151365/f) the other day. Let me know what you think 🙇‍♂️

kendallstrautman (approved):
💯 Thanks for fixing this

zchsh:
Breadcrumb: going to follow this up with [[image-config] throw error rather than warning for misuse](https://app.asana.com/0/1202097197789424/1202163585099211/f). Will likely involve backporting our `ImageConfig` work to `learn` as a temporary solution.

PR #363, opened 2022-04-22
feat: build collection-tutorial-list for collection views

zchsh:
## ✅ "Ready for Review" checklist

- [x] The Vercel preview link has been added to this PR's description
- [x] The Asana task has been added to this PR's description
- [x] This PR's link has been added to the "📐 Pull Request" field on the Asana task
- [x] This Vercel preview link has been added to the "🔍 Preview" field on the Asana task
- [x] The description template has been filled in below

---

## 🔗 Relevant links

- [Preview link](https://dev-portal-git-zscollection-tutorial-list-hashicorp.vercel.app/waypoint/tutorials/get-started-docker) 🔎
- [Asana task](https://app.asana.com/0/1202001387788447/1202149167439225/f) 🎟️

## 🗒️ What

Implements a `CollectionTutorialList` component, and uses it on all collection views.

## 🤷 Why

To continue building out our `collection-view` in dev-dot.

## 🛠️ How

- Uses existing `tutorial-card` component in a `display: grid` layout
- Accepts an `isOrdered` prop, as different collection types have different grid characteristics

## 📸 Design Screenshots

<details>

<summary>Ordered CollectionTutorialList</summary>

### Design spec

TO DO

### Implementation

TO DO

</details>

<details>

<summary>Unordered CollectionTutorialList</summary>

### Design spec

TO DO

### Implementation

TO DO

</details>

## 🧪 Testing

- [ ] Visit an ordered collection, such as [/waypoint/tutorials/get-started-docker][ordered-collection]
    - [ ] `CollectionTutorialList` should render in one column at all viewport sizes
- [ ] Visit an unordered collection, such as [/vault/tutorials/secrets-management][unordered-collection]
    - [ ] `CollectionTutorialList` should render in one column on mobile, and max out at 2 columns on larger viewports

> Note: you can also review [in Storybook][storybook], which I found helpful for testing responsive behaviour.

## 💭 Anything else?

Not at the moment!

[ordered-collection]: https://dev-portal-git-zscollection-tutorial-list-hashicorp.vercel.app/waypoint/tutorials/get-started-docker
[unordered-collection]: https://dev-portal-git-zscollection-tutorial-list-hashicorp.vercel.app/vault/tutorials/secrets-management
[storybook]: https://dev-portal-with-storybook.vercel.app/?path=/story/components-collectiontutoriallist--playground

PR #367, opened 2022-04-25
feat: match collection view to spec

zchsh:

---

## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][asana] 🎟️

## 🗒️ What

This PR finishes the UI implementation of `views/collection-view`. With recent work completed in #363 , the only remaining UI change needed is to remove the `Tutorials` heading and the sidecar to match design specs.

## 🤷 Why

To match the `collection-view` implementation to design specs.

## 🛠️ How

- Makes a change in `SidebarSidecarLayout` to allow `sidecarSlot` to be falsy
- Passes `null` to `sidecarSlot` in order to not render a sidecar content
- Removes `heading` generation for `collection-view` as it is now unused
- Makes related TypeScript updates to removal of `headings` from `collection-view` props

## 📸 Design Screenshots

<details>
<summary>Implementation before this PR</summary>
before
</details>

<details>
<summary>Design spec</summary>
design-spec
</details>

<details>
<summary>Implementation after this PR</summary>
after
</details>

## 🧪 Testing

- [ ] Visit a collection view, such as [`/vault/tutorials/encryption-as-a-service`][preview]
- [ ] Ensure the `Tutorials` heading is no longer rendered, and the sidecar area is empty

## 💭 Anything else?

Not at the moment!

[preview]: https://dev-portal-git-zsfinish-collection-view-hashicorp.vercel.app/vault/tutorials/encryption-as-a-service
[asana]: https://app.asana.com/0/1202001387788447/1202168098250852/f

### DIFF
--- a/src/layouts/sidebar-sidecar/index.tsx
+++ b/src/layouts/sidebar-sidecar/index.tsx
@@ -57,7 +57,7 @@ const SidebarSidecarLayoutContent = ({
   }
 
   const SidecarContent = (): ReactElement => {
-    if (sidecarSlot) {
+    if (typeof sidecarSlot !== 'undefined') {
       return sidecarSlot
     }
 
--- a/src/views/collection-view/index.tsx
+++ b/src/views/collection-view/index.tsx
@@ -28,7 +28,9 @@ function CollectionView({
       }
     >
       <CollectionMeta
-        heading={{ text: name, id: 'overview' }}
+        // Note: id is passed here because it is required by <Heading />,
+        // it's not used for #anchor linking since there is no sidecar.
+        heading={{ text: name, id: collection.id }}
         description={description}
         cta={{ href: getTutorialSlug(tutorials[0].slug, slug) }}
         numTutorials={tutorials.length}
--- a/src/views/collection-view/index.tsx
+++ b/src/views/collection-view/index.tsx
@@ -19,7 +19,7 @@ function CollectionView({
   return (
     <SidebarSidecarLayout
       breadcrumbLinks={layoutProps.breadcrumbLinks}
-      headings={layoutProps.headings}
+      sidecarSlot={null}
       sidebarSlot={
         <ProductCollectionsSidebar
           product={{ name: product.name, slug: product.slug }}
@@ -28,12 +28,11 @@ function CollectionView({
       }
     >
       <CollectionMeta
-        heading={{ text: name, id: layoutProps.headings[0].slug }}
+        heading={{ text: name, id: 'overview' }}
         description={description}
         cta={{ href: getTutorialSlug(tutorials[0].slug, slug) }}
         numTutorials={tutorials.length}
       />
-      <h2 id={layoutProps.headings[1].slug}>Tutorials</h2>
       <CollectionTutorialList
         isOrdered={ordered}
         tutorials={tutorials.map((t: ClientTutorialLite) =>
--- a/src/views/collection-view/server.ts
+++ b/src/views/collection-view/server.ts
@@ -26,7 +26,7 @@ export interface CollectionPageProps {
 
 export type CollectionLayout = Pick<
   SidebarSidecarLayoutProps,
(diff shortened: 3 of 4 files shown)


src/layouts/sidebar-sidecar/index.tsx
```diff
@@ -57,7 +57,7 @@ const SidebarSidecarLayoutContent = ({
   }
 
   const SidecarContent = (): ReactElement => {
-    if (sidecarSlot) {
+    if (typeof sidecarSlot !== 'undefined') {
```
zchsh, line 60: I think we want to pass `null` to `sidecarSlot` to keep it empty, so updated this check to be a bit more precise.

src/views/collection-view/index.tsx
```diff
@@ -28,12 +28,11 @@ function CollectionView({
       }
     >
       <CollectionMeta
-        heading={{ text: name, id: layoutProps.headings[0].slug }}
+        heading={{ text: name, id: 'overview' }}
```
zchsh, line 31: Hard-coded this at this level, as `slug` in the underlying `Heading` component is a required prop.

src/views/collection-view/index.tsx
```diff
@@ -28,12 +28,11 @@ function CollectionView({
...
         description={description}
         cta={{ href: getTutorialSlug(tutorials[0].slug, slug) }}
         numTutorials={tutorials.length}
       />
-      <h2 id={layoutProps.headings[1].slug}>Tutorials</h2>
```
zchsh, line 36: Have removed this heading entirely, to match the design spec.

src/views/collection-view/server.ts
```diff
@@ -26,7 +26,7 @@ export interface CollectionPageProps {
 
 export type CollectionLayout = Pick<
   SidebarSidecarLayoutProps,
-  'headings' | 'breadcrumbLinks'
```
zchsh, line 29: The `CollectionLayout` no longer uses `headings`, so I've dropped it from the type.

src/views/product-tutorials-view/server.ts
```diff
@@ -25,7 +26,10 @@ export interface ProductTutorialsPageProps {
   product: ProductTutorialsPageProduct
 }
 
-type ProductTutorialsLayout = CollectionLayout
+type ProductTutorialsLayout = Pick<
```
zchsh, line 29: `ProductTutorialsLayout` previously mirrored `CollectionLayout`, but now it does not, as the former needs `headings` while the latter has no use for them.

zchsh:
@kendallstrautman This handles the TODO mentioned in [this thread on #363](https://github.com/hashicorp/dev-portal/pull/363#discussion_r856669604).

---

Lower priority note on `helper` & `util` functions shared across multiple components:

As in #363 with `getReadableTime`, I considered trying to move these out of specific `component` folders, as many of them are shared... but wasn't sure where we should put them... maybe `lib`? `lib/learn-client`? or maybe not in `learn-client` since they aren't really "the client"? 🤔 In the end I figured since there are quite a few we'd potentially want to move around (_if_ it makes sense), I'd wait to get your take on what we should do, and _if_ it feels like re-organization is worth it, then we could establish a shared folder and move a bunch of stuff around in a separate PR.

src/views/collection-view/index.tsx
kendallstrautman: You could also use `id` or `slug` from the `collection` prop

zchsh: ah true, that might make it more clear that this now has nothing to do with sidecar - will update! 👍

zchsh:
@kendallstrautman Made a task for myself for the `lib/tutorials` change - 🎟️ [[chore] make lib/tutorials for shared use tutorial helpers](https://app.asana.com/0/1202097197789424/1202180048371484/f). Going to merg this in, & follow up with that later 👍

PR #370, opened 2022-04-26
feat: handle 404 tutorial views

kendallstrautman:

---

## 🔗 Relevant links

- [Preview link](https://dev-portal-git-kshandle-404-hashicorp.vercel.app/waypoint/tutorials) 🔎
- [Asana task](https://app.asana.com/0/1199634971449915/1202016460311408) 🎟️

## 🗒️ What

This PR adjusts the tutorial and collection page prop generation functions to return the 404 page if the collection or tutorial doesn't exist. The design implementation of the 404 page will occur later

## 🧪 Testing

- Visit this 404 link for a [tutorial](https://dev-portal-git-kshandle-404-hashicorp.vercel.app/waypoint/tutorials/get-started-docker/fake-tutorial) and [collection](https://dev-portal-git-kshandle-404-hashicorp.vercel.app/waypoint/tutorials/fake-collection) page
- Visit a valid tutorial and collection page, make sure they work as expected.


src/views/tutorial-view/utils/get-collection-context.ts
```diff
@@ -44,24 +41,24 @@ export async function getCurrentCollectionTutorial(
    * This type is only `TutorialLite` which doesn't have the tutorial content
    * so we only need the slug to make another request to get the full tutorial data in server.ts
    */
-  const currentTutorial = collection.tutorials.find((t) =>
+  const currentTutorial = collection?.tutorials.find((t) =>
     t.slug.endsWith(tutorialFilename)
   )
 
   if (!currentTutorial) {
-    throw Error(
```
kendallstrautman, line 52: Not throwing here, handling it as a 404 instead.

src/pages/vault/tutorials/[...tutorialSlug]/index.tsx
```diff
@@ -20,9 +20,16 @@ export function VaultTutorialPage({
...
   const product = vaultData as LearnProductData
-  return getTutorialPageProps(product, params.tutorialSlug)
+  const data = await getTutorialPageProps(product, params.tutorialSlug)
+
+  // If the tutorial doesn't exist, hit the 404
+  if (!data) {
+    return { notFound: true }
+  }
+
+  return { props: data, revalidate: 300 }
```
Comment (kendallstrautman) [inline src/pages/vault/tutorials/[...tutorialSlug]/index.tsx:32]: Not sure of the best `revalidate` number here. Or if that's config we should consolidate.

src/pages/vault/tutorials/[collectionSlug].tsx
```diff
@@ -10,10 +11,21 @@ import {
...
-}): Promise<{ props: CollectionPageProps }> {
+}): Promise<GetStaticPropsResult<CollectionPageProps>> {
   const { collectionSlug } = params
   const product = vaultData as LearnProductData
-  return await getCollectionPageProps(product, collectionSlug)
+
+  const props = await getCollectionPageProps(product, collectionSlug)
+
+  // If the collection doesn't exist, hit the 404
+  if (!props) {
```
Comment (kendallstrautman) [inline src/pages/vault/tutorials/[collectionSlug].tsx:21]: There's some repetition here, it could be lifted up into the server file but for some reason I like this being in the actual page file, since it so clearly shows how 404s are handled.

src/views/tutorial-view/server.ts
```diff
@@ -36,12 +36,23 @@ export interface TutorialPageProps {
...
   )
+
+  // the tutorial doesn't exist in collection - return 404
+  if (tutorialReference.dbSlug === null || collection.data === null) {
+    return null
+  }
+
   const fullTutorialData = await getTutorial(tutorialReference.dbSlug)
+  // double guard if tutorial doesn't exist after call - return 404
+  if (fullTutorialData === null) {
```
kendallstrautman, line 52: This is an extra safety guard, in most cases if the tutorial doesn't exist, we wouldn't get to this call. But I think this is still an important safety guard.

src/pages/waypoint/tutorials/[...tutorialSlug]/index.tsx
```diff
@@ -20,9 +20,16 @@ export function WaypointTutorialPage({
 
 export async function getStaticProps({
   params,
-}): Promise<{ props: TutorialPageProps }> {
+}): Promise<GetStaticPropsResult<TutorialPageProps>> {
   const product = waypointData as LearnProductData
-  return await getTutorialPageProps(product, params.tutorialSlug)
+  const data = await getTutorialPageProps(product, params.tutorialSlug)
```
Comment (zchsh) [inline src/pages/waypoint/tutorials/[...tutorialSlug]/index.tsx:25]: This is a _super_ minor nitpick, but noticed:
- in `getTutorialPageProps` we call the return value within the function `props`, and assign it to `const data`
- in `getCollectionPageProps` we call the return value within the function `data`, and assign it to `const props`
I think consistently using `props` might be more clear, but still feels pretty clear as-it so definitely feel free to ignore 👍

src/pages/vault/tutorials/[...tutorialSlug]/index.tsx
Comment (zchsh) [reply src/pages/vault/tutorials/[...tutorialSlug]/index.tsx:32]: Non-blocking out-of-curiosity question: Before this PR, we weren't returning a `revalidate` value at all, right? 🤔

I am a bit fuzzy on how `revalidate` & ISR works with relation to 404 pages, but if that's the case and we weren't returning `revalidate` previously, I wonder if it'd be possible to handle the 404 case without adding it? (not saying we _should_, just wondering if possible ☺️ )

zchsh (approved):
✅💯 This looks great to me!

I visited the `Testing` links, and they both render `404`s for me, which seems way better than the [`500 This Serverless Function has crashed`](https://developer.hashi-mktg.com/waypoint/tutorials/get-started-docker/fake-tutorial) upstream 🙌

I dig the repetition of `return { notFound: true }` in page files - +1 that it  feels more clear so very much worth it 👍 Especially given we have only 4 page files at the moment - _maybe_ once we have dozens of such files we'll want to abstract it, but even _if_ that's the case I think it's great to wait till it actually feels painful.

Had a super minor nit around `data` vs `prop` naming in `getCollectionPageProps` vs `getTutorialPageProps`, and a question on whether we need to add `revalidate` in this PR, but both are very much feel-free-to-ignore, non-blocking things 🚀

src/pages/vault/tutorials/[...tutorialSlug]/index.tsx
Comment (kendallstrautman) [reply src/pages/vault/tutorials/[...tutorialSlug]/index.tsx:32]: Ya fair, I can make a ticket for this to be something we address on another PR! To chat through how / if we want to add this.

src/pages/waypoint/tutorials/[...tutorialSlug]/index.tsx
Comment (kendallstrautman) [reply src/pages/waypoint/tutorials/[...tutorialSlug]/index.tsx:25]: Good call, I'll adjust to use `props`! 👌

PR #375, opened 2022-04-29
feat: implement BrandedCallout

zchsh:

---

## 🔗 Relevant links

- [Preview link][waypoint-tutorials] 🔎
- [Asana task][task] 🎟️ 

## 🗒️ What

Implements the `BrandedCallout` display component for `views/product-tutorial-view`.

## 🤷 Why

To finish building out the `product-tutorial-view`.

## 🛠️ How

- Add styling to the existing component
- Uses static imports with `next/image` to bring in product patterns. This required adding `declare module '*.png'` to `global.d.ts`

## 📸 Design Screenshots

<details>
<summary>⤵️ Design & Implementation</summary>

> Note: `box-shadow` looks a bit off, but implementation uses `hds-surface-low` which seems to correspond directly to designs, so I'm hesitant to try to eject from that to better match the Figma appearance.

### Design

design

### Implementation

vault-implementation

waypoint-implementation

</details>

## 🧪 Testing

- [ ] Visit [the Waypoint tutorials landing][waypoint-tutorials]
    - [ ] Ensure the `Sitemap` component looks as expected
- [ ] Visit [the Vault tutorials landing][vault-tutorials]
    - [ ] Ensure the `Sitemap` component looks as expected

## 💭 Anything else?

Not at the moment!

[waypoint-tutorials]: https://dev-portal-git-zsbranded-callout-hashicorp.vercel.app/waypoint/tutorials
[vault-tutorials]: https://dev-portal-git-zsbranded-callout-hashicorp.vercel.app/vault/tutorials
[task]: https://app.asana.com/0/1202097197789424/1202182325935196/f


src/views/product-tutorials-view/components/product-view-content/components/branded-callout/index.tsx
```diff
@@ -9,23 +35,40 @@ function BrandedCallout({
...
+    >
+      <div className={s.textContainer}>
+        <h2 id={headingSlug} className={s.heading}>
+          {heading}
+        </h2>
+        {subheading ? <p className={s.subheading}>{subheading}</p> : null}
+        <p className={s.cta}>
+          <StandaloneLink
+            text={cta.text}
+            href={cta.url}
```
kendallstrautman, line 55: We probably need to run this url through a formatter of some sort, as the links are still pointing to old Learn paths. We probably don't know whether its a tutorial or collection url though, so we'd need to parse that 🤔

zchsh:
@kendallstrautman I've taken a shot at the URL formatting for `BrandedCallout` links, with a new helper function `detect-and-reformat-learn-url`.

It gets a bit complex since I don't think we can assume that tutorial links will have an `?in` param defined, and in those cases, it seems we need to fetch the tutorial from the API to reformat to the `dev-dot` version of the link.

I think longer-term we could consider a different approach on the content side - for example, the `cta.url` value could be replaced by a `cta.collectionSlug` or `cta.tutorialSlug` or something, and then we'd have the tutorial or collection data we need thanks to the inline content resolution... but that felt very speculative; out-of-scope, and in some ways much more complex than having a stopgap URL correcter (that will go away at some point, I think, once authors start using the new URL formats).

Let me know what you think 🙇‍♂️ This PR does feel like it's growing a bit in scope, so also happy to change gears, undo the `detect-and-reformat-learn-url` work, and handle that in a separate PR if that's preferable.

src/views/product-tutorials-view/server.ts
```diff
@@ -85,11 +78,16 @@ export async function getProductTutorialsViewProps(
     allProductCollections,
     productSlug
   )
+  /**
+   * Process page data, reformatting as needed.
+   * Includes parsing headings, for use with the page's sidecar
+   */
+  const { pageData, headings } = await processPageData(rawPageData)
```
zchsh, line 85: I've moved the `headings` work into the `processPageData` function. I figure this might be a good place to do the ["trim down static props size"](https://app.asana.com/0/0/1202182325935203/f) work as well, in the future.

src/lib/learn-client/index.ts
```diff
@@ -1,6 +1,6 @@
 function getFetch() {
   // Note: purposely doing a conditional require here so that `@vercel/fetch` is not included in the client bundle
-  if (typeof window === 'undefined') {
+  if (typeof window === 'undefined' || typeof window.fetch !== 'function') {
```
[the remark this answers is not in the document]
kendallstrautman: Perhaps you want to mock that data call in the test? [I used nock ](https://github.com/hashicorp/dev-portal/blob/main/src/lib/remark-plugins/rewrite-tutorial-links/__tests__/rewrite-tutorial-links.test.ts#L91-L95)to handle this for the tutorial link rewrites work.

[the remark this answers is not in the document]
kendallstrautman: From what I see it looks like the test would actually fetch from the API currently? Which may not be what we want.

src/views/product-tutorials-view/helpers/detect-and-reformat-learn-url.ts
```diff
@@ -0,0 +1,69 @@
...
+import isAbsoluteUrl from 'lib/is-absolute-url'
+
+/**
+ * Given a URL string,  detect whether it is a collection or tutorial URL
+ * in the format used on learn.hashicorp.com.
+ *
+ * If it is, then reformat the URL to work with dev-dot's URL structure.
+ * Otherwise, return the URL unmodified.
+ */
+async function detectAndReformatLearnUrl(url: string): Promise<string> {
```
kendallstrautman, line 14: I'm seeing that we need to create a generic handler for this. This work is reminiscent of the [rewriting links logic](https://github.com/hashicorp/dev-portal/blob/main/src/lib/remark-plugins/rewrite-tutorial-links/utils/handle-learn-links.ts) in the remark plugin. 

I wonder if you could lean on the [tutorials-map endpoint ](https://github.com/hashicorp/dev-portal/blob/main/src/pages/api/tutorials-map.ts)to get the slug...Since it's already cached. Not mandatory, I'm just seeing how we can reuse some code here if it makes sense to.

src/views/product-tutorials-view/helpers/detect-and-reformat-learn-url.ts
```diff
@@ -0,0 +1,69 @@
...
+  // Don't mess with absolute URLs
+  if (isAbsoluteUrl(url)) {
+    return url
+  }
+  // Gather info on whether this is a collection or tutorial URL
+  const urlObject = new URL(url, 'https://learn.hashicorp.com')
+  const pathParts = urlObject.pathname.split('/')
+  const isTutorialUrl =
+    pathParts[1] === 'tutorials' && isThemeOrProduct(pathParts[2])
```
kendallstrautman, line 23: It could help to name what the pathPart is, `tutorialFilename` or `productSlug` etc. Since its difficult here to see how things are being combined.

src/lib/learn-client/index.ts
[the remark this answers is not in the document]
zchsh: Ah okay that makes sense - I also didn't realize https://github.com/hashicorp/dev-portal/blob/main/src/lib/remark-plugins/rewrite-tutorial-links/index.ts existed, seems like I might be able to re-use parts of that...

eg as a _very_ skateboard version we could construct a `Link` mdast node and make a named export of `handleRewriteTutorialsLink`; which would avoid doing any refactoring at all; maybe as a step beyond that we could split out a function that acts on `url` rather than `node.url` 🤔 

Apologies for the thrash here, I should have looked around a bit first 🙇‍♂️

src/views/product-tutorials-view/helpers/detect-and-reformat-learn-url.ts
```diff
@@ -0,0 +1,69 @@
...
+  if (isTutorialUrl) {
+    // For tutorial URLs, we need a collection slug
+    const tutorialSlugDb = `${pathParts[2]}/${pathParts[3]}`
+    // If there's an "?in=" param, we can use that as the collection slug
+    const inParam = urlObject.searchParams.get('in')
+    if (inParam) {
+      return getTutorialSlug(tutorialSlugDb, inParam)
+    } else {
+      // If not, we need to fetch the tutorial, & determine its default context
+      const tutorialData = await getTutorial(tutorialSlugDb)
```
kendallstrautman, line 35: We probably don't want to call the API for this tutorial if its a non-beta product right? Since we would just rewrite the url to externally point to Learn?

src/lib/remark-plugins/rewrite-tutorial-links/index.ts
```diff
@@ -66,32 +66,40 @@ export const rewriteTutorialLinksPlugin: Plugin = () => {
 }
 
 function handleRewriteTutorialsLink(node: Link | Definition) {
+  node.url = rewriteTutorialsLink(node.url, TUTORIAL_MAP)
+}
+
+export function rewriteTutorialsLink(
```
zchsh, line 72: This is a very light refactor to expose a `rewriteTutorialsLink` function that we can re-use elsewhere. In brief:
- changes `node.url` to new `let newUrl`, which is returned at the end of the function
- adds an argument for `TUTORIAL_MAP`, so that it can be provided without having to call the remark plugin
- everything else within `rewriteTutorialsLink` stays the same
- wraps `rewriteTutorialsLink` in `handleRewriteTutorialsLink` to keep the same clean call as above in each `visit` call

src/views/product-tutorials-view/helpers/__tests__/detect-and-reformat-learn-url.test.ts
```diff
@@ -0,0 +1,102 @@
+import nock from 'nock'
```
zchsh, line 1: Grabbed use of `nock` from `rewrite-tutorial-links` tests.

src/lib/learn-client/types.ts
```diff
@@ -261,5 +261,25 @@ export enum CompanyLogoOption {
 export function themeIsProduct(
   theme: ThemeOption | ProductOption
 ): theme is ProductOption {
-  return Object.values(ProductOption).includes(theme as ProductOption)
+  return isProductOption(theme)
+}
+/**
+ * Type guard to determine if a string is a ThemeOption or ProductOption
```
zchsh, line 267: I've added these type guards here in `lib` since they felt very generic. I'm using `isThemeOrProduct` in `detect-and-reformat-learn-url`.

src/views/product-tutorials-view/components/product-view-content/components/branded-callout/index.tsx
zchsh: Update here: have done a bunch of refactoring and now have `detect-and-reformat-learn-url` which uses `rewrite-tutorials-link` under the hood.

zchsh:
@kendallstrautman Thanks for the excellent feedback here, and the awesome work on `rewrite-tutorials-link` 🤩 

I've refactored `detect-and-reformat-learn-url` to use `rewrite-tutorials-link` under the hood. This required some pretty minor changes to the remark plugin file.

This made me think about whether we should move the `rewriteTutorialsLink` to its own file in `lib/learn` or something... but I figure that's non-urgent, maybe slightly outside the scope of this PR, and something we can handle in 🎟️ [[chore] make lib/tutorials for shared use tutorial helpers](https://app.asana.com/0/1202097197789424/1202180048371484/f).

Let me know what you think 🙇

src/lib/learn-client/index.ts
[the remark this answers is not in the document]
zchsh: ✅ Update here: have reverted this and other changes that were previously in place to allow the tests for `detect-and-reformat-learn-url` to fetch from the Learn API. Now using `nock` instead 👍

src/views/product-tutorials-view/helpers/detect-and-reformat-learn-url.ts
zchsh: Great call - updated to aim for this - leaning on `getTutorialsMap` & `rewriteTutorialsLink`, with some light refactoring in the latter to expose a more generic handler than can be called outside of remark plugin use.

zchsh:
Minor note here that just threw me off when double-checking thins: the Vault link is a `cloud` / HCP Vault tutorial, so it's not a beta product, and therefore links back to `learn.hashicorp.com` despite all the formatting 🤷

src/lib/learn-client/types.ts
kendallstrautman: lol reading these function names just makes me think we _really_ need to refactor / address the product / theme weirdness with Learn 😆

src/lib/remark-plugins/rewrite-tutorial-links/index.ts
```diff
@@ -108,19 +116,19 @@ function handleRewriteTutorialsLink(node: Link | Definition) {
...
-        node.url = `${nodePath}/tutorials`
+        newUrl = `${nodePath}/tutorials`
       }
 
-      if (!node.url) {
+      if (!newUrl) {
         // If the link wasn't found in the map, default to original link
         // Could be a typo, its up to the author to correct -- this feedback should help
-        node.url = nodePath
+        newUrl = nodePath
```
kendallstrautman, line 131: Hmmm....we definitely want the default behavior to still return the url.  We may want to return the `newUrl` in the catch statement to retain this?

src/lib/remark-plugins/rewrite-tutorial-links/index.ts
```diff
@@ -66,32 +66,40 @@ export const rewriteTutorialLinksPlugin: Plugin = () => {
...
     // external non-beta product links don't need to be rewritten. i.e. learn.hashicorp.com/consul
     if (!isBetaProduct && !isExternalLearnLink && !isAnchorLink) {
       // If its an internal link, rewrite to an external learn link
-      node.url = new URL(node.url, 'https://learn.hashicorp.com/').toString()
+      newUrl = new URL(newUrl, 'https://learn.hashicorp.com/').toString()
     }
 
     if (isBetaProduct) {
-      let nodePath = node.url // the path to be formatted - assumes to be absolute as current Learn impl does
+      let nodePath = newUrl // the path to be formatted - assumes to be absolute as current Learn impl does
```
kendallstrautman, line 102:
```suggestion
      let nodePath = newUrl // the path to be formatted - assumes to be absolute as current Learn impl does
```
```suggestion
      let nodePath = url // the path to be formatted - assumes to be absolute as current Learn impl does
```
It may just be my brain, but it seems easier to use `url` since `newUrl` could change and I need to read the code to check what `newUrl` may be at this point in the code.

src/lib/remark-plugins/rewrite-tutorial-links/index.ts
```diff
@@ -66,32 +66,40 @@ export const rewriteTutorialLinksPlugin: Plugin = () => {
...
+export function rewriteTutorialsLink(
+  url: string,
+  tutorialMap: Record<string, string>
+): string {
+  let newUrl = url
   try {
     // return early if non tutorial or collection link
-    if (!learnLink.test(node.url) && !docsLink) {
-      return
+    if (!learnLink.test(newUrl) && !docsLink.test(newUrl)) {
```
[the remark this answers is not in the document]
zchsh: This makes sense! I'd gone with a find-replace `node.url` → `newUrl` cause I was nervous about making changes beyond that, but can totally see how this is easier to reason around 👍

src/views/product-tutorials-view/helpers/detect-and-reformat-learn-url.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+async function detectAndReformatLearnUrl(url: string): Promise<string> {
+  // Don't mess with absolute URLs
+  if (isAbsoluteUrl(url)) {
+    return url
+  }
+  // Determine from path parts whether this is a collection or tutorial URL
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [_beforeSlash, part1, part2] = url.split('/')
+  const isTutorialUrl = part1 === 'tutorials' && isThemeOrProduct(part2)
```
kendallstrautman, line 23: There's another case here which is the product hub page link (i.e. /waypoint --> /waypoint/tutorials). We don't need to add it here but maybe we could add a few examples in the comments at the top of the file for exactly which urls will be rewritten here as a paper trail for when we use this function again later on.

kendallstrautman (approved):
This is great. Just left a few minor suggestions 💯 

Thanks for tacking on this new function, i'm sure it will be very useful. Learn links are a bit of a rats nest as you've now seen! haha Thanks for refactoring to reuse some of that logic ✨ It'll be a good pressure test! 

I saw that vault cloud collection link as well. I'm sure that's something we could flag for the the content team to update since ideally it would link to a dev portal collection.

src/views/product-tutorials-view/helpers/detect-and-reformat-learn-url.ts
kendallstrautman: For context, `rewriteTutorialLinks` _will_ handle the product hub page link if its passed.

src/views/product-tutorials-view/helpers/detect-and-reformat-learn-url.ts
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Given a URL string,  detect whether it is a collection or tutorial URL
+ * in the format used on learn.hashicorp.com.
+ *
+ * If it is, then reformat the URL to work with dev-dot's URL structure.
+ * Otherwise, return the URL unmodified.
+ */
+async function detectAndReformatLearnUrl(url: string): Promise<string> {
+  // Don't mess with absolute URLs
+  if (isAbsoluteUrl(url)) {
```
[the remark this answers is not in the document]
zchsh: ✅ great call - i added some tests for absolute urls 👍

src/views/product-tutorials-view/helpers/detect-and-reformat-learn-url.ts
zchsh: ah yeah great call... i think i can quickly patch this up, I think it'd be like `isHubPageUrl = isThemeOrProduct(part1)`, and then we'd be able to pass that right to `rewriteTutorialLinks` 🙌 will give that a shot, and add a test case!

src/lib/remark-plugins/rewrite-tutorial-links/index.ts
zchsh: ☝️ +1 that was my understanding! the stuff in `try` tries to execute, might get `caught`, but regardless of error or not, everything after the `try-catch` is always executed 🤖

src/views/product-tutorials-view/helpers/detect-and-reformat-learn-url.ts
zchsh: ✅ awesome, i think the above worked - now handles hub urls as well! going to merge this in 👍

PR #381, opened 2022-05-03
Adds opt-in / opt-out analytics tracking

BrandonRomano:
Includes Analytics Tracking plans for both opting in & opting out of the DevDot experience.

This will enable us to track the lifecycle of users who have opted in for the beta, which will enable us to perform analysis of data that is sandwiched between these two events.

These events will be ephemerally added into Dev Portal, only for the length of our public beta.

**Note:** this is only the specification, and is not the implementation of the analytics.


kendallstrautman (approved):
LGTM!

PR #382, opened 2022-05-03
feat: revise product landing view

zchsh:
## ✅ "Ready for Review" checklist

- [x] The Vercel preview link has been added to this PR's description
- [x] The Asana task has been added to this PR's description
- [x] This PR's link has been added to the "📐 Pull Request" field on the Asana task
- [x] This Vercel preview link has been added to the "🔍 Preview" field on the Asana task
- [x] The description template has been filled in below

---

## 🔗 Relevant links

- [Preview link][/vault] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Implements the revised product landing views, namely [/waypoint][/waypoint] & [/vault][/vault].

## 🤷 Why

Because we have new designs we need to implement.

## 🛠️ How

- [initial assembly PR] cuts back the existing implementation to the bare minimum
- [initial assembly PR] stubs in new JSON content, with schema and validation
- #390 
- #391
- #399
- #400
- #401
- #398
- #403
- #402
- #404
- #405

## Waypoint

product-landing-waypoint

## Vault

product-landing-vault

Figma for reference: 
- [Vault landing in Figma](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Dev-Portal?node-id=8395%3A110638)
- [Waypoint landing in Figma](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Dev-Portal?node-id=8395%3A108028)

## 🧪 Testing

- [ ] Visit [/waypoint][/waypoint], ensure the page looks and functions as expected
- [ ] Visit [/vault][/vault], ensure the page looks and functions as expected

## 💭 Anything else?

This PR started off stubbing in the most basic version of content. It then evolved into an assembly branch, the intent being to avoid replacing the existing, looks-okay-but-not-up-to-date landing pages with broken-looking stubs.

[task]: https://app.asana.com/0/1202097197789424/1202219063384066/f
[/waypoint]: https://dev-portal-git-zsstub-new-product-landing-hashicorp.vercel.app/waypoint
[/vault]: https://dev-portal-git-zsstub-new-product-landing-hashicorp.vercel.app/vault

PR #377, opened 2022-05-02
fix: tutorial landings table-of-contents 

zchsh:
## ✅ "Ready for Review" checklist

- [x] The Vercel preview link has been added to this PR's description
- [x] The Asana task has been added to this PR's description
- [x] This PR's link has been added to the "📐 Pull Request" field on the Asana task
- [x] This Vercel preview link has been added to the "🔍 Preview" field on the Asana task
- [x] The description template has been filled in below

---

## 🔗 Relevant links

- [Preview link][/vault/tutorials] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Fixes an issue on product tutorial landing views, such as [/waypoint/tutorials][/waypoint/tutorials] and [/vault/tutorials][/vault/tutorials], where headings were not reflected correctly in the sidecar table-of-contents due to lack of `scroll-margin-top`.

## 🤷 Why

To fix an issue on product tutorial landing views where the sidecar table-of-contents highlighting is not working as expected.

## 🛠️ How

- Adds a new global helper class, `.g-offset-scroll-margin-top`
- Uses `.g-offset-scroll-margin-top` for all heading elements on the product tutorial landing view
- Visually hides the top-level `"Overview"` page heading, to match mockups
    - Rendering this heading element seems necessary for our table-of-contents to work as expected
- Refactors some heading logic in `product-tutorials-view`
    - `BrandedCallout` does not actually need a heading, as it isn't part of the table-of-contents
    - When generating the table of contents from authored blocks, we're now more explicit about which blocks are intended for inclusion in the table of contents.

## 📸 Design Screenshots

Screenshot of Figma page layout:

design

Table of contents behaviour before this PR:

before

Table of contents behaviour after this PR:

after

## 🧪 Testing

- [ ] Visit [/waypoint/tutorials][/waypoint/tutorials] and [/vault/tutorials][/vault/tutorials], and ensure the sidecar is working as expected on both pages

## 💭 Anything else?

Nope!

[task]: https://app.asana.com/0/1202097197789424/1202205647216548/f
[/vault/tutorials]: https://dev-portal-git-zsfix-headings-tutorial-landing-hashicorp.vercel.app/vault/tutorials
[/waypoint/tutorials]: https://dev-portal-git-zsfix-headings-tutorial-landing-hashicorp.vercel.app/waypoint/tutorials

PR #390, opened 2022-05-04
feat: hook up product landing sidecar

zchsh:

---

## 🔗 Relevant links

- [Preview link][/waypoint] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Implements sidecar table-of-contents functionality for `views/product-landing`.

## 🤷 Why

To build out `views/product-landing`.

## 🛠️ How

- Adds a helper `transform-raw-content-to-prop`, which is hopefully clearly named.
    - This includes adding `headingSlug` elements where we need to include a heading in our table of contents.
    - This will later include fetching tutorial and collection data, and transforming it to card props
- Adds a helper `make-heading-slug-scope`, intent is to make it easy & clean to generate unique-on-the-page `headingSlug` values
- Adds all planned components as placeholders; this accounts for the majority of the lines-of-code
    - Adds prop types needed for the UI, which are referenced by `transform-raw-content-to-prop` (the collected component props types are the return type for that function!)
    - Some components & blocks (not all) render heading elements that get linked to from the table of contents

## 🧪 Testing

- [ ] Visit [/waypoint][/waypoint], and ensure the sidecar table-of-contents works as expected
- [ ] Visit [/vault][/vault], and ensure the sidecar table-of-contents works as expected

## 💭 Anything else?

Not at the moment! Excited to keep on truckin 🚂 

[task]: https://app.asana.com/0/1202097197789424/1202219063384067/f
[/waypoint]: https://dev-portal-git-zsproduct-landing-sidecar-hashicorp.vercel.app/waypoint
[/vault]: https://dev-portal-git-zsproduct-landing-sidecar-hashicorp.vercel.app/vault

### DIFF
--- a/src/views/product-landing/components/get-started-card/get-started-card.module.css
+++ b/src/views/product-landing/components/get-started-card/get-started-card.module.css
@@ -1,0 +1,9 @@
+.placeholder {
+  border: 1px solid magenta;
+  font-size: 12px;
+  white-space: pre-wrap;
+}
+
+.heading {
+  composes: g-offset-scroll-margin-top from global;
+}
--- a/src/views/product-landing/components/product-landing-blocks/index.tsx
+++ b/src/views/product-landing/components/product-landing-blocks/index.tsx
@@ -1,0 +1,39 @@
+/* eslint-disable react/no-array-index-key */
+import { ProductLandingBlocksProps, ProductLandingBlock } from './types'
+import { HeadingBlock, LinkedCards, TutorialCards } from './blocks'
+
+function ProductLandingBlocks({ blocks }: ProductLandingBlocksProps) {
+  return (
+    <>
+      {blocks.map((block: ProductLandingBlock, idx: number) => {
+        const { type } = block
+        switch (type) {
+          case 'heading':
+            return (
+              <HeadingBlock
+                key={idx}
+                heading={block.heading}
+                headingSlug={block.headingSlug}
+              />
+            )
+          case 'tutorial_cards':
+            return (
+              <TutorialCards key={idx} tutorialSlugs={block.tutorialSlugs} />
+            )
+          case 'linked_cards':
+            return <LinkedCards key={idx} cards={block.cards} />
+          default:
+            /**
+             * Note: any invalid block types are expected to have been caught
+             * by server-side validation against our content schema.
+             * At this point, we don't want to throw an error...
+             * so we render null for unrecognized block types.
+             */
+            return null
+        }
+      })}
+    </>
+  )
+}
+
+export default ProductLandingBlocks
--- a/src/views/product-landing/helpers/__tests__/make-heading-slug-scope.test.ts
+++ b/src/views/product-landing/helpers/__tests__/make-heading-slug-scope.test.ts
@@ -1,0 +1,24 @@
+import { makeHeadingSlugScope } from '../'
+
+describe('makeHeadingSlugScope', () => {
+  test('returns already-unique headings unmodified', () => {
+    const makeHeadingSlug = makeHeadingSlugScope()
+    expect(makeHeadingSlug('Foo Heading')).toBe('foo-heading')
+    expect(makeHeadingSlug('Bar Heading')).toBe('bar-heading')
+    expect(makeHeadingSlug('Baz Heading')).toBe('baz-heading')
+  })
+
+  test('appends duplicate headings with an index', () => {
+    const makeHeadingSlug = makeHeadingSlugScope()
+    expect(makeHeadingSlug('Foo Heading')).toBe('foo-heading')
+    expect(makeHeadingSlug('Foo Heading')).toBe('foo-heading-2')
+    expect(makeHeadingSlug('Foo headinG')).toBe('foo-heading-3')
+  })
+
+  test('detects duplicates within specific scopes', () => {
+    const makeHeadingSlugScope1 = makeHeadingSlugScope()
+    const makeHeadingSlugScope2 = makeHeadingSlugScope()
+    expect(makeHeadingSlugScope1('Foo Heading')).toBe('foo-heading')
+    expect(makeHeadingSlugScope2('Foo Heading')).toBe('foo-heading')
+  })
+})
--- a/src/views/product-landing/helpers/extract-headings.ts
+++ b/src/views/product-landing/helpers/extract-headings.ts
@@ -2,6 +2,16 @@ import { TableOfContentsHeading } from 'layouts/sidebar-sidecar/components/table
 import { ProductLandingViewProps } from '../types'
 import { ProductLandingBlock } from '../components/product-landing-blocks/types'
 
+/**
+ * Extracts headings from product-landing page content .
+ *
+ * Note: could potentially use `traverse` here to make this more generic.
+ * We could target any & all objects that have { heading, headingSlug }
+ * However, right now that felt more opaque and cryptic than being more
+ * explicit around which headings are being used and in what order.
+ * As well, I think a generic implementation should better account for "level".
+ * So, deferred creating a generic implementation, for now.
+ */
 export function extractHeadings(
   content: ProductLandingViewProps['content']
 ): TableOfContentsHeading[] {
--- a/src/views/product-landing/helpers/extract-headings.ts
+++ b/src/views/product-landing/helpers/extract-headings.ts
@@ -1,0 +1,47 @@
+import { TableOfContentsHeading } from 'layouts/sidebar-sidecar/components/table-of-contents'
+import { ProductLandingViewProps } from '../types'
+import { ProductLandingBlock } from '../components/product-landing-blocks/types'
+
+export function extractHeadings(
+  content: ProductLandingViewProps['content']
+): TableOfContentsHeading[] {
+  /**
+   * TODO: content is raw placeholder for now
+   * - Need to add "headings" that aren't placeholders.
+   *   These will be derived from content blocks
+   *   (mainly heading blocks, but also others, based on design intent)
+   * - Need to augment things like tutorialSlugs to fill in that data
+   *   Will close: https://app.asana.com/0/1201010428539925/1201654639085764/f
+   * - Need to build out the client-side components, whose props APIs will
+   *   inform what content structure we'll want to return for use on the client.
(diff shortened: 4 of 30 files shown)


src/views/product-landing/components/get-started-card/get-started-card.module.css
```diff
@@ -0,0 +1,9 @@
+.placeholder {
```
zchsh, line 1: I've added a _bunch_ of pretty identical placeholder components.

Intent here is mainly to have a clear picture around prop types for UI components, so that the logic in `transform-raw-content-to-prop` lines up nicely. Secondary intent is to nicely tee things up for subsequent PRs which will implement the non-placeholder UI for each of these components.

src/views/product-landing/components/get-started-card/get-started-card.module.css
```diff
@@ -0,0 +1,9 @@
...
+  border: 1px solid magenta;
+  font-size: 12px;
+  white-space: pre-wrap;
+}
+
+.heading {
+  composes: g-offset-scroll-margin-top from global;
```
zchsh, line 8: Note that stubbing out placeholder components is also helpful for getting the table-of-contents to function correctly off the bat. Re-using `g-offset-scroll-margin-top` here, which was introduced in #377 .

src/views/product-landing/components/product-landing-blocks/index.tsx
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable react/no-array-index-key */
+import { ProductLandingBlocksProps, ProductLandingBlock } from './types'
+import { HeadingBlock, LinkedCards, TutorialCards } from './blocks'
+
+function ProductLandingBlocks({ blocks }: ProductLandingBlocksProps) {
+  return (
+    <>
+      {blocks.map((block: ProductLandingBlock, idx: number) => {
```
zchsh, line 8: TLDR; is this component maps over the processed block props and renders the appropriate components.

Note that I still need to add `collection_cards` here - will handle that in an upcoming PR which will also fetch proper data for tutorials & collections (rather than just slugs as we have currently).

src/views/product-landing/helpers/__tests__/make-heading-slug-scope.test.ts
```diff
@@ -0,0 +1,24 @@
+import { makeHeadingSlugScope } from '../'
+
+describe('makeHeadingSlugScope', () => {
```
zchsh, line 3: Set up some tests here to try to validate that `makeHeadingSlugScope` isn't too wild of an idea 😅

src/views/product-landing/helpers/extract-headings.ts
```diff
@@ -0,0 +1,34 @@
+import { TableOfContentsHeading } from 'layouts/sidebar-sidecar/components/table-of-contents'
+import { ProductLandingViewProps } from '../types'
+import { ProductLandingBlock } from '../components/product-landing-blocks/types'
+
+export function extractHeadings(
```
zchsh, line 15: Extracts headings from processed content. I'm treating the content processing in `transform-raw-content-to-prop` as the single source of truth for valid headings.

src/views/product-landing/helpers/make-heading-slug-scope.ts
```diff
@@ -0,0 +1,31 @@
+import slugify from 'slugify'
+
+/**
+ * Returns a function that generates slugs from heading strings,
+ * while ensuring those slugs are unique across calls of the returned function
+ * by suffixing with a `-{number}` where needed.
+ */
+export function makeHeadingSlugScope(): (string) => string {
```
zchsh, line 8: Pretty excited about this, feels kinda basic in retrospect but it seems to work well ☺️ I don't _think_ there are any significant pitfalls to this approach (I was like "wait does this really work?" so added tests) but very open to changing course if there are 🙇

src/views/product-landing/helpers/transform-raw-content-to-prop.ts
```diff
@@ -0,0 +1,60 @@
+import { ProductData } from 'types/products'
+// raw content types
+import { ProductLandingContent, ProductLandingContentBlock } from '../schema'
+// component prop utilities & types
+import { makeHeadingSlugScope } from './'
+import { ProductLandingViewProps } from '../types'
+import { ProductLandingBlock } from '../components/product-landing-blocks/types'
+
+export async function transformRawContentToProp(
```
zchsh, line 9: TLDR; takes raw content, makes it into exactly the props needed to render the view.

Intent here is to pre-emptively tackle the "passing more data than necessary" to the client issue that I so far have failed to fix in `Product Tutorials Landing`. This function will also handle fetching tutorial and collection data (in a subsequent PR). My hope is that parts of this may be useful as reference when we do go to optimize the `Product Tutorials Landing` server stuff 👍

src/views/product-landing/index.tsx
```diff
@@ -1,44 +1,38 @@
...
+        cta={overview.cta}
+        image={overview.image}
+      />
+      <GetStartedCard
+        heading={get_started.heading}
+        headingSlug={get_started.headingSlug}
+        body={get_started.body}
+        ctas={get_started.ctas}
+      />
+      <ProductLandingBlocks blocks={blocks} />
```
zchsh, line 34: Less stubby! Everything being rendered is of course still very stubby. I have mixed feelings about the prop drilly happening here - may be one of those rare exceptions where `{ ...spread }` would be fine, but I figured better to err on the side of clarity 👍

zchsh:
@kendallstrautman This PR implements the sidecar table-of-contents functionality for the "Product Landing" view.

It also stubs in a _ton_ of component files, which inflates the diff quite a bit. I've tried to be thorough in self-review, let me know if there's any other context I can add 🙇 

Note: This targets `zs.stub-new-product-landing` / #382 as an assembly branch.

src/views/product-landing/helpers/make-heading-slug-scope.ts
kendallstrautman: Great idea using the closure! 🚀

src/views/product-landing/helpers/make-heading-slug-scope.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  /**
+   * Generate a heading slug, but avoid potential duplicates.
+   * Duplicates are determined based on USED_SLUGS in this scope.
+   */
+  function makeHeadingSlug(heading: string): string {
+    let suffix = 1
+    const baseSlug = slugify(heading, { lower: true })
+    let headingSlug = baseSlug
+    while (USED_SLUGS.indexOf(headingSlug) !== -1) {
```
kendallstrautman, line 22: Do you want to loop here? Or just use an `if` statement? 🤔

src/views/product-landing/helpers/transform-raw-content-to-prop.ts
kendallstrautman: Love it

kendallstrautman (approved):
🚀

src/views/product-landing/helpers/make-heading-slug-scope.ts
zchsh: I went for a `while` loop for those ✨ rare ✨ cases where there are many multiples of the same heading (eg [this tutorial](https://learn.hashicorp.com/tutorials/terraform/azure-change?in=terraform/azure-get-started#apply-your-changes-1) has multiple `Apply your changes` headers).

In such cases I _think_ the logic is simpler with `while` - though perhaps not quite as performant, not sure though, seems like performance might actually be better with `while` as we can avoid regex & `parseInt` work that would potentially be required to handle things with `if` rather than looping.

That being said I may not have conceived of the cleanest `if` approach... I might be overthinking it though... Here's my brain dump:

<details>

<summary>Possible `if` implementation</summary>

As a concrete example, consider a case where there are three identical `Apply your changes` headings. This is my thought process through trying to do things without a `while` loop:

1.  for the first heading, the initial `baseSlug` is `apply-your-changes`
    - No problemo, no duplicates yet, the returned `headingSlug` is `apply-your-changes`
    - USED_SLUGS is now `[ "apply-your-changes" ]`
2. for the second heading, the initial `baseSlug` is `apply-your-changes` again
    - We get a conflict
    - In this specific "second duplicate" case we only have one previous duplicate, so we could filter for exact duplicates of the `baseSlug` only, and tack `-1` onto the conflicting heading
    - For the first duplicate, this doesn't feel too complicated, we end up with `apply-your-changes-1`. Yay.
    - USED_SLUGS is now `[ "apply-your-changes", "apply-your-changes-1" ]`
3. for third heading, initial `baseSlug` is `apply-your-changes` once again 
    - We get a conflict, `"apply-your-changes"` is already in `USED_SLUGS`
    - The "filter for exact duplicates and tack something on" strategy falls down here, we'd end up with eg `apply-your-changes-1` again which could be a duplicate (in this case, and any "third duplicate" case, it would be guaranteed to be a duplicate!)
    - We could re-check for duplicates after the tack-on, but then we're back in a `while` loop. To stick with a single `if` we need to rethink, I think.
    - I think we'd want to detect any suffix we've previously added in all `USED_SLUGS` - maybe with a regex like `-\d+$`. 
        - In rare cases this might catch authored headings that happen to have the same suffix - that's fine, in fact good & necessary I think
    - We'd have multiple matches (`[ "apply-your-changes", "apply-your-changes-1" ]`), and we'd have to figure out which has the highest index, I think - otherwise we might generate a suffixed slug that already exists in `USED_SLUGS`.
    - So we'd need to sort these matches for the `${baseSlug}-\d+$` to get the one with the highest suffixed integer, probably using `parseInt` (simple `sort` also seems possible, maybe faster? though might not catch weird edge cases eg if an authored heading slugs out to `something-02` or whatnot)
    - After a bit of filtering, `parseInt`-ing, and then an increment of the highest detected integer suffic, we'd end up with `apply-your-changes-2`
    - I think this approach would work for fourth, fifth, `...n` headings

Definitely feels possible now that I've thought through it, but IMO the relatively minor performance gain (and possibly only more performant in cases where there are _many_ duplicates) is possibly not worth the extra complexity.

</details>

PR #393, opened 2022-05-04
First pass at new /waypoint/docs

ashleemboyer:

---

## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambnew-waypoint-docs-landing-hashicorp.vercel.app/waypoint/docs) 🔎
- [Asana task](https://app.asana.com/0/1201147515801429/1202288203881897/f) 🎟️

## 🗒️ What/Why

- Renames `src/pages/waypoint/docs/[[...page]].tsx` to `src/pages/waypoint/docs/[...page].tsx` so that this component no longer catches the `/waypoint/docs` route.
- Adds `src/pages/waypoint/docs/content.json` which currently houses the page subtitle and marketing content blocks.
- Adds `src/pages/waypoint/docs/index.tsx` for rendering the new `/waypoint/docs` component.
  - `DocsView` is rendered after the hero section so that the MDX content section has all the same capabilities and styles as other docs pages.
- Moves `TruncateMaxLines` to `src/components` for wider reuse
- Adds `hideSearch` prop to `DocsView` so that search can be rendered in a different place in the markup if needed when rendering `DocsView`.

## 🛠️ How

Chatted a bit via Slack about the approach and also modeled the work of changing `src/pages/waypoint/docs/[[...page]].tsx` to `src/pages/waypoint/docs/[...page].tsx` and adding `src/pages/waypoint/docs/index.tsx` off of Zach's past work for adding these pages to .io sites.

## 📸 Design Screenshots

<details>
<summary>expand for screenshot</summary>
<a href="https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/?node-id=8395%3A113819">Link to Figma frame</a>

</details>

## 🧪 Testing

- [ ] Go to https://dev-portal-git-ambnew-waypoint-docs-landing-hashicorp.vercel.app/waypoint/docs
- [ ] Check that the page content (aside from what's noted below) matches the design 
- [ ] Go to https://dev-portal-git-ambnew-waypoint-docs-landing-hashicorp.vercel.app/waypoint/docs/index
- [ ] The page should 404
- [ ] Try navigating to more pages under /waypoint/docs
- [ ] There should be no issues navigating to those pages

## 💭 Anything else?

Items for follow-up PRs:

- Handle rendering additional headings from `marketingContentBlocks` in sidecar
- Handle setting the color of hero section `CardLink` icons (I think I'd like to add an `IconCardLink` component that handles this)
- Abstract a `RootDocsPathLandingView` (or similarly named) view component
- Abstract a `CardGrid` component to handle the `card-grid` block type (this'll help clean up the markup a bit)
- Add support for description text to the `card-grid` block type (will be needed in /vault/docs)


src/pages/waypoint/docs/[...page].tsx
```diff
@@ -0,0 +1,59 @@
...
+})
+
+/**
+ * Wrapper for `generatedGetStaticProps` that first checks if the current page
+ * is `/waypoint/docs/index`. If that is the current page, then the user will be
+ * sent to the 404 page. Otherwise, `generatedGetStaticProps` will be invoked as
+ * usual.
+ */
+export async function getStaticProps(context: GetStaticPropsContext) {
+  if (context.params.page[0] === 'index') {
```
Comment (zchsh) [inline src/pages/waypoint/docs/[...page].tsx:31]: ⛏️ Nitpick-y thought:

TLDR; I think you could drop this extra layer, and export `getStaticProps` unmodified.

IMO there's no need to include this specific case in the scope of the custom docs landing work. I believe the underlying issue here needs to be fixed instead, as detailed in 🎟️ [[docs-loader] /docs/some-category/index should 404, but does not](https://app.asana.com/0/1202097197789424/1202289240521898/f). I _think_ the only way anyone would get to this working-but-it-shouldn't URL is by directly typing it in; and we have the same problem on live sites (eg see https://www.waypointproject.io/docs/index, https://www.waypointproject.io/docs/intro/index, https://www.waypointproject.io/docs/upgrading/index, etc etc.)

That being said, I suppose there's no harm in including this - though I think we'll want to remember to remove it once we do resolve the above task, so would personally lean towards omitting it (to have to remember about it later! ☺️ ).

src/pages/waypoint/docs/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+            </li>
+          ))}
+        </ul>
+      </div>
+      <div className={s.waypointMDXWrapper}>
+        <DocsView mdxSource={mdxSource} hideSearch />
+      </div>
+      <div>
+        {pageContent.marketingContentBlocks.map((block) => {
+          if (block.type === 'card-grid') {
```
zchsh, line 92: 💭 Thought: I found including `heading` as _part of a block_ rather than _as a separate block_ made things harder at the "build the sidecar" stage. Definitely something I started noticing _after_ building the dot-io docs landing pages (IIRC they made the same call, but I think I somewhat regret it).

The "it was hard" example: In [product-tutorials-view](https://github.com/hashicorp/dev-portal/blob/2428520b75ce77861a96d906d2329409b0364c4b/src/views/product-tutorials-view/components/product-view-content/index.tsx#L26), I found the existing block structure to be harder to work with. The idea that each block can contain nested elements felt like it made things particularly hard; but even without that, having headings _not_ be their own separate blocks felt like it made content less flexible than we really want it, and also made it more complicated to extract headings for the sidecar.

The "it felt better" example: in [product-landing-view](https://github.com/hashicorp/dev-portal/blob/2428520b75ce77861a96d906d2329409b0364c4b/src/views/product-landing/components/product-landing-blocks/index.tsx#L14), the flat array of authorable blocks make it more intuitive to write flexibly grouped content (eg multiple sets of cards "under" one heading), and it felt easier to reason about (IMO) when extracting headings, and when mapping over and dealing with CSS spacing between elements on the page.

☝️ Relatively hot take here, but wanted to note in case in sparks anything! Would mean some interesting changes, eg the concept of a "add support for description text" to the card-grid block (as mentioned in a comment below) would be re-conceptualized as supporting three simpler blocks: heading, text, and card grid. Take that with a big grain of salt, I might be off here, and I think the block structure already here in this PR is great either way 👍

src/pages/waypoint/docs/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+        <div className={s.pageHeroText}>
+          <IconTileLogo productSlug="waypoint" />
+          <div>
+            <Heading className={s.pageTitle} level={1} size={500} weight="bold">
+              Waypoint Documentation
+            </Heading>
+            <Text className={s.pageSubtitle}>{pageSubtitle}</Text>
+          </div>
+        </div>
+        <ul className={s.heroCardList}>
```
zchsh, line 69: There's [something similar in `product-landing`](https://github.com/hashicorp/dev-portal/blob/2428520b75ce77861a96d906d2329409b0364c4b/src/views/product-landing/components/icon-cards/index.tsx#L10) that uses `components/cards-grid-list`. Might be ripe for elevation & re-use here! (I probably should have lifted the `LinkedIconCard` into `components` initially 😅 )

zchsh (approved):
🙌 This looks great to me! 

Some breadcrumbs from previous work:
- `CardGrid` sounds a lot like [components/cards-grid-list](https://github.com/hashicorp/dev-portal/tree/main/src/components/cards-grid-list) - thought of this when I saw the PR description mention `CardGrid`, but then noticed you're already using it for the `card-grid` block which seems great! We might also be able to use for the "icon cards" section
- Related to icon cards: there's some quick [`IconCards` work in `views/product-landing`](https://github.com/hashicorp/dev-portal/tree/main/src/views/product-landing/components/icon-cards). Feel free to grab that if it's helpful when abstracting; or also feel free to ignore if it doesn't feel right (would be happy to swap it out if a new `components/icon-card-link` is built out as part of this work)

src/pages/waypoint/docs/[...page].tsx
Comment (ashleemboyer) [reply src/pages/waypoint/docs/[...page].tsx:31]: This sounds totally reasonable to me! Especially considering that it's an existing issue. I'll remove this in my next PR. 😊

src/pages/waypoint/docs/index.tsx
ashleemboyer: Agreed on this! It's partially why I decided to defer the Sidecar for a follow-up PR. I'm not 100% convinced either way just yet, but my mentality so far with having this block that has `title` and (upcoming) `description` properties is so that the responsibility of knowing some semantic HTML and heading element hierarchy isn't on the people that edit this page. So my goal thus far has been to enable flexibility in the content rendered but not so much flexibility that someone could accidentally break the accessibility of the page.

The main tradeoff IMO is increased complexity in the developer experience with (hopefully!) decreased complexity in the authoring experience. This is probably going to get more complicated when I work on /vault/docs next, because its content is less "flat" than Waypoint's currently. The `card-grid` in Waypoint's renders an `<h2>` heading because that's the appropriate heading given the MDX content that comes before it. On Vault however, the headings will need to be level 3 because they will be under the "Use Cases" heading which will be a level 2 heading.

<details>
<summary>Expand for /vault/docs design screenshot</summary>

</details>

But anyways! I'm going to see how implementing the Vault page goes, and if it doesn't go quickly enough I will probably compromise on these attempts in the short term and we can spend appropriate time on making the authoring experience more robust later. Totally appreciate your thoughts here and I'm glad to know we're pretty much on the same page!

src/pages/waypoint/docs/index.tsx
ashleemboyer: Sweet!! I'll definitely leverage these things and I'll lift them to the higher folder as well. 😊

PR #394, opened 2022-05-04
feat: warm .io cache on production deploy

dstaley:

---

## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Send an HTTP request to the most visited .io pages (on the .io domain) in order to cause them to be cached.

## 🤷 Why

Vercel caches statically generated pages on a domain basis, meaning .io pages that are built when the site is generated aren't cached under their respective .io domains.

## 🛠️ How

Adds a GitHub Actions workflow that's triggered when a production Vercel deploy occurs that requests the top 200 pages for each product (according to Fathom and the Marketing Content API).

## 🧪 Testing

- [ ] Clone locally
- [ ] `npm i`
- [ ] `npx --package @hashicorp/platform-tools@0.5.0 hc-tools ./scripts/warm-dotio-cache.ts`
- [ ] Ensure command executes successfully


.github/workflows/cache.yml
```diff
@@ -0,0 +1,18 @@
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version-file: '.nvmrc'
+      - name: Install npm@8
+        run: npm install --global npm@8
+      - name: Install dependencies
+        run: npm ci
+      - name: Cache .io sites
+        run: npx --package @hashicorp/platform-tools@0.5.0 hc-tools ./scripts/warm-dotio-cache.ts
```
BRKalow, line 18: nit: `@hashicorp/platform-tools` should be installed from the `npm ci` call above, so I believe we could do:
```suggestion
        run: npx hc-tools ./scripts/warm-dotio-cache.ts
```

build-libs/proxy-config.js
```diff
@@ -34,7 +34,7 @@ const proxyConfig = {
     ],
   },
   packer: {
-    domain: 'https:/www.packer.io',
+    domain: 'https://www.packer.io',
```
BRKalow, line 37: 👀 😬

scripts/warm-dotio-cache.ts
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Returns a string representing a date that's `daysAgo` in the past.
+ * Format: YYYY-MM-DD HH:MM:SS
+ */
+function getDateRange(daysAgo: number): string {
+  const d = new Date()
+  d.setTime(d.getTime() - 24 * 60 * 60 * 1000 * daysAgo)
+  return d.toISOString().substring(0, 19).split('T').join(' ')
+}
```
BRKalow, line 30: How stable is this? It would be nice if we could make the timestamp generic so we can leverage caching on the `static_paths` endpoint.

dstaley: We could omit the time portion, which would basically give us a stable value for 24 hours. So every request today would have resulted in `2022-04-06 00:00:00`

```js
`${d.toISOString().substring(0, 10)} 00:00:00`
```

BRKalow: That seems like a good approach to me!

BRKalow (approved):
This is awesome! Very excited for this. This should be a great stop-gap up to the full dev portal migration.

dstaley:
alrighty, let's deploy and see what happens :eyes:

PR #425, opened 2022-05-12
fix: footer overflow issues

zchsh:

---

## 🔗 Relevant links

- [Preview link][preview-sidebar-sidecar] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Fixes an issue where the top border of our `footer` component caused horizontal overflow and scroll in both our core layout, and in our sidebar-sidecar layout.

## 🤷 Why

Horizontal scroll is unintentional, unexpected to visitors, and quite janky.

#### Context on design needs

In our designs, the footer needs to be aligned with main content, but its top border needs to span all the way across the main content and sidecar area:

CleanShot 2022-05-13 at 12 13 46@2x

#### Context on current situation

Various approaches were attempted without changing markup, and I hit dead ends in all of them. In brief, I think the fundamental challenge with the current implementation's markup is that `overflow: hidden` and `position: sticky` [don't work together](https://css-tricks.com/dealing-with-overflow-and-position-sticky/):

CleanShot 2022-05-12 at 16 54 11@2x

#### Proposed solution, implemented in this PR

To work around this, markup as well as styles have been changed. A new `.mainAreaAndFooter` element creates a flex "column" layout, with two rows:

- the first row, `.mainArea`, contains the main area and handles the sidecar stickiness. It cannot have `overflow: hidden` because this would mess with sidecar's `position: sticky`.
- the second row, `.footerArea`, handles the footer positioning and overflow issues. It contains the footer and a `footerSpaceForSidecar` element. This area _must_ have `overflow: hidden` to prevent footer top border from overflowing

2022-05-12-sidebar-sidecar-footer-overflow-fix-summary

This requires lifting some CSS custom properties in order to share things like the sidecar width and the main content area max-width between the now-separate `.mainArea` and `.footerArea`. But, it addresses the overflow issue without breaking the `sticky` behaviour of the sidecar.

### Fix for base layout (as on home page, for example)

- Sets `overflow: hidden` on the footer container
- This was the easy one, very much a footnote compared to the sidebar-sidecar fix

## 🧪 Testing

- [ ] Visit the [home page][preview-home], and ensure there is no horizontal overflow. As well, ensure the footer looks as expected.
- [ ] Visit a page with `sidebar-sidecar-layout`, such as [/waypoint/tutorials][preview-sidebar-sidecar]
    - [ ] On both desktop and mobile sized viewports...
        - [ ] Ensure there is no horizontal overflow
        - [ ] Ensure the footer looks and functions as expected
    - [ ] On desktop viewports...
        - [ ] Ensure the left boundary of items in the top-level navigation bar aligns with the left boundary of the content area within the layout
        - [ ] Ensure the footer's left and right boundaries align with the content area's left and right boundaries
        - [ ] Ensure the main content area and sidecar still look and function as expected, with a focus on correct alignment

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/1202097197789424/1202271065589112/f
[preview-home]: https://dev-portal-git-zsfix-footer-overflow-hashicorp.vercel.app
[preview-sidebar-sidecar]: https://dev-portal-git-zsfix-footer-overflow-hashicorp.vercel.app/waypoint/tutorials

### DIFF
--- a/src/components/navigation-header/navigation-header.module.css
+++ b/src/components/navigation-header/navigation-header.module.css
@@ -69,7 +69,7 @@
   display: flex;
   justify-content: left;
   width: calc(
-    var(--dev-dot-sidebar-width) + 2 * var(--main-area-padding-left-right) -
+    var(--dev-dot-sidebar-width) + var(--main-area-padding-left) -
       var(--header-padding-left-right) -
       var(--header-dropdown-menu-padding-left-right)
   );
--- a/src/layouts/base-new/base-new-layout.module.css
+++ b/src/layouts/base-new/base-new-layout.module.css
@@ -54,6 +54,7 @@ for both --navigation-header-height and --alert-bar-height.
 
 .footer {
   margin-top: auto;
+  overflow: hidden;
   padding-left: 24px;
   padding-right: 24px;
 }
--- a/src/layouts/core-dev-dot-layout/core-dev-dot-layout.module.css
+++ b/src/layouts/core-dev-dot-layout/core-dev-dot-layout.module.css
@@ -1,7 +1,13 @@
 .root {
   /* Widths/spacing needed in multiple stylesheets */
   --dev-dot-sidebar-width: 320px;
-  --main-area-padding-left-right: 24px;
+
+  /* Note: "right" value is not used in multiple stylesheets,
+     so is declared in sidebar-sidecar-layout.module.css */
+  --main-area-padding-left: 24px;
+  @media (--dev-dot-desktop) {
+    --main-area-padding-left: 48px;
+  }
 
   /* Mobile and tablet breakpoints */
   --mobile-width-breakpoint: 728px;
--- a/src/layouts/sidebar-sidecar/index.tsx
+++ b/src/layouts/sidebar-sidecar/index.tsx
@@ -93,9 +93,9 @@ const SidebarSidecarLayoutContent = ({
         >
           <SidebarContent />
         </motion.div>
-        <div className={s.mainArea}>
-          <div className={s.main}>
-            <main id="main">
+        <div className={s.mainAreaAndFooter}>
+          <div className={s.mainArea}>
+            <main id="main" className={s.mainContent}>
               {breadcrumbLinks && <BreadcrumbBar links={breadcrumbLinks} />}
               {children}
               {githubFileUrl && (
@@ -106,13 +106,16 @@ const SidebarSidecarLayoutContent = ({
                 />
               )}
             </main>
+            <div className={s.sidecar}>
+              <SidecarContent />
+            </div>
+          </div>
+          <div className={s.footerArea}>
             <Footer
               className={s.footer}
               openConsentManager={openConsentManager}
             />
-          </div>
-          <div className={`${s.sidecar} g-hide-on-mobile g-hide-on-tablet`}>
-            <SidecarContent />
+            <div className={s.footerSpaceForSidecar}></div>
           </div>
         </div>
       </div>
--- a/src/layouts/sidebar-sidecar/index.tsx
+++ b/src/layouts/sidebar-sidecar/index.tsx
@@ -95,7 +95,7 @@ const SidebarSidecarLayoutContent = ({
         </motion.div>
         <div className={s.mainAreaAndFooter}>
           <div className={s.mainArea}>
-            <main id="main" className={s.mainContent}>
+            <main id="main" className={s.main}>
               {breadcrumbLinks && <BreadcrumbBar links={breadcrumbLinks} />}
               {children}
               {githubFileUrl && (
--- a/src/layouts/sidebar-sidecar/index.tsx
+++ b/src/layouts/sidebar-sidecar/index.tsx
@@ -115,7 +115,6 @@ const SidebarSidecarLayoutContent = ({
               className={s.footer}
               openConsentManager={openConsentManager}
             />
-            <div className={s.footerSpaceForSidecar}></div>
           </div>
         </div>
       </div>
--- a/src/layouts/sidebar-sidecar/sidebar-sidecar-layout.module.css
+++ b/src/layouts/sidebar-sidecar/sidebar-sidecar-layout.module.css
@@ -107,7 +107,7 @@ for further details.
   padding-right: var(--main-area-padding-right);
 }
 
-.mainContent {
+.main {
   align-self: flex-start;
   flex-grow: 1;
   min-width: 0;
--- a/src/layouts/sidebar-sidecar/sidebar-sidecar-layout.module.css
+++ b/src/layouts/sidebar-sidecar/sidebar-sidecar-layout.module.css
@@ -130,7 +130,6 @@ for further details.
 
   @media (--dev-dot-desktop) {
     display: block;
-    border: 1px solid magenta;
     align-self: flex-start;
     flex-shrink: 0;
     margin-left: var(--sidecar-margin-left);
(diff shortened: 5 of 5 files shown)


src/components/navigation-header/navigation-header.module.css
```diff
@@ -69,7 +69,7 @@
   display: flex;
   justify-content: left;
   width: calc(
-    var(--dev-dot-sidebar-width) + 2 * var(--main-area-padding-left-right) -
+    var(--dev-dot-sidebar-width) + var(--main-area-padding-left) -
```
zchsh, line 72: This maintains alignment of the nav header items on desktop with the main content area.

CleanShot 2022-05-12 at 16 28 25@2x

Note on the previous approach for context:
- `--main-area-padding-left` was always set to `24px`
- on mobile & tablet, `.main` had `0` padding-left, while `.mainArea` had `--main-area-padding-left`, resulting in a net padding-left of `24px`
- on desktop, _where this rule is relevant for alignment_, `.main` and `.mainArea` _both_ had padding-left set to `--main-area-padding-left`. This resulted in total padding-left of `48px`, creating the need to double the value here to achieve correct alignment.

In this PR, I've lightly refactored the use of `padding-left` to try to simplify the model:
- `--main-area-padding-left` is set to `24px` on mobile & tablet, and changes to `48px` on desktop
- so this rule can use the value directly rather than doubling it

src/layouts/base-new/base-new-layout.module.css
```diff
@@ -54,6 +54,7 @@ for both --navigation-header-height and --alert-bar-height.
 
 .footer {
   margin-top: auto;
+  overflow: hidden;
```
zchsh, line 57: This is the one-line fix for footer overflow issues on the home page.

src/layouts/core-dev-dot-layout/core-dev-dot-layout.module.css
```diff
@@ -1,7 +1,13 @@
 .root {
   /* Widths/spacing needed in multiple stylesheets */
   --dev-dot-sidebar-width: 320px;
-  --main-area-padding-left-right: 24px;
+
+  /* Note: "right" value is not used in multiple stylesheets,
+     so is declared in sidebar-sidecar-layout.module.css */
+  --main-area-padding-left: 24px;
```
zchsh, line 7: As noted in this comment, `--main-area-padding-left` is shared across components, but there's no need to share `--main-area-padding-right`, so I have kept it local to `sidebar-sidecar-layout`.

src/layouts/sidebar-sidecar/index.tsx
```diff
@@ -93,9 +93,9 @@ const SidebarSidecarLayoutContent = ({
         >
           <SidebarContent />
         </motion.div>
-        <div className={s.mainArea}>
-          <div className={s.main}>
-            <main id="main">
+        <div className={s.mainAreaAndFooter}>
```
zchsh, line 96: This `.mainAreaAndFooter` element creates a flex "column" layout, with two rows:
- first row: main area + sidecar, _cannot_ have `overflow: hidden` because this messes with `sidecar`'s `position: sticky`
- second row: footer + footerSpaceForSidecar, _must_ have `overflow: hidden` to prevent footer top border from overflowing

src/layouts/sidebar-sidecar/index.tsx
```diff
@@ -93,9 +93,9 @@ const SidebarSidecarLayoutContent = ({
...
+          <div className={s.mainArea}>
+            <main id="main" className={s.main}>
```
zchsh, line 98: I've avoided creating an extra wrapper div by moving the `.main` className to the `<main>` element.

src/layouts/sidebar-sidecar/sidebar-sidecar-layout.module.css
```diff
@@ -18,8 +18,44 @@ for further details.
...
+  which serves as the border, and which overflows the footer on both sides.
+  
+  The footer container ensures this border overflow does not cause scroll
+  using overflow hidden. We place an empty sidecar placeholder element
+  within this container, next to the footer, to ensure spacing is consistent.
+
+  This method requires two separate "rows" for the main content + sidecar area,
+  and for the footer area. Having to share values between them is not ideal.
+  But, it sidesteps issues in mixing overflow: hidden with position: sticky.
+  Details: https://css-tricks.com/dealing-with-overflow-and-position-sticky/
```
zchsh, line 48: As in other "weird" work, I've tried to be extra explicit about _why_ this approach was taken, with the main intent here being that with this documentation, someone in the future may be able to come up with a more elegant solution.

src/layouts/sidebar-sidecar/sidebar-sidecar-layout.module.css
```diff
@@ -50,47 +86,82 @@ for further details.
...
-  @media (--dev-dot-tablet-down) {
-    padding-left: 0;
-    padding-right: 0;
-    max-width: none;
+  /* 50vh padding-bottom ensures that all
+  jump-nav items have room to be jumped to */
+  padding-bottom: 50vh;
+
+  @media (--dev-dot-desktop) {
+    max-width: var(--main-area-max-width);
```
zchsh, line 120: I've inverted the media queries here to be mobile-first.

src/layouts/sidebar-sidecar/sidebar-sidecar-layout.module.css
```diff
@@ -50,47 +86,82 @@ for further details.
...
+  padding-right: var(--main-area-padding-right);
 }
 
 .main {
   align-self: flex-start;
   flex-grow: 1;
   min-width: 0;
-  max-width: calc(780px + (var(--main-area-padding-left-right) * 2));
-  padding-left: var(--main-area-padding-left-right);
-  padding-right: var(--main-area-padding-left-right);
```
zchsh, line 70: I've removed duplicate `padding` settings. The intent here is to make the left and right padding of the main content area easier to reason about; but it did require a few changes in related files.

BRKalow:
@zchsh Could we put a `border-top` on the footerArea instead of using the pseudo element?

zchsh:
@BRKalow I think we probably could - however I think we'd then need to remove the border from within `footer`, and declare an identical border in `base-new-layout`. The intent with the `::before` element was to be able to define footer's top-border once within `footer`, rather than having the consumer of the component define it... but can definitely see how that creates a compromise, in that consumers have to set `overflow: hidden` on whatever container they render `footer` into 🤔 

Let me know if that helps - definitely happy to adjust, & also add a top border to the home page, just wanted to make sure the compromise there was clear!

src/layouts/sidebar-sidecar/index.tsx
```diff
@@ -106,13 +106,16 @@ const SidebarSidecarLayoutContent = ({
...
+          </div>
+          <div className={s.footerArea}>
             <Footer
               className={s.footer}
               openConsentManager={openConsentManager}
             />
-          </div>
-          <div className={`${s.sidecar} g-hide-on-mobile g-hide-on-tablet`}>
-            <SidecarContent />
+            <div className={s.footerSpaceForSidecar}></div>
```
alexcarpenter, line 118: **Nitpick ⛏️:**  I think this fill could be done via CSS vs needing the extra empty element in the dom.

CleanShot 2022-05-13 at 12 30 31@2x

Applying the same styles to the `footerArea` after element and adding gap since were in a flex container.

```css
.footerArea {
  gap: var(--sidecar-margin-left);
  
  &::after {
    content: '';
    display: block;
    width: var(--sidecar-width);
    background: red;
  }
}
```

zchsh: ooh nice! ✅  updated, I dig this.

this made me notice we can use `gap` for spacing the main area with the actual sidecar as well, so updated that too 👍

alexcarpenter (approved):
👏🏼 nice fix on this one.

PR #429, opened 2022-05-13
fix: roll back subnanv

zchsh:
This PR fixes an issue where `@hashicorp/react-subnav`'s `overflow: hidden` means the mobile menu would not be visible.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -46,7 +46,7 @@
         "@hashicorp/react-search": "^6.4.0",
         "@hashicorp/react-section-header": "^5.0.4",
         "@hashicorp/react-sentinel-embedded": "^1.0.4",
-        "@hashicorp/react-subnav": "^9.3.5",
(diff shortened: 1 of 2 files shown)


BRKalow (approved):
Oh no!

zchsh:
@BRKalow Sadly the previous subnav change is actually not great, it needs to apply `overflow: hidden` more judiciously (only when _not_ showing mobile nav). Working on an upstream fix in `react-components` but this addresses the immediate live brokenness.

BRKalow:
Should we power merge?

alexcarpenter:
@zchsh @BRKalow was there any follow up to subnav for this issue? Installing the latest version includes the bug currently.

zchsh:
@alexcarpenter Ah dang there might not have been? I'll look into this now, thanks for the ping 💖

PR #430, opened 2022-05-13
feat: useScrollPercentageAnalytics

BRKalow:

---

## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1202114367927919/1202207336073929/f) 🎟️

## 🗒️ What

Implement hook to add scroll percentage tracking per our analytics spec.

## 🛠️ How

Went with a `scroll` event listener, with wrapping the dimension calculations in `requestAnimationFrame`. This should ensure that we aren't negatively impacting scrolling performance. We also remove the scroll listener once all of the events are fired for that page, so that's another perf gaurd.

## 🧪 Testing

- [ ] Open the network tab
- [ ] Navigate to any page
- [ ] Observe the calls made to segment's API with a `Page Scrolled` event


src/hooks/use-scroll-percentage-analytics.ts
```diff
@@ -0,0 +1,68 @@
+import { useEffect } from 'react'
+import { canTrackAnalytics } from 'lib/analytics'
+
+const SCROLL_PERCENTAGE_THRESHOLDS = [25, 50, 75, 90]
+
+// TODO: do we care about this potentially updating?
+const windowHeight = typeof window !== 'undefined' ? window.innerHeight : 0
```
BRKalow, line 7: It seems safe to only calculate this once, happy to adjust though if it feels like an edge case we should handle.

BrandonRomano: Yeah this is a good catch!  I feel like probably we should update this for accuracy.

BrandonRomano (commented):
Agree with your motivation to update the `windowHeight` if that changes, could we do that?

src/hooks/use-scroll-percentage-analytics.ts
BrandonRomano: If this ends up being a big lift, we don't *absolutely* need to do this; because I think it's a smaller % of the data (and  it will come out in the wash).  I am just concerned that I am not right about it being a smaller %, and users are resizing their browsers often (I feel like I do this).

BrandonRomano (approved):

src/hooks/use-scroll-percentage-analytics.ts
```diff
@@ -0,0 +1,70 @@
+import { useEffect } from 'react'
+import { canTrackAnalytics } from 'lib/analytics'
+
+const SCROLL_PERCENTAGE_THRESHOLDS = [25, 50, 75, 90]
+
+const getPercentageScrolled = (documentHeight, windowHeight) => {
+  const scrollOffset = window.scrollY + windowHeight
+
+  return Math.round(scrollOffset / documentHeight) * 100
+}
```
zchsh, line 10: Possibly related to percentages firing early: should `documentHeight` take `windowHeight` into account rather than scrollOffset?

For example, if window height is 1000, document height is 1200, scrollY of 20, expected 20 / 200 = 10% scrolled:

```js
const scrollOffset = window.scrollY + windowHeight // will be 1020
return Math.round(scrollOffset / documentHeight) * 100 // will be 1020 / 1200 = 0.85
```

I think this might be more correct:

```suggestion
const getPercentageScrolled = (documentHeight, windowHeight) => {
  const totalScrollableDistance = documentHeight - windowHeight
  const scrolledDistance = window.scrollY
  return Math.round(scrolledDistance / totalScrollableDistance) * 100
}
```

zchsh (changes requested):

src/hooks/use-scroll-percentage-analytics.ts
zchsh: ah right that makes total sense 🤦 i think the naming of the event threw me off - so if a visitor loads a 1200 pixel tall page in a 1000 pixel tall viewport, and hasn't scrolled at all, we'll report 75%. But that represents 75% of page content "seen" (or was visible), rather than 75% of the possible amount of scrolling they could do.

zchsh (approved):

PR #443, opened 2022-05-18
feat: toast component & learn opt-in welcome toast

zchsh:

---

## 🔗 Relevant links

- [Preview - `/waypoint/tutorials?optInFrom=learn`][preview-tutorials-waypoint] 🔎
- [Preview - `/vault/tutorials?optInFrom=learn`][preview-tutorials-vault] 🔎
- [Preview - `/swingset/components/toast`][preview-swingset] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Implements `components/toast`, and uses it to welcome visitors who have just opted into the beta.
## 🤷 Why

We want to surface a toast notification to visitors when they first opt-in to the beta.

## 🛠️ How

- Implements toast components with [react-hot-toast]
    - Implementation is very similar to [@hashicorp/react-toast], and also draws inspiration from toasts in [hashicorp/learn], both of which also use [react-hot-toast]
    - Aims to match [HDS Toast component][hds-toast] 
- Uses `components/toast` to implement an opt-in toast to welcome visitors to the beta

## 📸 Design Screenshots

CleanShot 2022-05-22 at 23 20 03@2x

toast-stickersheet

## 🧪 Testing

- Visit any tutorials page, such as [/waypoint/tutorials?optInFrom=learn][preview-tutorials-waypoint]
    - You should see a "Welcome" toast pop up
    - This toast should be announced with `aria-live="polite"`. The action buttons on the toast should not be announced.
- The behaviour above should be consistent across tutorial URLs with an `?optInFrom` parameter
    - Product tutorial page - other page is [/vault/tutorials?optInFrom=learn][preview-tutorials-vault]
    - Collection page - [/waypoint/tutorials/get-started-docker?optInFrom=learn](https://dev-portal-git-zstoast-components-hashicorp.vercel.app/waypoint/tutorials/get-started-docker?optInFrom=learn)
    - Tutorial page - for example [/waypoint/tutorials/get-started-docker/get-started-intro?optInFrom=learn](https://dev-portal-git-zstoast-components-hashicorp.vercel.app/waypoint/tutorials/get-started-docker/get-started-intro?optInFrom=learn)
- Visit [the Swingset page for Toast][preview-swingset] and confirm the component looks to spec
    - Some small elements, eg the `×`-close-button, were hard to track down in spec, so I matched what I could see at the [HDS Toast Component page][hds-toast]

## 💭 Anything else?

- Intent is to build in this PR what we need for toasts for June 20
    - Apart from the downloads view `code-block` toast in this PR, I think this is only opt-in toasts
- Intent is _not_  to fully match the [HDS Toast Component][hds-toast] spec, _yet_
    - For example, arbitrary content such as italics, bold, and inline links are not _yet_ supported in the `description` prop. Intent is to implement that functionality later, if we need it. (May be sooner rather than later, I did see some design iterations that use inline links within the opt-in toasts, but I believe those are not the most current).

[preview-tutorials-waypoint]: https://dev-portal-git-zstoast-components-hashicorp.vercel.app/waypoint/tutorials?optInFrom=learn
[preview-tutorials-vault]: https://dev-portal-git-zstoast-components-hashicorp.vercel.app/vault/tutorials?optInFrom=learn
[preview-downloads]: https://dev-portal-git-zstoast-components-hashicorp.vercel.app/waypoint/downloads
[preview-downloads-vault]: https://dev-portal-git-zstoast-components-hashicorp.vercel.app/vault/downloads
[preview-swingset]: https://dev-portal-git-zstoast-components-hashicorp.vercel.app/swingset/components/toast
[react-hot-toast]: https://react-hot-toast.com
[@hashicorp/react-toast]: https://github.com/hashicorp/react-components/tree/main/packages/toast
[hashicorp/learn]: https://github.com/hashicorp/learn
[hds-toast]: https://design-system-components-hashicorp.vercel.app/components/toast
[task]: https://app.asana.com/0/1201903760348480/1202114097062999/f


package.json
```diff
@@ -94,7 +94,7 @@
     "@hashicorp/react-button": "^6.0.3",
     "@hashicorp/react-call-to-action": "^4.1.1",
     "@hashicorp/react-checkbox-input": "^5.0.1",
-    "@hashicorp/react-code-block": "^4.5.0",
+    "@hashicorp/react-code-block": "5.1.0-canary-20220518102202",
```
zchsh, line 97: Note: this is a pre-release of https://github.com/hashicorp/react-components/pull/591. Intention is to merge the `react-components` PR first, get a stable release out, and then bump to stable here before merging.

src/components/toast/components/close-button/close-button.module.css
```diff
@@ -0,0 +1,49 @@
...
+    top: 0;
+    width: 100%;
+    z-index: -1;
+  }
+
+  &:hover {
+    /* Note: use of alpha-100 differs from HDS toast implementation
+       (https://github.com/hashicorp/design-system/blob/f1d2475d86acb769833504cfb13b3531b387ae7e/packages/components/app/styles/components/alert.scss#L149).
+       Have done this to avoid using a non-token value. May need to adjust. */
+    --background-color: var(--token-color-palette-alpha-100);
```
zchsh, line 41: See comment here, open to other approaches. Intent here is to try to get _something_ approximately right rendering; I poked in [the HDS Figma file for Alert & Toast](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/HDS-Product---Components?node-id=1377%3A11987) but didn't see a hover state (relatively quick look so I may be missing something). Given lack of clear spec this seems very much close enough IMO.

src/components/toast/components/close-button/types.ts
```diff
@@ -0,0 +1,14 @@
...
+   */
+  onClick?: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
+  /**
+   * Optional class to be added to the button element.
+   */
+  className?: string
+  /**
+   * A label that describes the buttons action. Applied as aria-label value.
+   */
+  ariaLabel: string
```
zchsh, line 13: Note that `CloseButton` as well as this types file is mostly cribbed from https://github.com/hashicorp/react-components/tree/main/packages/close-button - with minor style adjustments for dev-dot, and a simplified props API that cuts out what we have no concrete use case for (before June 20, at least).

src/pages/_app.tsx
```diff
@@ -60,6 +61,7 @@ export default function App({ Component, pageProps, layoutProps }) {
                   <Layout {...allLayoutProps} data={allLayoutProps}>
                     <Component {...pageProps} />
                   </Layout>
+                  <Toaster />
```
zchsh, line 64: We render `Toaster` once at the app level, it manages the toast 🍞

src/views/product-downloads-view/components/downloads-section/toast-on-copy.tsx
```diff
@@ -0,0 +1,40 @@
...
+ * Given a copySuccess state,
+ * as well as an OS and optional package manager name,
+ *
+ * Display a "successful copy" toast if the copySuccess state is true,
+ * or a "failed to copy" toast if copySuccess is false.
+ *
+ * If copySuccess is "null" (as it is in its initial state),
+ * this is fine, but it does not merit toast.
+ */
+export default function toastOnCopy(
```
zchsh, line 15: `toastOnCopy`, as its name promises, displays a toast notification when the `copySuccess` state changes to either `true` (successful copy!) or `false` (failed copy).

zchsh:
@BrandonRomano @ashleemboyer This PR implements toasts components, and demos them with the install view's "copy install command" case.

Some brief notes:
- This PR relies on an underlying change in `code-block`, in https://github.com/hashicorp/react-components/pull/591. Probably makes sense to review that around the same time as this PR since they're so related. (Def feel free to review that as well Ashlee; didn't explicitly tag ya since I know there's lots else going on!)
- I tried tracking down a spec for the exact copy to show in the "copy install command" toast, but no luck yet (I might not be looking hard enough). Going to follow up in the #proj-dev-portal channel.

src/components/toast/index.tsx
```diff
@@ -13,8 +20,10 @@ function toast({
   description,
   icon,
   title,
-}: Omit<ToastDisplayProps, 'onDismiss'>) {
-  return reactHotToast((t: Toast) => {
+  autoDismiss = true,
+}: Omit<ToastDisplayProps, 'onDismiss'> & toastOptions) {
+  const toastFn = autoDismiss ? reactHotToast : reactHotToast.loading
```
alexcarpenter, line 25: there also is the ability to pass `duration: Infinity` as a toast option to prevent it from auto dismissing. https://react-hot-toast.com/docs/toast#default-durations

Not sure if one way or the other is better. Went the duration route on the new consent manager setup so far. https://github.com/hashicorp/react-components/pull/590/files#diff-675b091c60a57fc2d30f112facaabff26bf5ddeaa0c067a5cbcccf01c6e4fa52R109

zchsh: @alexcarpenter Since you're here, any thoughts on `role="dialog"` and related dialog behaviour for toasts with interactive content? We're running into a related conundrum here, since we may actually want this "install command copied" toast to have a link to a getting started guide 🤔

alexcarpenter: Not sure how that'd get announced to the user since currently `role="status"` gets applied so that when it's rendered, the contents are read to the user. If that was updated to use `role="dialog"` I don't think it would be read to the user. Not sure the recommendation for interactive content within a toast outside of maybe not auto dismissing it?

https://user-images.githubusercontent.com/825855/169566127-f84ecbb1-86eb-4988-8972-ccde2c0eb6d4.mp4

alexcarpenter: Dialog role would interrupt the users flow vs role status does not effect flow. Since with a dialog, you'd move focus into the dialog on show.

zchsh: I think the concern with interactive toasts with `role="status"` is that the interactive content will be announced, but the interactivity will be difficult to access through keyboard navigation.

For example, imagine there's some button or action that, after receiving the toast notification, I want to activate. With the dismiss button I guess 🤷‍♂️ cause the notification will be auto-dismissed soon, and assistive tech users will receive the notification and the "Dismiss" action and hopefully ignore it. But imagine it's a critical action, or even just a link or "accept consent" button that I receive, and then want to get to and activate. How do I do that?

CleanShot 2022-05-20 at 13 12 46

On the react-hot-toast website, the toasts are rendered in the DOM before the footer, it seems. So it seems I'd have to furiously tab to the end of the page to interact with toast content. And once I'm done interacting with the toast, I'd have to tab back to my original position. Seems like a lot of tabbing!

With that in mind, it seems like interactive content in toasts maybe does merit `role="dialog"` and interruption 🤔 Otherwise, it seems really difficult to surface the interactive content to keyboard users.

Does that track? Definitely thinking out loud here, very open to input on this 🙇‍♂️

alexcarpenter: Yeah, I am tracking with your logic here. I am thinking it's not a one size fits all solution for this. Because I wouldn't assume clicking copy to clipboard button would auto focus me into the toast as a dialog even if there was a link in there to learn more within it. I am wondering if there is some alternative method to move to the toast via screen reader that we might not be aware of?

zchsh: ☝️+1 on "alternative method". My immediate thought was "well react-hot-toast said it's accessible, so i must be doing something wrong!" 🤣 

This is a copy of a similar msg in dev-dot-engineers, but I think there are broadly four use cases:

1. ✅ Interruptive, text-only toast - `role="alert"`. `react-hot-toast` does this fine.
2. ✅ Non-urgent, text-only toast - `role="status"`. `react-hot-toast` does this fine too.
3. ✅ ( ❓)  Interruptive, interactive "toast" - `role="dialog"`. `react-hot-toast` fails here on its own... but I think we can hack around a bit to basically make a dialog that looks like a toast and renders in the bottom corner
    - Have started on this here in this PR! Basically hacking around Reach dialog, to prevent it from rendering into a portal, but otherwise using it. Definitely very broken right now (I need to figure out focus management, and remove duplicate dialog roles), but maybe a useful path.
4. ❓ Non-interruptive, interactive "toast" - it feels unclear how to do this, there doesn't seem to be much blessed accessibility guidelines on this. But we clearly have use cases - consent-manager seems like one? Maybe my problem is I've been searching for like "interactive toast patterns", and instead should search for "non-interruptive dialog pattern" 🤔

zchsh:
@ashleemboyer I've revised this PR quite a bit to try to sidestep the interactivity issues a bit... TLDR; I don’t think the opt-in toasts strictly _require_ interactivity, even though they include it. More detailed rationale:

- One action is “Continue”, which seems the same as ignoring or dismissing the toast. Users with assistive technology, who cannot view and click the toast, will the toast announced with `aria-live="polite"`. It seems there will be little need for users of assistive tech to dismiss the toast, as after being announced, it is relatively unobtrusively located at the very end of the DOM. And will disappear on the next page navigation (`?optInFrom` won't be preserved in the URL).
- The other action is “Leave Beta”. This action also appears at the start of the `main` area. The toast already indicates that the “Leave Beta” action can be found “in the top right of the content area”. With some slight modification to the screen reader announced contents, to change that to the less visually biased “beginning of the main content area”, visitors using assistive technology should be able to access the “Leave Beta” action with few barriers, as we’ll announce that the action is at the start of the main content, and we have a “skip to main content” link at the start of the DOM.

This definitely doesn't feel ideal to me, but thinking it might be a decent stopgap for June 20. Let me know if this feels reasonable to you 🙇‍♂️

src/components/toast/components/close-button/close-button.module.css
ashleemboyer: This seems fine to me! Also reasonable to ask about this in #team-design-systems. 😊

src/components/toast/components/close-button/close-button.module.css
```diff
@@ -0,0 +1,49 @@
...
+  color: var(--color);
+  cursor: pointer;
+  isolation: isolate;
+  display: grid;
+  height: var(--size);
+  margin: 0;
+  padding: 0;
+  place-items: center;
+  position: relative;
+  width: var(--size);
```
ashleemboyer, line 22: 🎨 Can these be alphabetized?

src/components/toast/components/close-button/close-button.module.css
```diff
@@ -0,0 +1,49 @@
+.closeButton {
+  composes: g-focus-ring-from-box-shadow from global;
+
+  --background-color: transparent;
+  --border-color: transparent;
+  --color: var(--token-color-foreground-faint);
+  --size: 24px;
```
ashleemboyer, line 7: 🙌🏻 Great idea to build the larger size into the element rather than just the `::before` pseudoelement!

src/components/toast/components/close-button/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+  return (
+    <button
+      onClick={onClick}
+      className={classNames(s.closeButton, className)}
+      aria-label={ariaLabel}
+    >
+      <IconX16 />
+    </button>
+  )
+}
```
ashleemboyer, line 24: 🎨 Can all the props be alphabetized?

```suggestion
export default function CloseButton({
  ariaLabel,
  className,
  onClick,
}: CloseButtonProps) {
  return (
    <button
      aria-label={ariaLabel}
      className={classNames(s.closeButton, className)}
      onClick={onClick}
    >
      <IconX16 />
    </button>
  )
}
```

src/components/toast/components/close-button/types.ts
ashleemboyer, line 13: 🎨 Can these be alphabetized and also have spaces between each for more separation?

I have the same request for:

- `src/components/toast/components/toast-display/types.ts`
- `toastOptions` currently defined in `src/components/toast/index.tsx`

```suggestion
  /**
   * A label that describes the buttons action. Applied as aria-label value.
   */
  ariaLabel: string

  /**
   * Optional class to be added to the button element.
   */
  className?: string

  /**
   * A function that will be called when the button is clicked.
   */
  onClick?: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
```

src/components/toast/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+   * Intended to allow re-focusing of elements that trigger interactive toasts.
+   */
+  onDismissCallback?: () => void
+  /**
+   * Optionally auto-dismiss the toast during client-side navigation.
+   * Defaults to true. Set to false to persist toast across routes.
+   * For non-client-side navigation, toast is always dismissed.
+   */
+  dismissOnRouteChange?: boolean
+}
```
ashleemboyer, line 34: 🎨 Can this be moved to a new `src/components/toast/types.ts` file?

src/components/toast/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+   * Defaults to 4000. Set to "false" for show-till-dismissed toast.
+   */
+  autoDismiss?: number | boolean
+  /**
+   * Flag that there is interactive content in the toast.
+   * If the toast contains interactive content,
+   * then we treat it with `role="dialog"`.
+   * Note this defaults to `true` if `actions` are provided, `false` otherwise.
+   */
+  isInteractive?: boolean
```
ashleemboyer, line 22: 🚧 Looks like this can be removed since it's not used in `toast`!

src/components/toast/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+   */
+  dismissOnRouteChange?: boolean
+}
+
+/**
+ * Wraps our ToastDisplay component in react-hot-toast.
+ */
+function toast({
+  renderActions,
+  children,
```
ashleemboyer, line 41: 🚧  I don't think we should allow `children` yet. I think it provides too much ability to render something inaccessible.

src/components/toast/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+  autoDismiss = AUTO_DISMISS_DEFAULT,
+  dismissOnRouteChange = true,
+}: Omit<ToastDisplayProps, 'dismissSelf'> & toastOptions) {
+  // Determine the auto-dismiss duration
+  const duration =
+    autoDismiss == false
+      ? Infinity
+      : autoDismiss == true
+      ? AUTO_DISMISS_DEFAULT
+      : autoDismiss
```
ashleemboyer, line 56: 🚧 Can this be updated to be a series of `if`s? I'm having a hard time understanding this very quickly.

src/components/toast/components/toast-display/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+   * no title and no description (except when using custom content).
+   */
+  if (IS_DEV && !title && !description && !children) {
+    throw new Error(
+      `Toast must be provided either a "title" or a "description", or you must use "children" to render custom content in the Toast. Please ensure toast has a non-empty string for its "title" or "description", or provide custom "children" content to render.`
+    )
+  }
+
+  const DescriptionElem =
+    typeof description === 'string' ? 'p' : description ? 'div' : null
```
ashleemboyer, line 32: 🚧 I think we should only allow a `string` `description` for now, so it's not so easy to render something inaccessible.

src/components/toast/components/toast-display/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+        {children ? children : null}
+        {typeof renderActions == 'function' ? (
+          <div className={s.actions}>{renderActions({ dismissSelf })}</div>
+        ) : null}
+      </div>
+      <div className={s.closeArea}>
+        <CloseButton onClick={dismissSelf} ariaLabel="Dismiss notification" />
+      </div>
+    </div>
+  )
```
ashleemboyer, line 53: 🚧 What are your thoughts on updating this to use variables instead of ternaries with `null` falsy branches? I think it'll make the markup easier to parse.

Here's what I'm thinking (includes `children` for now, but as noted in another comment I'm not sure we want to allow this yet):

```tsx
  let iconElement
  if (icon) {
    iconElement = <div className={s.iconArea}>{icon}</div>
  }

  let titleElement
  if (title) {
    titleElement = <p className={s.title}>{title}</p>
  }

  let descriptionElement
  if (description) {
    descriptionElement = <p className={s.description}>{description}</p>
  }

  let actionsElement
  if (typeof renderActions == 'function') {
    actionsElement = <div className={s.actions}>{renderActions({ dismissSelf })}</div>
  }

  return (
    <div className={classNames(s.root, s[`color-${color}`])}>
      {iconElement}
      <div className={s.contentArea}>
        {titleElement}
        {descriptionElement}
        {children}
        {actionsElement}
      </div>
      <div className={s.closeArea}>
        <CloseButton onClick={dismissSelf} ariaLabel="Dismiss notification" />
      </div>
    </div>
  )
```

Not sure I love the `Element` suffixes, maybe `Slot` would be a better one?

src/components/opt-in-out/helpers/make-beta-welcome-toast.tsx
```diff
@@ -0,0 +1,43 @@
...
+        <Button
+          color="tertiary"
+          icon={<IconArrowUpRight16 />}
+          iconPosition="trailing"
+          onClick={handleOptOut}
+          size="small"
+          text="Leave beta"
+        />
+      </>
+    ),
```
ashleemboyer, line 40: 🙋🏻 What led you to the approach of making this a function rather than array of `ReactNode`s or a single `ReactNode` via a `Fragment`?

ashleemboyer (changes requested):

src/components/opt-in-out/helpers/make-beta-welcome-toast.tsx
zchsh: Great question! 💖 The intent here was to make the `dismissSelf` function, which is declared within the context of `components/toast`, available to the actions.

I'd had a single `ReactNode` as the prop type previously, but that seemed to limit the ability to make `dismissSelf` available to the `actions`.

I hadn't considered the array of `ReactNode` too much, it's not a pattern I've seen often! Seems neat though, might align with the layout intent of the slot here (which kinda does want an array of actions to render in a flex layout; rather than having to trust the consumer to set that up correctly).

src/components/opt-in-out/types.ts
```diff
@@ -6,6 +6,16 @@ export enum PlatformOptionTitles {
 
 export type OptInPlatformOption = keyof typeof PlatformOptionTitles
 
+/**
+ * Type guard to determine if a string is an OptInPlatformOption
+ */
+export function isOptInPlatformOption(
+  string: string
+): string is OptInPlatformOption {
```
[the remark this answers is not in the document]
zchsh: Broadly I think of this as a type guard thing and didn't know the specific name 😅 Looks like it's called a "type predicate", specifically for "user defined type guards" (ref: <https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates>).

Random related thread: in another PR I learned about another return type when doing some validation work - [asserts arg is Type](https://github.com/hashicorp/dev-portal/blob/b890a667bb4857d6b2fa5a5830358f2d295cd48e/src/views/product-landing/helpers/validate-against-schema.ts#L7), which is kinda cool cause it doesn't require an `if` branch when using the function (useful for validation, which throws an error; less useful in this case I think).

src/components/toast/components/toast-display/index.tsx
zchsh: This is a hard one for me! I definitely appreciate that ternaries don't feel great all the time... however, I think there's another side of the coin (very much an opinion), which is that I find the grouping of JSX elements into the return statement of a component makes it easier to see the elements _and_ structure of the component in one go:

CleanShot 2022-05-24 at 09 43 58@2x

For me, this is the "render the markup" area of the component, and I know I'll see all the `className` etc passed into this area.

We had [a Slack thread on this a while back](https://hashicorp.slack.com/archives/C02T80JTNS2/p1649872956699629), and I don't remember exactly where we landed (if we landed anywhere?), but I would really appreciate being able to use ternaries in JSX components in order to keep the JSX grouped. I think once I pare back the description functionality, they'll all be one-line ternaries as well. It seems hard to both keep the JSX grouped, and avoid ternaries completely... I suppose maybe the `RenderIf` helper component mentioned in that thread might be useful?

src/components/toast/index.tsx
zchsh: Sure thing! Will declare `duration` before the return 👍

src/components/toast/index.tsx
zchsh: ✅ 🔪 Removed, thanks for flagging this! 💖

src/components/toast/components/toast-display/index.tsx
zchsh: ✅ Have pared this back to only support a plain text string for `description`

zchsh:
@ashleemboyer I've removed interactivity from the toast, dropped support for `children`, and pared back `description` to support plain text only. I've left the `actions` prop in even though it isn't used (yet!), but happy to remove if we feel strongly that it shouldn't be included.

I've also done my best to address the formatting feedback. The one thing I haven't switched up is the ternary conditional renders - I think establishing a strict practice for avoiding ternaries in the JSX render-slash-return portion of a componet is definitely worthy of consideration, but feels outside the scope of this PR.

src/components/toast/index.tsx
```diff
@@ -0,0 +1,76 @@
+import { useRouter } from 'next/router'
+import { useCallback, useEffect } from 'react'
+import { default as reactHotToast, Toast } from 'react-hot-toast'
+import Toaster from './components/toaster'
+import ToastDisplay from './components/toast-display'
+import { ToastDisplayProps, ToastColor } from './components/toast-display/types'
+import { ToastOptions } from './types'
+
+const AUTO_DISMISS_DEFAULT = 4000
```
kendallstrautman, line 9: Is this in seconds?

src/components/toast/components/toaster/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+        inset: 24,
+      }}
+      toastOptions={{
+        style: {
+          margin: 0,
+          padding: 0,
+          maxWidth: 404,
+          width: '100%',
+          backgroundColor: 'transparent',
+          boxShadow: 'none',
```
kendallstrautman, line 26: It looks like in the hds spec there is a box-shadow 🤔 
Screen Shot 2022-05-24 at 9 32 34 AM

src/components/opt-in-out/types.ts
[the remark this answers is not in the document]
ashleemboyer: Wow! These are snazzy. Thank you for sharing the links! 🥰

src/components/opt-in-out/helpers/make-beta-welcome-toast.tsx
ashleemboyer: This makes sense! Thanks for the extra context. 😊

PR #446, opened 2022-05-18
chore: adds beta opt in analytics event

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksopt-in-analytics-event-hashicorp.vercel.app/waypoint/tutorials) 🔎
- [Asana task](https://app.asana.com/0/1199634971449915/1202207336073938) 🎟️

## 🗒️ What

This PR adds a new `useOptInAnalyticsTracking` hook, which can be called per view to handle the firing of the opt in event. The hook checks for the presence of the platform opt-in cookie and fires the event if it hasn't already been done. Once the event is fired, a new cookie is created to note that this opt-in event was tracked.

### DIFF
--- a/src/layouts/base-new/index.tsx
+++ b/src/layouts/base-new/index.tsx
@@ -3,16 +3,17 @@ import createConsentManager from '@hashicorp/react-consent-manager/loader'
 import DevAlertBanner from 'components/dev-alert-banner'
 import Footer from 'components/footer'
 import NavigationHeader from 'components/navigation-header'
-import { OptInPlatformOption } from 'components/opt-in-out/types'
-import { handleOptInAnalytics } from 'hooks/use-opt-in-analytics-tracking'
 import CoreDevDotLayout from 'layouts/core-dev-dot-layout'
-import { useRouter } from 'next/router'
 import s from './base-new-layout.module.css'
 
 interface BaseNewLayoutProps {
   /** Defaults to true. If true, the global footer will be shown at the bottom of the page. */
   showFooter?: boolean
 }
+
+const { ConsentManager, openConsentManager } = createConsentManager({
+  preset: 'oss',
+})
 
 const BaseNewLayout: React.FC<BaseNewLayoutProps> = ({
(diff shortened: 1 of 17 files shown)


src/layouts/base-new/index.tsx
```diff
@@ -40,7 +44,12 @@ const BaseNewLayout: React.FC<BaseNewLayoutProps> = ({
           )}
         </div>
       </CoreDevDotLayout>
-      <ConsentManager />
+      <ConsentManager
+        onAcceptAll={() => {
+          handleOptInAnalytics(query['optInFrom'] as OptInPlatformOption)
+          console.log('accepted!!')
```
kendallstrautman, line 50: TODO - it doesn't seem like this function is firing as expected. I need to debug within the consent manager 🤔

ashleemboyer: I got curious and checked this out myself, it looks like you have to pass the props through the object that's passed to `createConsentManager`! I tried it out and that worked for me. Although, I was expecting your approach here to work as well, and it feels weird that it doesn't work.

kendallstrautman: Oh interesting! Okay I'll try that. Ty 😄

ashleemboyer: Sure thing! 😊

src/views/collection-view/index.tsx
```diff
@@ -20,6 +21,7 @@ function CollectionView({
   layoutProps,
   product,
 }: CollectionPageProps): React.ReactElement {
+  useOptInAnalyticsTracking('learn')
```
kendallstrautman, line 24: The analytics hook is called only on the views that handle opt-out UI. We could eventually lift this up, but this feels appropriate for now.

BrandonRomano (approved):
This looks good to me; I don't have any concerns about perf from that hook as it should be minimal; and it's temporary so this is good from my perspective!

PR #449, opened 2022-05-19
chore: hide TOC if only one heading

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-kshide-sidecar-one-item-hashicorp.vercel.app/http://localhost:3000/waypoint/tutorials/get-started-docker/get-started-intro) 🔎
- [Asana task](https://app.asana.com/0/1199634971449915/1202213086993158) 🎟️

## 🗒️ What
As per design's request this PR adds a check for the number of headings in the table of contents component (default for sidecar). If there's only one heading, the component doesn't render.

### DIFF
--- a/src/layouts/sidebar-sidecar/components/table-of-contents/index.tsx
+++ b/src/layouts/sidebar-sidecar/components/table-of-contents/index.tsx
@@ -11,15 +11,13 @@ const TABLE_OF_CONTENTS_LABEL_ID = 'table-of-contents-label'
 const TableOfContents = ({ headings }: TableOfContentsProps): ReactElement => {
   const { isDesktop } = useDeviceSize()
   const hasOneHeading = headings.length === 1
+  const enableActiveSection = hasOneHeading ? false : isDesktop
 
   /**
    * @TODO (2022-5-17) update the second argument to useActiveSection. Sidecar
    * now goes away at 1280px rather than the isDesktop width.
    */
-  const activeSection = useActiveSection(
-    headings,
-    hasOneHeading ? false : isDesktop
-  )
+  const activeSection = useActiveSection(headings, enableActiveSection)
 
   // Don't render if only one item
   if (hasOneHeading) {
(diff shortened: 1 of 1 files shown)


src/layouts/sidebar-sidecar/components/table-of-contents/index.tsx
```diff
@@ -17,6 +17,11 @@ const TableOfContents = ({ headings }: TableOfContentsProps): ReactElement => {
    */
   const activeSection = useActiveSection(headings, isDesktop)
 
+  // Don't render if only one item
+  if (headings.length === 1) {
+    return null
+  }
```
ashleemboyer, line 23: 🚧  I think `useActiveSection`'s second argument also needs to be updated as well so that the hook is disabled when `headings.length === 1` or `isDesktop` is `true`.

ashleemboyer (changes requested):

BrandonRomano (approved):
Looks good -- I like placing this at component level which will ensure this is never accidentally misused.  Saw a few threads with design having strong opinions that this sidecar is _never_ used with only one item.

src/layouts/sidebar-sidecar/components/table-of-contents/index.tsx
```diff
@@ -10,12 +10,19 @@ const TABLE_OF_CONTENTS_LABEL_ID = 'table-of-contents-label'
 
 const TableOfContents = ({ headings }: TableOfContentsProps): ReactElement => {
   const { isDesktop } = useDeviceSize()
+  const hasOneHeading = headings.length === 1
+  const enableActiveSection = hasOneHeading ? false : isDesktop
```
ashleemboyer, line 14: 🎨 Up to you if you want to make this change or not, but I'd personally find the following quite a bit easier to read & understand than a ternary!

```suggestion
  const enableActiveSection = !hasOneHeading && isDesktop
```

ashleemboyer (approved):

src/layouts/sidebar-sidecar/components/table-of-contents/index.tsx
kendallstrautman: I'm gonna leave it as is for now 👐

PR #456, opened 2022-05-20
refactor: memoize collection api calls

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksmemoize-learn-api-calls-hashicorp.vercel.app/) 🔎

## 🗒️ What

This PR memoizes collection related API calls where appropriate. For example, calls for `getAllCollections` are often identical for all collections with the same product so it makes sense to cache this result.


thiskevinwang (approved):
Looks straightforward to me!

I'd be curious to see what timings/metrics this changes (A "before n' after" as evidence for _how beneficial_ this memoization is)

src/views/tutorial-view/utils/get-collection-context.ts
```diff
@@ -25,6 +26,13 @@ interface CurrentCollectionTutorial {
   }
 }
 
+const moizeOpts: Options = {
+  isPromise: true,
+  maxSize: Infinity,
```
thiskevinwang, line 31: Any particular reason to not use `isDeepEqual: true,` here, but use it in the other 2 spots?

kendallstrautman: In this case we only pass a string as an arg to this call, as opposed to an object with nested keys (with `getAllCollections`) where we need the deep equality check.

kendallstrautman:
@thiskevinwang I agree on the metrics. I was trying to get a sense of that by just looking at build times but it didn't seem super _significant_ 🤷 Do you have any recos on how to try and measure the effect of this?

PR #458, opened 2022-05-20
fix: no mobile scroll when menu is open

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksfix-mobile-no-scroll-hashicorp.vercel.app/vault/docs) 🔎
- [Asana task](https://app.asana.com/0/1199634971449915/1202227071718207) 🎟️

## 🗒️ What

This PR fixes an issue where if one scrolled down the page and opened the sidebar, it may not render. Also if the sidebar was open and visible, it could be scrolled past. 

## 🛠️ How

I created a hook `useNoScrollBody` that accepts a trigger param, when this trigger is true, the body won't be scrollable. I also adjusted the positioning of the mobile sidebar a bit. 

## 🧪 Testing

- [Visit a page ](https://dev-portal-git-ksfix-mobile-no-scroll-hashicorp.vercel.app/vault/docs) in simulator or adjust your browser window to <tablet view. 
- Open the sidebar and scroll down. You should only be able to scroll to the bottom of the sidebar contents. 
- Close the sidebar and scroll a ways down the page, open it again to see that it renders in the correct position. 

- Also the search results component was updated to use this shared hook. Please [visit a docs page](https://dev-portal-git-ksfix-mobile-no-scroll-hashicorp.vercel.app/vault/docs) with search on a mobile view, type something in so the results panel shows up - there should be no regressions and the body shouldn't be scrollable.


src/layouts/sidebar-sidecar/index.tsx
```diff
@@ -82,12 +83,26 @@ const SidebarSidecarLayoutContent = ({
...
   }
 
+  /**
+   * Handles the top positioning of the sidebar if a user has scrolled
+   * and prevents scrolling on the rest of the page body
+   */
+  useEffect(() => {
+    if (sidebarIsOpen) {
+      setScrollDelta(window.scrollY)
+      document.body.style.overflow = 'hidden'
```
kendallstrautman, line 93: Open to a suggestion for a better way to handle this 😸

src/layouts/sidebar-sidecar/sidebar-sidecar-layout.module.css
```diff
@@ -75,9 +75,9 @@ for further details.
   @media (--dev-dot-tablet-down) {
     bottom: 0;
     left: -150vw;
-    position: absolute;
-    top: 0;
+    position: absolute; /** Top position set dynamically based on scroll pos in component */
     width: 100vw;
+    padding-bottom: 70px;
```
kendallstrautman, line 80: We needed a bit extra space to allow the whole component to be scrollable. We can fiddle with this number.

src/layouts/sidebar-sidecar/index.tsx
ashleemboyer: Was just coming here to share what Bryce did for the search component! I can't remember where we chatted, but he said this code would be good to lift up to be reusable: https://github.com/hashicorp/dev-portal/blob/main/src/components/algolia-search/components/panel.tsx#L97-L109

```jsx
/**
 * Ensures the body can't be scrolled while the search panel is open on mobile viewports.
 */
useEffect(() => {
  const initialValue = document.body.style.overflow
  if (!isDesktop) {
    document.body.style.overflow = 'hidden'
  }

  return () => {
    document.body.style.overflow = initialValue
  }
}, [isDesktop])
```

kendallstrautman: Ah cool! I'll lift this up then, ty 😄

alexcarpenter: Can we use position fixed on the sidebar and overflow hidden on the body?  That'd reduce the need to do any scroll page offset calculations.

https://user-images.githubusercontent.com/825855/169607308-97afb2ad-fef1-4c09-95a6-82e702578231.mp4

alexcarpenter: `--sticky-bars-height` doesn't account for text wrapping, which is why the top of the sidebar is a little cut off.

ashleemboyer: I think the alpha callout with the wrapping text is going away for beta, but I'm not 100% sure. @BrandonRomano do you know? We will probably be able to make `--sticky-bars-height` less brittle and have less maintenance to do on it if so!

src/components/algolia-search/components/panel.tsx
```diff
@@ -97,16 +97,7 @@ export default forwardRef(function Panel<THit extends Hit<unknown>>(
...
-    const initialValue = document.body.style.overflow
-    if (!isDesktop) {
-      document.body.style.overflow = 'hidden'
-    }
-
-    return () => {
-      document.body.style.overflow = initialValue
-    }
-  }, [isDesktop])
+  useNoScrollBody(!isDesktop)
```
zchsh, line 100: 🙌 props for finding places that do the same thing and backfilling the abstraction 💖

src/layouts/sidebar-sidecar/sidebar-sidecar-layout.module.css
zchsh: 💭 Thought, very much non-blocking and outside the scope of this PR: I think this is a fundamental issue with alert bar height mismatch. I think we might be able to remove the need for this if we tackle this other task unrelated to the contents of this PR:
- 🎟️ [[sidebar slash top nav] fix mismatch in CSS height of top bars vs reality, especially on mobile](https://app.asana.com/0/1202114367927919/1202316169578550/f)

Not sure exactly what to do in the scope of this PR - I think it meets the criteria of scroll lock. So more a random related mention than anything else 👍

zchsh (approved):
✅ The scroll lock behaviour, and the hook, both look great to me! 💯 🚀 

📌  Very much outside the scope of this PR, but noticed a couple of other mobile issues and wanted to tag em here to check whether my assessment makes sense:
- 🎟️ [[top nav] fix width issue on smallest mobile](https://app.asana.com/0/1202114367927919/1202316169578542/f) (noticed cause I went to test on mobile, and couldn't open the mobile menu)
- 🎟️ [[sidebar slash top nav] fix mismatch in CSS height of top bars vs reality, especially on mobile](https://app.asana.com/0/1202114367927919/1202316169578550/f) (might reduce or remove the need for the 70px "magic"-ish number in this PR

src/layouts/sidebar-sidecar/sidebar-sidecar-layout.module.css
ashleemboyer: This ticket's also very relevant, and I think it'll reduce the complexity of the brittle height variable we have:

🎟️ [[alpha vs beta feedback] when should we drop the "alpha" feedback callout?](https://app.asana.com/0/1202110981600689/1202285743002882/f)

src/hooks/use-no-scroll-body.ts
```diff
@@ -0,0 +1,27 @@
...
+    } else if (!trigger && document.body.style.overflow === 'hidden') {
+      document.body.style.overflow = initialValue
+    }
+
+    // reset on component unmount
+    return () => {
+      document.body.style.overflow = initialValue
+    }
+  }, [trigger])
+}
```
ashleemboyer, line 27: ✨

src/layouts/sidebar-sidecar/index.tsx
```diff
@@ -82,6 +83,12 @@ const SidebarSidecarLayoutContent = ({
     },
   }
 
+  /**
+   * Handles the top positioning of the sidebar if a user has scrolled
+   * and prevents scrolling on the rest of the page body
+   */
```
ashleemboyer, line 89: 🚧 Since this hook only affects `document.body.style.overflow`, can this comment be updated to reflect that?

```suggestion
  /**
   * Prevents scrolling on the rest of the page body
   */
```

src/layouts/sidebar-sidecar/sidebar-sidecar-layout.module.css
ashleemboyer: 🗒️ This is just a breadcrumb in case it comes up next week, but I'll be adding the version select (which is fixed to the bottom left side of the page, "within" the sidebar) so some tweaks might be needed on this value then.

ashleemboyer (approved):
Approving so a re-review isn't needed after the requested change, but there's just one comment in here I think needs updated and the rest looks good! Nicely done 🙌🏻

src/layouts/sidebar-sidecar/sidebar-sidecar-layout.module.css
kendallstrautman: Sounds good. I'll link to this convo in those tickets 👌

src/layouts/sidebar-sidecar/index.tsx
kendallstrautman: Ahh good catch, ty 😄

src/components/algolia-search/components/panel.tsx
kendallstrautman: That was very much @ashleemboyer's idea 🙇 👏

BRKalow: TY ❤️

PR #465, opened 2022-05-24
feat: adds rewriteTutorialLinksPlugin to docs rendering

BRKalow:

---

## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Add the `rewriteTutorialLinks` plugin which @kendallstrautman created for learn content, to docs content rendering.

## 🤷 Why

This should allow us to handle any existing learn links in docs content by rewriting them to point at the correct dev dot path.


src/data/vault.json
```diff
@@ -145,7 +145,7 @@
       "os": "linux"
     }
   ],
-  "basePaths": ["docs", "api-docs", "intro"],
+  "basePaths": ["docs", "api-docs", "intro", "tutorials"],
```
BRKalow, line 148: Not sure if we feel like this conflates the intent of this property, but it does allow our current content URL rewrites to "just work" for tutorials links that are authored in the context of dev dot.

BrandonRomano (approved):
LGTM: Maybe just wait for @kendallstrautman also to verify your Q you asked inline RE: it's intent to be used here too.

kendallstrautman (approved):
Looks great! 

One thing I did notice is that older links that are treated as redirects on the original site are 404'ing. For example there's a bunch of old [redirects to learn](https://github.com/hashicorp/vault/blob/main/website/redirects.js#L714-L729) from .io that are broken. See one at the bottom of this[ docs page](https://dev-portal-git-brkfeat-rewrite-learn-links-docs-hashicorp.vercel.app/vault/docs/concepts/dev-server#use-case), and another 'Hashicorp Tutorials' link [on this page.](https://dev-portal-git-brkfeat-rewrite-learn-links-docs-hashicorp.vercel.app/vault/docs/install#package-manager) If a user is opted in and the redirect is in effect, it _should_ keep them within dev portal for a link like this. We don't need to address that on this PR but just wanted to note it.

BRKalow:
@kendallstrautman good call. I'd say we need to do a content sweep to make sure that any links like this are actually fixed. Ideally they would have been fixed when the redirects were added.

It'd be preferable if we didn't have to bring forward redirects for .io sites to dev dot, but not sure if it's realistic.

kendallstrautman:
@BRKalow do you think this should just be handled at the content level?

BRKalow:
@kendallstrautman Probably, but I think we'll need this to support old versions too.

src/data/vault.json
BRKalow, line 148:
```suggestion
  "basePaths": ["docs", "api-docs", "intro", "tutorials", "downloads"],
```

PR #468, opened 2022-05-24
style: fix card link height

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksfix-grid-item-height-hashicorp.vercel.app/waypoint) 🔎
- [Asana task](https://app.asana.com/0/1199634971449915/1202316221241102) 🎟️

## 🗒️ What

Fixes a bug where cards in a grid list didn't fill the entire height. 

CleanShot 2022-05-20 at 21 28 55@2x


src/components/cards-grid-list/cards-grid-list.module.css
```diff
@@ -19,6 +19,11 @@
   margin: 0;
   padding: 0;
 
+  /** Ensures the grid items fill the full row height */
+  & > * {
+    display: flex;
+  }
```
ashleemboyer, line 25: 🚧 I'm not sure that this selector is the approach we want to take, because I think we're tend to be pretty averse where possible to non-classname selectors? The approach we've used so far with making card heights fill the space is by adding a class to the container holding `children` that has `height: 100%`. There was a PR where we discussed trying to make cards smarter to auto-fill their height, but I can't remember which one or what we landed on. @zchsh do you remember? We had the discussion with Alex (not directly tagging yet) when he was working on the home page.

kendallstrautman: Cool, yeah the first approach I took was to just add `height: 100%` to the card link `root`. I can refactor back to that if it makes more sense!

kendallstrautman: Cool, yeah the first approach I took was to just add `height: 100%` to the card link `root`. I can refactor back to that if it makes more sense!

ashleemboyer: I think adding `height: 100%` to `CardLink` will break some things. Let me grab an example of what I mean from other places we've handled this. 😊

ashleemboyer: Here's an example from the `FeaturedTutorialsSection ` component (rendered by `ProductDownloadsView`):

- The `.card` class: https://github.com/hashicorp/dev-portal/blob/main/src/views/product-downloads-view/components/featured-tutorials-section/featured-tutorials-section.module.css#L28-L33
- Passing the `.card` class to `CardLink` via `className`: https://github.com/hashicorp/dev-portal/blob/main/src/views/product-downloads-view/components/featured-tutorials-section/index.tsx#L41-L54

ashleemboyer: Just did some checking of setting `height: 100%` on `CardLink`'s `root` class, and I think what I was thinking of before about it potentially breaking things was actually on `Card`'s `root` class! I remember now that that was one of the ideas for making these components automatically grow when in a grid. So as long as we ID all the places this change could affect, test them on the preview url, and they don't break -- sounds like a fine approach to me!

kendallstrautman: Sounds good, i'll update and QA ❇️

kendallstrautman: @ashleemboyer made that update and did a quick sweep of that pages I know use this card link component. Care to give it a second set of eyes?

ashleemboyer: Taking a look now! 👀

ashleemboyer (approved):
Woo! Seriously, killin' with these one liners 🤩

src/components/card-link/card-link.module.css
```diff
@@ -4,6 +4,7 @@ style applied to anchor elements.
 */
 .root {
   color: black;
+  height: 100%;
```
zchsh, line 7: 🙌 I was too slow to review this in time, but kudos for going for the root cause here 💯

PR #469, opened 2022-05-24
fix: update tutorials anchor styles

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksupdate-tutorials-anchor-styles-hashicorp.vercel.app/waypoint/tutorials/get-started-docker/get-started-install) 🔎
- [Asana task](https://app.asana.com/0/1199634971449915/1202325853922364) 🎟️

## 🗒️ What

Updates the tutorials mdx content to have the correct link styles. 

## 🛠️ How

This PR creates a new global `MdxInlineLink` component that has the appropriate mdx content styles applied. Both docs and tutorials now use this component. 

## 🧪 Testing

- [Visit a tutorial,](https://dev-portal-git-ksupdate-tutorials-anchor-styles-hashicorp.vercel.app/waypoint/tutorials/get-started-docker/get-started-install) ensure it matches the[ docs page anchors](https://dev-portal-git-ksupdate-tutorials-anchor-styles-hashicorp.vercel.app/waypoint/docs). 

## 💭 Anything else?

I think there is potential for MdxInlineLink to use `MaybeInternalLink` eventually but right now the docs anchor component is doing some logic to decide whether it uses `next/link` and refactoring that feels outside the scope of this PR. I like that this is a super simple styled anchor for now.


src/components/docs-anchor/components/styled-anchor/index.tsx
```diff
@@ -0,0 +1,17 @@
+import classNames from 'classnames'
+import { AnchorHTMLAttributes } from 'react'
+import s from './styled-anchor.module.css'
+
+export default function StyledAnchor(
```
kendallstrautman, line 5: I refactored the docs link to extract out just the styling of the anchor element so this could be shared. Another approach I considered was just `composing` the styles from the docs stylesheet in a custom component defined on the tutorials `MDX_COMPONENTS` list. I thought it might be nice to make this abstraction here, as just composing seems like it could introduce potential for drift. Super open to input here. 

This also made me realize that we could do more optimization for anchor links on tutorial content that is an internal versus external link and use `next/link` like in docs.

ashleemboyer: I think with the goal being to have something reusable in many areas of the site, we should place this component in the main `src/components` file! ✨

We could name it something like `MDXInlineLink` which I think:
- aligns with how we want to name and create MDX components ([Convert global-ish docs content CSS to MDX custom components](Convert global-ish docs content CSS to MDX custom components))
- conveys it's intended use (in MDX content) & what component it's an abstraction of (`InlineLink`)

kendallstrautman: Sounds good! i'll lift this up to there 👌

src/components/docs-anchor/index.tsx
```diff
@@ -48,15 +42,15 @@ const DocsAnchor = ({
   if (adjustedHref !== href) {
     return (
       <Link href={adjustedHref}>
```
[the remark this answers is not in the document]
zchsh: 📌  Random related thing: IIRC `next/link` _can_ be used for external links (it seems to work I think?) but doesn't seem to be recommended (wording [in next/link docs](https://next-js.ir/docs/api-reference/next/link) is:

> External URLs, and any links that don't require a route navigation using /pages, don't need to be handled with Link; use the anchor tag for such cases instead.

I wonder whether the "don't need" wording there has significant performance implications... or whether we could drop `MaybeInternalLink` altogether, and use `next/link` all the time to avoid messy logic? 🤔

src/components/mdx-inline-link/index.tsx
```diff
@@ -0,0 +1,14 @@
+import classNames from 'classnames'
+import { AnchorHTMLAttributes } from 'react'
+import s from './mdx-inline-link.module.css'
+
+// styled anchor element for tutorial and docs content
+export default function MdxInlineLink(
```
zchsh, line 6: 🙌 This is great, feels like a nice step towards more unified custom MDX components

zchsh (approved):

src/components/docs-anchor/index.tsx
```diff
@@ -1,8 +1,7 @@
 import Link from 'next/link'
```
ashleemboyer, line 1: 🚧 I can't leave a comment on the specific line, but I think we only want to replace the `InlineLink` component reference here with the new `MSXInlineLink` component, and leave the other manual `<a>` elements as they are. `InlineLink` is the one that applies the custom styling for within docs content.

ashleemboyer (changes requested):
Looking good so far! 🙌🏻 This is really close, there are just a few things that need changed/discussed first.

src/components/docs-anchor/index.tsx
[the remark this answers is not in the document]
ashleemboyer: I think @dstaley can speak more to this because he did some related work on one of our properties (maybe dotcom?), but AFAIK it's OK for us to always use `next/link`! And we actually have [a ticket in the backlog to remove `MaybeInternalLink`](https://app.asana.com/0/1202097197789424/1202129912162168/f) when there's time. 😊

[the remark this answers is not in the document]
zchsh: Oh other thought here - do we need `passHref` here on `next/link`?

ref: <https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-custom-component-that-wraps-an-a-tag>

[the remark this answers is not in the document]
ashleemboyer: Hmm that `code` example is a good point. I wonder how many of these manual `<a>` elements need to be replaced with `InlineLink`/`MDXInlineLink`, and if `InlineLink` needs to be updated to replace the `text` prop with `children`? Would mean not using `Text` inside of `InlineLink`, but I think we could probably use CSS composition in place of that. Thoughts?

src/components/docs-anchor/index.tsx
kendallstrautman: The other `a` tags had custom styles being pass as well via the `s.anchor` class in `passableProps`

src/components/docs-anchor/index.tsx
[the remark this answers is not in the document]
kendallstrautman: @zchsh The goal with this PR is just to address style inconsistencies with tutorials anchor links, so I'm hesitant to make more changes to this docs-anchor without having full context to how this original docs anchor component was set up. Should we make a follow up task?

src/components/docs-anchor/index.tsx
ashleemboyer: I think we can totally make a follow-up task as you've asked in the other comment thread after these changes. 😊

kendallstrautman: Do you mean after I swap `InlineLink` for `MdxInlineLink`?

ashleemboyer: Yeah! So to summarize what I'm thinking should be done:

- Undo other changes in this file where `MdxInlineLink` is added
- Update `MdxInlineLink` to render `InlineLink`
- Replace [the `InlineLink` reference in this file](https://github.com/hashicorp/dev-portal/blob/main/src/components/docs-anchor/index.tsx#L37-L42) with `MdxInlineLink`

src/components/docs-anchor/index.tsx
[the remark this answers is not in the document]
zchsh: @kendallstrautman For the style issues, absolutely! We've got 🎟️ [[code-link] Visited style refinement](https://app.asana.com/0/1202097197789424/1202331881539237/f) which I think covers it.

For `passHref`, I think it should be added in this PR, as otherwise it's a regression (previously `href`s were rendered into the DOM without the need for `passHref` since we had plain `<a />` element rather than "custom" component).

I think all we need is:

```suggestion
      <Link href={adjustedHref} passHref>
```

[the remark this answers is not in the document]
ashleemboyer: @zchsh do our suggested changes conflict? Here's what I'm suggesting in another thread: https://github.com/hashicorp/dev-portal/pull/469#discussion_r882979173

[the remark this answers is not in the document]
kendallstrautman: So y'all think I should keep the `a` with the custom style defs? I'm just so confused now. Maybe I should just duplicate the styles for tutorials and not touch this docs component.

src/components/docs-anchor/index.tsx
kendallstrautman: Maybe I shouldn't have even made mdxInlineLink if we just want it to be a wrapper for inline link?

src/components/docs-anchor/index.tsx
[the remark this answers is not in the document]
kendallstrautman: @zchsh @ashleemboyer I reverted updates to the docs anchor component and added a TODO to refactor and use this new mdx link component.

package.json
```diff
@@ -134,6 +134,8 @@
     "@reach/visually-hidden": "^0.16.0",
     "@react-aria/ssr": "^3.1.0",
     "@react-aria/utils": "^3.11.0",
+    "@types/cors": "^2.8.12",
+    "@types/google-spreadsheet": "^3.2.2",
```
kendallstrautman, line 138: So the type checker would pass. This may be merged in on another PR as well.

src/components/docs-anchor/index.tsx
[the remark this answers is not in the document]
zchsh: ☝️ I'm good with that - re-affirming my approval on this PR ✅ apologies for the confusion!

src/views/product-downloads-view/components/downloads-section/index.tsx
```diff
@@ -184,21 +184,21 @@ const NotesSection = ({ selectedRelease }) => {
         You can find the{' '}
         <InlineLink
           href={`https://releases.hashicorp.com/${name}/${version}/${shasums}`}
-          text={`SHA256 checksums for ${currentProduct.name} ${version}`}
           textSize={200}
-        />{' '}
+        >
+          {`SHA256 checksums for ${currentProduct.name} ${version}`}
```
ashleemboyer, line 189: 🚧 This no longer needs the template literal.

```suggestion
          SHA256 checksums for {currentProduct.name} {version}
```

src/components/inline-link/index.tsx
```diff
@@ -1,24 +1,27 @@
...
+  children,
+  ...rest
 }: InlineLinkProps): ReactElement => {
-  const classes = classNames(s.root, className)
+  const classes = classNames(
+    s.root,
+    className,
+    `hds-typography-body-${textSize}`,
+    `hds-font-weight-${textWeight}`
+  )
```
ashleemboyer, line 20: 🚧 Can we apply `className` last so it's easy to customize any of the CSS properties?

```suggestion
  const classes = classNames(
    s.root,
    `hds-typography-body-${textSize}`,
    `hds-font-weight-${textWeight}`
    className,
  )
```

src/components/inline-link/types.ts
```diff
@@ -25,4 +21,9 @@ export interface InlineLinkProps {
    * The `weight` passed to the inner `Text` component.
    */
   textWeight?: TextProps['weight']
+
+  /**
+   * The text to render in the anchor
```
ashleemboyer, line 26: 🚧 Can this be updated for accuracy and to still mention <a> element?

```suggestion
   * The content to render within the `<a>` element.
```

src/components/inline-link/types.ts
```diff
@@ -1,4 +1,5 @@
 import { TextProps } from 'components/text'
+import React from 'react'
```
ashleemboyer, line 2: 🚧 Can this be updated to have the `react` import first and also to only import `ReactNode`?

```suggestion
import { ReactNode } from 'react'
import { TextProps } from 'components/text'
```

src/components/inline-link/docs.mdx
```diff
@@ -15,79 +15,46 @@ Internally, this component uses two other components:
 | ------------ | -------------------------------------------------------------------------------- | --------- | ------------- | ------------------- |
 | `className`  | A string of one or more class names. Applied last to the rendered `<a>` element. | No        | undefined     | string              |
 | `href`       | The destination of the link.                                                     | Yes       | N/A           | string              |
-| `text`       | The text rendered within the `<a>` element.                                      | Yes       | N/A           | string              |
-| `textSize`   | The `size` passed to the inner `Text` component.                                 | No        | undefined     | TextProps[`size`]   |
-| `textWeight` | The `weight` passed to the inner `Text` component.                               | No        | undefined     | TextProps[`weight`] |
+| `children`   | The anchor text to render within the link.                                       | Yes       | N/A           | ReactNode           |
```
ashleemboyer, line 18: 🚧 Can this be updated for accuracy and to still mention `<a>` element? The latter was originally done to make it clear the exact element that is rendered.

```suggestion
| `children`   | The content to render within the `<a>` element.                                  | Yes       | N/A           | ReactNode           |
```

src/components/inline-link/docs.mdx
```diff
@@ -15,79 +15,46 @@ Internally, this component uses two other components:
...
   <br />
-  <InlineLink
-    href="/"
-    text="link with textWeight=bold"
-    textWeight="bold"
-  />
+  <InlineLink href="/" textWeight="bold">link with textWeight=bold</InlineLink>
 </>
-`}</LiveComponent>
+`}</LiveComponent>{' '}
```
ashleemboyer, line 60: 🙋🏻 Was this change intentional?

ashleemboyer (changes requested):
Looking great! 🙌🏻 Tested the affected pages, just have a few small change requests.

src/views/product-downloads-view/components/downloads-section/index.tsx
```diff
@@ -184,21 +184,21 @@ const NotesSection = ({ selectedRelease }) => {
...
+          verify the checksums signature file
+        </InlineLink>{' '}
         which has been signed using{' '}
-        <InlineLink
-          href="https://www.hashicorp.com/security"
-          text="HashiCorp's GPG key"
-          textSize={200}
-        />
+        <InlineLink href="https://www.hashicorp.com/security" textSize={200}>
+          {"HashiCorp's GPG key"}
```
ashleemboyer, line 200: Didn't notice this one until now, but it can be updated too!

```suggestion
          HashiCorp's GPG key
```

ashleemboyer (approved):

PR #475, opened 2022-05-25
feat: revise dev-dot 404 pages

zchsh:

---

## 🔗 Relevant links

- [Preview - dev-dot 404, no product context][/generic-yeehaw] 🔎
- [Preview - dev-dot Vault 404][/vault/yeehaw] 🔎
- [Preview - dev-dot Waypoint 404][/waypoint/yeehaw] 🔎
- [Preview - dev-dot Vault versioned 404][/vault/docs/v1.5.x/browser-support] 🔎
- [Preview - dev-dot Waypoint versioned 404][/waypoint/docs/v0.2.x/projects] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Implements revised 404, versioned 404, and generic error views for dev-dot.

## 🤷 Why

Current dev-dot error views do not match design specifications; and use tokens and styles from the marketing design system.

## 🛠️ How

<details>

<summary>➡️ Pre-work (also in self-review)</summary>

- Adds `components/inline-link-button`
     - We need a link-looking element that navigates the browser back a page
     - This element should have `onClick={router.back}`, so should be a `<button />`, I think, but that looks like an inline link
     - This PR adds `components/inline-link-button` to fill this need
- Prevents "missing product data" errors
    - for context, in `components/navigation-header`, previous logic was to assume we should try to render product-specific `RightSideHeaderContent`, unless the `currentPath` is the home page `/`
    - with this PR, we introduce a 404 page that needs to render the navigation header, but is not the home page. this created a need to modify the logic, as otherwise, we'd try to render `LeftSideHeaderContent` and fail.
    - This PR modifies the logic to check if `currentProduct` data exists before trying to render elements that depend on that data
- Exports `productConfig`, which is a dictionary of `ProductData`
    - Exported from `lib/cms`, since that's where the dictionary lived already
    - The `ProductData` dictionary is used in `pages/_error.tsx`'s `getServerSideProps`
    - We use it to pass the appropriate product data to the error view, so that we can render specific navigation header contents depending on product context
- Adds an `isProductSlug` type guard
    - Used in `pages/_error.tsx`'s `getServerSideProps` to ensure we're correctly dynamically detecting product context from URL segments

</details>

- Handles layout switching `pages/_error.tsx`
    - We already handle this, though this PR updates to use the `base-new` layout
    - Side note: we should probably delete the old base layout, I think? Not urgent for June 20 though.
        - 🎟️ [[cleanup] deprecate or delete src/layouts/base](https://app.asana.com/0/1202097197789424/1202349999400579/f)
- Handles view content switching in `views/error-view-switcher`
    - The `pages/_error.tsx` page uses this view switcher
    - We pass the view switcher `statusCode` and `isProxiedDotIo`, and it uses this (along with versioned URL detection) to determine which error view to render
- Organizes dot-io error views in `views/_proxied-dot-io`
    - Retains but refactors existing views (which had logic that is now handled by `error-view-switcher`)
    - Refactored to set up a display-only `views/_proxied-dot-io/versioned-error`
    - Generic 404 view, and fallback error view, continues to be handled by `@hashicorp/react-error-view`
    - Side note: text contrast on generic 404 view seems off (but correct on versioned view). Looks like we need to update this in `react-components`.
        - 🎟 [[react-error-view] Fix text contrast for generic vault 404 page](https://app.asana.com/0/1100423001970639/1202349999400572/f)
- Organizes dev-dot error views in `views/error-views`
    - Sets up shared `views/error-views/components` for use in all dev-dot error views
    - Adds `versioned-404`, `404`, and `fallback` error views, each of which compose the shared components

## 📸 Design Screenshots

CleanShot 2022-05-26 at 18 19 45@2x

## 🧪 Testing

- [ ] Test the error pages on dev-dot
    - Note that the `go back` link on non-versioned 404 pages will do nothing if there is no browser history
    - [ ] Test the 404 error page outside product context, such as [/generic-yeehaw][/generic-yeehaw]
    - [ ] Test a Vault 404 page, such as [/vault/yeehaw][/vault/yeehaw]
    - [ ] Test a versioned Vault 404 page, such as [/vault/docs/v1.5.x/browser-support][/vault/docs/v1.5.x/browser-support]
    - [ ] Test a Waypoint 404 page, such as [/waypoint/yeehaw][/waypoint/yeehaw]
    - [ ] Test a versioned Waypoint 404 page, such as [/waypoint/docs/v0.2.x/projects][/waypoint/docs/v0.2.x/projects]
- [ ] Test the error pages on dot-io sites
    - You'll need to select a specific product context using the preview mode dropdown in the lower left
    - They should look the same as they did previously
    - [ ] Test a regular 404 error page, such as [/another-yeehaw][/another-yeehaw]
    - [ ] Test a versioned 404 error page, for example as [/docs/v1.5.x/browser-support][/docs/v1.5.x/browser-support]
        - Note: for this specific URL, the "go to latest version" link on this page is only expected to work for Vault.
        - The  "go back to docs" link should work on all products (that have a `/docs` route, which I think is all of em!)

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/1202114367927919/1202114360365288/f
[/yeehaw]: https://dev-portal-git-zsrevise-404-page-hashicorp.vercel.app/yeehaw
[/generic-yeehaw]: https://dev-portal-git-zsrevise-404-page-hashicorp.vercel.app/generic-yeehaw
[/another-yeehaw]: https://dev-portal-git-zsrevise-404-page-hashicorp.vercel.app/another-yeehaw
[/vault/yeehaw]: https://dev-portal-git-zsrevise-404-page-hashicorp.vercel.app/vault/yeehaw
[/waypoint/yeehaw]: https://dev-portal-git-zsrevise-404-page-hashicorp.vercel.app/waypoint/yeehaw
[/vault/docs/v1.5.x/browser-support]: https://dev-portal-git-zsrevise-404-page-hashicorp.vercel.app/vault/docs/v1.5.x/browser-support
[/waypoint/docs/v0.2.x/projects]: https://dev-portal-git-zsrevise-404-page-hashicorp.vercel.app/waypoint/docs/v0.2.x/projects
[/docs/v1.5.x/browser-support]: https://dev-portal-git-zsrevise-404-page-hashicorp.vercel.app/docs/v1.5.x/browser-support


src/components/inline-link-button/index.tsx
```diff
@@ -0,0 +1,23 @@
+import { ReactElement } from 'react'
+import classNames from 'classnames'
+import { InlineLinkButtonProps } from './types'
+import s from './inline-link-button.module.css'
+
+/**
+ * A <button> version of "components/inline-link",
+ * to support onClick navigation actions such as window.history.back()
+ */
+function InlineLinkButton({
```
zchsh, line 10: From designs, we seem need a link-looking element that can call `router.back()`. So I figured it made sense to add a `<button />` styled like an inline link. Styles are quite duplicative of `components/inline-link`, but felt better to copy and paste than to mash together a more complex abstraction.

src/components/navigation-header/index.tsx
```diff
@@ -35,12 +34,13 @@ const RightSideHeaderContent = () => {
...
-  const LeftSideHeaderContent =
-    currentPath === '/' ? HomePageHeaderContent : ProductPageHeaderContent
+  const currentProduct = useCurrentProduct()
+  const LeftSideHeaderContent = !currentProduct
+    ? HomePageHeaderContent
+    : ProductPageHeaderContent
 
   // TODO: menu for the home page, which does not use SidebarSidecarLayout
-  const shouldShowRightSide = !isDesktop && currentPath !== '/'
+  const shouldShowRightSide = !isDesktop && currentProduct
```
zchsh, line 43: I've modified the logic here to check if `currentProduct` data exists before trying to render elements that depend on that data.

Previously we made this decision purely on whether the URL was the home page - which meant for URLs not the home page, that don't have clear product context (eg `/yeehaw`), we'd have a client-side error when trying to render product-specific `RightSideHeaderContent`.

src/lib/cms.ts
```diff
@@ -67,7 +67,7 @@ function proxiedRivetClient(productSlug: ProductSlug) {
 }
 
 export default instance
-export { client, proxiedRivetClient }
+export { productConfig, client, proxiedRivetClient }
```
zchsh, line 70: Exports `productConfig`, which is a dictionary of `ProductData`, as it's needed in `pages/_error.tsx`'s `getServerSideProps` to render the product-specific top navigation (where possible).

src/lib/products.ts
```diff
@@ -18,6 +18,16 @@ const productSlugsToNames: { [slug in ProductSlug]: ProductName } = {
   vagrant: 'Vagrant',
 }
 
+/**
+ * Type guard to determine if a string is a ProductSlug
+ *
+ * TODO: should we define ProductSlug as an enum,
+ * so that we can use its values directly here?
+ */
+function isProductSlug(string: string): string is ProductSlug {
```
zchsh, line 27: This `isProductSlug` type guard is used in `pages/_error.tsx`'s `getServerSideProps` to ensure we're correctly dynamically detecting product context from URL segments.

src/pages/_error.jsx
```diff
@@ -1,14 +1,16 @@
-import BaseLayout from 'layouts/base'
+import BaseLayout from 'layouts/base-new'
 import proxiedLayouts from 'layouts/_proxied-dot-io/dict'
 import { getProxiedProductSlug } from 'lib/env-checks'
-import { VersionedErrorPage } from 'views/_proxied-dot-io/versioned-error'
+import ErrorViewSwitcher from 'views/error-view-switcher'
```
zchsh, line 4: See notes in PR description for a more high-level view of `views/error-view-switcher` and how it's meant to work.

zchsh:
@BRKalow This PR implements dev-dot specific 404 views. Tagged you here as I feel like you might have the most context on the versioned 404 error page 🙇‍♂️ 

@ashleemboyer Tagged you here since this feels learn-related. Feel free to opt out of review, this is a bit of a chonker! I do think the addition of `components/inline-link-button` is likely of interest - it felt right given `router.back()` needs to be an `onClick`, yet also felt like navigation so felt reasonable that it's styled like a link. But happy to take another approach if there's something else we'd rather do - eg I guess we could expand `components/inline-link` to support `onClick`, and then `e.preventDefault` on click followed by `router.back()`? 🤔 

Side note for review:
- Vercel seems to do some clever caching for 404 pages, unfortunately this creates a review quirk
    - once _anyone_ visits a 404 page on the preview links with a specific product context cookie set, we'll all see that same product 404 rendered no matter what context you later switch to.
    - for example, I've visited [/yeehaw](https://dev-portal-git-zsrevise-404-page-hashicorp.vercel.app/yeehaw) with no cookie set, [/another-yeehaw](https://dev-portal-git-zsrevise-404-page-hashicorp.vercel.app/another-yeehaw) with Vault set, and [/another-yeehaw-consul](https://dev-portal-git-zsrevise-404-page-hashicorp.vercel.app/another-yeehaw-consul) with Consul set
    - I _believe_ this is a preview-specific issue, and may be less relevant in production, where each product has its own domain name (and therefore separate Vercel cache, I think?). But _definitely_ wanted to flag in case that assumption is incorrect. cc @BRKalow

src/lib/products.ts
BRKalow: @zchsh and type guards, name a more dynamic duo 😛

src/views/error-view-switcher/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+   *
+   * One solution might be localize `@hashicorp/react-error-view`
+   * in the dev-portal project. However, we may not be ready to do that,
+   * as the package may be used elsewhere?
+   */
+  const { asPath } = useRouter()
+  const [isMounted, setIsMounted] = useState(false)
+
+  /**
+   * Note from Bryce on this useEffect:
```
BRKalow, line 39: 👉 🤡  it's me

src/views/error-view-switcher/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const versionMatches = VERSION_PATTERN.exec(asPath)
+  const versionInPath = versionMatches?.groups?.version
+  const pathWithoutVersion = asPath.replace(VERSION_PATTERN, '')
+  const pathBeforeVersion = asPath.substring(0, asPath.indexOf(versionInPath))
+  const is404 = statusCode == 404
+  const isVersioned404 = versionInPath && is404
+
+  /**
+   * Switch between proxied dot-io and dev-dot error views
+   */
```
BRKalow, line 67: thought: not sure if this would be easier to grok, but we could use the `if` to determine like a `type` variable and then use an actual switch statement. I guess it's more stylistic than anything, but might make it easier to see what's rendered for each scenario. (Optional really, just a thought)

src/views/error-views/dev-dot-404/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+              productSlug: 'waypoint',
+              icon: <IconWaypointColor16 />,
+            },
+            {
+              url: '/',
+              text: 'HashiCorp Developer',
+              productSlug: 'hashicorp',
+              icon: <IconHome16 />,
+            },
+          ].map(
```
BRKalow, line 56: nit: this feels like a bit of a future-proof hedge, which is okay, but I wonder if it's any better than just rendering the components with these props?

BRKalow (commented):
Really nicely done here! This one is tricky. I believe the caching thing you are seeing will not be an issue in production for the reason you mentioned.

ashleemboyer (changes requested):
🚧 I think we need a new `CTALink` instead of a new `InlineLinkButton`. Have started a thread in Slack to discuss in more depth!

src/views/error-views/dev-dot-404/index.tsx
zchsh: yeah great call - I'm very much on the fence! Intent here was that we'll almost certainly want to later source this data from somewhere else, so I figured separating it from the markup was a decent first step (also makes it more clear our data source would either need to handle JSX, or we'll need to make an icon dict).

I very weakly lean towards leaving it as-is, but happy to adjust if needed 🙇‍♂️

src/views/error-view-switcher/index.tsx
zchsh: great call - i'll... `switch` it up 😎

src/components/navigation-header/index.tsx
```diff
@@ -35,12 +34,13 @@ const RightSideHeaderContent = () => {
  */
 const NavigationHeader = () => {
   const { isDesktop } = useDeviceSize()
-  const currentPath = useCurrentPath({ excludeHash: true, excludeSearch: true })
-  const LeftSideHeaderContent =
-    currentPath === '/' ? HomePageHeaderContent : ProductPageHeaderContent
+  const currentProduct = useCurrentProduct()
+  const LeftSideHeaderContent = !currentProduct
+    ? HomePageHeaderContent
+    : ProductPageHeaderContent
```
ashleemboyer, line 40: 🚧 Could this be flipped around to not have the `!`? I missed it immediately and it doesn't feel necessary with a ternary since it's easy to swap things around.

```suggestion
  const LeftSideHeaderContent = currentProduct
    ? ProductPageHeaderContent
    : HomePageHeaderContent
```

src/views/error-views/dev-dot-404/index.tsx
BRKalow: Makes sense, let's leave it!

src/views/error-views/dev-dot-404/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+      <ErrorViewH1>
+        We couldn&apos;t find the page you&apos;re looking&nbsp;for.
+      </ErrorViewH1>
+      <ErrorViewParagraph>
+        Please check the url you entered for typos,{' '}
+        <InlineLinkButton onClick={router.back}>go back</InlineLinkButton> to
+        the page you came from, or go to one of the pages below.
+      </ErrorViewParagraph>
+      <div className={s.cards}>
+        <CardsGridList>
```
ashleemboyer, line 36: 💭 I think you may want to use `IconCardLinkGridList` here, which should reduce the length of this code a bit I think!

src/components/inline-link-button/index.tsx
ashleemboyer: 🚧 I know we're still having a convo with design about this in Slack, but just to reiterate here for posterity:I feel very strongly that we do not place an element that looks identical to an inline link, inline with text, but code it as a button. This is a very large accessibility concern; larger than a link a that looks like a button. Our safest bet in this case is to use plain text and iterate on the page + a "go back" element that behaves like the browser back button.

src/components/button-link/button-link.module.css
```diff
@@ -52,6 +52,10 @@
     color: var(--token-color-foreground-disabled);
     cursor: not-allowed;
   }
+
+  &:hover:not(:disabled) {
+    text-decoration: underline;
+  }
```
ashleemboyer, line 58: 🚧 I believe we want this underline to show all the time based on the HDS component handles rendering links that look like buttons:

image

I was involved in the RFC that held the discussion about this, and the rationale is so that there is always a visual distinction between buttons and links-that-look-like-buttons.

ashleemboyer (changes requested):

src/views/error-views/dev-dot-404/index.tsx
zchsh: Had tried initially, unfortunately `IconCardLinkGridList` supports only a top-level `productSlug`, rather than individual `productSlug` values per card.

I felt it was easier and almost as clean to compose things this way. But happy to update `IconCardLinkGridList` to support say `defaultProductSlug` at the top level, and optional `productSlug` properties per `IconCard` if this feels like a blocker!

src/components/button-link/button-link.module.css
zchsh: Neat, apologies for missing this earlier, thanks for re-linking the screenshot! 👍

zchsh: ✅ Here's the updated appearance:

CleanShot 2022-05-27 at 08 43 30@2x

src/views/error-view-switcher/index.tsx
zchsh: ✅ Switched it up! The `if` does feel cleaner, and nice having an explicit type for `type` so that the options are clear up front 👍

src/lib/products.ts
zchsh: @BRKalow you converted me, I'm a TypeScript fan now ahah 🤣 I look back on my former disdain for TypeScript with a mix of guilt and nostalgia 🙇‍♂️ Next thing you know I'll be all-in working on Swingset 💖 😝

src/components/inline-link-button/index.tsx
zchsh: @ashleemboyer I've updated for `go back` to be plain text, for now at least, and have removed the `inline-link` component!

Definitely looking forward to getting more design input on this front (see Slack thread), but I'm very on board with having unlinked text here; the error page design still feels _way_ better than before since it has nice card-links.

zchsh:
@ashleemboyer Ready for re-review, I think!

1. I've switched the `go back` text from a "router.back() button that looks like an inline link" to plain text. Fully on board with scaling this back since there are a11y concerns; I think the design with card links below is a big improvement regardless 👍 cc @braydenlove (hopefully this is okay as an MVP! Definitely open to iterating further)
2. I've added a persistent (not just `:hover`) underline to `ButtonLink` (for the versioned 404 docs view use case).

src/views/error-view-switcher/index.tsx
```diff
@@ -63,29 +63,36 @@ function ErrorView({
...
-     * Dev dot error views
-     */
     if (isVersioned404) {
+      type = 'dev-dot-versioned-404'
+    } else if (is404) {
+      type = 'dev-dot-standard-404'
+    } else {
+      type = 'dev-dot-fallback'
+    }
+  }
```
BRKalow, line 88: Clean ✨  self-documenting and very easy to grok how the type is determined _and_ what is rendered for each type. 👏

src/components/inline-link-button/index.tsx
ashleemboyer: Agreed! It looks much better. 😊

src/views/error-views/dev-dot-404/index.tsx
ashleemboyer: Ah ok this makes sense! ✨ 

🎨 I'll leave it up to you if you want to make a change here or not! I think we probably want to make the change at some point.

---

💭 This is what I'd probably do to make `IconCardLinkGridList` more reusable:

- Make `productSlug` optional
- Update the `IconCardLink`s rendered there to accept `productSlug || card.productSlug`

💭 Another possible optimization: make `IconCardLink`'s `productSlug` default to `hashicorp`.

ashleemboyer (approved):
🥇

src/components/navigation-header/index.tsx
zchsh: 💖 Great suggestions here, and appreciate the option to defer! ✅ Updated with the above changes, markup does feel much cleaner ✨

src/components/icon-card-link-grid-list/index.tsx
```diff
@@ -4,19 +4,19 @@ import { IconCardLinkGridListProps, IconCard } from './types'
 
 function IconCardLinkGridList({
   cards,
-  productSlug,
+  productSlug = 'hcp',
```
ashleemboyer, line 7: This isn't quite what I was suggesting. This is now optional which is great, but I think this default value should be placed in `IconCardLink`'s `productSlug` property instead (has a wider impact). That would mean also making `IconCardLink`'s `productSlug` optional, and then the `IconCard` interface could be updated to this:

```tsx
export type IconCard = Pick<IconCardLinkProps, 'icon' | 'productSlug' | 'text' | 'url'>
```

and then `IconCardLinkGridList`'s would be effectively optional as well. Does that make sense?

ashleemboyer (changes requested):

src/components/icon-card-link-grid-list/index.tsx
zchsh: yep, makes snese! ✅ have moved the optional prop down a level

zchsh:
@ashleemboyer I've updated the `productSlug` prop to be optional on `IconCardLink`, as suggested. Let me know if that looks alright!

braydenlove (approved):
Awesome work @zchsh!!!! From design's perspective this is good to go.

ashleemboyer (approved):
Looks great!! 🙌🏻 Very much appreciate the `IconCardLink` side quest 📜🛡️⚔️💎

PR #478, opened 2022-05-25
chore: fix margin-bottom on alert component

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksfix-alert-bottom-margin-hashicorp.vercel.app/vault/docs/concepts/username-templating#examples) 🔎
- [Asana task](https://app.asana.com/0/1199634971449915/1202309283074066) 🎟️

## 🗒️ What

Fixes an issue where the alert component had no bottom margin. 

## 🧪 Testing

- Visit this PR and verify that the bottom margin is properly applied. See screenshot below for the previous view 
Screen Shot 2022-05-19 at 2 15 01 PM
- Also click around a few more docs pages to verify that the alert spacing looks fine.

### DIFF
--- a/src/components/dev-dot-content/dev-dot-content.module.css
+++ b/src/components/dev-dot-content/dev-dot-content.module.css
@@ -20,6 +20,7 @@
       border-radius: 6px;
       color: var(--token-color-palette-neutral-600);
       padding: 16px;
+      margin-bottom: 20px;
 
       & code {
         background-color: white;
--- a/src/components/dev-dot-content/dev-dot-content.module.css
+++ b/src/components/dev-dot-content/dev-dot-content.module.css
@@ -20,7 +20,7 @@
       border-radius: 6px;
       color: var(--token-color-palette-neutral-600);
       padding: 16px;
-      margin-bottom: 20px;
+      margin: 20px 0;
 
       & code {
         background-color: white;


src/components/dev-dot-content/dev-dot-content.module.css
```diff
@@ -20,6 +20,7 @@
       border-radius: 6px;
       color: var(--token-color-palette-neutral-600);
       padding: 16px;
+      margin-bottom: 20px;
```
kendallstrautman, line 23: I'm not sure what the recommended vertical rhythm is for these docs components? I chose `20px` to match the bottom margin of the table component below but please lmk if that value should be different.

ashleemboyer: So far, we've mostly added `margin-top` to elements inside of docs content. Brayden and I worked together closely on spacing between different kinds of content awhile ago, so you could check this value with him if you feel it's necessary!

In this case since we're already setting `margin-bottom` on the table component, I'd probably recommend setting the `margin-top` there as well so the relationship in spacing values is clearer.

zchsh (approved):

src/components/dev-dot-content/dev-dot-content.module.css
zchsh: (crossed wires there, +1 on Ashlee's take on setting both margin-top and margin-bottom. Once we have a more unified system, I think we'll be able to pare it back to margin-top only if we really want, though setting both gives us some nice flexibility due to margin-collapse which may come in handy in meeting design spacing needs).

ashleemboyer: > Once we have a more unified system, I think we'll be able to pare it back to margin-top only if we really want

+1!

Also, @zchsh and I are always commenting at almost the same time. 🤣

braydenlove: We initially were using 16px but 20px is also pretty nice. Let's not over engineer this and move forward, can refine later as needed.

braydenlove (approved):
Nice work @kendallstrautman, I clicked through everything in Vault and about half of Waypoint and it looks like you fixed everything!

src/components/dev-dot-content/dev-dot-content.module.css
braydenlove: @ashleemboyer @zchsh sorry I saw these after I put my comment in! Let's work on a more unified system for how we handle margins (post June 20th) :)

kendallstrautman: Awesome! i'll set `margin-top` to 20px ⛵

PR #486, opened 2022-05-26
Adds opt-out-feedback API endpoint

BrandonRomano:

- [x] ~The Vercel preview link has been added to this PR's description~
- [x] ~This Vercel preview link has been added to the "🔍 Preview" field on the Asana task~

---

## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1202276096167515/1202114007623817/f) 🎟️

## 🗒️ What

Adds in an API endpoint for us to be able to send opt-out feedback to a Google Sheet.

## Testing

```sh
curl -X POST \
  https://dev-portal-git-bropt-out-feedback-hashicorp.vercel.app/api/opt-out-feedback \
  -H 'Content-Type: application/json' \
  -d '{
	"segment_anonymous_id": "random-uuid-here",
	"primary_opt_out_reason": "I can'\''t find the content I need",
	"details": "some-details-here",
	"opt_out_page_url": "https://developer.hashicorp.com/vault/api-docs"
}'
```

[Link to Google Sheet](https://docs.google.com/spreadsheets/d/15KyBwiEui6YfatLkN0QXcee3TR-pew_sVhnb9nk0DWo/edit?usp=sharing) (only accessible w/ HashiCorp email)

https://user-images.githubusercontent.com/2105067/170577072-f68b6214-e35a-4c5d-bb15-1a5ad80b3dd2.mp4

### DIFF
--- a/src/pages/api/opt-out-feedback.ts
+++ b/src/pages/api/opt-out-feedback.ts
@@ -1,0 +1,118 @@
+import { NextApiRequest, NextApiResponse } from 'next'
+import {
+  GoogleSpreadsheet,
+  GoogleSpreadsheetRow,
+  GoogleSpreadsheetWorksheet,
+} from 'google-spreadsheet'
+
+const FEEDBACK_SHEET_ID = process.env.FEEDBACK_SHEET_ID
+const FEEDBACK_SERVICE_EMAIL = process.env.FEEDBACK_SERVICE_EMAIL
+const FEEDBACK_PRIVATE_KEY = process.env.FEEDBACK_PRIVATE_KEY
+
+export default async function handler(
+  request: NextApiRequest,
+  response: NextApiResponse
+) {
+  // Only allow POST for this endpoint
+  if (request.method !== 'POST') {
+    response.status(404).json({
+      meta: {
+        status_code: 404,
+        status_text: 'Not Found',
+      },
+    })
+    return
+  }
+
+  // Validate Request Headers
+  if (request.headers['content-type'] !== 'application/json') {
+    response.status(400).json({
+      meta: {
+        status_code: 400,
+        status_text: 'Bad Request',
+      },
+      errors: [
+        {
+          msg: 'expects header Content-Type: application/json',
+        },
+      ],
+    })
+    return
+  }
+
+  // Ensure `primary_opt_out_reason` is specified
+  if (!request.body.primary_opt_out_reason) {
+    response.status(400).json({
+      meta: {
+        status_code: 400,
+        status_text: 'Bad Request',
+      },
+      errors: [
+        {
+          msg: 'Required body key `primary_opt_out_reason` is not provided',
+        },
+      ],
+    })
+    return
+  }
+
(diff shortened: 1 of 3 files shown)


src/pages/api/opt-out-feedback.ts
```diff
@@ -0,0 +1,118 @@
...
+  GoogleSpreadsheet,
+  GoogleSpreadsheetRow,
+  GoogleSpreadsheetWorksheet,
+} from 'google-spreadsheet'
+
+const FEEDBACK_SHEET_ID = process.env.FEEDBACK_SHEET_ID
+const FEEDBACK_SERVICE_EMAIL = process.env.FEEDBACK_SERVICE_EMAIL
+const FEEDBACK_PRIVATE_KEY = process.env.FEEDBACK_PRIVATE_KEY
+
+export default async function handler(
```
BRKalow, line 12: Maybe not a huge deal as it could be sniffed pretty easily, but do we want to require an API key via a header here too?

BrandonRomano: Good question. 

 Maybe a better protection is actually wiring up CORS / header sniffing so the request can only be sent from developer.hashicorp.com 🤔 ?

We did something like this on Learn's feedback form (which I drew inspiration from), and that seemingly was enough to prevent abuse: https://github.com/hashicorp/learn/blob/master/pages/api/feedback.ts#L62-L69

BRKalow (approved):
Implementation looks great! Nice job cranking this out.

src/pages/api/opt-out-feedback.ts
BRKalow: Yeah, good call, CORS would be appropriate here. It wouldn't necessarily stop someone from making requests from a non-browser env if we only used the proper CORS header (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin), but your origin check is effective too.

I think any approach here would be adequate!

BrandonRomano: I think I'll do CORS (cause we will need it to allow the requests) + the origin check (to add a barrier to abuse).

src/pages/api/opt-out-feedback.ts
```diff
@@ -0,0 +1,118 @@
...
+  // Ensure `primary_opt_out_reason` is specified
+  if (!request.body.primary_opt_out_reason) {
+    response.status(400).json({
+      meta: {
+        status_code: 400,
+        status_text: 'Bad Request',
+      },
+      errors: [
+        {
+          msg: 'Required body key `primary_opt_out_reason` is not provided',
```
kendallstrautman, line 52: On the form, we have the opt out reason as optional, do we want to make this required?

kendallstrautman (approved):
👏

src/pages/api/opt-out-feedback.ts
BrandonRomano: Oh!  Okay yeah so *both* are optional?   We definitely want to treat it as such, then.

PR #492, opened 2022-05-27
fix: prevent opt-in-slot shrink, truncate breadcrumbs

zchsh:

---

## 🔗 Relevant links

- [Preview link][/waypoint/tutorials/application-deployment/gitops-helm-deployment?optInFrom=learn] 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1202227071718235/f) 🎟️

## 🗒️ What

Fixes an issue where:
1. the opt-in-out slot in sidebar-sidecar layout shrank, causing text-wrapping within a button
2. long items in breadcrumb-bar were not truncated, causing the above issue to happen more frequently, especially at smaller viewport sizes

Also updates:
- Styles to better match HDS spec (implements `:hover`, `:focus`, and `:active` styles)
- Learn pages, docs landing, and product landing to correctly show the active breadcrumb item

## 🤷 Why

- Text wrapping within button doesn't match spec
- Breadcrumb truncation matches design intent
    - Note: the component is currently already responsive, in that items wrap, however it can be a bit overwhelming when many of the items have very long titles

## 🛠️ How

- Adds `flex-shrink: 0` to the opt-in slot in sidebar-sidecar layout, to prevent button text wrapping
- Adds CSS truncation to `breadcrumb-bar` items

## 📸 Screenshots

Note: we don't have a specific design spec for truncation, I'm working off the [HDS component](https://design-system-components-hashicorp.vercel.app/components/breadcrumb) as a guide, and have eyeballed the truncation point.

### Before

before

### With `flex-shrink: 0` on opt-in slot

after-no-flex-shrink-only

### With `flex-shrink: 0` and item truncation

after-no-flex-shrink-and-truncation

### Example where truncation prevents wrap at larger viewports

> Note that a large percentage of visitors have viewports at or wider than 1440px wide. So this case is rather common.

truncation-prevents-wrap-example-before

truncation-prevents-wrap-example-after

## 🧪 Testing

- [ ] Visit [/waypoint/tutorials/application-deployment/gitops-helm-deployment?optInFrom=learn][/waypoint/tutorials/application-deployment/gitops-helm-deployment?optInFrom=learn]
    - Ensure truncation is working as expected
- [ ] Browse other pages with breadcrumb bars, such as other docs and learn pages, and ensure they continue to render correctly even when truncation does not come into play

## 💭 Anything else?

Nope!

[/waypoint/tutorials/application-deployment/gitops-helm-deployment?optInFrom=learn]: https://dev-portal-git-zstruncate-breadcrumbs-hashicorp.vercel.app/waypoint/tutorials/application-deployment/gitops-helm-deployment?optInFrom=learn


package.json
```diff
@@ -134,6 +134,8 @@
     "@reach/visually-hidden": "^0.16.0",
     "@react-aria/ssr": "^3.1.0",
     "@react-aria/utils": "^3.11.0",
+    "@types/cors": "^2.8.12",
+    "@types/google-spreadsheet": "^3.2.2",
```
zchsh, line 138: Note: this is an unrelated change, I was getting typescript errors due to the lack of these types packages and couldn't commit anything as a result. Not sure how these slipped through - I think possibly in #486 ?

BRKalow (approved):
Good catch!

src/components/breadcrumb-bar/breadcrumb-bar.module.css
```diff
@@ -18,6 +18,8 @@
 }
 
 .listItem {
+  display: flex;
```
ashleemboyer, line 21: 🚧 / 🙋🏻 -- Was this added only to get the text to truncate?

zchsh: Yep! This was done cause `breadcrumbText` would otherwise render as `display: inline` (since it's a span), which means the `max-width` doesn't seem to work 👍

braydenlove (approved):
This looks good Zach, thanks. Assuming that the breadcrumb styling refinements will come in a seperate PR and that this one is just figuring out the truncation functionaltity?

src/components/breadcrumb-bar/breadcrumb-bar.module.css
ashleemboyer: Ok got it! Thank you for the extra background, I thought that's what was going on. I think rather than setting `display: flex` on the `.listItem` class, we should set `display: block` on the `.breadcrumbText` class. This will isolate changes to the single class that needs them, and I think it'll be clearer at a glance what CSS is needed for the desired effect.

ashleemboyer (changes requested):

src/components/breadcrumb-bar/breadcrumb-bar.module.css
ashleemboyer: Well shoot, that'll mess up the `::after` elements and then `display: flex` would be needed anyways on `.listItem`. Nevermind! 😎

ashleemboyer (approved):

src/components/breadcrumb-bar/breadcrumb-bar.module.css
zchsh: I'll add a lil docs on this, your second comment sums it up great! 💖

zchsh:
@braydenlove Ah great flag, ty! I'll add those style changes onto this PR 💖 one sec!

src/components/breadcrumb-bar/breadcrumb-bar.module.css
ashleemboyer: Nice! Greatly appreciated 🙌🏻

zchsh:
@braydenlove I've updated the styles here, let me know if they look alright!

@ashleemboyer When making the style tweaks, I noticed the "current page" state was only reflected within docs page. So added in a fix to reflect "current page" for `/{product}`, `/{product}/docs`, as well as all tutorial views.

src/components/breadcrumb-bar/breadcrumb-bar.module.css
```diff
@@ -26,10 +33,32 @@
 }
 
 .breadcrumbText {
-  color: var(--token-color-palette-neutral-500);
+  composes: g-focus-ring-from-box-shadow from global;
```
zchsh, line 36: @ashleemboyer Update since last review: these are the updates for breadcrumb link styles.

src/views/product-landing/server.ts
```diff
@@ -66,7 +66,7 @@ async function generateStaticProps({
       headings: extractHeadings(content),
       breadcrumbLinks: [
         { title: 'Developer', url: '/' },
-        { title: product.name, url: `/${product.slug}` },
+        { title: product.name, url: `/${product.slug}`, isCurrentPage: true },
```
zchsh, line 69: @ashleemboyer Update since last review: in this server file and a couple others, setting `isCurrentPage` on the correct breadcrumb as needed.

I realized as I was doing this we could likely simplify the logic here: if we can trust our breadcrumb bar generation, then the last item in a breadcrumb bar should be marked as current page (which we could do easily on the front end, no need to pass `isCurrentPage` through). Make a task for that as it felt outside the scope of this PR: 🎟️ [[breadcrumb-bar] simplify isCurrentPage logic - always highlight last item, rather than based on href](https://app.asana.com/0/1202097197789424/1202354347457831/f)

zchsh:
@braydenlove Great catches, updated! 💖 ✅

braydenlove:
@zchsh the spacing looks so good thank you. The Selected Page styling looks to be lost now though (could be an updating PR thing or something so ignore me if it is still working).

image
image

zchsh:
dang it, great call - should be fixed now

braydenlove:
Nice looks great to me, approved from the design side.

src/components/breadcrumb-bar/index.tsx
```diff
@@ -39,18 +39,19 @@ function BreadcrumbBar({
       <ol className={s.listRoot}>
         {links.map(({ title, url, isCurrentPage }) => {
           const cleanTitle = title.replace(/<[^>]+>/g, '')
-          const Elem = url ? InternalLink : 'span'
+          const shouldLink = url && !isCurrentPage
+          const Elem = shouldLink ? InternalLink : 'span'
```
ashleemboyer, line 43: 🚧 I very much disagree with this change, and disagree with the design system component doing this. Can we wait to make this breaking change until we know more about why the DS team decided on this? [The APG pattern we based this component off](https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/) of uses a link in the last item so it's reachable via keyboard/assistive technology and I'm not confident we want to break that.

src/components/breadcrumb-bar/utils/get-docs-breadcrumbs.ts
```diff
@@ -161,7 +161,11 @@ function getDocsBreadcrumbs({
   return [
     { title: 'Developer', url: '/' },
     { title: productName, url: `/${productPath}` },
-    { title: baseName, url: `/${productPath}/${basePath}` },
+    {
+      title: baseName,
+      url: `/${productPath}/${basePath}`,
+      isCurrentPage: pathParts.length == 0,
+    },
```
ashleemboyer, line 168: 🙌🏻 Great catch!

src/views/tutorial-view/utils/get-tutorials-breadcrumb.ts
```diff
@@ -16,22 +16,28 @@ export function getTutorialsBreadcrumb({
   collection,
```
ashleemboyer, line 16: 🙌🏻 Great catches here too!

ashleemboyer (changes requested):

zchsh:
@ashleemboyer @braydenlove Based on Ashlee's feedback, I've made the `current` breadcrumb link work again.

The intent with this change is to match [Web Accessibility Initiative authoring practices guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/), but at the cost of compromising on full adherence to [the HDS spec](https://design-system-components-hashicorp.vercel.app/components/breadcrumb). I've made a follow-up task to reconcile the differences:
- 🎟️ [[breadcrumb-bar] resolve HDS spec (don't link current page item) vs WAI Authoring Practices Guide (link it)](https://app.asana.com/0/1202097197789424/1202355385649681/f)

src/components/breadcrumb-bar/index.tsx
zchsh: Made a task for this in 🎟️ [[breadcrumb-bar] resolve HDS spec (don't link current page item) vs WAI Authoring Practices Guide (link it)](https://app.asana.com/0/1202097197789424/1202355385649681/f) ☺️ 

I did notice the WAI guidelines seem to imply the current page being a link with `aria-current="page"` as optional:

> If the element representing the current page is not a link, aria-current is optional.

braydenlove:
You all do whatever, I am just trying to align with the design systems team.

ashleemboyer (approved):

src/components/breadcrumb-bar/index.tsx
ashleemboyer: Thank you for making the ticket! Just want to make sure we have time to discuss and document in the code/Swingset page. 😊

PR #497, opened 2022-05-27
feat: usePreloadNextDynamic, lazy load dialog for opt-out flow

BRKalow:

---

## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Adds a new hook `usePreloadNextDynamic`, which preloads the assets for a `next/dynamic` instance during idle time. Can be used to lazily preload components which we expect to need in the near-term.

With vanilla `next/dynamic`, see how the request comes before the red line (which is the load event):
Screen Shot 2022-05-27 at 1 48 10 PM

With `usePreloadNextDynamic`, see the request come after the load event during "idle" time:
Screen Shot 2022-05-27 at 1 48 33 PM

## 🤷 Why

The dialog component is like 10kb gzipped and not needed for initial load, so deferring it!

### DIFF
--- a/src/components/opt-in-out/index.tsx
+++ b/src/components/opt-in-out/index.tsx
@@ -1,11 +1,11 @@
 import { useCallback, useEffect, useState } from 'react'
 import { useRouter } from 'next/router'
+import dynamic from 'next/dynamic'
 import Cookies from 'js-cookie'
 import { IconSignOut16 } from '@hashicorp/flight-icons/svg-react/sign-out-16'
 import '@reach/dialog/styles.css'
 import Button from 'components/button'
 import OptOutForm from './components/opt-out-form'
-import Dialog from 'components/dialog'
 import { getLearnRedirectPath } from './helpers/get-learn-redirect-path'
 import {
   PlatformOptionRedirectData,
@@ -15,6 +15,9 @@ import {
 import { safeAnalyticsTrack } from 'lib/analytics'
 import makeBetaWelcomeToast from './helpers/make-beta-welcome-toast'
 import { getIoRedirectPath } from './helpers/get-io-redirect-path'
+import usePreloadNextDynamic from 'hooks/use-preload-next-dynamic'
+
+const Dialog = dynamic(() => import('components/dialog'))
 
 // Could these go in the config? or I could source the base urls elsewhere
 export const PLATFORM_OPTIONS: PlatformOptionRedirectData = {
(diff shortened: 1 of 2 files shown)


src/components/opt-in-out/index.tsx
```diff
@@ -92,13 +98,19 @@ export default function OptInOut({ platform, redirectPath }: OptInOutProps) {
...
+          isOpen={showDialog}
+          label="Opt out form"
+        >
+          <OptOutForm
+            onSubmit={handleOptOut}
+            onDismiss={closeDialog}
+            platform="learn"
+          />
+        </Dialog>
+      ) : null}
```
BRKalow, line 113: This is behind the conditional because if you just render a `next/dynamic` component, it will load the assets right away. In this case, we don't want that to happen, so we add the `isDialogPreloaded` check and only render the component if someone clicks the button OR the component is loaded.

thiskevinwang: Does this impact any animation/transitions on `Dialog`?

BRKalow: Nope, that's why `showDialog` is part of the condition. If someone tries to open it, it'll render as normal so the animations will work.

src/hooks/use-preload-next-dynamic.ts
```diff
@@ -0,0 +1,44 @@
...
+) => {
+  const [isPreloaded, setIsPreloaded] = useState(false)
+
+  useEffect(() => {
+    if (isPreloaded) {
+      return
+    }
+
+    const load = () => {
+      // @ts-expect-error - this dose not exist on next/dynamic's type, but it is there!
```
thiskevinwang, line 22:
```suggestion
      // @ts-expect-error - this does not exist on next/dynamic's type, but it is there!
```

src/components/opt-in-out/index.tsx
```diff
@@ -92,13 +98,19 @@ export default function OptInOut({ platform, redirectPath }: OptInOutProps) {
         iconPosition="trailing"
         onClick={openDialog}
       />
-      <Dialog onDismiss={closeDialog} isOpen={showDialog} label="Opt out form">
-        <OptOutForm
-          onSubmit={handleOptOut}
+      {isDialogPreloaded || showDialog ? (
```
thiskevinwang, line 101: Open thought: I wonder if a wrapper component would a more ergonomic interface for this?

```
<PreloadNextDynamic show={boolean}>
```

BRKalow: What would the implementation / usage look like?

thiskevinwang: Hmm, maybe nevermind! I was thinking we could abstract away the conditionals. But looking at it a bit closer, there's not much net improvement / additional interface is still required...

```diff
- {isDialogPreloaded || showDialog ? (
+ <PreloadNextDynamic component={Dialog}>

- ) : null}
+ </PreloadNextDynamic>
```

BRKalow: Ah word gotcha, yeah that could be an option too! I think in conjunction with the hook that would give some nice flexibility.

I think preloading is generally a pretty imperative action, so entrenching it into component/rendering flow isn't always what we want.

PR #498, opened 2022-05-27
feat: API data for install vuew tutorial cards

zchsh:

---

## 🔗 Relevant links

- [Asana task](url) 🎟️


src/data/vault-install.json
```diff
@@ -1,34 +1,22 @@
 {
   "featuredTutorials": [
     {
-      "title": "Getting Started",
-      "description": "A quick tour of common Vault features to jump-start.",
-      "href": "https://learn.hashicorp.com/collections/vault/getting-started"
+      "collectionSlug": "vault/getting-started"
```
zchsh, line 4: I've used an array of objects here so that:
1. A mix of collections and tutorials can be used (to match previous data)
2. So that the data works nicely with existing "extract inline content" utilities, which look for `collectionSlug` and `tutorialSlug` keys

src/pages/vault/downloads/index.tsx
```diff
@@ -3,25 +3,33 @@ import { GetStaticProps } from 'next'
...
 }
 
 export const getStaticProps: GetStaticProps = async () => {
-  const product = vaultData as ProductData
-
-  return generateStaticProps(product)
+  return await generateStaticProps(
+    vaultData as ProductData,
+    installData as unknown as RawProductDownloadsViewContent
+  )
```
zchsh, line 32: Rather than passing `installData` directly to the view component, we now do some server-side stuff first - namely, fetching the tutorial and collection data based on authored slugs.

src/views/product-downloads-view/components/featured-tutorials-section/featured-tutorials-section.module.css
```diff
@@ -1,49 +1,7 @@
...
 }
 
-/*
-TODO: these @media queries come from the numbers defined in src/pages/style.css
-but I wasn't able to get the @media --dev-dot-mobile and @media --dev-dot-tablet
-queries to work like they do in that file. The code below will eventually need
-to be refactored, this is a temporary copy/paste to get the new install view
-out.
-*/
 .cardGrid {
```
zchsh, line 5: Now using `CardsGridList`, which handles the styling needed here.

src/views/product-landing/components/product-landing-blocks/blocks/collection-cards/collection-cards.module.css
```diff
@@ -1,5 +0,0 @@
-.placeholder {
```
zchsh, line 1: These `product-landing-blocks` CSS files are unrelated to this PR, I noticed them when elevating the `getInlineContentMaps` utility. They need to be cleaned up; they're not used anywhere.

zchsh:
@kendallstrautman @ashleemboyer This PR updates the `/{product}/downloads` views' `FeaturedTutorial` cards to source from the Learn API.

---

One thing I want to call out - there's some ambiguity in the use of the word "tutorials" for the thing we previously called "learn" (the entire content type, i guess) vs the specific "tutorial" documents vs "collections" (part of the tutorials platform; but not a tutorial document type).

This manifests here in that the designs show only `TutorialCards`, but our data for these pages mixes `TutorialCards` and `CollectionCards`. This implementation aims for the latter. We could simplify it if we narrow down to _either_ `TutorialCards` or `CollectionCards` - I'm going to check with design on what the original intent was 👍

ashleemboyer:
Taking a look now! 😊

src/data/vault-install.json
ashleemboyer: This makes sense to me! 👍🏻 Making the property names match the expected key names is helpful too.

src/pages/vault/downloads/index.tsx
ashleemboyer: Good call! I think that's the right approach.

🙋🏻 Why does `installData` need to be first cast as `unknown` before `RawProductDownloadsViewContent`? It's not needed for `vaultData`+ `ProductData`, so it's not clear to me yet why it's needed for `installData` + `RawProductDownloadsViewContent`.

src/views/product-downloads-view/types.ts
```diff
@@ -1,40 +1,77 @@
...
+
 export interface ProductDownloadsViewProps {
   latestVersion: string
   pageContent: {
     doesNotHavePackageManagers?: boolean
-    featuredTutorials?: FeaturedTutorial[]
+    featuredTutorialCards?: FeaturedTutorialCard[]
     packageManagerOverrides?: PackageManager[]
     sidecarMarketingCard: SidecarMarketingCardProps
   }
```
ashleemboyer, line 48: 💭 I think we probably want to update this to extend `RawProductDownloadsViewContent`, to make the relationship between `RawProductDownloadsViewContent` and `ProductDownloadsViewProps['pageContent']` clearer.

```tsx
// this would be new
interface ProductDownloadsViewPageContentProp
  extends RawProductDownloadsViewContent {
  featuredTutorialCards?: FeaturedTutorialCard[]
}

export interface ProductDownloadsViewProps {
  latestVersion: string
  pageContent: ProductDownloadsViewPageContentProp
  releases: ReleasesAPIResponse
}
```

src/views/product-downloads-view/types.ts
```diff
@@ -1,40 +1,77 @@
...
+ */
+export type FeaturedTutorialContent =
+  | {
+      collectionSlug: string
+      tutorialSlug: never
+    }
+  | {
+      tutorialSlug: string
+      collectionSlug: never
+    }
```
ashleemboyer, line 18: I think these kinds of types are so fun! 🙌🏻

🚧 There is one thing I think needs updated, and maybe we're not seeing an error pop up because the content is coming from a JSON file and not a TS file. I'm fairly certain the properties set to `never` also need to be optional.

🎨 Would also prefer the properties to be alphabetical in both parts, but totally up to you!

```suggestion
/**
 * Raw page content, as structured in .json files
 */
export type FeaturedTutorialContent =
  | {
      collectionSlug: string
      tutorialSlug?: never
    }
  | {
      collectionSlug?: never
      tutorialSlug: string
    }
```

src/views/product-downloads-view/components/featured-tutorials-section/featured-tutorials-section.module.css
ashleemboyer: This was probably satisfying to remove 😎

BRKalow:
This is a very exciting pattern! Yay content APIs 😍

src/views/product-downloads-view/index.tsx
```diff
@@ -89,10 +89,12 @@ const ProductDownloadsViewContent = ({
...
         <OfficialReleasesSection />
-        {featuredTutorials && (
-          <FeaturedTutorialsSection featuredTutorials={featuredTutorials} />
-        )}
       </DevDotContent>
+      {featuredTutorialCards ? (
+        <FeaturedTutorialsSection
+          featuredTutorialCards={featuredTutorialCards}
+        />
+      ) : null}
```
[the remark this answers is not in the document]
ashleemboyer: 💭 If I recall correctly, I think `DevDotContent` is used here to make the permalink headings have the correct `scroll-margin-top` set on them.

🚧 I just tested if this link would be affected by this change: https://dev-portal-git-zsinstall-view-tutorial-cards-hashicorp.vercel.app/waypoint/downloads#featured-tutorials. It is, so could you make a few more changes to go along with these updates?

- Update the `.sectionHeading` class to compose the `g-offset-scroll-margin-top` helper
- Add an Asana ticket to the backlog for removing `DevDotContent` from this view, noting that the `g-offset-scroll-margin-top` helper will be needed for headings
- Add a TODO in this file referencing the Asana ticket

Last two are only if you agree of course, but as far as I can tell `DevDotContent` isn't needed in this view anymore!

src/views/product-downloads-view/components/featured-tutorials-section/index.tsx
```diff
@@ -30,30 +25,16 @@ const FeaturedTutorialsSection = ({
...
       </Heading>
       <div className={s.cardGrid}>
-        {featuredTutorials.map(({ title, description, href }) => {
-          return (
-            /**
-             * TODO: these will more than likely be replaced by a future
-             * `LearnTutorialCard` component.
-             *
-             * ref: https://app.asana.com/0/1201010428539925/1201654639085737/f
-             */
```
ashleemboyer, line 40: 🎉

src/pages/vault/downloads/index.tsx
```diff
@@ -3,25 +3,33 @@ import { GetStaticProps } from 'next'
...
+}: ProductDownloadsViewStaticProps): ReactElement => {
   return (
     <ProductDownloadsView
       latestVersion={latestVersion}
-      pageContent={installData}
+      pageContent={pageContent}
       releases={releases}
     />
   )
 }
```
ashleemboyer, line 26: 💭 Do you think we can update these page components to do just `export default ProductDownloadsView` and update `ProductDownloadsView` to set its `layout` to `CoreDevDotLayout`? Not at all required for this PR, just wondering if we want to update it while we're here or not. Totally up to you!

ashleemboyer (changes requested):
This is looking great so far! 🙌🏻

---

> One thing I want to call out - there's some ambiguity in the use of the word "tutorials" for the thing we previously called "learn" (the entire content type, i guess) vs the specific "tutorial" documents vs "collections" (part of the tutorials platform; but not a tutorial document type).
>
> This manifests here in that the designs show only TutorialCards, but our data for these pages mixes TutorialCards and CollectionCards. This implementation aims for the latter. We could simplify it if we narrow down to either TutorialCards or CollectionCards - I'm going to check with design on what the original intent was 👍

Once you've got clarity on this from design, if it turns out that we want to show both tutorials and collections, I think we need to update the language from "tutorials" to "learn content". Would definitely help bring clarity to the code. 👍🏻

src/pages/vault/downloads/index.tsx
zchsh: Great call, absolutely! ✅ Updated.

src/pages/vault/downloads/index.tsx
zchsh: Ah good callout, I meant to ask about this.

I'm not sure why it's needed, but TypeScript gets real mad if I don't do this:

CleanShot 2022-05-31 at 13 38 03@2x

I figured we must be setting up types for this data somewhere? But couldn't find anything. So I'm unsure why casting to unknown is needed.

src/views/product-downloads-view/types.ts
zchsh: Personally I find this adds another layer of complexity to some already complex types - I think to be fully accurate we'd need to do:

```ts
interface ProductDownloadsViewPageContentProp
  extends Omit<RawProductDownloadsViewContent, 'featuredTutorials'> {
  featuredTutorialCards?: FeaturedTutorialCard[]
}

export interface ProductDownloadsViewProps {
  latestVersion: string
  pageContent: ProductDownloadsViewPageContentProp
  releases: ReleasesAPIResponse
}
```

Strong opinion weakly held, but I think it's easier to figure out the shape of props here without these extra layers.

src/views/product-downloads-view/index.tsx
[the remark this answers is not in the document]
zchsh: Ah cool this makes a lot of sense - I forgot about `offset-scroll-margin-top`!

- ✅ Have updated to use `g-offset-scroll-margin-top` on `.sectionHeading`
- ✅ Have added a task: 🎟️ [[install view] remove DevDotContent, use g-offset-scroll-margin-top for headings instead](https://app.asana.com/0/1202097197789424/1202370169937866/f) and linked to it in a comment 👍

src/views/product-downloads-view/server.ts
```diff
@@ -0,0 +1,77 @@
...
+        const collectionData = inlineContent.inlineCollections[collectionSlug]
+        return { type: 'collection', ...formatCollectionCard(collectionData) }
+      } else if (typeof tutorialSlug == 'string') {
+        const tutorialData = inlineContent.inlineTutorials[tutorialSlug]
+        const defaultContext = tutorialData.collectionCtx.default
+        const tutorialLiteCompat = { ...tutorialData, defaultContext }
+        return { type: 'tutorial', ...formatTutorialCard(tutorialLiteCompat) }
+      }
+    }
+  )
```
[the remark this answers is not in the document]
zchsh: great question - I'm not sure the original term was finessed all that much, it's a holdover from [the learn repo](https://github.com/hashicorp/learn/blob/79fa1971f60b30038d2ee8da0ed056497994c900/pages/consul/index.jsx#L10). I think the intent is "stuff that authors reference in their content, without having to type out the full data". 

I'm definitely open to renaming, but would prefer to handle outside this PR - there's a bunch of places we'd need to change it! Would personally lean towards deferring the rename till dev-dot is in GA (since related utilities in `hashicorp/learn` still have to be maintained)

src/views/product-downloads-view/components/featured-tutorials-section/featured-tutorials-section.module.css
zchsh: Haha, twas indeed 😬🔪

zchsh:
@ashleemboyer Thanks for the feedback! 💖 

- Have added in `g-offset-scroll-margin-top` to fix the heading issue (and added related TODO comment)
- Have confirmed with design, mix is good, and have renamed to `featuredLearnCards` (& made some similar type renames) for clarity

src/views/product-downloads-view/components/featured-tutorials-section/index.tsx
```diff
@@ -1,23 +1,18 @@
 import { ReactElement } from 'react'
-import { IconPlay16 } from '@hashicorp/flight-icons/svg-react/play-16'
-import { useCurrentProduct } from 'contexts'
-import getTruncatedText from 'lib/get-truncated-text'
-import { FeaturedTutorial } from 'views/product-downloads-view/types'
-import CardLink from 'components/card-link'
+import { FeaturedTutorialCard as FeaturedLearnCard } from 'views/product-downloads-view/types'
```
ashleemboyer, line 2: 🚧 Noticed that this had an alias added to it in a subsequent commit. Can we just rename the new `FeaturedTutorialCard` to `FeaturedLearnCard` in `src/views/product-downloads-view/types.ts` since it's more accurate?

src/pages/vault/downloads/index.tsx
ashleemboyer: (sorry didn't mean to resolve!)

src/views/product-downloads-view/server.ts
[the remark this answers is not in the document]
ashleemboyer: This makes sense! Totally reasonable to wait to rename until GA when there's less to maintain. 💖

src/views/product-downloads-view/types.ts
ashleemboyer: I see what you're saying! `ProductDownloadsViewPageContentProp` would also reduce some duplication, but fine with deferring that for now.

---

Just for the sake of discussion (no action needed), I think these two have the same output as the code currently stands so I think it becomes a matter of preference?

```ts
interface ProductDownloadsViewPageContentProp
  extends RawProductDownloadsViewContent {
  featuredTutorialCards?: FeaturedTutorialCard[]
}

interface ProductDownloadsViewPageContentProp
  extends Omit<RawProductDownloadsViewContent, 'featuredTutorials'> {
  featuredTutorialCards?: FeaturedTutorialCard[]
}
```

src/views/product-downloads-view/components/featured-tutorials-section/index.tsx
zchsh: Absolutely, this was a slip up, VSCode's "rename symbol" isn't as good as I thought it was 😬 
✅ Should be fixed now!

ashleemboyer (commented):
Looks great! ✨

src/views/product-downloads-view/types.ts
zchsh: ☝️ Great question! Definitely optional 💭 thought here, but here's my take:

---

In the first example above, without `Omit`, I think it's possible that `ProductDownloadsViewPageContentProp` would have an optional property `featuredTutorialContent` (fka `featuredTutorials`)? But I think we might want to make sure that property isn't present in the type, for clarity and to avoid confusion.

To be more concrete, without `Omit`, it's possible to include `featuredLearnContent` slugs, even though they won't be used on the front end:

without-omit

With `Omit` this gets flagged (it won't be used on the front-end, only `featuredLearnCards` will be, so it feels worth flagging to me): 

with-omit

---

More broadly, I think what's meant by `output` is maybe a key there. Personally, one "output" of type files for me is of course what's actually happening when types are being checked (and `Omit` feels important to me there). 

Another though is being able to look at the type file itself, and at type annotations, and hover or go-to-definitions etc and get meaningful information as quickly as I can (so that I don't get thrown out of whatever context I'm in). I find the extra layer means I have to "go to definition" an extra time to get meaningful information:

CleanShot 2022-05-31 at 23 18 45@2x

Where being slightly repetitive and explicit surfaces things more quickly:

CleanShot 2022-05-31 at 23 21 19@2x

Definitely agreed it feels like a matter of preference though... I think in this case, unlike with `Omit`, the type checking will shake out the exact same. Personally I find a bit of repetition is worth it for types that feel a bit more easy-to-dig-through. I think it's about the same amount of type-safe either way, so personally I would rather leave the types explicit and keep the abstraction and engineering of types to a minimum.

zchsh:
@ashleemboyer I'm making an assumption based on the text comment of your last review, but I think you may have hit `Comment` rather than `Approve` 😁 I might be missing something though, if there's any more blocking feedback here let me know and I'll do my best to address 🙇‍♂️

ashleemboyer (approved):
Ahh! I totally meant to check Approve when I commented last. Sorry about that!! Hate when that happens 😆

PR #509, opened 2022-05-31
Add version to docs sidebar overview item href

ashleemboyer:
## ✅ "Ready for Review" checklist

- [x] The Vercel preview link has been added to this PR's description
- [x] The Asana task has been added to this PR's description
- [x] This PR's link has been added to the "📐 Pull Request" field on the Asana task
- [x] This Vercel preview link has been added to the "🔍 Preview" field on the Asana task
- [x] The description template has been filled in below

---

## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambadd-version-to-overview-item-href-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1201903760348480/1202364251355937/f) 🎟️

## 🗒️ What

- Added some comments to separate imports in `Sidebar` in my first approach, thought they were nice so I left them in
- Updated `SidebarSidecarLayout`'s server code to determine if there is a version being viewed, and build `overviewItemHref` based on that

## 🤷 Why

- All docs `Sidebar` nav data includes versions in it, but the `overviewItemHref` had not been coded to handle that yet
- This approach taken will help in the next PR for updating the breadcrumb to handle looking at a past version

## 🛠️ How

- First, the `context.params.page` array is checked to see if there is a version in it matching any version in the `versions` `VersionSelectItem` array passed to `SidebarSidecarLayout`
- Then, if a version is found, `overviewItemHref` has the found version appended to its `href`

### Learn pages

- [ ] Vault tutorials landing: [/vault/tutorials](https://dev-portal-git-ambadd-version-to-overview-item-href-hashicorp.vercel.app/vault/tutorials)
- [ ] A Vault collection: [/vault/tutorials/getting-started](https://dev-portal-git-ambadd-version-to-overview-item-href-hashicorp.vercel.app/vault/tutorials/getting-started)
- [ ] Waypoint tutorials landing: [/waypoint/tutorials](https://dev-portal-git-ambadd-version-to-overview-item-href-hashicorp.vercel.app/waypoint/tutorials)
- [ ] A Waypoint collection: [/waypoint/tutorials/get-started-docker](https://dev-portal-git-ambadd-version-to-overview-item-href-hashicorp.vercel.app/waypoint/tutorials/get-started-docker)

### Docs pages

For the steps below for the following pages:

- [/vault/docs](https://dev-portal-git-ambadd-version-to-overview-item-href-hashicorp.vercel.app/vault/docs)
[...]

PR #510, opened 2022-05-31
Add version to docs breadcrumb

ashleemboyer:

---

## 🔗 Relevant links

- [Asana task](url) 🎟️


src/layouts/sidebar-sidecar/server.ts
```diff
@@ -200,6 +201,24 @@ export function getStaticGenerationFunctions<
...
+            (version) => pathPart === version.version
+          )
+          if (matchingVersion) {
+            versionPathPart = pathPart
+            indexOfVersionPathPart = index
+            return true
+          }
+        })
+      }
+
```
ashleemboyer, line 221: 🗒️ This will go away when #509 is merged

src/layouts/sidebar-sidecar/server.ts
```diff
@@ -218,17 +237,39 @@ export function getStaticGenerationFunctions<
...
+        generatedBaseName = `${baseName} ${versionPathPart}`
+        generatedBasePath = `${basePath}/${versionPathPart}`
+        filteredPathParts = pathParts.filter(
+          (_, index) => index !== indexOfVersionPathPart
+        )
+      } else {
+        generatedBaseName = baseName
+        generatedBasePath = basePath
+        filteredPathParts = pathParts
+      }
```
BRKalow, line 263: 💭  It seems like this could all be handled in `getDocsBreadcrumbs` based on whether or not `version` is defined.

src/layouts/sidebar-sidecar/server.ts
```diff
@@ -218,17 +237,39 @@ export function getStaticGenerationFunctions<
...
 
+      /**
+       * Generate the arguments sent to `getDocsBreadcrumbs` based on whether or
+       * not there is a version in the current path.
+       */
+      let generatedBaseName
+      let generatedBasePath
+      let filteredPathParts
+      if (indexOfVersionPathPart >= 0) {
+        generatedBaseName = `${baseName} ${versionPathPart}`
```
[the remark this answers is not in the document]
ashleemboyer: Yes, I included it with the root path breadcrumb item because it doesn't seem like it's needing to be its own item. It also wasn't accurate to make it the last breadcrumb item, since it's not the last item in our URL paths for versioned pages. What are your thoughts on this?

src/layouts/sidebar-sidecar/server.ts
ashleemboyer: I see what you're saying! I felt it made more sense to me to change the arguments passed in to `getDocsBreadcrumbs` than to essentially make `getDocsBreadcrumbs` modify the arguments passed to it, since that's typically dangerous to do.

src/layouts/sidebar-sidecar/server.ts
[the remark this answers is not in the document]
BRKalow: Makes sense! I feel like this could be something that comes up in QA or feedback based on the designs, should we propose this adjustment before we go ahead with it?

src/layouts/sidebar-sidecar/server.ts
ashleemboyer: If this is a blocker for you, happy to update it!

src/layouts/sidebar-sidecar/server.ts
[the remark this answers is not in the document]
ashleemboyer: Got the OK via [Slack](https://hashicorp.slack.com/archives/C01KCU4HDPY/p1654118539129629?thread_ts=1654117966.138619&cid=C01KCU4HDPY)!

src/layouts/sidebar-sidecar/server.ts
BRKalow: I think given that the version might need to be added on to different path segments, based on the convo in slack, it feels like it would be worthwhile to abstract this into `getDocsBreadcrumbs`! 😊 

Let's make the change! 💪

ashleemboyer: Can do!

src/components/breadcrumb-bar/utils/get-docs-breadcrumbs.ts
```diff
@@ -178,29 +178,47 @@ interface GetDocsBreadcrumbsOpts extends GetPathBreadcrumbsOpts {
...
   productPath: string
   versions?: VersionSelectItem[]
+  indexOfVersionPathPart?: number
 }
 
 function getDocsBreadcrumbs({
-  basePath,
   baseName,
+  basePath,
+  indexOfVersionPathPart,
```
BRKalow, line 187: nit: `indexOfVersionPathPart` can be derived from `pathParts` and `version`.

zchsh: 💭 ⛏️  if we want to avoid passing this from server, and avoid iterating, maybe we could use [findIndex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex), eg

```ts
const indexOfVersionPathPart = pathParts.findIndex(version)
```

src/components/breadcrumb-bar/utils/get-docs-breadcrumbs.ts
```diff
@@ -148,28 +177,49 @@ interface GetDocsBreadcrumbsOpts extends GetPathBreadcrumbsOpts {
...
+  version,
 }: GetDocsBreadcrumbsOpts): BreadcrumbLink[] {
+  /**
+   * Generate the arguments sent to `getDocsBreadcrumbs` based on whether or
+   * not there is a version in the current path.
+   */
+  let generatedBaseName
+  let filteredPathParts
+  if (indexOfVersionPathPart >= 0) {
+    generatedBaseName = `${baseName} ${version}`
```
zchsh, line 201: ⛏️ Nitpick: not sure if this is a case of design specs being out of date, but it looks like the `version` is a distinct path item:

CleanShot 2022-06-01 at 17 53 48@2x

IMO non-blocking since we probably need to sort out the UX of where the top-level `basePath` item should then link? I guess to the most recent version:

zchsh (approved):
✅ Looks good to me, ran through the testing steps!

Had a couple of nitpicks but nothing blocking 👍

src/components/breadcrumb-bar/utils/get-docs-breadcrumbs.ts
ashleemboyer: Bryce raised this as well here: https://github.com/hashicorp/dev-portal/pull/510#discussion_r887222493

ashleemboyer: No worries! 💖

PR #517, opened 2022-06-01
Use transparent outlines; enforce it

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambuse-transparent-outlines-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1201987349274776/1202105084146685/f) 🎟️

## 🗒️ What

- Adds another custom stylelint rule named `no-removed-outlines`
- Adds tests for new rule
- Specifies `length-zero-no-unit` rule, which makes validation in `no-removed-outlines` simpler
- Fixes cases that break the rule
- Renamed the other custom rule's file to match the rule name

## 🤷 Why

As noted in the error message, removed outlines don't work in high contrast modes. Setting a transparent outline gets us the same result as a removed outline, while also getting us support in high contrast modes. This rule helps us maintain this stance.

## 🛠️ How

I followed the same documentation as last time, and additionally read through [the Testing section](https://stylelint.io/developer-guide/plugins/#testing) for adding tests.

## 🧪 Testing

Nothing to really manually test here! We shouldn't see any changes in the app. We just need the PR checks to pass.


stylelint-rules/__tests__/no-removed-outlines.test.js
```diff
@@ -0,0 +1,43 @@
...
+
+    expect(warnings).toHaveLength(1)
+    expect(parseErrors).toHaveLength(0)
+
+    const [{ line, column, text }] = warnings
+    expect(line).toBe(1)
+    expect(column).toBe(testCase.indexOf('outline') + 1)
+    expect(text).toBe(messages.error)
+  })
+})
```
BRKalow, line 43: 👏

stylelint-rules/no-removed-outlines.js
```diff
@@ -0,0 +1,67 @@
...
+const stylelint = require('stylelint')
+const { report, ruleMessages, validateOptions } = stylelint.utils
+
+const ruleName = 'digital-plugin/no-removed-outlines'
+const messages = ruleMessages(ruleName, {
+  error:
+    'Do not remove outlines. Set them to be transparent instead. Removing outlines makes focus indicators inaccessible in high contrast modes.',
+})
+
+module.exports = stylelint.createPlugin(
```
BRKalow, line 15: very cool 🤩

BRKalow (approved):
@ashleemboyer a README in the `stylelint-rules` folder would be great, with a short blurb about each custom rule.

Nice job, this is an ideal way to enforce accessibility best practices 🥰

ashleemboyer:
Thanks @BRKalow! I'll add a `README` in a quick-follow PR. I also want to add tests for the other rule, and the `module.exports` in the other rule need to be moved after the `createPlugin` as well.

PR #527, opened 2022-06-02
[Install view] Mobile download links

ashleemboyer:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1202022787106817/1201654639085767/f) 🎟️

## 🗒️ What

- Adds `MobileStandaloneLink`
- Updates a `StandaloneLink` on the install view to use `MobileStandaloneLink`
- Updates `MobileDownloadStandaloneLink` to render `MobileStandaloneLink`
- Renames `DownloadStandaloneLink` to `MobileDownloadStandaloneLink`

## 🤷 Why

- The icon-only-on-mobile was in the designs from the start, just had not been implemented right away
- The addition of `MobileStandaloneLink` allows us to not force every `StandaloneLink` to behave in this way
  - The `StandaloneLink` API was also going to be too confusing if an `isMobile` prop (or something like that) was going to be added

## 📸 Design Screenshots

<details>
  <summary>Install view @ mobile width</summary>
  <a href="https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/?node-id=10144%3A145458">Figma link</a>
  CleanShot 2022-06-02 at 16 58 02@2x
  </summary>
</details>

## 🧪 Testing

- [ ] Go to [/waypoint/downloads](https://dev-portal-git-ambmobile-download-links-hashicorp.vercel.app/waypoint/downloads) or [/vault/downloads](https://dev-portal-git-ambmobile-download-links-hashicorp.vercel.app/vault/downloads)
- [ ] Make sure the affected elements are text + icon at tablet and higher widths
- [ ] Make sure the affected elements are icon-only at mobile widths

PR #530, opened 2022-06-02
feat: add give feedback button to top nav

zchsh:

---

## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

This PR adds a `Give beta feedback` button to the top navigation bar.

## 🤷 Why

- To receive feedback during the beta
- So that we can remove the alpha feedback bar ASAP, as it causes layout challenges

## 🛠️ How

- Uses `ButtonLink` on table and desktop
- Uses a custom component for the icon-only button on mobile
    - This custom component duplicates styles from `button-link` to serve the needs of this PR
    - Longer term, we likely want to align with HDS on icon size & button size options. In short, in [HDS Button](https://design-system-components-hashicorp.vercel.app/components/button), buttons with 36px height usually only support 16px icon size. We're diverging from the HDS spec here to meet the dev-portal-specific design spec.

## 📸 Design Screenshots

CleanShot 2022-06-02 at 17 46 05@2x

## 🧪 Testing

- [ ] Visit any page on the site, such as [the home page][preview], [/waypoint][/waypoint], or [/vault/tutorials/getting-started/getting-started-intro][/vault/tutorials/getting-started/getting-started-intro]
    - Ensure the `Give feedback` CTA in the to header looks as expected, on mobile and desktop
    - Test that it links to [the beta feedback form][form-url]
    - Ensure the rest of the top nav continues to look and function as expected, including on mobile

## 💭 Anything else?

Note: another option considered here is passing customizations through to `ButtonLink` via a new `className` prop (eg could set `--padding-top-bottom` and `--padding-left-right`).

However, in past approaches like this, we tend to create brittleness and eventually ossification in any component customized in this way without clear forward intention, as said component has to account for all possible customizations built on top of it through CSS overrides passed through `className`. There are more detailed notes on this rationale in our CSS modules adoption RFC, [[Digital RFC] Establish conventions for style additions with CSS modules](https://docs.google.com/document/d/1GPN1pt1xPktT9lYkGZZZ-o-SkagOZR1WIsU2X250s-Q/edit#heading=h.j4iia1mkqs9o).

With this in mind, the duplicate code here feels safer for future maintainability.

[task]: https://app.asana.com/0/1202114367927919/1202348169653878/f
[preview]: https://dev-portal-git-zsadd-give-feedback-to-nav-hashicorp.vercel.app/
[/waypoint]: https://dev-portal-git-zsadd-give-feedback-to-nav-hashicorp.vercel.app/waypoint
[/vault/tutorials/getting-started/getting-started-intro]: https://dev-portal-git-zsadd-give-feedback-to-nav-hashicorp.vercel.app/vault/tutorials/getting-started/getting-started-intro
[form-url]: https://forms.gle/fnHLuNahLEhjuKvE6


zchsh:
@BRKalow @ashleemboyer @braydenlove This PR adds the `Give beta feedback` button to the top nav.

The "icon only" button in the mobile designs doesn't fit into the HDS-based implementation specs cleanly, so I've made a quick fork. I've opted for copy-pasting things, rather than trying to shim in an abstraction (such as passing `className` or `style` through), as from past experience the shims tend to be much more painful to deal with later (further rationale is linked to in PR description).

As well, note that `ButtonLink` retains its underline for now - we have a task 🎟️ [Remove button link text underline ](https://app.asana.com/0/1202114367927919/1202355923167674/f) to address that separately.

Let me know if there's any other context I can add here 🙇

zchsh:
@braydenlove ooh great call! have updated to open in a new tab 💯 💖

braydenlove:
I am out tomorrow but will review with Mike first thing Monday if that is ok?

braydenlove:
I haven't tested it on all the browsers and devices but it looks good as far as what I have seen. If you are alright with refinement tickets coming next week if I missed something I am comfortable going with option 1. I will take a more thorough look next week? How does that sound? We could pack any changes with updating the icon as well.

src/components/button-link/index.tsx
```diff
@@ -3,16 +3,22 @@ import s from './button-link.module.css'
...
+ * copied from `Button`.
  **/
 interface ButtonLinkProps
-  extends Pick<ButtonProps, 'color' | 'size' | 'text' | 'ariaLabel'> {
+  extends Pick<
+    ButtonProps,
+    'color' | 'size' | 'text' | 'ariaLabel' | 'icon' | 'iconPosition'
+  > {
   href: string
+  target?: NativeLinkProps['target']
```
ashleemboyer, line 21: 💭 Should we do [what `StandaloneLink` does](https://github.com/hashicorp/dev-portal/blob/main/src/components/standalone-link/index.tsx) and have an `openInNewTab` prop that handles setting both `target` and `rel` appropriately on the rendered `<a>`?

src/components/button-link/index.tsx
```diff
@@ -21,14 +27,37 @@ export default function ButtonLink({
...
+    throw new Error(
+      '`Button` must have either `text` or an `icon` with accessible labels.'
+    )
+  }
+
+  if (isIconOnly && !hasLabel) {
+    throw new Error(
+      'Icon-only `Button`s require an accessible label. Either provide the `text` prop, or `ariaLabel`.'
+    )
+  }
```
ashleemboyer, line 50: Can these messages be updated to reference `ButtonLink` instead of `Button`?

zchsh:
@braydenlove That sounds perfect, thanks for taking a look so quickly on short notice 🙇 Definitely happy to refine further if there's additional feedback (for this PR of course, but also for any others!) 💯

src/components/navigation-header/components/give-feedback-button/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+    return (
+      <ButtonLink
+        target="_blank"
+        href={FORM_URL}
+        text="Give beta feedback"
+        icon={<IconExternalLink16 />}
+        iconPosition="trailing"
+      />
+    )
+  }
```
ashleemboyer, line 36: Can we set the ARIA label for both of these components to be `"Give beta feedback (opens in new tab)"`? That will give non-sighted screen reader users the affordance they need to know how the link behaves.

ashleemboyer (commented):
I noticed this is flashing if you view at a mobile width and then refresh. I just ran into this as well on #527 and had to use CSS to fix. A quick fix here might be to render both components where you're checking `isMobile` and wrap them with a CSS class that does a media query?

<details>
  <summary>GIF (hidden because it's flashy)</summary>
  
</details>

src/components/button-link/index.tsx
zchsh: ✅  Great call! Done, I took the same approach as `StandaloneLink` 💯 Thanks for the quick reference ✨

src/components/button-link/index.tsx
zchsh: ✅ Ah great catch, updated!

src/components/navigation-header/components/give-feedback-button/index.tsx
zchsh: ✅ Absolutely - updated to `"Give beta feedback (opens in new tab)"` for both links 👍

zchsh:
@ashleemboyer Thanks for the review & feedback! 💖 I've updated to try to address:
- Now shows & hides with CSS, not `useDeviceSize()`
- Updates `ButtonLink` to take `openInNewTab` prop, similar ensures both links have both `target` & `rel` set as we want (swiped from `StandaloneLink`)
- Fixes comment in `ButtonLink`
- Updates `aria-label` on both links to `"Give beta feedback (opens in new tab)"`

src/components/button-link/index.tsx
```diff
@@ -3,16 +3,22 @@ import s from './button-link.module.css'
 import { ButtonProps } from 'components/button/types'
 import classNames from 'classnames'
 
+type NativeLinkProps = JSX.IntrinsicElements['a']
+
```
ashleemboyer, line 7: This appears unused now!

```suggestion
```

src/components/navigation-header/components/give-feedback-button/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+          className={classNames(s.iconButton, s.primary)}
+          target="_blank"
+          rel="noreferrer"
+        >
+          <IconMessageSquareFill24 />
+        </a>
+      </span>
+      <span className={s.textButtonContainer}>
+        <ButtonLink
+          openInNewTab={true}
```
ashleemboyer, line 28: 🎨 We should be able to pass this prop without explicitly setting the value. 😊

```suggestion
          openInNewTab
```

ashleemboyer (approved):
Looks great! Thanks for making those last changes 💖

PR #532, opened 2022-06-02
fix: remove color: inherit from code elements in inline alerts

BRKalow:

---

## 🔗 Relevant links

- [Preview link](https://dev-portal-git-brkfix-code-snippet-visited-color-hashicorp.vercel.app/waypoint/docs/platforms/kubernetes/kubectl-deploy) 🔎
- [Asana task](https://app.asana.com/0/1202114367927919/1202331881539237/f) 🎟️

## 🗒️ What

Adjust styling so that links in code snippets in inline alerts have the correct coloring for the visited state.

## 🧪 Testing

- [ ] visit https://dev-portal-git-brkfix-code-snippet-visited-color-hashicorp.vercel.app/waypoint/docs/platforms/kubernetes/kubectl-deploy
- [ ] Click a link in an inline alert that is rendered as an inline code snippet
- [ ] Navigate back
- [ ] Observe that the link you clicked has the correct visited link styles (purple text color)


src/components/dev-dot-content/dev-dot-content.module.css
```diff
@@ -29,7 +29,6 @@
         background-color: white;
         border: 1px solid #cce3fe;
         border-radius: 5px;
-        color: inherit;
```
BRKalow, line 32: Did some quick testing and it looks okay to remove this. Visited color works correctly (fixes the initial ticket) and the color still looks fine when the code snippet is not a link.

braydenlove (approved):
From a design perspective this is looking good. Tested on Firefox/Safari/Chrome and Edge. Thanks!

ashleemboyer (commented):
Went through the test steps and noticed that the `:visited:hover` style doesn't match other links in the page. These are currently handled by [`InlineLink`](https://github.com/hashicorp/dev-portal/blob/main/src/components/inline-link/inline-link.module.css), so I think we need to make use of that component here? I think it's pure luck that [the `DocsAnchor` component](https://github.com/hashicorp/dev-portal/blob/main/src/components/docs-anchor/docs-anchor.module.css) has the `:visited` color correctly hardcoded as a hex value.

BRKalow:
Ah yep, good catch. I think what I'll try and do is use the .root class from InlineLink and apply it to the anchors rendered by DocsAnchor. Seems like a good path forward without changing behavior.

ashleemboyer:
> Ah yep, good catch. I think what I'll try and do is use the .root class from InlineLink and apply it to the anchors rendered by DocsAnchor. Seems like a good path forward without changing behavior.

@BRKalow sounds good to me! Do you think we could add a `TODO` in `DocsAnchor` noting this? I don't think we'll want it to be that way forever, but it's a good holdover until we have more time to QA links in docs content!

BRKalow:
@ashleemboyer thanks again for catching that, I've updated `DocsAnchor` to use `InlineLink`'s styles for now!

ashleemboyer (approved):
This screenshot is just a breadcrumb for later, but noticed that links with HTML in them don't render with the font weight `InlineLink`s use. But this is because the component applies the font weight, not the CSS, so I think this'll be resolved along with the TODO you left! Also glad to see `docs-anchor.module.css` removed, I think it was causing a bit of confusion. 🥇

CleanShot 2022-06-03 at 10 41 33@2x

ashleemboyer:
> @ashleemboyer what do you think about using `composes:` to declare the `hds-*` classes for `InlineLink` in CSS instead of in the component? Would keep all of the styling concerns in the CSS file, and address that issue too.

@BRKalow I think if all `InlineLink` components had the same font weight, that would work. But the ones on the downloads pages for example are a lighter weight than the ones we see in content. I think removing the text props from `InlineLink` will take some refactoring of references to `InlineLink`, which feels out of scope for this PR.

PR #535, opened 2022-06-02
feat: use next/image for docs images with known dimensions

dstaley:

---

## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

This PR removes our remark plugin that was previously used to rewrite image URLs to the Next.js image optimization endpoint in favor of using the native `next/image` element for image urls containing a `width` and `height` parameter (an indication of images being sourced from our content API static asset endpoint).

In order to support the `layout="raw"` option, this PR also updates Next to the next (lol) point release, `12.1.1`.

## 🤷 Why

Previously, we rewrote the URLs for images when rendering content, and manually appended the optimization endpoint URL. By using a native `next/image` element, we no longer need to maintain that logic, and are one step closer to enabling some of the more complex features like blurhashes.

## 🛠️ How

Parse the searchParams of `url` passed to our custom `Image` component, and if `width` and `height` are present, render a `next/image` instead of a `img`.

## 🧪 Testing

- Go to preview URL:
- Confirm that the first image in the documentation renders correctly


next.config.js
```diff
@@ -102,5 +102,10 @@ module.exports = withSwingset({
       dangerouslyAllowSVG: true,
       contentSecurityPolicy: "default-src 'self'; script-src 'none'; sandbox;",
     },
+    experimental: {
```
BRKalow, line 105: 🧑‍🔬

src/layouts/sidebar-sidecar/server.ts
```diff
@@ -36,30 +34,6 @@ const BASE_PATHS_TO_NAMES = {
...
-          params.set('w', '1920')
-          // By default, next/image uses a quality setting of 75.
-          params.set('q', '75')
-
-          node.url = `/_next/image?${params.toString()}`
-        }
-      })
-    }
-  }
-}
```
BRKalow, line 61: Do you think we should still do this for images without dimensions specified? Or just consider the next/image usage as a progressive enhancement, especially as more content continues to get uploaded with dimensions in the image URLs.

dstaley: My thinking here was that we could manually trigger a content sync for other versions, so it wouldn't be worth the effort to maintain this remark plugin alongside the new component-based approach.

That being said, if there's some reason we can't easily sync old versions, I think I'd rather do this URL rewrite in the `Image` component as opposed to a remark plugin given how much easier it would be for others to discover.

BRKalow (approved):

PR #538, opened 2022-06-02
Fix DocsVersionSwitcher position

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambfix-version-switcher-position-hashicorp.vercel.app/) 🔎

## 🗒️ What

Fixes the DocsVersionSwitcher position. This also appears to fix it on narrower views! 🎉

## 🤷 Why

It would move up as the Sidebar is scrolled.

## 🛠️ How

Moved the `overflow-y: auto` declaration to the `DocsVersionSwitcher` wrapper's sibling element. That element had `position: relative`, so `DocsVersionSwitcher`'s wrapper was `position: fixed` relative to that sibling `<div>`.

## 🧪 Testing

Follow these steps at multiple viewport breakpoints:

- [ ] Go to [/waypoint/docs](https://dev-portal-git-ambfix-version-switcher-position-hashicorp.vercel.app/waypoint/docs) (has a long sidebar)
- [ ] Scroll the sidebar
- [ ] The `DocsVersionSwitcher` should stay put
- [ ] Go to [/vault/api-docs](https://dev-portal-git-ambfix-version-switcher-position-hashicorp.vercel.app/vault/api-docs) (has a short sidebar)
- [ ] The `DocsVersionSwitcher` should stay put


zchsh (approved):
✅ This looks great to me in terms of positioning!

I am seeing an issue with the last items being hidden behind the version switcher. IMO fine to address in a follow-up PR if you want 👍

last-item-hidden

There's still an issue with the top bar affecting the sidebar position (or rather height), but I think this'll _mostly_ fix itself if you pull in from `main` (where said bar has been removed!):

height-of-alert-bar-affecting-sidebar-position

PR #540, opened 2022-06-03
fix: key error in sidebar-nav-menu-item due to incorrect id used for key

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Noticed a react key error being logged in development. Tracked it down and it looks like we were using the wrong `id` value when rendering nested `SidebarNavMenuItem`s.


ashleemboyer (approved):
Good eyes! 👀

PR #543, opened 2022-06-03
Small polishing fixes

ashleemboyer:
## 🗒️ What

- Removes old `70px` padding for mobile sidebar since that's fixed
- Adds `150px` padding to the bottom of sidebar content
  - This makes it so the version switcher doesn't cut off the content anymore
  - This is also nice so that sidebar content isn't always at the bottom of a window/screen
- Fixes the link of the third docs breadcrumb item when a past version is selected

## 🧪 Testing

- [ ] Go to /waypoint/docs
- [ ] On desktop, make sure the end of sidebar content isn't cut off by the version switcher
- [ ] On mobile, make sure everything is placed correctly
- [ ] Make sure the sidebar looks correct on mobile/desktop on these pages:
  - [ ] /waypoint
  - [ ] /waypoint/tutorials
  - [ ] /waypoint/tutorials/get-started-docker
  - [ ] /waypoint/tutorials/get-started-docker/get-started-intro

### DIFF
--- a/src/layouts/sidebar-sidecar/sidebar-sidecar-layout.module.css
+++ b/src/layouts/sidebar-sidecar/sidebar-sidecar-layout.module.css
@@ -73,7 +73,6 @@ for further details.
   @media (--dev-dot-tablet-down) {
     left: -150vw;
     width: 100vw;
-    padding-bottom: 70px;
     height: 100%;
     position: fixed;
     top: unset; /** We want to pull the top relative to the parent on mobile */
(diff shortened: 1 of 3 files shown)


kendallstrautman (approved):

src/layouts/sidebar-sidecar/sidebar-sidecar-layout.module.css
```diff
@@ -84,7 +83,10 @@ for further details.
   height: 100%; /* for positioning the version switcher */
   overflow-y: auto; /* for positioning the version switcher */
   position: relative;
-  padding: 24px;
+  padding-bottom: 150px;
```
BRKalow, line 86: 💭 What do you think about changing this to `84px` so the padding all the way around the content is largely the same? It feels _off_ with significantly more bottom padding.

Screen Shot 2022-06-03 at 3 31 36 PM
.

ashleemboyer: I think it's better with more padding. It provides very clear separation from the version switcher and on other pages it keeps the content from being so close to the bottom of the window/screen.

PR #548, opened 2022-06-03
chore: fix search results on blur

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksclose-search-panel-on-blur-hashicorp.vercel.app/waypoint/docs) 🔎
- [Asana task](https://app.asana.com/0/1202276096167515/1202390857238416) 🎟️

## 🗒️ What

Fixes the issue where the search panel would stay open when the mobile sidebar menu is opened. The default `onBlur` is overridden for the `autocomplete` props so we should validate that this doesn't break other behavior. 

## 🧪 Testing

- Visit a docs page, enter some search input, open the hamburger menu. The search panel should close.

### DIFF
--- a/src/components/algolia-search/index.tsx
+++ b/src/components/algolia-search/index.tsx
@@ -113,6 +113,9 @@ export default function AlgoliaSearch<THit extends Hit<unknown>>({
   return (
     <div
       className={classNames(s.root, className)}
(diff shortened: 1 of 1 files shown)


BRKalow (approved):
👏

PR #550, opened 2022-06-05
feat: instrument DataDog

zchsh:
- feat: add DataDog, via ConsentManager
- perf: prefetch DataDog script

## 🔗 Relevant links

- [Asana task](url) 🎟️

### DIFF
--- a/src/layouts/base-new/index.tsx
+++ b/src/layouts/base-new/index.tsx
@@ -1,6 +1,10 @@
+import Head from 'next/head'
 import usePageviewAnalytics from '@hashicorp/platform-analytics'
 import createConsentManager from '@hashicorp/react-consent-manager/loader'
-import { devDotConsentManagerServices } from 'lib/consent-manager-services/dev-dot'
+import {
+  devDotConsentManagerServices,
+  DATADOG_SCRIPT_URL,
+} from 'lib/consent-manager-services/dev-dot'
 import Footer from 'components/footer'
 import NavigationHeader from 'components/navigation-header'
 import useScrollPercentageAnalytics from 'hooks/use-scroll-percentage-analytics'
@@ -29,6 +33,9 @@ const BaseNewLayout: React.FC<BaseNewLayoutProps> = ({
 
   return (
     <>
+      <Head>
+        <link rel="prefetch" href={DATADOG_SCRIPT_URL} />
+      </Head>
       <CoreDevDotLayout>
         <div className={s.root} data-layout="base-new">
           <div className={s.header}>
--- a/src/layouts/base-new/index.tsx
+++ b/src/layouts/base-new/index.tsx
@@ -1,5 +1,6 @@
 import usePageviewAnalytics from '@hashicorp/platform-analytics'
 import createConsentManager from '@hashicorp/react-consent-manager/loader'
+import { devDotConsentManagerServices } from 'lib/consent-manager-services/dev-dot'
 import Footer from 'components/footer'
 import NavigationHeader from 'components/navigation-header'
 import useScrollPercentageAnalytics from 'hooks/use-scroll-percentage-analytics'
@@ -13,6 +14,7 @@ interface BaseNewLayoutProps {
 
 const { ConsentManager, openConsentManager } = createConsentManager({
   preset: 'oss',
+  otherServices: [...devDotConsentManagerServices],
 })
 
 const BaseNewLayout: React.FC<BaseNewLayoutProps> = ({
--- a/src/lib/consent-manager-services/dev-dot.ts
+++ b/src/lib/consent-manager-services/dev-dot.ts
@@ -1,0 +1,40 @@
+import { ConsentManagerService } from '@hashicorp/react-consent-manager/types'
+
+const DATADOG_SCRIPT_URL =
+  'https://www.datadoghq-browser-agent.com/datadog-rum-v4.js'
+const DATADOG_CLIENT_TOKEN = 'TODO-need-access-to-datadog'
+const DATADOG_APP_ID = 'TODO-need-access-to-datadog'
+const DATADOG_SITE = 'TODO-need-access-to-datadog'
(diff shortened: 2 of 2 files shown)


src/layouts/base-new/index.tsx
```diff
@@ -27,6 +33,9 @@ const BaseNewLayout: React.FC<BaseNewLayoutProps> = ({
 
   return (
     <>
+      <Head>
+        <link rel="prefetch" href={DATADOG_SCRIPT_URL} />
```
zchsh, line 37: Intent here is to prefetch the DataDog RUM script so that, if consent is given, the script will load more quickly.

src/lib/consent-manager-services/dev-dot.ts
```diff
@@ -0,0 +1,48 @@
...
+   d=o.createElement(u);d.async=1;d.src=n
+   n=o.getElementsByTagName(u)[0];n.parentNode.insertBefore(d,n)
+})(window,document,'script','${DATADOG_SCRIPT_URL}','DD_RUM')
+  DD_RUM.onReady(function() {
+    DD_RUM.init({
+      clientToken: '${DATADOG_CLIENT_TOKEN}',
+      applicationId: '${DATADOG_APP_ID}',
+      site: '${DATADOG_SITE}',
+      env: '${ENV}', 
+      trackInteractions: false,
```
zchsh, line 42: Wanted to check on whether we want to enable `trackInteractions`. DataDog's snippets show it set to `true`, however they also explicitly note:

> The trackInteractions parameter enables the automatic collection of user clicks in your application. Sensitive and private data contained on your pages may be included to identify the elements interacted with.

I've set to `false` for now as I'm not 100% sure on the potential for sensitive and private data with dev-portal.

BRKalow:
I believe we're currently working to get DD to stop collecting IP addresses, which would allow us to not have to implement this behind consent manager. In that case, maybe we remove the consent manager gate here in preparation for that?

I think we'll also want to set the `sampleRate` fairly low. Maybe not yet as we don't have real traffic, but definitely before the launch. (https://docs.datadoghq.com/real_user_monitoring/browser/#configuration)

mhonnatti:
@zchsh @BRKalow confirming that DataDog needs to be outside the consent manager.

src/lib/datadog/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+      env: '${ENV}', 
+      version: '${COMMIT_SHA}',
+      trackInteractions: false,
+    })
+  })`,
+}
+
+function DataDogScriptTag() {
+  return (
+    <Script id={dataDogService.name} strategy="afterInteractive">
```
mhonnatti, line 48: considering that we will be using Datadog for web vitals and console error logging, should this be `beforeInteractive` instead of `afterInteractive`? @BRKalow thoughts?

BRKalow: @mhonnatti Yeah, that seems reasonable for now. We'll need to shift this to `_document` then, per https://nextjs.org/docs/basic-features/script#beforeinteractive

src/lib/consent-manager-services/dev-dot.ts
mhonnatti: I'm okay with starting this way! We'll need to enable it eventually for session replays

mhonnatti:
@zchsh since you still don't have Datadog access, here's the snippet that Datadog generated for us:

`
import { datadogRum } from '@datadog/browser-rum';

datadogRum.init({
    applicationId: '4486817b-44b8-4e6c-b6a4-cef83e48cf46',
    clientToken: 'pub750ceadc5a9c79a6d3da59b534eb5108',
    site: 'datadoghq.com',
    service:'developer.hashicorp.com',
    env:'env',
    // Specify a version number to identify the deployed version of your application in Datadog 
    // version: '1.0.0',
    sampleRate: 100,
    premiumSampleRate: 100,
    trackInteractions: true,
    defaultPrivacyLevel:'mask-user-input'
});
    
datadogRum.startSessionReplayRecording();
`
We've already addressed `trackInteractions`. `site` looks like it needs to be updated, `premiumSampleRate` needs to be added

zchsh:
@mhonnatti @BRKalow I've updated this PR with config values (thanks Manish!), and have made the following changes as well:
- Decoupled from `ConsentManager`, now loads script directly via `next/script`
- Moved to `_document.tsx`, and changed `strategy` to `beforeInteractive`
- Added `version` config, set to the commit SHA from Vercel
- For `sampleRate`, have left at `100` for now for testing purposes. Happy to adjust before merging if there's a specific number we want (eg `10`?). But figure we might also be fine to leave this for a follow-up PR before launch (after I can more thoroughly confirm this is working by logging into Datadog)

Thanks for both of your help here, and let me know what ya'll think 🙇

src/lib/datadog/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+      site: 'datadoghq.com',
+      service: '${DATADOG_SERVICE}',
+      env: '${ENV}', 
+      version: '${COMMIT_SHA}',
+      sampleRate: 100,
+      premiumSampleRate: 100,
+      trackInteractions: false,
+      defaultPrivacyLevel:'mask-user-input'
+    })
+    DD_RUM.startSessionReplayRecording()
```
zchsh, line 41: I'm unclear what this will do with `trackInteractions: false`, but figured might as well add it in so that we don't have to remember to do so if / when we set `trackInteractions: true`

mhonnatti (approved):

zchsh:
@mhonnatti @BRKalow apologies for the delay here, this should be ready for re-review! Have updated as follows:

- Sticking with `afterInteractive`
- Added variable config per environment, and using a separate Datadog app for prod vs non-prod
- Moved script tag to `dev-dot` specific layout file, so that this doesn't run on `dot-io`s

mhonnatti (approved):
🤘

PR #551, opened 2022-06-05
fix: better fallback title for breadcrumb edge case

zchsh:
## 🔗 Relevant links

- [Asana task](url) 🎟️

### DIFF
--- a/src/components/breadcrumb-bar/utils/get-docs-breadcrumbs.ts
+++ b/src/components/breadcrumb-bar/utils/get-docs-breadcrumbs.ts
@@ -1,6 +1,7 @@
 import { VersionSelectItem } from '@hashicorp/react-docs-page/server/loaders/remote-content'
 import { NavData } from '@hashicorp/react-docs-sidenav'
 import { BreadcrumbLink } from '..'
+import findAllFallbackTitles from './find-all-fallback-titles'
 
 const IS_DEV = process.env.NODE_ENV !== 'production'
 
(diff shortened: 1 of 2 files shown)


src/components/breadcrumb-bar/utils/get-docs-breadcrumbs.ts
```diff
@@ -105,14 +106,16 @@ function getPathMatchedNode(navNodes, pathString, basePath) {
...
   if (IS_DEV) {
     console.warn(
       `Missing breadcrumb path under "${basePath}": Found zero matches for "${pathString}". Please ensure there is a node (or index-less category) with the path "${pathString}" in the provided navData.`
     )
   }
-  const lastPathPart = pathString.split('/').pop()
-  return { title: lastPathPart }
+  // Get a fallback title by searching all navNodes for title text that,
+  // in slug form, matches the last path part. At worst, path will be used.
+  const fallbackTitle = getFallbackTitle(pathString, navNodes)
```
zchsh, line 117: `getFallbackTitle` is the approach I took initially, though it fits an even edgier case than I thought. It might not be that useful (the edge case may _never_ come up), so I've added a comment to flag that.

src/components/breadcrumb-bar/utils/get-docs-breadcrumbs.ts
```diff
@@ -152,11 +155,14 @@ function findPathMatchedNodes(navNode, pathString, depth) {
         return Boolean(r.path)
       })
       if (routesWithPaths.length) {
-        const inferredPathParts = routesWithPaths[0].path.split('/')
-        inferredPathParts.pop()
-        const inferredPath = inferredPathParts.join('/')
-        if (inferredPath == pathString) {
-          return [{ title: navNode.title }]
+        // Iterate over all child nodes, match may not be in first node
+        for (let i = 0; i < routesWithPaths.length; i++) {
```
zchsh, line 159: This iteration is necessary because on https://developer.hashicorp.com/vault/docs/get-started/developer-qs, we have links in the sidebar that _start_ their lives as external `href` links but end up with `path` values that are _external to docs_. This messes with the expectation that `{ title, path }` nodes will be internal to docs, which was previously true.

I didn't realize this initially, which is why I took the initial approach of traversing nav-data (in `getFallbackTitle`) if a `title` match can't be found.

BRKalow (approved):
Looks god. It feels like this breadcrumb logic would be a great candidate for some unit tests, especially with newly discovered edge cases like this.

Nice sleuthing.

PR #561, opened 2022-06-07
fix: back link error triggered by ISR

BRKalow:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1202114367927919/1202391122486283/f) 🎟️

## 🗒️ What

Fixes infrequent error occurring when the product landing pages attempt to revalidate. They read from the filesystem, which won't have the expected files during ISR, causing an error. It's not possible to change this page independent of a deployment, so there's no need for revalidate here.


kendallstrautman (approved):
🌟

PR #567, opened 2022-06-07
fix: adjust filter input box shadow

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksfix-input-outline-hashicorp.vercel.app/waypoint/docs) 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1202317183612796) 🎟️

## 🗒️ What

Fixes an issue where the sidebar filter input box shadow wasn't rendering on iOS browsers.  

At first I thought it was related to this known issue. See this note from [can-i-use](https://caniuse.com/?search=box-shadow), apparently this is a known issue with iOS safari and we can't use the `blur-radius` as 0. I updated it to 1px and it didn't seem to fix the issue. 

Screen Shot 2022-06-07 at 3 55 07 PM

> [Also see this stack overflow discussion.](https://stackoverflow.com/questions/28245254/box-shadow-inset-doesnt-work-in-ios)

~~I ended up moving the box shadow up to the container div and that seems to solve it. Perhaps there some iOS input issue with box shadow. When I would inspect the input element, the `box-shadow` was computing to `none`. Visually this still seems consistent. Open to alternative fixes though!~~ Ended up using `appearance: none` and this solved it.

## 🧪 Testing

Fire up simulator and [visit the preview link](https://dev-portal-git-ksfix-input-outline-hashicorp.vercel.app/waypoint/docs). Hit the hamburger menu to open the sidebar and the filter input box shadow outline should render.

### DIFF
--- a/src/components/sidebar/components/sidebar-filter-input/sidebar-filter-input.module.css
+++ b/src/components/sidebar/components/sidebar-filter-input/sidebar-filter-input.module.css
@@ -18,8 +18,6 @@
   height: var(--input-height);
   margin-bottom: 16px;
   position: relative;
-  border-radius: 5px;
-  box-shadow: var(--token-surface-inset-box-shadow);
 }
 
 .filterIcon {
@@ -47,11 +45,14 @@
 }
 
 .filterInput {
+  appearance: none;
+  border-radius: 5px;
+  box-shadow: var(--token-surface-inset-box-shadow);
   composes: g-focus-ring-from-box-shadow from global;
   background-color: transparent;
   border: none;
   color: var(--token-color-palette-neutral-700);
-  font-size: 14px;
+  font-size: 16px;
   line-height: 18px;
   padding-bottom: 9px;
   padding-left: calc(var(--input-padding) * 2 + var(--input-icon-size));
--- a/src/components/sidebar/components/sidebar-filter-input/sidebar-filter-input.module.css
+++ b/src/components/sidebar/components/sidebar-filter-input/sidebar-filter-input.module.css
@@ -49,7 +49,7 @@
   background-color: transparent;
   border: none;
   border-radius: 5px;
-  box-shadow: inset 0 0 0 1px rgba(101, 106, 118, 0.3),
+  box-shadow: inset 0 0 1px 1px rgba(101, 106, 118, 0.3),
     inset 0 1px 2px 1px rgba(101, 106, 118, 0.1);
   color: var(--token-color-palette-neutral-700);
   font-size: 14px;
--- a/src/components/sidebar/components/sidebar-filter-input/sidebar-filter-input.module.css
+++ b/src/components/sidebar/components/sidebar-filter-input/sidebar-filter-input.module.css
@@ -45,12 +45,12 @@
 }
 
 .filterInput {
-  appearance: none;
-  border-radius: 5px;
-  box-shadow: var(--token-surface-inset-box-shadow);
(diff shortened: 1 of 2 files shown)


src/styles/sizing.css
```diff
@@ -1,5 +1,5 @@
 :root {
-  --input-height: 38px;
+  --input-height: 36px;
```
kendallstrautman, line 2: This seems to align with [the design spec](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/DevDot?node-id=7502%3A73814) and also matches the search input height.

src/components/sidebar/components/sidebar-filter-input/sidebar-filter-input.module.css
```diff
@@ -18,6 +18,8 @@
   height: var(--input-height);
   margin-bottom: 16px;
   position: relative;
+  border-radius: 5px;
+  box-shadow: var(--token-surface-inset-box-shadow);
```
kendallstrautman, line 22: lmk if there are concerns to moving this box shadow up to the wrapping div!

src/components/sidebar/components/sidebar-filter-input/sidebar-filter-input.module.css
```diff
@@ -48,9 +50,6 @@
   composes: g-focus-ring-from-box-shadow from global;
   background-color: transparent;
   border: none;
-  border-radius: 5px;
-  box-shadow: inset 0 0 0 1px rgba(101, 106, 118, 0.3),
-    inset 0 1px 2px 1px rgba(101, 106, 118, 0.1);
   color: var(--token-color-palette-neutral-700);
   font-size: 14px;
```
BRKalow, line 54: Wanna change the font size here to 16px? This will prevent iOS from zooming in a bit when focusing the input.

BRKalow:
Looks good on my device!

ashleemboyer:
Reviewing now! 👀

ashleemboyer (changes requested):
Did some searching on this `box-shadow` issue with `input`s on iOS and found something! Looks like we need to remove the default styling and then our custom styles should be applied as expected.

https://stackoverflow.com/a/23211766

Tested and confirmed with a CodeSandbox and my own iOS device: https://codesandbox.io/s/sharp-mcnulty-73v4g0?file=/styles.css

kendallstrautman:
@ashleemboyer awesome that worked. Thanks for finding that.

src/components/sidebar/components/sidebar-filter-input/sidebar-filter-input.module.css
```diff
@@ -45,14 +45,14 @@
 }
 
 .filterInput {
+  appearance: none;
+  border-radius: 5px;
+  box-shadow: var(--token-surface-inset-box-shadow);
   composes: g-focus-ring-from-box-shadow from global;
   background-color: transparent;
   border: none;
```
ashleemboyer, line 53: We've been putting `composes` first inside of blocks like this and then placing CSS properties after that, can this be updated to do that here as well??

```suggestion
  composes: g-focus-ring-from-box-shadow from global;
  appearance: none;
  background-color: transparent;
  border: none;
  border-radius: 5px;
  box-shadow: var(--token-surface-inset-box-shadow);
```

BRKalow (approved):

PR #569, opened 2022-06-08
feat: adds initial metadata setup to render proper meta tags

BRKalow:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1202114367927919/1202293883440482/f) 🎟️

## 🗒️ What

- Adds a new `HeadMetadata` component which handles rendering the proper meta tags for any given page.
- Introduces a new pattern where any given page can return a `metadata` prop, which will be used to build up the final meta tags. Currently supports `title` and `description`. I chose not to use `next/head` at each page to avoid potential boilerplate and duplication. We want some control over the final structure of the page title to ensure consistency.

### DIFF
--- a/src/components/head-metadata/index.tsx
+++ b/src/components/head-metadata/index.tsx
@@ -1,0 +1,42 @@
+import HashiHead from '@hashicorp/react-head'
+import { useCurrentProduct } from 'contexts'
+import { HeadMetadataProps } from './types'
+
+/**
+ * Builds up the the necessary meta tags for the site. Rendered in `_app`, where it receives `pageProps.metadata` as props
+ */
+export default function HeadMetadata(props: HeadMetadataProps) {
+  const { name: productName } = useCurrentProduct()
+
+  const titleParts = [__config.dev_dot.meta.title]
+  const description = props.description ?? __config.dev_dot.meta.description
+
+  if (productName) {
+    titleParts.unshift(productName)
+  }
+
+  if (props.title) {
+    titleParts.unshift(props.title)
+  }
+
+  // Ultimately we want to build up a title that looks like {specified page title} | {product} | {root title title}
+  const title = titleParts.join(' | ')
+
+  return (
+    // TODO: OpenGraph image to be passed as the image prop here
+    <HashiHead
+      title={title}
+      siteName={title}
+      pageName={title}
+      description={description}
+    >
+      <link rel="icon" href="/favicon.ico" sizes="any" />
+      <link rel="icon" href="/icon.svg" type="image/svg+xml" />
+      <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+      <meta
+        name="google-site-verification"
+        content="zRQZqfAsOX-ypXfU0mzAIzb5rUvj5fA4Zw2jWJRN-JI"
+      />
+    </HashiHead>
+  )
+}
--- a/src/components/head-metadata/index.tsx
+++ b/src/components/head-metadata/index.tsx
@@ -6,7 +6,7 @@ import { HeadMetadataProps } from './types'
  * Builds up the the necessary meta tags for the site. Rendered in `_app`, where it receives `pageProps.metadata` as props
  */
 export default function HeadMetadata(props: HeadMetadataProps) {
-  const { name: productName } = useCurrentProduct()
+  const { name: productName } = useCurrentProduct() ?? {}
 
   const titleParts = [__config.dev_dot.meta.title]
   const description = props.description ?? __config.dev_dot.meta.description
--- a/src/layouts/core-dev-dot-layout/index.tsx
+++ b/src/layouts/core-dev-dot-layout/index.tsx
@@ -1,17 +1,18 @@
-import { DatadogHeadTag, DatadogScriptTag } from 'lib/datadog'
 import Head from 'next/head'
+
+import { DatadogHeadTag, DatadogScriptTag } from 'lib/datadog'
 import s from './core-dev-dot-layout.module.css'
 
-const CoreDevDotLayout: React.FC = ({ children }) => (
-  <>
-    <Head>
-      <link rel="icon" href="/favicon.ico" sizes="any" />
-      <link rel="icon" href="/icon.svg" type="image/svg+xml" />
-      <DatadogHeadTag />
-    </Head>
-    <div className={s.root}>{children}</div>
-    <DatadogScriptTag />
-  </>
-)
+const CoreDevDotLayout: React.FC = ({ children }) => {
+  return (
+    <>
+      <Head>
+        <DatadogHeadTag />
+      </Head>
+      <div className={s.root}>{children}</div>
+      <DatadogScriptTag />
+    </>
+  )
+}
 
 export default CoreDevDotLayout
--- a/src/pages/_app.tsx
+++ b/src/pages/_app.tsx
@@ -1,12 +1,12 @@
 import React, { useEffect } from 'react'
 import dynamic from 'next/dynamic'
 import { Toaster } from 'components/toast'
-import Head from 'next/head'
 import { SSRProvider } from '@react-aria/ssr'
 import { ErrorBoundary } from 'react-error-boundary'
 import '@hashicorp/platform-util/nprogress/style.css'
 import useAnchorLinkAnalytics from '@hashicorp/platform-util/anchor-link-analytics'
 import CodeTabsProvider from '@hashicorp/react-code-block/provider'
+import HashiHead from '@hashicorp/react-head'
 import {
   AllProductDataProvider,
   CurrentProductProvider,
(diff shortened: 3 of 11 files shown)


BRKalow:
Still need to do a sweep of pages to ensure they have the proper data. 👀

src/components/head-metadata/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  if (productName) {
+    titleParts.unshift(productName)
+  }
+
+  if (props.title) {
+    titleParts.unshift(props.title)
+  }
+
+  // Ultimately we want to build up a title that looks like {specified page title} | {product} | {root title title}
```
BRKalow, line 22:
```suggestion
  // Ultimately we want to build up a title that looks like {specified page title} | {product} | {root title}
```

src/components/head-metadata/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const titleParts = [__config.dev_dot.meta.title]
+  const description = props.description ?? __config.dev_dot.meta.description
+
+  if (productName) {
+    titleParts.unshift(productName)
+  }
+
+  if (props.title) {
+    titleParts.unshift(props.title)
+  }
```
ashleemboyer, line 20: I had to look up `unshift` to double-check what it does, could comments be added above each of these with a brief note on the behavior?

```suggestion
  // add `productName` to beginning of `titleParts` array
  if (productName) {
    titleParts.unshift(productName)
  }

  // add `props.title` to beginning of `titleParts` array
  if (props.title) {
    titleParts.unshift(props.title)
  }
```

src/components/head-metadata/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  // Ultimately we want to build up a title that looks like {specified page title} | {product} | {root title}
```
ashleemboyer, line 22: 💭 What do you think about putting this comment in the block comment above `HeadMetadata` or creating a helper function just for building the title?

src/components/head-metadata/index.tsx
```diff
@@ -0,0 +1,47 @@
+import HashiHead from '@hashicorp/react-head'
+import { useCurrentProduct } from 'contexts'
+import { HeadMetadataProps } from './types'
+
+/**
+ * Builds up the the necessary meta tags for the site. Rendered in `_app`, where it receives `pageProps.metadata` as props
+ */
+export default function HeadMetadata(props: HeadMetadataProps) {
+  const { name: productName } = useCurrentProduct() ?? {}
```
ashleemboyer, line 9: 🙋🏻 Was there a page where `currentProduct` wasn't defined an issue came up here?

src/layouts/core-dev-dot-layout/index.tsx
```diff
@@ -1,6 +1,7 @@
 import { useRouter } from 'next/router'
 import { DatadogHeadTag, DatadogScriptTag } from 'lib/datadog'
 import Head from 'next/head'
+
```
ashleemboyer, line 4: 🙋🏻 Is this extra space intentional?

src/pages/_app.tsx
```diff
@@ -1,12 +1,12 @@
 import React, { useEffect } from 'react'
 import dynamic from 'next/dynamic'
 import { Toaster } from 'components/toast'
-import Head from 'next/head'
 import { SSRProvider } from '@react-aria/ssr'
 import { ErrorBoundary } from 'react-error-boundary'
 import '@hashicorp/platform-util/nprogress/style.css'
 import useAnchorLinkAnalytics from '@hashicorp/platform-util/anchor-link-analytics'
 import CodeTabsProvider from '@hashicorp/react-code-block/provider'
+import HashiHead from '@hashicorp/react-head'
```
ashleemboyer, line 9: 🙋🏻 It doesn't look like this is used below, accidental import?

src/views/collection-view/server.ts
```diff
@@ -94,6 +94,10 @@ export async function getCollectionPageProps(
 
   return {
     props: stripUndefinedProperties({
+      metadata: {
+        title: collection.shortName,
+        description: collection.description,
```
ashleemboyer, line 99: 🙋🏻 Are these properties on `collection` always defined or are we leveraging `stripUndefinedProperties` to set values to `null`? Same question for `src/views/tutorial-view/server.ts`.

src/views/product-downloads-view/server.ts
```diff
@@ -60,6 +60,9 @@ async function generateStaticProps(
    */
   return {
     props: stripUndefinedProperties({
+      metadata: {
+        title: 'Install',
+      },
```
ashleemboyer, line 65: 📌 Do we need a `TODO` here and in `src/views/product-tutorials-view/server.ts` to specify `description`?

src/components/head-metadata/index.tsx
BRKalow: The homepage is the main one I believe!

src/layouts/core-dev-dot-layout/index.tsx
BRKalow: It was, just to separate the imports here from the CSS import below.

src/views/collection-view/server.ts
BRKalow: I think they should always be defined, but yes we would rely on `stripUndefinedProperties` if they happen to be undefined. 👍

src/views/product-downloads-view/server.ts
BRKalow: I think it's adequately captured by this ticket, so I don't feel too inclined to add a TODO here: https://app.asana.com/0/1202114367927919/1202293883440482/f

README.md
```diff
@@ -169,6 +169,23 @@ Calls to `window.analytics.track()` are logged in development for easy iteration
...
+  return {
+    props: {
+      metadata: {
+        title: 'My Page', // Will be joined with the root site title
+        description: 'This is a cool page',
+      },
+    },
+  }
+}
+```
```
BrandonRomano, line 187: Neat, I really dig this interface!

BrandonRomano: Also nice docs 👌

src/components/head-metadata/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+      description={finalDescription}
+      image={ogImageUrl}
+    >
+      <link rel="icon" href="/favicon.ico" sizes="any" />
+      <link rel="icon" href="/icon.svg" type="image/svg+xml" />
+      <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+      <meta
+        name="google-site-verification"
+        content="zRQZqfAsOX-ypXfU0mzAIzb5rUvj5fA4Zw2jWJRN-JI"
+      />
```
BrandonRomano, line 49: I believe that the `HashiHead`'s `image` prop is insufficient alone to render images properly on all social networks.  Let me verify this real quick...

BrandonRomano: Yep won't render on Twitter correctly.. Lemme find what you gotta add here

CleanShot 2022-06-09 at 11 45 44@2x

BRKalow: I think you are right, we've been seeing some issues with twitter share cards and io sites as a result of this. Lemme check the validator.

BRKalow: Good catch @BrandonRomano I think we also need `twitter:image`: https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary-card-with-large-image

BrandonRomano: Here's some usage that I think illustrates all the other metadata props we have to add here for it to work on all networks:

https://github.com/hashicorp/how-hashicorp-works/blob/b8c2b775b88615d875e1d2adf92f10e829aecea2/layouts/base/index.jsx#L27-L50

src/components/head-metadata/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  const title = titleParts.join(' | ')
+
+  const finalDescription = description.replace(
+    '{product}',
+    productName ?? 'HashiCorp'
+  )
+
+  const ogImageUrl = `/og-image/${productSlug ?? 'base'}.jpg`
```
BrandonRomano, line 32: I think this is pretty low level to be including this logic, but I am fine with it as it will be easy to work away from in the event we need something more flexible.

I think the two scenarios I am thinking about:
- Adding share cards for other products, we'll probably have to dig in this code to see how they are referenced as it may not be super obvious as we roll out the other products
- In the event we decide to roll something with a different set of logic than just per-product / base share cards, this will have to be changed.

Just thinking out loud here, for the most part.

BrandonRomano (changes requested):
Added some sample code to ensure that the share cards work on all networks, notably Twitter is annoying.

Here's some quick links for further testing:
- https://cards-dev.twitter.com/validator
- https://developers.facebook.com/tools/debug/
- https://www.linkedin.com/post-inspector/

src/components/head-metadata/index.tsx
BRKalow: Good call. I think this is a reasonable initial implementation, we can thread through an `image` property much like we do `title` and `description`, from pages when we get to the point where we need more granular customization. I'll add a note to the README section explaining where we pull images from.

BrandonRomano: Rad, that more than satisfies me 👌 we can wait for the need to arise to refactor if need be if we need more flexability.

BrandonRomano (approved):

PR #576, opened 2022-06-08
chore: add cookie mgr sidebar sidecar

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksadd-cookie-mgr-sidebar-sidecar-hashicorp.vercel.app/waypoint/docs) 🔎
- [Asana task](https://app.asana.com/0/1202276096167515/1202407108957088) 🎟️

## 🗒️ What

This PR creates a new HOC footer that is set up to open the cookie manager. All pages should render this footer item, as requested by legal. This is then used in the sidebar sidecar layout, instead of making this driven by props. 

### Note: 

The width of the footer in this view isn't wide enough so it breaks to two lines. I'll add a note on [this ticket](https://app.asana.com/0/1202276096167515/1202074017129359) to address that as well, asking design now. cc @zchsh 
Screen Shot 2022-06-08 at 2 00 00 PM

## 🧪 Testing

Visit the preview and verify that all docs, tutorial pages render the cookie manager footer item.


src/layouts/base-new/index.tsx
```diff
@@ -15,6 +16,12 @@ const { ConsentManager, openConsentManager } = createConsentManager({
   preset: 'oss',
 })
 
+// To be consumed by Sidebar-Sidecar layout where the footer should be attached
+// to the same consent manager, but needs to render within the style of a different layout
+export function FooterWithCookieManager(props: FooterProps) {
+  return <Footer openConsentManager={openConsentManager} {...props} />
+}
```
kendallstrautman, line 23: Open to changing where this lives. But given we instantiate the consent manager in this file, I think it makes sense here.

zchsh (approved):
✅  🙌 This looks great to me! Tested and am seeing the ~link~ button 😬  on all views.

Definitely an interesting design challenge here with the restricted footer (thinking of 🎟️ [Update footers to align with new designs](https://app.asana.com/0/1201987349274776/1202074017129359/f)). I wonder if we might want to revisit 🎟️ [[footer] could we move the footer in sidebar-sidecar layout for consistency, implementation simplicity, and future-friendliness?](https://app.asana.com/0/1202110981600689/1202276670208195/f)

src/layouts/base-new/index.tsx
kendallstrautman: oh thats great! I didn't realize

src/layouts/sidebar-sidecar/index.tsx
```diff
@@ -4,6 +4,7 @@ import { motion, useReducedMotion } from 'framer-motion'
 
 // HashiCorp imports
 import { IconInfo16 } from '@hashicorp/flight-icons/svg-react/info-16'
+import { open as openDialog } from '@hashicorp/react-consent-manager'
```
kendallstrautman, line 7:
```suggestion
import { open as openConsentManager } from '@hashicorp/react-consent-manager'
```

PR #585, opened 2022-06-10
chore: make display helper classes use !important

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Adjust the global helper display classes to use `!important`. This is to ensure that the values apply. I think it makes sense to use here as if we're explicitly using the helper class we should always expect it to be applied, regardless of specificity.

## 🤷 Why

Fixes a bug with the mobile menu toggle showing up on large viewports 🤦

### DIFF
--- a/src/pages/style.css
+++ b/src/pages/style.css
@@ -114,19 +114,19 @@ Mobile & tablet visibility helpers, specific to dev-dot.
 */
 @media (--dev-dot-mobile) {
   .g-hide-on-mobile {
-    display: none;
+    display: none!important;
   }
 }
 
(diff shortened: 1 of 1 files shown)


ashleemboyer:
Taking a look! 👀

ashleemboyer (approved):
> I think it makes sense to use here as if we're explicitly using the helper class we should always expect it to be applied, regardless of specificity.

Agree, looks good to me! 👍🏻

zchsh:
Accidentally did duplicate work of this in #590 - merged there, so closing here 👍

PR #329, opened 2022-04-17
[DO NOT MERGE] Merge Sidebars

ashleemboyer:
## "Ready for Review" checklist

- [x] The Vercel preview link has been added to this PR's description
- [x] The Asana task has been added to this PR's description
- [x] This PR's link has been added to the "📐 Pull Request" field on the Asana task
- [x] This Vercel preview link has been added to the "🔍 Preview" field on the Asana task
- [x] The description template has been filled in below

---

## Relevant links

- [Preview link](https://dev-portal-git-ambmerge-sidebars-hashicorp.vercel.app/) 🔎
- Asana tasks 🎟️
  - [Update tutorial-view and collection-view to use main Sidebar via sidebarProps](https://app.asana.com/0/1202022787106807/1202132849199133/f)
  - [Sidebar: hr  is not a valid child of ul](https://app.asana.com/0/1202022787106807/1202111234165817/f)

## What

The file diff looks large at first, but the majority of the files changed are due to the removal of `resourcesNavData`.

1. `SidebarNavLink` now checks for `path` on an `item`
2. `SidebarNavSubmenu` no longer conditionally renders its `<ul>` element
3. Moved `<li>` wrapper from `SidebarNavLink` and `SidebarNavSubmenu` to `SidebarNavMenuItem`
4. Moved "Resources" section handling to `Sidebar`
5. Removed references to `resourcesNavData`
6. Updated `Sidebar`'s `addItemMetadata` `isActive` check to look at all three link properties (`path`, `fullPath`, `href`) and check if the link is internal before doing the `isActive` calculation
7. Added `visuallyHideTitle` prop to `Sidebar`
8. Added `g-screen-reader-only` CSS helper class
9. Updated `prepareNavDataForClient` and `prepareNavNodeForClient` to generate unique IDs for docs Sidebar nav items
10. Refactored `format-sidebar-sections` to be `generate-sidebar-nav-items`. Also renamed `formatCollectionToListItem` to `formatCollectionToMenuItem`.
11. Updated `getCollectionPageProps` to use new `generate-sidebar-nav-items`
12. Updated `collection-view` to use the main `Sidebar`
13. Removed `CollectionViewSidebar`
14. Moved `tutorial-view` `sidebarProps` creation to `src/views/tutorial-view/server.ts`
15. Updated `formatTutorialToMenuItem` and `formatCollectionToMenuItem` to:
  - No longer check `isActive`
  -  Set `path` instead of `href`
[...]

PR #589, opened 2022-06-10
Miscellaneous Sidebar adjustments/fixes

ashleemboyer:
## 🗒️ What

- Updates `prepareNavDataForClient` to generate unique IDs for each item (including dividers and headings) so we don't have `React` unique `key` warnings and so that we don't have to rely on an array `index` (which breaks rendering when a user is searching the sidebar)
- Updates reference to `prepareNavDataForClient` in `sidebar-sidecar`'s `server` code
- Adds suffixes to the `<button>` and `<ul>` element `id`s in `SidebarNavSubmenuItem` for more verbose/clear `id` names
- Updates `Sidebar` to not render the `Overview` menu item if the search input has a search value since this item won't be searched for
- Fixes a typo I noticed in the `filerInputWrapper` class name (was missing a `t`)

## 🛠️ How

From #329, where I originally did the work for generating unique IDs:

> There are block comments above `prepareNavDataForClient` and `prepareNavNodeForClient` that give more detail, but the gist is that they were updated help each other keep track of how many nodes have been prepared so that IDs could be generated based on an index as if the nested data had been flattened and traversed.

## 🧪 Testing

- [ ] Go to a docs page
- [ ] Make sure there are no key warnings for Sidebar in the console
- [ ] Make sure Sidebar searching still works as expected
- [ ] Make sure Sidebar submenus still work as expected

### DIFF
--- a/src/components/sidebar/components/sidebar-nav-menu-item/index.tsx
+++ b/src/components/sidebar/components/sidebar-nav-menu-item/index.tsx
@@ -183,12 +183,15 @@ const SidebarNavSubmenuItem = ({ item }: SidebarNavMenuItemProps) => {
     }
   }
 
+  const buttonId = `${item.id}-button`
+  const listId = `${item.id}-list`
   return (
     <>
       <button
-        aria-controls={item.id}
+        aria-controls={listId}
         aria-expanded={isOpen}
         className={s.sidebarNavMenuItem}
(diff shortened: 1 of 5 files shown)


zchsh (approved):
🙌 This looks good to me, not seeing any console.warns 👍

PR #590, opened 2022-06-10
fix: homepage menu issues

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Fixes two issues on the homepage:
- Mobile menu icon shown on large viewports, even though intent is not to show it
- Mobile menu not layering correctly when open

## 🛠️ How

- Adds `!important` to global helper classes, this seems necessary to account for potential loading order differences with classes applied through CSS modules
- Adds `z-index` to allow mobile menu to layer above content on homepage, even though it appears earlier in the DOM

## 🧪 Testing

- [ ] Visit [the homepage][preview]
    - The mobile menu icon should not appear on desktop viewports
    - The mobile menu should be layered correctly when open on mobile viewports

[task]: https://app.asana.com/0/1202114367927919/1202428126706761/f
[preview]: https://dev-portal-git-zsfix-home-mobile-menu-hashicorp.vercel.app/

### DIFF
--- a/src/pages/style.css
+++ b/src/pages/style.css
@@ -114,19 +114,19 @@ Mobile & tablet visibility helpers, specific to dev-dot.
 */
 @media (--dev-dot-mobile) {
   .g-hide-on-mobile {
-    display: none;
+    display: none !important;
   }
 }
 
 @media (--dev-dot-tablet) {
   .g-hide-on-tablet {
-    display: none;
+    display: none !important;
   }
 }
 
 @media (--dev-dot-desktop) {
   .g-hide-on-desktop {
-    display: none;
+    display: none !important;
   }
 }
 
--- a/src/views/homepage/index.tsx
+++ b/src/views/homepage/index.tsx
@@ -48,7 +48,7 @@ const HomePageMobileMenu = () => {
 
 const HomePageContent = () => {
   return (
-    <>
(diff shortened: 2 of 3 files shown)


alexcarpenter (approved):
👍🏼

src/pages/style.css
```diff
@@ -114,19 +114,19 @@ Mobile & tablet visibility helpers, specific to dev-dot.
 */
 @media (--dev-dot-mobile) {
   .g-hide-on-mobile {
-    display: none;
+    display: none !important;
```
zchsh, line 117: `!important` is _generally_ no good, in this case though it seems we need it.

This is maybe a good example of a caveat of using global helper classes, vs inlining these styles 🤷 For `.mobileMenuButton` specifically, if this style was inline, it might be more clear why this breaks (we compose and then basically override with `display: flex`).

src/views/homepage/index.tsx
```diff
@@ -48,7 +48,7 @@ const HomePageMobileMenu = () => {
 
 const HomePageContent = () => {
   return (
-    <>
+    <div className={s.homepageContent}>
```
zchsh, line 51: We need the mobile menu to stack above `homepageContent`, class here to make that possible.

PR #593, opened 2022-06-10
feat: add search no results

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksadd-search-no-results-hashicorp.vercel.app/waypoint/docs) 🔎
- [Asana task](https://app.asana.com/0/1202276096167515/1202364251355949) 🎟️

## 🗒️ What

Adds a basic 'no results' component for the search panel when no hits exist. 

## 🧪 Testing

Visit a docs page, type in a long word until you see the no results state. 

## 💭 Anything else?

Kept this simple without specific guidance from design. The navigation bar does feel a bit odd when this renders, should I remove it entirely? The `esc` note still feels helpful. Also, feels like a bit of an edge case.


src/components/algolia-search/algolia-search.module.css
```diff
@@ -147,13 +147,15 @@
...
-.itemLink {
+.itemLinkLayout {
   position: relative;
   display: block;
   padding: 8px;
   padding-right: 36px;
   border-radius: 5px;
+}
 
+.itemLinkHover {
```
kendallstrautman, line 158: Separated the hover styles into own class so the item layout could be shared.

BRKalow:
@kendallstrautman it seems like the search results panel isn't closing now after clicking out of the search input, I'm guessing due to:

```ts
{(autocompleteState.isOpen || autocompleteState.query) && (
```

is `isOpen` false if the query returns no results?

kendallstrautman:
@BRKalow ah dang, that makes sense. Yes `isOpen` is false if there are no results. I'll have to do a bit more digging to see a better way to handle this.

kendallstrautman:
@BRKalow @ashleemboyer I reworked this to use the `shouldPanelOpen` function, and it seems to have fixed the focus issue mentioned. And I haven't seen the 'ghost ui' issue with the navigation ui. Lmk y'all thoughts!

Regarding that search input issue where the panel doesn't open again if you scale things down unless it is cleared. That appears to be unrelated to these changes and I observed it on production currently. So for now, I'm not going to fix that as it is an edge case.

src/components/algolia-search/components/panel.tsx
```diff
@@ -110,12 +117,11 @@ export default forwardRef(function Panel<THit extends Hit<unknown>>(
         style={panelPositionStyle}
       >
         <div className={s.panelLayout}>
-          {collections.map((collection, index) => {
-            const { source, items } = collection
-
-            return (
-              <section key={`source-${index}`}>
-                {items.length > 0 && (
+          {showResults ? (
```
kendallstrautman, line 120: Adjusted this so it will work if / when we have multiple collections. Previously it was just assuming there was only one collection.

src/components/algolia-search/lib/get-items-exist.ts
```diff
@@ -0,0 +1,18 @@
+import { AutocompleteCollection, BaseItem } from '@algolia/autocomplete-core'
+
+/**
+ * If any collection has hit 'items', return true. This is used to help
+ * determine whether the 'no results' state should show
+ */
+export function getItemsExist(collections: AutocompleteCollection<BaseItem>[]) {
```
kendallstrautman, line 7: Maybe `getHasResults` is a better name?... 💭 I don't love this name.

ashleemboyer:
Will re-review in the AM! 😊

src/components/algolia-search/lib/get-items-exist.ts
ashleemboyer: 💭 What do you think of just `hasResults`?

kendallstrautman:
I added a ticket re: that ui bug with the nav items after scaling up. I saw it again. https://app.asana.com/0/1202097197789424/1202489876954744

ashleemboyer (approved):
LGTM! 🙌🏻

BRKalow (approved):
Looks great!

PR #601, opened 2022-06-14
fix: adds fallback for navigationHeaderHeight and use observer.disconnect

brkalow:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1202114367927919/1202439054132401/f) 🎟️

## 🗒️ What

Adds a fallback in `getFullNavHeaderHeight` to address an intermittent error where reading the CSS custom property was not returning a valid value. Also using `observer.disconnect()` instead of iterating through the header to ensure intersection observers from past pages are not still observing.

The error can be seen if you open up the console, navigate to vault's product landing page, then navigate from there to waypoint's product landing page.

```
1970-d5d641d9ae986b6c.js:1 Uncaught TypeError: Cannot read properties of null (reading 'getBoundingClientRect')
    at 1970-d5d641d9ae986b6c.js:1:5744
    at Set.forEach (<anonymous>)
    at IntersectionObserver.<anonymous> (1970-d5d641d9ae986b6c.js:1:5691)
```

### DIFF
--- a/src/components/algolia-search/algolia-search.module.css
+++ b/src/components/algolia-search/algolia-search.module.css
@@ -66,6 +66,10 @@
     outline: none;
   }
 
+  &:placeholder-shown {
+    text-overflow: ellipsis;
+  }
+
   /* Remove native appearence */
   &::-webkit-search-decoration,
   &::-webkit-search-cancel-button,
@@ -86,7 +90,8 @@
   margin-right: 4px;
 
   &[hidden] {
-    display: none;
+    visibility: hidden; /* Retain space so the input doesn't overflow */
+    pointer-events: none;
   }
 }
 


src/layouts/sidebar-sidecar/components/table-of-contents/use-active-section.ts
```diff
@@ -138,14 +154,7 @@ export function useActiveSection(
...
 
     return () => {
-      headings.forEach((section) => {
-        const el = document
-          .getElementById('main')
-          ?.querySelector(`#${section.slug}`)
-        if (el) {
-          observer.unobserve(el)
-        }
-      })
```
BRKalow, line 148: This was a funky one, because we were iterating through each heading and unobserving like this, on navigation from a page with a sidecar to another page with a sidecar, an old observer was still firing if there were headings with the same slug on the new page. Switched to `.disconnect()` here to avoid that.

src/layouts/sidebar-sidecar/components/table-of-contents/use-active-section.ts
```diff
@@ -18,6 +20,20 @@ const getFullNavHeaderHeight = () => {
...
+   */
+  if (Number.isNaN(navigationHeaderHeight)) {
+    if (process.env.NODE_ENV === 'development') {
+      console.warn(
+        'Unable to read --navigation-header-height, falling back to static value.'
+      )
+    }
+
+    return NAVIGATION_HEADER_HEIGHT_FALLBACK
+  }
```
ashleemboyer, line 35: 💭 Thoughts on putting this in `getCSSVariableFromDocument` instead and adding a `fallbackValue` option?

BRKalow: Sure! Updated in https://github.com/hashicorp/dev-portal/pull/601/commits/3f041cba52c18731f34b727b410cfac7acd5e703

src/lib/get-css-variable-from-document.ts
```diff
@@ -1,13 +1,37 @@
...
+    value = parseInt(value, 10)
+  }
+
+  if (options.fallback && (value === '' || Number.isNaN(value))) {
+    if (process.env.NODE_ENV === 'development') {
+      console.warn(
+        `Unable to read ${variableName}, falling back to static value.`
+      )
+    }
+    return options.fallback
```
ashleemboyer, line 34: ⛏️ Should this set `value` instead of early return?

```suggestion
    value = options.fallback
```

ashleemboyer (approved):

PR #604, opened 2022-06-14
fix: home page nav mobile overflow

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Fixes an issue where the top nav contents would overflow on very small viewports.

## 🛠️ How

Adjusts height of dev-dot logo on mobile to match design spec.

## 📸 Design Screenshots

CleanShot 2022-06-13 at 16 15 08@2x

## 🧪 Testing

- [ ] Visit the [home page][preview], and ensure the top nav does not overflow on mobile viewports
    - The logo should be smaller on mobile viewports
    - All other top nav appearance and functionality should remain the same

[task]: https://app.asana.com/0/1202114367927919/1202439831418472/f
[preview]: https://dev-portal-git-zsfix-top-nav-mobile-overflow-hashicorp.vercel.app

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,13 +12,17 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - run: npm ci
-      - uses: mattallty/jest-github-action@v1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
         with:
-          coverage-comment: false
+          node-version-file: '.nvmrc'
+
+      - run: npm ci
+      - run: npm run test:ci
+
   lint:
+    name: 'Lint'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


src/components/navigation-header/components/home-page-content/home-page-content.module.css
```diff
@@ -1,13 +1,16 @@
 .siteLogo {
   display: flex;
 
-  /* Note: padding here necessary to match height of .productLogo
+  /* Note: min-height necessary to match height of .productLogo
      in navigation-header.module.css */
-  padding-top: 4px;
-  padding-bottom: 4px;
+  min-height: 40px;
```
zchsh, line 6: Since the height of the `svg` can vary, I'm using `min-height` here to achieve the same effect as before.

Without this adjustment, the height of the top nav bar varies on the home page due to the different logo size. An alternate approach might be to set exact padding values with a media query; however this seemed more opaque and less maintainable.

alexcarpenter: Should the logo be vertically centered?

CleanShot 2022-06-15 at 10 04 48@2x

BrandonRomano (approved):

braydenlove (approved):
Nice, approved!

PR #610, opened 2022-06-14
[home page] Update Learn cards

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambupdate-certification-cards-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202114367927919/1202439144025013/f) 🎟️

## 🗒️ What

- Updates the Learn cards on the home page to:
  - have their data loaded server-side via a slug
  - render the card based on the type of content (tutorial or collection)

## 🤷 Why

- The data was previously hard-coded
- Tutorial cards were being shown for Collection content

## 🛠️ How

- Added a `server.ts` file for `HomePageView` that loads featured learn data in the same way as the install views do
- Updated `HomePageView` to accept a new `featuredLearnCards` prop that's generated server side
- Updated `src/views/homepage/components/learn-section/index.tsx` to render the correct kind of card based on the content

## 🧪 Testing

- [ ] Go to the home page
- [ ] Make sure there are now collection cards rendering in the learn section

## 💭 Anything else?

It would be really nice to abstract this functionality into `lib` helpers instead of sourcing them from the install view and learn card helpers. Soon! 🤩

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -1,5 +1,20 @@
+import { GetStaticProps } from 'next'
 import BaseNewLayout from 'layouts/base-new'
 import HomePageView from 'views/homepage'
+import { generateStaticProps } from 'views/homepage/server'
+
+export const getStaticProps: GetStaticProps = async () => {
+  return generateStaticProps({
+    featuredLearnContent: [
+      {
+        collectionSlug: 'vault/associate-cert',
+      },
+      {
+        collectionSlug: 'vault/ops-pro-cert',
+      },
+    ],
+  })
+}
 
 HomePageView.layout = BaseNewLayout
 export default HomePageView
--- a/src/views/homepage/components/learn-section/index.tsx
+++ b/src/views/homepage/components/learn-section/index.tsx
@@ -1,24 +1,25 @@
 import type { CSSProperties, ReactElement } from 'react'
 import slugify from 'slugify'
 import { IconArrowRight16 } from '@hashicorp/flight-icons/svg-react/arrow-right-16'
+import { FeaturedLearnCard } from 'views/product-downloads-view/types'
+import CollectionCard from 'components/collection-card'
 import Heading from 'components/heading'
 import StandaloneLink from 'components/standalone-link'
 import TutorialCard from 'components/tutorial-card'
-import { TutorialCardProps } from 'components/tutorial-card/types'
 import s from './learn-section.module.css'
 
 interface LearnSectionProps {
   media: ReactElement
   heading: string
   description: ReactElement
-  tutorials: Array<TutorialCardProps>
+  learnCards: FeaturedLearnCard[]
 }
 
 export default function LearnSection({
   media,
   heading,
   description,
-  tutorials,
+  learnCards,
 }: LearnSectionProps) {
   return (
     <section className={s.learnSection}>
@@ -51,17 +52,16 @@ export default function LearnSection({
           className={s.tutorialsList}
           style={
             {
-              '--tutorials-count': tutorials.length,
+              '--tutorials-count': learnCards.length,
             } as CSSProperties
           }
         >
-          {tutorials.map((tutorial: TutorialCardProps) => {
+          {learnCards.map((cardProps: FeaturedLearnCard) => {
+            const { id, type } = cardProps
             return (
-              <li
-                key={slugify(tutorial.heading, { lower: true })}
-                className={s.tutorialsListItem}
-              >
-                <TutorialCard {...tutorial} />
+              <li key={id} className={s.tutorialsListItem}>
+                {type === 'collection' && <CollectionCard {...cardProps} />}
+                {type === 'tutorial' && <TutorialCard {...cardProps} />}
               </li>
             )
           })}
--- a/src/views/homepage/server.ts
+++ b/src/views/homepage/server.ts
@@ -1,0 +1,50 @@
+import { stripUndefinedProperties } from 'lib/strip-undefined-props'
+import { getInlineContentMaps } from 'lib/tutorials/get-inline-content-maps'
+import {
+  FeaturedLearnCard,
+  FeaturedLearnContent,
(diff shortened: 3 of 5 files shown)


src/views/homepage/components/learn-section/index.tsx
```diff
@@ -51,17 +52,16 @@ export default function LearnSection({
...
+            const { id, type } = cardProps
             return (
-              <li
-                key={slugify(tutorial.heading, { lower: true })}
-                className={s.tutorialsListItem}
-              >
-                <TutorialCard {...tutorial} />
+              <li key={id} className={s.tutorialsListItem}>
+                {type === 'collection' && <CollectionCard {...cardProps} />}
+                {type === 'tutorial' && <TutorialCard {...cardProps} />}
```
zchsh, line 64: ⛏️ Nitpick: This feels a little odd to me since it feels technically possible that nothing would render here. One approach could be to do like `const CardComponent = type === 'collection' ? CollectionCard : TutorialCard` in the `map` body, and then render `<CardComponent {...cardProps} />`.

For context: I think we had [a Slack thread](https://hashicorp.slack.com/archives/C02T80JTNS2/p1649872956699629), where it felt like we had alignment on wanting to avoid `&&` for conditional rendering (here's [an article I like on the pitfalls of &&](https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx))... though I think we still haven't set anything in stone, so definitely feel free to ignore this if it doesn't feel productive 👍

zchsh (dismissed):
🙌 ✅  This looks great to me!

Had some _very_ nitpick-y things, so definitely feel free to ignore 👍 🚀

src/pages/index.tsx
```diff
@@ -1,5 +1,20 @@
...
+import { generateStaticProps } from 'views/homepage/server'
+
+export const getStaticProps: GetStaticProps = async () => {
+  return generateStaticProps({
+    featuredLearnContent: [
+      {
+        collectionSlug: 'vault/associate-cert',
+      },
+      {
+        collectionSlug: 'vault/ops-pro-cert',
```
kendallstrautman, line 13: We should probably adjust all the data in this view to be editable via a json file, to follow similar patterns. I'm just thinking of authors needing to come and update content in this file, whereas an abstracted data file would be ideal. Of course, this is a long term thought, [not necessary for this PR. ](https://app.asana.com/0/1202276096167515/1202445483168876)

src/pages/index.tsx
```diff
@@ -1,5 +1,20 @@
+import { GetStaticProps } from 'next'
 import BaseNewLayout from 'layouts/base-new'
 import HomePageView from 'views/homepage'
+import { generateStaticProps } from 'views/homepage/server'
+
+export const getStaticProps: GetStaticProps = async () => {
+  return generateStaticProps({
+    featuredLearnContent: [
```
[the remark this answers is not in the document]
kendallstrautman: I'd agree for hedging against flexibility here. We could have something like an inline collections array of slugs. But also, this works for now.

src/views/homepage/server.ts
```diff
@@ -0,0 +1,50 @@
...
+const generateStaticProps = async ({
+  featuredLearnContent = [],
+}: GenerateStaticPropsOptions): Promise<GenerateStaticPropsOptionsResult> => {
+  // Gather tutorials and collections based on slugs used
+  const { inlineCollections, inlineTutorials } = await getInlineContentMaps(
+    featuredLearnContent
+  )
+
+  // Transform featured tutorial and collection entries into card data
+  const featuredLearnCards: FeaturedLearnCard[] = featuredLearnContent.map(
```
kendallstrautman, line 23: If we keep this here we could note the duplication of this code in the install view, just as a 🔖 to be aware of the spread.

kendallstrautman (approved):
My vote would be to remove the tutorial card option from this implementation and just do collection cards. But approving as to not block this work 👌

braydenlove (approved):
Nice job, approved.

ashleemboyer:
@kendallstrautman @zchsh @braydenlove I've updated this PR to now only show/allow collection cards! I've also added a `content.json` file as a head start to working on this new task: [Refactor homepage data to json file](https://app.asana.com/0/1202114367927919/1202445483168876/f). This is ready for re-review ✨

src/views/homepage/components/learn-section/index.tsx
```diff
@@ -51,17 +52,15 @@ export default function LearnSection({
           className={s.tutorialsList}
           style={
             {
-              '--tutorials-count': tutorials.length,
+              '--tutorials-count': collectionCards.length,
```
kendallstrautman, line 55: ⛏️ Nit: Perhaps we should update this css var? `--collection-cards-count`?  (non blocking)

src/views/homepage/server.ts
```diff
@@ -0,0 +1,41 @@
...
+      '`HomePageView` requires `collectionSlugs` to be defined in `src/pages/content.json` but none were provided.'
+    )
+  }
+
+  // Fetch the collections specified
+  const collections = await getCollections(collectionSlugs)
+
+  // Transform collection entries into card data with `collectionSlugs` order
+  const collectionCards = collectionSlugs.map((collectionSlug: string) => {
+    const thisCollection = collections.find(
```
kendallstrautman, line 27: Might be worth a comment here re: sorting 🤷‍♀️

kendallstrautman (approved):
Looks great! Ty for reworking this 🙇

src/views/homepage/components/learn-section/index.tsx
ashleemboyer: Good point! Glazed right over this one 😆

zchsh (approved):
🙌  +1 on the extra optional push here for making some tough calls and doing a super clean rework! 💯

braydenlove (approved):
_`APPROVED`_ ✅

PR #614, opened 2022-06-14
fix: dev portal routes visible under io domains for same product

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Fixes this: https://www.vaultproject.io/vault

Exposing dev portal routes for the same product under its .io domain. Not something we want to do! Added a test case to validate that you can't hit a dev portal route for the same product. I also removed an unused block of redirect generation.

### DIFF
--- a/build-libs/proxy-settings.js
+++ b/build-libs/proxy-settings.js
@@ -105,7 +105,7 @@ function getDevPortalRoutesToProxy(product) {
     })
     .filter((ent) => ent.isDirectory() && ent.name !== '_proxied-dot-io')
     .map(({ name }) => ({
-      proxiedRoute: `/:path((?!${product})${name}.*)`,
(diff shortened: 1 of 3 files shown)


build-libs/proxy-settings.js
```diff
@@ -105,7 +105,7 @@ function getDevPortalRoutesToProxy(product) {
     })
     .filter((ent) => ent.isDirectory() && ent.name !== '_proxied-dot-io')
     .map(({ name }) => ({
-      proxiedRoute: `/:path((?!${product})${name}.*)`,
```
BRKalow, line 108: Honestly, I have no idea _why_ I did it this way the first time 🤔 🙃

kendallstrautman (approved):

PR #618, opened 2022-06-15
chore: npm test directly instead of jest GH action

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Remove use of the jest runner action and run jest directly. This action was obscuring output and wasn't really providing benefit, so I'm removing it to align with how we run tests elsewhere. Also running tests with `--runInBand` to see if that addresses some flaky failures.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
           node-version-file: '.nvmrc'
 
       - run: npm ci
-      - run: npm test -- -i
+      - run: npm test:ci
 
   lint:
     name: 'Lint'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
           node-version-file: '.nvmrc'
 
       - run: npm ci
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -12,13 +12,17 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - run: npm ci
-      - uses: mattallty/jest-github-action@v1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
thiskevinwang, line 18: Curious why this action needs `GITHUB_TOKEN`... seems strange!

thiskevinwang (approved):
🚀

.github/workflows/ci.yml
BRKalow: I think it uses the checks API to report results back to specific line numbers and such: https://github.com/mattallty/jest-github-action

Kinda cool but didn't seem to work too well.

PR #619, opened 2022-06-15
feat: extract home page content to json

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Consolidates much of the copy and other content on the home page to be authorable via a single `.json` file.

## 🤷 Why

To provide a consolidated authoring interface for home page content. From #610, "tutorial" section cards have to be authored in a `slug` format and then transformed, which causes a kind of content spread.

## 🛠️ How

- creates Joi schema and coupled Typescript types for home page content, and validates against it in `getStaticProps`
- refactors home page components, to accept content props rather than using inline content, and to create separate `types.ts` files to reference in home page prop types
- creates home page content in `src/pages/content.json` to match schema and replicate existing content in JSX

## 🧪 Testing

- [ ] Visit [the home page][preview], and ensure content looks as expected

[task]: https://app.asana.com/0/1202114367927919/1202445483168876/f
[preview]: https://dev-portal-git-zshome-content-all-json-hashicorp.vercel.app/


src/pages/index.tsx
```diff
@@ -1,12 +1,10 @@
...
 import { generateStaticProps } from 'views/homepage/server'
-import pageContent from './content.json'
 
 export const getStaticProps: GetStaticProps = async () => {
-  return generateStaticProps(pageContent)
+  const contentJsonFile = 'src/pages/content.json'
+  return generateStaticProps(contentJsonFile)
 }
 
-HomePageView.layout = BaseNewLayout
```
zchsh, line 11: This is already set in `views/homepage/index.tsx`.

src/views/homepage/components/merchandising-slots/slots/hashiconf-global-slot/index.tsx
```diff
@@ -2,9 +2,10 @@ import Image from 'next/image'
 import VisuallyHidden from '@reach/visually-hidden'
 import CardLink from 'components/card-link'
 import Text from 'components/text'
+import { HashiConfGlobalSlotProps } from './types'
 import s from './hashiconf-global-slot.module.css'
 
-function HashiConfGlobalSlot() {
+function HashiConfGlobalSlot({ description }: HashiConfGlobalSlotProps) {
```
zchsh, line 8: For merchandising slots, it didn't feel feasible to expose much content, as it feels pretty tightly coupled to the presentation here (which I think is kind of a good thing on more "polished" pages - trades off some content flexibility for highly intentional polish 🙌 )

src/views/homepage/index.tsx
```diff
@@ -45,122 +38,55 @@ const HomePageMobileMenu = () => {
...
   )
 }
 
-function HomePageView({ collectionCards }: HomePageViewProps): ReactElement {
+function HomePageView({ content }: HomePageProps): ReactElement {
   return (
     <div className={s.homepage}>
       <HomePageMobileMenu />
-      <HomePageContent collectionCards={collectionCards} />
+      <HomePageContent {...content} />
```
zchsh, line 89: This is one of the few cases where props spreading felt fine to me; thanks to TypeScript, and since the `HomePageContent` component is in the same file.

src/views/product-landing/helpers/index.ts
```diff
@@ -1,4 +1,3 @@
-export * from './validate-against-schema'
```
zchsh, line 1: Moved to `src/lib` as now used for multiple views - here in `product-landing`, and now for `home` as well.

BrandonRomano:
Tagging @alexcarpenter here as well for review -- I feel like he might have some good opinions on this, if he has the time to take a peek. If no no worries!

src/views/homepage/components/merchandising-slots/slots/hashiconf-global-slot/index.tsx
alexcarpenter: Yeah, these merch slots are pretty bespoke, and still unsure there is any pattern to follow that would allow JSON content to easily modify these that wouldn't be too overwhelming.

alexcarpenter (commented):
No real comments for me, makes sense. First time seeing the validation schema setup, nice!

src/views/homepage/helpers/transform-raw-content.ts
```diff
@@ -0,0 +1,38 @@
...
+  const { hero, navNotice, merchandising, learnSection, preFooter } =
+    authoredContent
+
+  // For the learnSection, transform collectionSlugs to card data
+  const { collectionSlugs, ...restLearnSection } = learnSection
+  const { inlineCollections } = await getInlineContentMaps(learnSection)
+  const collectionCards = collectionSlugs.map((slug: string) => {
+    const collectionData = inlineCollections[slug]
+    return formatCollectionCard(collectionData)
+  })
```
ashleemboyer, line 23: Why was this part changed from what was just approved and merged in #610?

```suggestion
  // Transform collection entries into card data with `collectionSlugs` order
  const collectionCards = collectionSlugs.map((collectionSlug: string) => {
    const thisCollection = collections.find(
      (collection: Collection) => collection.slug === collectionSlug
    )
    return formatCollectionCard(thisCollection)
  })
```

zchsh: @ashleemboyer Great question, it's not a very meaningful change - functions same as it did previously. Apologies for the thrash!

Intent here was to match patterns in other pages to fill in tutorial and collection card data (such as tutorial product landing pages and the install view).

This pattern is intended to be slightly more flexible - eg if other learn content were to be added in `learnSection` or even in some other section on the home page, we'd traverse the home page content with `getInlineContentMaps` once and be able to fill in all sections (rather than having to individual reason about tutorial and collection fields, and individually fetch with `getCollections` / `getTutorials`).

Let me know if that seems reasonable, happy to adjust this if there are any strong opinions on why we should take a different approach here 🙇‍♂️ 💖

ashleemboyer: I think until we know we need to fetch both tutorials and collections, that we should leave it as-is and not use `getInlineContentMaps` in this instance. I had opted against it in #610 because the only extra thing it does for us here is re-format the response from `getCollections` into an object from an array which doesn't feel necessary here when we're not mixing collection and tutorial cards in the same section.

BRKalow: Pulling from an object instead of calling `.find()` here feels like a nice improvement. If it aligns with how we fetch learn data on other pages, the pattern will be more discoverable and consistent if we do need to adjust it down the road.

If @zchsh's approach is functionally equivalent, I don't think we need to revert here.

src/views/homepage/content-schema.ts
```diff
@@ -0,0 +1,130 @@
+import Joi from 'joi'
```
BRKalow, line 1: 👍  I dig this!

src/views/homepage/helpers/transform-raw-content.ts
BRKalow: I don't think there was anything wrong with the initial approach 😃, but code evolves and a new pair of eyes might see opportunities for refactoring / adjustment on another pass! 🧑‍🤝‍🧑

PR #626, opened 2022-06-16
Fixing Dialog/Opt-out Dialog visual issues

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambopt-out-dialog-hashicorp.vercel.app/waypoint/docs?optInFrom=waypoint-io) 🔎
- Asana tasks 🎟️
  - [[text area] Text area in opt-out dialog not showing border on iOS devices](https://app.asana.com/0/1202114367927919/1202440311005396/f)
  - [[dialog] Refine opt-out beta feedback at smaller viewports](https://app.asana.com/0/1202114367927919/1202440311005385/f)
  - [[dialog] background color visual bugs](https://app.asana.com/0/1202114367927919/1202440311005416/f)

## 🗒️ What

- Fixes "missing" border on opt-out dialog `textarea` for iOS devices
- Changes the placement of `Dialog`s on viewports <= `728px` wide
- Fixes the look of the `Dialog` background

## 🛠️ How

- The "missing" border was fixed by setting `appearance: none` on the `textarea`. We also had to do this for `input` elements. It will be very beneficial to create UI components soon that handle this for us. The Design Systems team is currently working on form elements, so I imagine we'll work on this soon. Especially since we'll be adding more form elements with the development of authentication and more.
- To update the placement of `Dialog` on various viewports, I added a wrapper `<div>` around the `DialogContent` element and updated styles for both elements. The approach I took was for the wrapper to have `padding` that restrains its contents, rather than add margins to the `DialogContent` element because that would require more math to calculate the width of the element.
- To fix the strange-looking background behind `Dialog`s, I set `background: none` on the `AnimatedDialogOverlay` element. Setting `background: none` on the `DialogOverlay` did nothing, and I made note of this in a comment in the CSS.

## 🧪 Testing

- [ ] Go to the preview link: [/waypoint/docs?optInFrom=waypoint-io](https://dev-portal-git-ambopt-out-dialog-hashicorp.vercel.app/waypoint/docs?optInFrom=waypoint-io)
- [ ] Click the "Leave Beta" button towards the top of the page
- [ ] With an iOS device, make sure the `textarea` border is visible
- [ ] Ensure the placing of the dialog is desirable in both narrow and wide viewports
- [ ] Ensure the background behind the dialog looks correct

### DIFF
--- a/src/components/dialog/index.tsx
+++ b/src/components/dialog/index.tsx
@@ -31,9 +31,11 @@ export default function Dialog({
               exit={{ y: 50 }}
               transition={{ min: 0, max: 100, bounceDamping: 8 }}
             >
-              <DialogContent className={s.content} aria-label={label}>
-                {children}
-              </DialogContent>
+              <div className={s.contentWrapper}>
+                <DialogContent className={s.content} aria-label={label}>
+                  {children}
+                </DialogContent>
+              </div>
             </m.div>
           </DialogOverlay>
         </AnimatedDialogOverlay>
--- a/src/components/dialog/index.tsx
+++ b/src/components/dialog/index.tsx
@@ -14,14 +14,14 @@ export default function Dialog({
     <AnimatePresence>
       {isOpen && (
         <AnimatedDialogOverlay
-          className={s.dialogOverlay}
+          className={s.animatedDialogOverlay}
           initial={{ opacity: 0 }}
           animate={{ opacity: 1 }}
           exit={{ opacity: 0 }}
           onDismiss={onDismiss}
         >
           <DialogOverlay
-            className={s.overlay}
+            className={s.dialogOverlay}
             isOpen={isOpen}
             onDismiss={onDismiss}
           >
--- a/src/components/dialog/dialog.module.css
+++ b/src/components/dialog/dialog.module.css
@@ -3,10 +3,24 @@
   z-index: 100;
 }
 
-.content {
-  box-shadow: var(--token-surface-overlay-box-shadow);
-  border-radius: 6px;
+.contentWrapper {
+  align-items: flex-start;
+  display: flex;
+  height: 100vh;
+  justify-content: center;
   padding: 24px;
+  width: 100vw;
+
+  @media (--dev-dot-tablet-up) {
+    padding-top: 10vh;
+  }
+}
+
+.content {
+  border-radius: 6px;
+  box-shadow: var(--token-surface-overlay-box-shadow);
+  margin: 0;
   max-width: 600px;
-  min-width: 300px;
+  padding: 24px;
+  width: 100%;
 }
--- a/src/components/dialog/dialog.module.css
+++ b/src/components/dialog/dialog.module.css
@@ -1,5 +1,18 @@
-.overlay {
-  background: var(--token-color-border-strong);
+/*
+***
+* This class is necessary so avoid issues with the layer behind the dialog
+* looking strange. Setting `background: none` on the `dialogOverlay` class does
+* not work.
+* 
+* ref: https://app.asana.com/0/1202114367927919/1202440311005416/f
+***
+*/
+.animatedDialogOverlay {
+  background: none;
+}
+
+.dialogOverlay {
+  background-color: var(--token-color-border-strong);
   z-index: 100;
 }
 
--- a/src/components/opt-in-out/components/opt-out-form/index.tsx
+++ b/src/components/opt-in-out/components/opt-out-form/index.tsx
@@ -1,4 +1,4 @@
-import { FormEvent, useState, ChangeEvent, useRef } from 'react'
+import { FormEvent, useState, ChangeEvent } from 'react'
 import { IconX16 } from '@hashicorp/flight-icons/svg-react/x-16'
 import Button from 'components/button'
 import Text from 'components/text'
--- a/src/components/opt-in-out/components/opt-out-form/opt-out-form.module.css
+++ b/src/components/opt-in-out/components/opt-out-form/opt-out-form.module.css
@@ -74,6 +74,7 @@
 
 .optionalText {
   composes: hds-typography-body-200 hds-font-weight-regular from global;
+  appearance: none; /* resets iOS styles */
   border-radius: 5px;
   border: unset;
   box-shadow: var(--token-surface-base-box-shadow);
--- a/src/components/opt-in-out/components/opt-out-form/opt-out-form.module.css
+++ b/src/components/opt-in-out/components/opt-out-form/opt-out-form.module.css
@@ -74,15 +74,15 @@
 
 .optionalText {
   composes: hds-typography-body-200 hds-font-weight-regular from global;
-  width: 100%;
-  height: 100px;
-  display: flex;
   border-radius: 5px;
-  color: var(--token-color-foreground-primary);
-  padding: 12px;
-  box-shadow: var(--token-surface-base-box-shadow);
   border: unset;
+  box-shadow: var(--token-surface-base-box-shadow);
+  color: var(--token-color-foreground-primary);
+  display: flex;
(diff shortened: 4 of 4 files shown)


zchsh (approved):
✅ Code looks good to me, and tested on my iPhone, and seeing the `textarea` border 👍 💯

braydenlove:
text area looks good on iOS!

Curious if it is expected behavior to have the x (close icon) automatically focused after you click the Leave Beta button? Not a huge deal just curious if that is what we want! 
focused close icon on launch

zchsh:
@braydenlove Great question - i think when a dialog is opened, the first focus-able element is auto-focused. In this case, the close button is first in the document order, so focusing it is expected... but I am also curious whether it'd be preferable to have the buttons focused first.

Or perhaps for dialogs generally - is it preferable to have the close button always be the _last_ focus-able element? Rationale being, if someone opens the dialog, presumably they want to take some action within it, or at least see what actions are available, rather than immediately being drawn to (or even announcing) the close button.

braydenlove:
@zchsh ah cool thanks for the explanation that makes sense. I see your logic as well, ultimately I will lean on you and @ashleemboyer to figure out the best tab order! We can always go back and refine as well or wait for the DST dialog specs.

ashleemboyer:
> @zchsh ah cool thanks for the explanation that makes sense. I see your logic as well, ultimately I will lean on you and @ashleemboyer to figure out the best tab order! We can always go back and refine as well or wait for the DST dialog specs.

@braydenlove what @zchsh said! The first focusable item is focused automatically. That's standard behavior in accessible modals, and it's also what the Reach UI component does for us. There are more details on this in the `Note` under [keyboard interaction for the Dialog (Modal) APG pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialogmodal/#keyboard-interaction-7). It mentions "the most appropriate focus placement will depend on the nature and size of the content", so we are able to change this behavior on a case-by-case basis. But that is out-of-scope of this PR which just handles fixing the linked tickets. Happy to revisit this in a separate discussion & PR, preferably after our deadline. 😊

ashleemboyer:
> Awesome job with the margin refinement at small viewports 💅🏼.
> 
> Last question I have before approving. When I am looking at the dialog overlay I am still a bit confused (_although you did fix the odd contrast stuff we were seeing before 🤠_)!
> 
> I see that `--token-color-border-strong: #3b3d4566;` is using the 8 digit hex where `66` = 40% opacity. Where I get lost is the overlay on `[data-reach-dialog-overlay]`. It doesn't seem to visually affect the `--token-color-border-strong: #3b3d4566;` but I want to make sure I cover all the bases.
> 
> ```
> [data-reach-dialog-overlay] {
>     background: rgba(0,0,0,.33);
>     }
> ```
> 
> I am not totally sure where that value came `rgba` value came from. Curious if we can update that to the `rgba` version of `--token-color-border strong`? Which would be:
> 
> ```
> [data-reach-dialog-overlay] {
>     background: rgba(59, 61, 69, 0.4);
> }
> ```
> 
> It is my understanding that newer versions of chrome automatically updates colors with opacity to that 8 digit hex that includes opacity ([reference](https://gist.github.com/lopspower/03fb1cc0ac9f32ef38f4)) but I want to double check that browser support of those 8 digit hex codes is widely accepted? I am getting in the weeds, mostly curious about that random `rgba` overlay on data-reach-dialog-overlay.

That RGBA value is being set by the Reach component. Do you want it overwritten? We should be able to do so by setting `background: none` on the `.dialogOverlay` class.

braydenlove:
@ashleemboyer yeah I think if we could override it to `background: none;` or update it to the `rgba` color for `--token-color-border-strong` it might remove this two color flash I am seeing on Safari. I added a video showing said flash to the [asana ticket](https://app.asana.com/0/0/1202440311005416/1202460430635688/f).

ashleemboyer:
> @ashleemboyer yeah I think if we could override it to `background: none;` or update it to the `rgba` color for `--token-color-border-strong` it might remove this two color flash I am seeing on Safari. I added a video showing said flash to the [asana ticket](https://app.asana.com/0/0/1202440311005416/1202460430635688/f).

@braydenlove Oh, interesting! Let me look at this a little bit longer. I'm guessing it has something to do with the animation Reach handles.

BRKalow:
@ashleemboyer I think the main issue here is that we're actually rendering two overlays (`AnimatedDialogOverlay` and `DialogOverlay`). If we removed the nested `DialogOverlay`, we will only have one overlay element being rendered. This also fixes some transition jankiness.

We'll need to move the `isOpen` prop onto `AnimatedDialogOverlay`

braydenlove:
@ashleemboyer this is super strange, had to put on my detective cap here. I trust that you all will know the best way to handle background vs background color here. I do however think that the background value on [data-reach-dialog-overlay] has something to do with that Safari flash!

braydenlove:
Hmmm the safari background color flash actually might be related to how macOS handles double scrollbars and or -webkit transitions?

 _"The problem occurs on OSX if you don't have the iOS-style scrollbars activated. The iOS-style scrollbars overlay the page instead of indenting it, so the issue is not visible. But in OS/Browser environments that use 'static' scrollbars you get the issue."_ [link](https://github.com/twbs/bootstrap/issues/8768#issuecomment-22117599) to old thread.

More recent thread that essentially states that the screen flicker on safari is caused by -webkit animations happening at different times... [link](https://stackoverflow.com/questions/24465619/opacity-change-during-a-transition-flickers-in-safari).

Not the end of the world if we have a tiny flicker, just digging in!

src/components/dialog/index.tsx
```diff
@@ -9,33 +9,26 @@ export default function Dialog({
   onDismiss,
   label,
 }: DialogProps) {
-  const AnimatedDialogOverlay = m(DialogOverlay)
+  const shouldReduceMotion = useReducedMotion()
+  const AnimatedDialogOverlay = motion(DialogOverlay)
```
BRKalow, line 13: just noticed this definition is inside the component body, can we move it out? Otherwise it recreates a new component on every render.

ashleemboyer: Good call! Updating

ashleemboyer: https://github.com/hashicorp/dev-portal/pull/626/commits/27c2dd1fbeb64f5b7498577dca4f2f38e5d03e51 ✅

src/components/dialog/index.tsx
```diff
@@ -1,41 +1,35 @@
 import { DialogOverlay, DialogContent } from '@reach/dialog'
-import { AnimatePresence, m } from 'framer-motion'
+import { AnimatePresence, m as motion, useReducedMotion } from 'framer-motion'
```
BRKalow, line 2: Worth noting that `m` and `motion` are functionally different, this rename could cause confusion: https://www.framer.com/docs/guide-reduce-bundle-size/#how-to-reduce-bundle-size

BRKalow (approved):
👍

src/components/dialog/index.tsx
ashleemboyer: I'm finding just `m` to be confusing. Is there another name that's not `motion` that can help provide clarity?

ashleemboyer: @BRKalow what do you think about `slimMotion`?

BRKalow: Works for me 👍

ashleemboyer: Boom, done: https://github.com/hashicorp/dev-portal/pull/626/commits/b7e53c815d14c5d6dcea2be879ba42a9880e3332 💥

braydenlove (approved):
Awesome work, ship it!!!!!!!

PR #635, opened 2022-06-17
Add `tokensSource` option to `digital-plugin/no-undeclared-hds-color-tokens`; Add tests; Add README

ashleemboyer:
## 🗒️ What

- Adds `tokensSource` option to `digital-plugin/no-undeclared-hds-color-tokens` rule
- Adds tests for `digital-plugin/no-undeclared-hds-color-tokens`
- Adds a `README` in `stylelint-rules`

## 🤷 Why

- The `tokensSource` option is a step for making the `digital-plugin/no-undeclared-hds-color-tokens` rule more flexible to both read from more sources and eventually check more token types
- Tests were added for the other new rule in #517, so `digital-plugin/no-undeclared-hds-color-tokens` also needed some tests
- `README` is for making usage clear and also for sharing resources on the "why" behind these rules

## 🧪 Testing

- [ ] All PR checks should pass
- [ ] The [README](https://github.com/hashicorp/dev-portal/blob/amb.document-stylelint-rules/stylelint-rules/README.md) should be clear and understandable


stylelint-rules/no-undeclared-hds-color-tokens.js
```diff
@@ -1,33 +1,53 @@
...
+const isBoolean = (value) => {
+  const result = typeof value === 'boolean' || value instanceof Boolean
+  return result
+}
+
+// TODO move to src/lib?
+const isString = (value) => {
+  const result = typeof value === 'string' || value instanceof String
+  return result
+}
```
ashleemboyer, line 23: I assume we want these somewhere else (built-in Stylelint rules use these but they're not exposed), but wanted to raise it as a question first. 💭

BRKalow: Ideally the code for these stylelint rules would be self contained, so if we want to abstract them out they could live in `stylelint-rules/lib`. This will make the code more portable in the future if we want to publish these as npm packages, for example!

stylelint-rules/README.md
```diff
@@ -0,0 +1,63 @@
...
+HashiCorp Developer leverages some custom Stylelint rules for linting CSS. Each is documented below. See [`.stylelintrc.js`](/.stylelintrc.js) for the current Stylelint configuration.
+
+## Table of Contents
+
+- [`no-removed-outlines`](#no-removed-outlines)
+- [`no-undeclared-hds-color-tokens`](#no-undeclared-hds-color-tokens)
+
+## `no-removed-outlines`
+
+The `no-removed-outlines` rule checks for outlines that have been removed using the `outline`, `outline-style`, or `outline-width` properties. Completely removing outlines can lead to no focus indicators in Window High Contrast Mode (WHMC). Setting `outline: transparent` is one easy way to visually remove the default `outline` but still have it present in WHMC.
```
BRKalow, line 12: ⛏️ should this be WHCM (Windows High Contrast Mode)?

```suggestion
The `no-removed-outlines` rule checks for outlines that have been removed using the `outline`, `outline-style`, or `outline-width` properties. Completely removing outlines can lead to no focus indicators in Window High Contrast Mode (WHCM). Setting `outline: transparent` is one easy way to visually remove the default `outline` but still have it present in WHCM.
```

stylelint-rules/README.md
```diff
@@ -0,0 +1,63 @@
...
+  '@hashicorp/design-system-tokens/dist/devdot/css/tokens.css'
+const tokensFile = require.resolve(tokensFileName)
+const tokensFileContent = fs.readFileSync(tokensFile).toString()
+
+module.exports = {
+  plugins: ['./stylelint-rules/no-undeclared-hds-color-tokens.js'],
+  rules: {
+    'digital-plugin/no-undeclared-hds-color-tokens': [
+      true,
+      { tokensSource: tokensFileContent },
```
BRKalow, line 59: 💭 It would be cool if this rule could also support accepting a path and reading the file contents. It seems like lint rule configuration generally does not require consumers to instrument additional code to properly configure rules. This way the error message could also reference the source file path for additional clarity.

I do like that it can also accept a source blob too though!

stylelint-rules/__tests__/no-undeclared-hds-color-tokens.test.js
```diff
@@ -0,0 +1,49 @@
...
+const testCases = [
+  ['--token-color-this-is-made-up', 1, 0],
+  ['--token-color-this-is-also-made-up', 1, 0],
+  ['--token-color-this-is-yet-another-made-up', 1, 0],
+  ['--token-color-not-made-up', 0, 0],
+  ['--token-color-should-be-found', 0, 0],
+  ['--token-color-one-more', 0, 0],
+]
+
+describe(ruleName, () => {
```
BRKalow, line 26: 👏

BRKalow (commented):
Cool stuff!

stylelint-rules/README.md
ashleemboyer: I started to take this route, and even wanted to allow sending an array of file paths, but found that the files would be read several times. I think it's once for each CSS file being linted. The way I tested this was by placing a `console.log` in `function ruleFunction`, just before `return function lint`. Thinking back on it, we might be able to create a variable outside of the rule, and then define it on the first run to be the file contents of provided file paths? Want me to try that?

zchsh (approved):
🙌 The `README` looks good to me!

I can't seem to get this rule to work locally, but pretty sure it's an issue with my local VSCode setup that I need to figure out 😅 If other stylelint rules are working (eg `comment-empty-line-before`), is there any additional config required for this rule to work?

ashleemboyer:
> 🙌 The `README` looks good to me!
> 
> I can't seem to get this rule to work locally, but pretty sure it's an issue with my local VSCode setup that I need to figure out 😅 If other stylelint rules are working (eg `comment-empty-line-before`), is there any additional config required for this rule to work?

@zchsh there shouldn't be anything extra to do! I'll send you a message in Slack and see if we can debug it 😊

PR #641, opened 2022-06-20
chore: reject invalid requests early

thiskevinwang:
## 🔗 Relevant links

Related PR:
- https://github.com/hashicorp/react-components/pull/622

- [Asana task](url) 🎟️

### DIFF
--- a/src/layouts/sidebar-sidecar/server.ts
+++ b/src/layouts/sidebar-sidecar/server.ts
@@ -18,6 +18,8 @@ import {
   generateTopLevelSidebarNavData,
 } from 'components/sidebar/helpers'
 
+import { isInvalidURI } from './utils/is-invalid-uri'
+
 /**
  * @TODO update the basePaths inside of `src/data/${productSLug}.json` files to
  * be arrays of objects that look like:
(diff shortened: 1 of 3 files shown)


BRKalow:
@thiskevinwang Can you elaborate a bit on the approach here and explain how we consider a URL to be "invalid"?

thiskevinwang:
@BRKalow the approach here is on the naive side:

Given a path like...
`/docs/upgrade%25'%20AND%202*3*8=6*8%20AND%20'zVVl'!='zVVl%25/upgrade-specific`...n

...I'm decoding it once...
`/docs/upgrade%' AND 2*3*8=6*8 AND 'zVVl'!='zVVl%/upgrade-specific`

...and if it still contains `%` or whitespace, I'm considering that _invalid_, and returning an early `404`

src/layouts/sidebar-sidecar/server.ts
```diff
@@ -118,6 +120,13 @@ export function getStaticGenerationFunctions<
       const pathParts = (ctx.params.page || []) as string[]
       const headings = [] // populated by anchorLinks plugin below
 
+      // catch invalid URIs early
+      if (isInvalidURI(pathParts.join('/'))) {
+        return {
+          notFound: true,
+        }
+      }
```
BRKalow, line 128: 💭 Can we expand on the approach / heuristic here for how we determine an invalid URL / why we are doing this? I think that will be a helpful breadcrumb for the future!

thiskevinwang: Actually thinking about this more, i'm curious the reasoning behind catching **only** `404`'s and returning `notFound: true`...

https://github.com/hashicorp/dev-portal/blob/f8cf9c78dfd42045d09d0744b9606693c7de0c6e/src/layouts/sidebar-sidecar/server.ts#L163-L166

This class of bad requests cause our Content API to return a `Vercel 400`, so maybe it'd be simpler to include that in the detected status codes

PR #643, opened 2022-06-20
feat(product sites): add canonical tags to .io site docs pages

BRKalow:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1100423001970639/1202461448678534/f) 🎟️

## 🗒️ What

Adds canonical tags to our .io site docs pages as a hedge against any potentially duplicated content, as well as a general best-practice for SEO.

Screen Shot 2022-06-20 at 1 12 32 PM


src/components/_proxied-dot-io/common/docs-page/lib/get-canonical-url.ts
```diff
@@ -0,0 +1,47 @@
...
+export function getCanonicalUrlForDocsPage({
+  baseRoute,
+  currentPath,
+  productSlug,
+}: {
+  baseRoute: string
+  currentPath: string
+  productSlug: string
+}) {
+  const canonicalUrl = new URL(
```
zchsh, line 32: ⛏️ Nitpick: could we construct `const canonicalUrl` after determining `productDomain` (and possibly returning early)... in order to avoid the "stub in domain"?

zchsh (approved):
🙌 ✅ This looks good to me!

🙋 I feel relatively new & unfamiliar with SEO-related stuff - it seems like these canonical tags point to the pages they're already on. Is the intent here to signal that dot-io docs pages are the canonical versions, and that HashiCorp Developer pages with the same content are duplicates (at least during beta?), but in some different way than adding these same canonical tags on corresponding `developer.` pages? Or is there some other expected effect of these canonical tags?

src/components/_proxied-dot-io/common/docs-page/lib/get-canonical-url.ts
BRKalow: Yeah, good call!

BRKalow:
@zchsh Yeah I think this is generally a hedge against any unexpected SEO implications of the developer beta. I think because the developer pages have `noindex` on them we are still fairly insulated from any impact, which is also why we haven't added the canonical tags to those pages.

PR #644, opened 2022-06-20
fix: full URLs for all product meta open graph images

zchsh:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-zsabsolute-og-images-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1100423001970639/1202407276817529/f) 🎟️

## 🗒️ What

Fixes an issue where Open Graph image URLs were not absolute. Using absolute URLs allows services like Twitter to display the Open Graph images in share cards as expected.


src/data/consul.json
```diff
@@ -16,7 +16,7 @@
     "image": "/consul/img/og-image.png",
     "icon": [
       {
-        "href": "/consul/_favicon.ico"
+        "href": "https://www.consulproject.io/consul/_favicon.ico"
```
BRKalow, line 19: I think this change should be on the `image` attribute above 👀

src/data/sentinel.json
```diff
@@ -12,7 +12,7 @@
   "metadata": {
     "title": "Sentinel by HashiCorp",
     "description": "Sentinel is a language and framework for policy built to be embedded in existing software to enable fine-grained, logic-based policy decisions.",
-    "image": "/sentinel/img/og-image.png",
+    "image": "https://docs.hashicorp.com/sentinel/img/og-image.png",
```
zchsh, line 15: Note: I'm not sure the open graph image for Sentinel will work yet, as I can't seem to access anything in Sentinel's public folder (favicon.svg is the only thing I can test, though) 🤔

Do we need some extra config to expose folders? Eg I can access https://www.consul.io/consul/_favicon.ico, but not https://docs.hashicorp.com/sentinel/favicon.svg . Seems potentially related to `vault-public` / `waypoint-public` naming? cc @BRKalow

BRKalow: @zchsh I think there might be something funky going on with sentinel as we have a redirect in place to send all docs.hashicorp.com traffic to docs.hashicorp.com/sentinel 💭

zchsh: Ah that makes sense... do you think it makes sense to merge this PR as-is, and follow up on Sentinel?

BRKalow: Yeah, that sounds great to me 😃

BRKalow (approved):
🙇

PR #663, opened 2022-06-23
Fixing `no-undeclared-hds-color-tokens` stylelint rule

ashleemboyer:
## 🗒️ What

- Updates our `stylelint-rules/no-undeclared-hds-color-tokens` rule to see if the value of a property _contains_ a reference to a token, rather than _starts with_.

## 🤷 Why

- Some CSS property shorthands allow colors to be set as the second, third, etc. item in the property declaration, such as `border: 1px solid black`.

## 🛠️ How

- Chose to use regex for this one because `String.match` is really handy for plucking out parts of a string. And we need token names to be able to check against the `tokensSource` and then report a useful error.

## 🧪 Testing

Nothing to manually test here! I made the tests a little more thorough to cover this case. So as long as the PR checks pass I think we're ok!

## 💭 Anything else?

Shout out to @zchsh for finding this tricky bug! 🙌🏻


stylelint-rules/no-undeclared-hds-color-tokens.js
```diff
@@ -48,9 +48,11 @@ const plugin = stylelint.createPlugin(
       // https://postcss.org/api/#atrule-walkdecls
       postcssRoot.walkDecls((decl) => {
         // Only look at nodes with a value starting with a token reference
-        if (decl.value.startsWith('var(--token-color')) {
+        const matches = decl.value.match(/var\(--token-color-.*\)/g)
+        const usesHDSColorToken = matches?.length >= 1
+        if (usesHDSColorToken) {
           // Slice out the name of the token
-          const token = decl.value.slice('var('.length, -1)
+          const token = matches[0].slice('var('.length, -1)
```
BRKalow, line 55: ⛏️ we could use a [named capture group](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Groups_and_Ranges) here to make accessing the token name a bit easier.

BRKalow (approved):
Left a suggestion, but approving regardless. Feel free to take it or leave it!

zchsh (approved):
✅ This looks great to me, tested locally and working! 🙌 

💭 📌  Random future thought: if we have a value with multiple uses of color `var`, such as eg a `background` linear-gradient, would we need to account for that as well? [ie not just `match[0]` but possibly iterate over all matches]

ashleemboyer:
@zchsh @BRKalow thank you both for the input! 💖 I decided to go back and use `String.matchAll` and iterate over the results to:

- handle when multiple tokens can be used in a single property (such as gradients)
- make it easier to pull out the name of the token (rather than using `String.slice`)

zchsh (approved):
🔥 🥇 nicely done! tested a gradient thing locally 💯 

CleanShot 2022-06-23 at 16 44 43@2x

PR #665, opened 2022-06-24
fix: escape query selector to ensure it doesn't break

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Fixing this error: https://developer.hashicorp.com/vault/docs/platform/k8s/helm/examples/standalone-tls

This is caused by a query selector which has a leading digit, which is not valid as a CSS selector. Using the built-in [`CSS.escape()`](https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape) method to properly escape the selector.

### DIFF
--- a/src/layouts/sidebar-sidecar/components/table-of-contents/use-active-section.ts
+++ b/src/layouts/sidebar-sidecar/components/table-of-contents/use-active-section.ts
@@ -131,7 +131,7 @@ export function useActiveSection(
     headings.forEach((section) => {
       const el = document
         .getElementById('main')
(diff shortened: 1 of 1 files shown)


zchsh (approved):

BrandonRomano (approved):

BRKalow:
@zchsh The slugs are actually a-ok, I think we just need to ensure that any time we're building a query selector which has a leading digit we use this method 👍 The highlighting and such all still works with this in place.

PR #669, opened 2022-06-28
refactor(consul): use case implementations

alexcarpenter:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-acconsul-use-cases-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/0/1202452636091706) 🎟️

## 🗒️ What

Refactors use cases on the homepage and the nav to be fetched via dato cms multiple links fields vs just pulling in all use cases into the nav.

## 🤷 Why

Content authors need easier control of ordering and rendering of use case in the nav and on the homepage.

## 🛠️ How

- Creates new `Consul Nav` model to all choosing up to 5 use cases to be displayed in the nav
- Updates homepage model to allow choosing use cases via multiple links field vs the current blocks approach. Creates less content duplication and improved content maintanance.

## 📸 Design Screenshots

CleanShot 2022-07-13 at 09 26 45@2x

CleanShot 2022-07-13 at 09 27 08@2x

## 💭 Anything else?

---
  - https://app.asana.com/0/0/1202452636091706


BRKalow (approved):
LGTM!

src/layouts/_proxied-dot-io/consul/index.tsx
```diff
@@ -40,67 +37,67 @@ function ConsulIoLayout({ children, data }: Props): React.ReactElement {
 			/>
 
 			<Min100Layout footer={<Footer openConsentManager={openConsentManager} />}>
-				<ProductMetaProvider product={productData.slug as Products}>
```
BRKalow, line 43: Can we add back the `ProductMetaProvider`? It's necessary for certain components to render with the right product context.

alexcarpenter: Whoops, good catch https://github.com/hashicorp/dev-portal/pull/669/commits/8eaf40fe490275150324b7c5211bfac3389381a1

src/layouts/_proxied-dot-io/consul/index.tsx
```diff
@@ -51,21 +51,23 @@ function ConsulIoLayout({ children, data }: Props): React.ReactElement {
...
-									},
-									...useCaseNavItems.map((item) => {
-										return {
-											text: item.text,
-											url: `/use-cases/${item.url}`,
-										}
-									}),
-								].sort((a, b) => a.text.localeCompare(b.text)),
-							},
+							consulNav.useCases.length > 0
```
BRKalow, line 54: Is there any possibility `consulNav` might be undefined? this would error in that case. 💭

alexcarpenter: I don't think so as this would return an empty array if useCases were empty https://github.com/hashicorp/dev-portal/pull/669/files#diff-dc15fdaf1595690d3b8934305132caed4e9a54c929a5e2465af1041e7d41e32aR1-R7

alexcarpenter:
```suggestion
							consulNav && consulNav.useCases?.length
```

I can do some defensive coding here though.

BRKalow (approved):
LGTM

PR #670, opened 2022-06-28
Turning on Prettier useTabs option & formatting files

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambuse-tabs-instead-of-spaces-hashicorp.vercel.app/) 🔎
- Relevant Asana Task: [Switch to tabs for accessibility concerns](https://app.asana.com/0/1120751276730768/1201390817461565/f) 🎟️

## 🗒️ What

- Adds two npm scripts in `package.json`:
  - `prettier:check`
  - `prettier:write`
- Updates `prettier.config.js` to set:
  - `useTabs: true`
  - `tabWidth: 2`
- Adds `.prettierignore` file
- Applies new prettier configuration to all files not in `.prettierignore`
- Adds `.git-blame-ignore-revs` file

## 🤷 Why

- We discussed tabs vs. spaces way back in November in a weekly team meeting. The resolution was to test out using tabs instead of spaces, so that is the purpose of this PR.
- Some resources on how tabs are more accessible than spaces:
  - [Why we should default to Tabs instead of Spaces for an 'accessible first' environment](https://dev.to/alexandersandberg/why-we-should-default-to-tabs-instead-of-spaces-for-an-accessible-first-environment-101f)
  - [Tabs vs. Spaces: It's an Accessibility Issue](https://adamtuttle.codes/blog/2021/tabs-vs-spaces-its-an-accessibility-issue/)
  - [Nobody talks about the real reason to use Tabs over Spaces](https://www.reddit.com/r/javascript/comments/c8drjo/nobody_talks_about_the_real_reason_to_use_tabs/)
- Some resources on configuring custom tab widths:
  - [VS Code `editor.tabSize` setting](https://code.visualstudio.com/docs/editor/codebasics?ref=hackernoon.com#_indentation)
  - [Managing your tab size rendering preference](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-personal-account-settings/managing-your-tab-size-rendering-preference) in GitHub
- Adding the new `.prettierignore` file made it easier to run `prettier --write .` without unintended side effects.
- The new `prettier:check` and `prettier:write` npm scripts are purely for convenience of not having to remember what the prettier cli commands look like.
- The new `.git-blame-ignore-revs` file is so the `git blame` ignores the commit where 800+ files were changed, since formatting changes aren't nearly as valuable in a blame as functionality changes. This approach is taken from the [Ignore commits in the blame view](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) section of GitHub's [Viewing a file](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file) article.

## 🛠️ How

1. Updated the `prettier.config.js` file
2. Created the `.prettierignore` file
3. Added `prettier:write` npm script
4. Ran `prettier:write` script
5. Later added `.git-blame-ignore-revs` file after some research

## 🧪 Testing

- [ ] Check out this branch locally and try modifying some files like you usually do
- [ ] Note if there are any suspicious issues you run into
- [ ] Modify your `editor.tabSize` setting
- [ ] Make sure the changes visually apply as you expect in files

## 💭 Anything else?

Nope!


.git-blame-ignore-revs
```diff
@@ -0,0 +1,4 @@
+# https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view
```
zchsh, line 1: 🙌  This is neat, TIL!

PR #671, opened 2022-06-29
fix(middleware): update maxAge for opt-in cookies

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Next.js pre 12.2 assumes `maxAge` is in milliseconds here, even though the cookie spec expects seconds. Updating it so our opt-in cookies on developer are long-lived. @kendallstrautman and I had noticed that we weren't seeing the opt-out CTA on developer, even though we had opted in recently.


kendallstrautman (approved):

kendallstrautman:
Thanks for catching this 🙇‍♀️

PR #676, opened 2022-07-05
fix: use host in get-deployed-url

zchsh:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-zspatch-local-dev-img-url-issue-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1202555939973908/f) 🎟️

## 🗒️ What

Fixes an issue related to `@hashicorp/react-head`, where aggressive validation of image URLs passed to the component would cause errors in local development.

## 🛠️ How

Gets the `host` value in `_app.tsx` `getInitialProps`.

An alternate solution might be to hard-code the `PORT` and read that from `process.env`, as suggested [here](https://github.com/hashicorp/dev-portal/pull/664#discussion_r911359822). However, that would mean we'd sacrifice NextJS's dynamic port selection (which automatically chooses a port other than `3000` if that port is already in use).

### DIFF
--- a/src/components/head-metadata/index.tsx
+++ b/src/components/head-metadata/index.tsx
@@ -38,7 +38,9 @@ export default function HeadMetadata(props: HeadMetadataProps) {
 		productName ?? 'HashiCorp'
 	)
 
-	const ogImageUrl = `${getDeployedUrl()}/og-image/${productSlug ?? 'base'}.jpg`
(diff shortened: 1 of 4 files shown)


BrandonRomano (approved):
💯

PR #678, opened 2022-07-05
chore: remove internal links from README

zchsh:
## 🔗 Relevant links

- [Asana task](url) 🎟️


zchsh:
@mhonnatti Combed through the `README` for internal links, and removed them. Some notes and thoughts that came up:

- Would it be preferable to mention internal documents by unambiguous titles (such as RFCs in Google Docs), and only remove the links, rather than removing mentions entirely? (I wasn't exactly sure on the intent here, but seemed quick enough to dive in that it would be easier to sort out in PR comments)
- Should we remove the [setting up environment variables](https://github.com/hashicorp/dev-portal/blob/main/README.md#setting-up-environment-variables) section, since it links to an "internal" Vercel project? (This seems critical to project setup, and it seems like it might imply that the project can be developed locally by folks without access to that Vercel project, which I think is incorrect... so my hot take was that we should keep the section, both for our own convenience and for clarity in what is required for local development)

zchsh:
@ashleemboyer Tagging you here to see if you have any thoughts on what we might need to do to finalize this - I thought your input in Slack re: splitting sections of `docs/README.md` would be a great follow-up 🙌

PR #684, opened 2022-07-07
refactor: central docs URL adjusts, use MdxA

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

- Consolidated docs URL adjustment to server-side functions
- Adds an `MdxA` component for link rendering in MDX content
- Replaces `DocsAnchor` with `MdxA`

## 🤷 Why

- To ensure consistency in link styling in MDX
- To make it easier to reason about URL adjustments in docs content, by centralizing logic

## 🛠️ How

- Refactors server-side `productUrlAdjusters` to do work previously done client-side by `DocsAnchor`
- Uses link adjusting remark plugin with `productUrlAdjusters` to docs view server code (in `layout/sidebar-sidecar`)

## 🧪 Testing

- Visit docs pages for beta products, links should appear and function as expected, with correct URLs
    - [vault][/vault/docs] - ⭐ focus area for review, since already in Beta
    - [waypoint][/waypoint/docs] - ⭐ focus area for review, since already in Beta
- Visit docs pages for all other products, links should appear and function as expected, with correct URLs
    - [boundary][/boundary/docs]
    - [consul][/consul/docs]
    - [hcp][/hcp/docs]
    - [nomad][/nomad/docs]
    - [packer][/packer/docs]
    - [sentinel][/sentinel/docs]
    - [terraform][/terraform/docs]
    - [vagrant][/vagrant/docs]

## 💭 Anything else?

Not at the moment.

[task]: https://app.asana.com/0/1202097197789424/1202569127502494/f
[preview]: https://dev-portal-git-zsshared-mdx-anchor-hashicorp.vercel.app
[/boundary/docs]: https://dev-portal-git-zsshared-mdx-anchor-hashicorp.vercel.app/boundary/docs
[/consul/docs]: https://dev-portal-git-zsshared-mdx-anchor-hashicorp.vercel.app/consul/docs
[/hcp/docs]: https://dev-portal-git-zsshared-mdx-anchor-hashicorp.vercel.app/hcp/docs
[/nomad/docs]: https://dev-portal-git-zsshared-mdx-anchor-hashicorp.vercel.app/nomad/docs
[/packer/docs]: https://dev-portal-git-zsshared-mdx-anchor-hashicorp.vercel.app/packer/docs
[/sentinel/docs]: https://dev-portal-git-zsshared-mdx-anchor-hashicorp.vercel.app/sentinel/docs
[/terraform/docs]: https://dev-portal-git-zsshared-mdx-anchor-hashicorp.vercel.app/terraform/docs
[/vagrant/docs]: https://dev-portal-git-zsshared-mdx-anchor-hashicorp.vercel.app/vagrant/docs
[/vault/docs]: https://dev-portal-git-zsshared-mdx-anchor-hashicorp.vercel.app/vault/docs
[/waypoint/docs]: https://dev-portal-git-zsshared-mdx-anchor-hashicorp.vercel.app/waypoint/docs


src/components/dev-dot-content/mdx-components/mdx-a/index.tsx
```diff
@@ -0,0 +1,7 @@
+import InlineLink from 'components/inline-link'
+
+function MdxA(props) {
+	return <InlineLink {...props} textWeight="medium" />
```
zchsh, line 4: `MdxA` is a very light wrapper around `InlineLink`. The consolidated URL adjustments, and a minor tweak within `InlineLink`, are what make this possible.

src/components/docs-anchor/index.tsx
```diff
@@ -1,60 +0,0 @@
-import Link from 'next/link'
-import classNames from 'classnames'
-import { useCurrentProduct } from 'contexts'
-import InlineLink from 'components/inline-link'
-import s from 'components/inline-link/inline-link.module.css'
-
-const DocsAnchor = ({
```
zchsh, line 7: `DocsAnchor` has been removed. There's quite a bit going on here, so refer to comments on each block in `DocsAnchor` for details.

src/components/docs-anchor/index.tsx
```diff
@@ -1,60 +0,0 @@
...
-	const currentProduct = useCurrentProduct()
-	const passableProps = {
-		...rest,
-		// TODO: we probably want to make use of the InlineLink component in the future instead of just adopting its styles here. Tackle that when we have bandwidth to VQA lots of links.
-		className: classNames(rest.className, s.root),
-	}
-
-	if (!href) {
-		return <a {...passableProps}>{children}</a>
-	}
```
zchsh, line 21: This case mainly applies to "target" elements generated by our `anchor-links` remark plugin. I've modified the `InlineLink` component to fallback to using `#` as the `href` if `href` is falsy. This is a temporary measure, we should likely not generate `<a />` tags without `href`. Asana task to address this in the context of `anchor-links`: 🎟️ [Write RFC for using ids in anchor-links plugin rather than injecting HTML](https://app.asana.com/0/1100423001970639/1202488740132127/f).

src/components/docs-anchor/index.tsx
```diff
@@ -1,60 +0,0 @@
...
-	// Authors write content as if it only exists for their product,
-	// eg Waypoint content contains links that start with "/docs".
-	// We need to be adjust these links, eg to start with "/waypoint/docs".
-	let adjustedHref = href
-	// TODO: infer this condition dynamically
-	if (
-		currentProduct.basePaths.some((basePath) => href.startsWith(`/${basePath}`))
-	) {
-		adjustedHref = `/${currentProduct.slug}${href}`
-	}
```
zchsh, line 32: URL adjustment is now handled in our docs view server code (currently in `layouts/sidebar-sidecar`). So this adjustment is no longer needed client-side.

src/components/docs-anchor/index.tsx
```diff
@@ -1,60 +0,0 @@
...
-
-	// Render inline link if it's for sure a string and not a heading permalink
-	if (typeof children === 'string' && children !== '»') {
-		return (
-			<InlineLink href={adjustedHref} textWeight="medium">
-				{children}
-			</InlineLink>
-		)
-	}
```
zchsh, line 41: I think the intent here was to only render `InlineLink` for links with string children.

The new `MdxA` component renders `InlinkLink` in this way, so serves the same purpose for this case.

src/components/docs-anchor/index.tsx
```diff
@@ -1,60 +0,0 @@
...
-
-	// Render a next/link to avoid a full reload if we know for sure the link is an internal docs link.
-	// TODO: This heuristic can likely be tweaked to be more inclusive of other internal links
-	if (adjustedHref !== href) {
-		return (
-			<Link href={adjustedHref}>
-				<a {...passableProps}>{children}</a>
-			</Link>
-		)
-	}
```
zchsh, line 51: As far as I can tell, rendering this `<Link />` & `<a />` markup with `s.root` imported from `inline-link` and applied in the same way is equivalent to rendering `InlineLink` itself (minus the `hds-typography` and `hds-font-weight` classes, which it seems we would want to apply regardless).

So, I think always rendering `InlineLink` for this case makes sense - and we now do this with the `MdxA` component.

src/components/docs-anchor/index.tsx
```diff
@@ -1,60 +0,0 @@
...
-
-	return (
-		<a href={adjustedHref} {...passableProps}>
-			{children}
-		</a>
-	)
```
zchsh, line 57: This last case was originally intended to handle absolute URLs, I think... though as mentioned in a TODO comment above, it might also catch some internal relative URLs.

We've recently discussed how `next/link` can handle absolute URLs without any problems - so as with other cases in this file, rendering `InlineLink` here seems equivalent (only subtlety being that we'll now be adding `hds-typography` and `hds-font-weight` classes from `InlineLink`). The new `MdxA` renders `InlineLink`, so meets this case as well.

src/components/inline-link/index.tsx
```diff
@@ -22,24 +22,24 @@ const InlineLink = ({
...
 	 */
-	const hasHref = Boolean(href)
-	if (!hasHref) {
-		if (IS_DEV) {
-			console.error('[InlineLink] a falsy `href` was given. Returning `null`.')
-		}
-
-		return null
-	}
+	const hrefWithFallback = href || '#'
```
zchsh, line 39: Expanded the comment here. This change allows `Link` to render without error. I think the proper long-term solution here would be to avoid generating `<a />` elements that do not have an `href` - but that feels outside the scope of this PR (would involve `anchor-link` plugin refactoring, I think).

src/layouts/sidebar-sidecar/server.ts
```diff
@@ -132,6 +134,15 @@ export function getStaticGenerationFunctions<
...
 					[anchorLinks, { headings }],
 					rewriteTutorialLinksPlugin,
+					/**
+					 * Rewrite docs content links, which are authored without prefix.
+					 * For example, in Waypoint docs authors write "/docs/some-thing",
+					 * we need this to be "/waypoint/docs/some-thing" for Dev Dot.
+					 */
+					[
+						remarkPluginAdjustLinkUrls,
+						{ urlAdjustFn: productUrlAdjusters[product.slug] },
```
zchsh, line 144: This change applies docs `productUrlAdjusters` across all docs routes. Previously this remark plugin was added on a route-by-route basis (other changes in this PR remove those route-by-route additions).

We can control how URLs are adjusted on a per-product basis (and even per-base-path basis) within `./utils/product-url-adjusters` (We already had product-specific adjustments, eg for Vault's `/api` → `/api-docs` URLs).

src/layouts/sidebar-sidecar/utils/_local_platform-docs-mdx/index.tsx
```diff
@@ -58,7 +58,7 @@ function _defaultComponents() {
 		CodeBlockConfig,
 		CodeTabs,
 		pre,
-		a: DocsAnchor,
+		a: MdxA,
```
zchsh, line 61: Both docs and tutorials content now use `MdxA` as the MDX custom component for `<a />` tags.

src/layouts/sidebar-sidecar/utils/product-url-adjusters.ts
```diff
@@ -1,124 +1,174 @@
...
+ * every product during initial migration to HashiCorp Developer.
+ *
+ * If we want to remove the need for these URL adjustmenets, then
+ * at some point in the future, we could apply these as remark transforms
+ * using "remarkPluginAdjustLinkUrls" to the original source content
+ * (across ALL versions!), and ask authors to use the adjusted URL format
+ * going forward. Asana task:
+ * https://app.asana.com/0/1202097197789424/1202569127502507/f
+ */
+const productUrlAdjusters: Record<ProductSlug, (url: string) => string> = {
```
zchsh, line 29: The changes in this file are likely the most significant part of the refactor:
- For most products, there's not too much change - the `rewriteDocsUrl` function does the job that `DocsAnchor` `adjustedHref` did on the client, but at content processing time
- For Sentinel, the before-after comparison of what's going on is a little more complex; but hopefully a little easier to grok now that URL adjustment is happening all at once (rather than split between server and client)

src/layouts/sidebar-sidecar/utils/product-url-adjusters.ts
```diff
@@ -1,124 +1,174 @@
...
+	},
+	hcp: (url) => rewriteDocsUrl(url, hcpData as ProductData),
+	nomad: (url) => rewriteDocsUrl(url, nomadData as ProductData),
+	packer: (url) => rewriteDocsUrl(url, packerData as ProductData),
+	sentinel: rewriteSentinelDocsUrls,
+	terraform: (url) => rewriteDocsUrl(url, terraformData as ProductData),
+	vagrant: (url) => rewriteDocsUrl(url, vagrantData as ProductData),
+	vault: (url) => {
+		const withApiDocsFix = rewriteApiToApiDocs(url)
+		const withPrefix = rewriteDocsUrl(withApiDocsFix, vaultData as ProductData)
```
zchsh, line 44: Vault and Consul already had `rewriteApiToApiDocs` applied. This was applied before rendering on the client. So, the ordering of URL adjustments is retained:
1. fix `/api` → `/api-docs`
2. prefix with product slug (previously done client-side)

src/pages/consul/api-docs/[[...page]].tsx
```diff
@@ -2,7 +2,6 @@ import consulData from 'data/consul.json'
 import { ProductData } from 'types/products'
 import { getStaticGenerationFunctions } from 'layouts/sidebar-sidecar/server'
 import DocsView from 'views/docs-view'
-import { consulUrlAdjuster } from 'layouts/sidebar-sidecar/utils/product-url-adjusters'
```
Comment (zchsh) [inline src/pages/consul/api-docs/[[...page]].tsx:5]: All docs routes now have URL adjustments applied, so these per-route imports are no longer necessary.

src/views/tutorial-view/utils/mdx-components.tsx
```diff
@@ -54,7 +54,7 @@ const MDX_COMPONENTS = {
 	ol: MdxOrderedList,
 	ul: MdxUnorderedList,
 	li: MdxListItem,
-	a: (props) => <InlineLink {...props} textWeight="medium" />,
+	a: MdxA,
```
zchsh, line 57: `MdxA` is equivalent, intent here is to consolidate MDX content components.

zchsh:
@ashleemboyer This PR consolidated URL adjustment to a remark plugin, and gets us fully consistent on how `<a />` are rendered in MDX contexts (across docs and tutorials).

Bit of a larger PR, but I had trouble thinking of a good way to split it up, so tried to make up for that with thorough review comments. Let me know what you think 🙇‍♂️

src/components/docs-anchor/index.tsx
ashleemboyer: 🗒️ This case handled at least two things:

1. The heading permalinks injected by `anchor-links`, that have the `»` character as `children`
2. Anchor elements with non-string `children`, such as an anchor element that contains a `code` element

I think 2. is covered fine by `MdxA`. Maybe we want to add a child selector for styling `code` elements in `MdxA`? 💭 

As for 1., we shouldn't render `InlineLink` for injected permalinks that have the `»` character. We should render those as-is, without applying anything to them (other than the CSS that's currently hiding them). Since this would need to be considered for both Learn and Docs content, I think we have two options.

- Update `MdxA` to check `typeof children === 'string' && children === '»'` and render a plain anchor if true. Otherwise, render `InlineLink`.
- Create another component that checks `typeof children === 'string' && children === '»'` and renders a plain anchor if true. Otherwise, renders `MdxA`.

I think we should also consider rendering plain anchor elements for the additional anchors injected by `anchor-links` with empty `href`s. We definitely want to make sure those stay not-visible, so we probably shouldn't be applying any CSS to them like `InlineLink` does.

src/components/inline-link/index.tsx
```diff
@@ -22,24 +20,24 @@ const InlineLink = ({
...
+	 * For now, we still want to render these elements, as they're needed
+	 * for some anchor links to function.
 	 *
 	 * FUTURE TODOs
 	 *   - update the plugin to no longer add empty <a> elements
-	 *   - update this to throw an error
+	 *   - update this to throw an error if href is falsy
+	 *
+	 * Asana task:
+	 * https://app.asana.com/0/1100423001970639/1202488740132127/f
```
ashleemboyer, line 35: This is the link to the new task, can we reference it instead?

```suggestion
	 * https://app.asana.com/0/1100423001970639/1202560954440296/f
```

src/layouts/sidebar-sidecar/utils/product-url-adjusters.ts
```diff
@@ -1,124 +1,183 @@
...
+ * Authors write content as if it only exists for their product.
+ * For example, Waypoint content contains links that start with "/docs",
+ * but we need to be adjust these links to start with "/waypoint/docs".
+ *
+ * NOTE on /api to /api-docs:
  * We must fix /api URLs, which should really be /api-docs. This would ideally
  * be done in the source MDX. However, we need to support all past versions, so
  * even if we corrected the latest source, we'd likely still want to run this on
  * older content.
+ */
```
ashleemboyer, line 64: Can these two comments be moved to the block comments above their associated functions? That will put them as close as possible to where their information is useful.

src/layouts/sidebar-sidecar/utils/product-url-adjusters.ts
```diff
@@ -1,124 +1,183 @@
...
+		/**
+		 * All other products need their docs routes prefixed, as within docs
+		 * content, authors write links as if URLs are on the dot-io sites.
+		 * For example, in Waypoint content, authors write URLs like
+		 * "/docs/some-waypoint-page", which need to be adjusted to be
+		 * "/waypoint/docs/some-waypoint-page".
+		 */
+		default:
+			return (url: string) => rewriteDocsUrl(url, productData)
+	}
```
ashleemboyer, line 50: The comments here are breaking up the `switch` in a way that's harder to understand. Could this be updated to be a series of `if`s instead? That will make the comment+`if` statement sections form more visual blocks with clearer distinctions between each.

```typescript
/**
 * Sentinel comment
 */
if (productData.slug === 'sentinel') {
  return (url: string) => rewriteSentinelDocsUrls(url, productData)
}

/**
 * Consul/Vault comment
 */
if (productData.slug === 'consul' || productData.slug === 'vault') {
  return (url: string) => {
    const withApiDocsFix = rewriteApiToApiDocs(url)
    const withProductPrefix = rewriteDocsUrl(withApiDocsFix, productData)
    return withProductPrefix
  }
}

/**
 * Default case comment
 */
return return (url: string) => rewriteDocsUrl(url, productData)
```

src/lib/remark-plugin-adjust-link-urls.ts
```diff
@@ -8,7 +8,7 @@ const remarkPluginAdjustLinkUrls: Plugin = ({
 	urlAdjustFn: (url: string) => string
 }): Transformer => {
 	return function transformer(tree) {
-		visit(tree, 'link', (node: Link) => {
+		visit(tree, ['link', 'definition'], (node: Link) => {
```
ashleemboyer, line 11: 🙋🏻 Why is this update needed?

zchsh: Great question - it's needed to adjust URLs for all ways that links can be written in markdown. 

In detail: We currently only target `mdast` [`link` nodes only](https://github.com/syntax-tree/mdast#link), where the URL is written inline (like `[link text](https://www.example.com)`. We were not adjusting the URLs [`linkReference` nodes](https://github.com/syntax-tree/mdast#linkreference) (written like `[link text][reference-id]`), which reference [`definition` nodes](https://github.com/syntax-tree/mdast#definition). 

Related: we do something similar in the [`rewriteTutorialLinksPlugin`](https://github.com/hashicorp/dev-portal/blob/aa5aec86ace8004bf7b60d344308efd8af87b21e/src/lib/remark-plugins/rewrite-tutorial-links/index.ts#L75).

src/layouts/sidebar-sidecar/utils/product-url-adjusters.ts
zchsh: ✅ updated

src/layouts/sidebar-sidecar/utils/product-url-adjusters.ts
zchsh: ✅ great call, moved these to the associated functions

zchsh:
@ashleemboyer I've updated to move those comments, and switch the `switch` to `if` blocks.

src/components/docs-anchor/index.tsx
zchsh: @ashleemboyer as far as I can tell, since we were already passing `className` imported from `InlineLink` here, the only difference here in switching to `MdxA` is:

1. `Link` rather than `<a />` is used
2. Additional `hds-` classes are applied.

I believe the CSS to hide the `»` links is still functional, so I don’t think either of these differences are significant. If that sounds right, I’d love to avoid the additional conditional here, and defer the `»` styling concerns to a different PR (eg, when we implement and show those elements more intentionally - it seems likely this will be through the MdxHeading component and won’t collide with MdxA). Would you be on board with deferring this concern, as long as the `»` links are still hidden?

zchsh: @ashleemboyer Based on discussion in Slack, have gone with `1.` that you suggested. Let me know if that looks alright to you! 🙇

PR #687, opened 2022-07-08
Address Tabs flashing with useLayoutEffect

ashleemboyer:
## 🔗 Relevant links

- [Asana task](url) 🎟️

### DIFF
--- a/src/components/tabs/hooks/use-has-overflow.ts
+++ b/src/components/tabs/hooks/use-has-overflow.ts
@@ -19,7 +19,7 @@ const MUTATION_OPTS = { attributes: true, childList: true, subtree: true }
  *
  * https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver
  */
-export default function useOverflowRef<T extends HTMLElement>(
+export default function useHasOverflow<T extends HTMLElement>(
 	ref: MutableRefObject<T>
 ): boolean {
 	const smallestWidthBeforeOverflow = useRef<number>(null)
--- a/src/components/tabs/index.tsx
+++ b/src/components/tabs/index.tsx
@@ -1,4 +1,4 @@
-import { ReactElement, useState } from 'react'
+import { ReactElement, useRef, useState } from 'react'
 import classNames from 'classnames'
 import { Tab, TabButtonControls, TabDropdownControls } from './components'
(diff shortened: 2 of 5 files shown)


src/components/tabs/hooks/use-has-overflow.ts
```diff
@@ -0,0 +1,139 @@
...
+ * 	- Chrome 26+
+ * 	- Edge 12+
+ * 	- Firefox 14+
+ * 	- Internet Explorer 11+
+ * 	- Safari 17+
+ *
+ * https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver
+ */
+export default function useHasOverflow<T extends HTMLElement>(
+	ref: MutableRefObject<T>
```
kendallstrautman, line 23: Could you share a bit more on this change to accept a ref rather than handling it in the hook?

src/components/tabs/index.tsx
```diff
@@ -33,7 +33,8 @@ const Tabs = ({
 	/**
 	 * Track whether tabs are overflowing, so we can switch to a select.
 	 */
-	const [hasOverflow, overflowRef] = useOverflowRef<HTMLDivElement>()
+	const ref = useRef<HTMLDivElement>()
+	const hasOverflow = useHasOverflow<HTMLDivElement>(ref)
```
BRKalow, line 37: 💭 thought / suggestion: IMO, it feels more intuitive to have the hook return a ref for use, as opposed to passing one in, as now the consumers of the hook _need_ to import and declare a ref to use the hook. It would be functionally equivalent to declare the ref inside `useHasOverflow` and return it in a tuple similar to how the previous API looked.

ashleemboyer: @BRKalow Oh! I think I see what you're saying now. I thought the suggestion was to go back to exactly what the API was before. Previously it returned a callback function intended to be passed as an element's `ref` prop, not an actual ref. You're suggesting to return a ref, not a callback like before, is that right?

BRKalow: @ashleemboyer Ah word, yeah just returning a ref object, not a ref callback, is just fine!

BRKalow (approved):
Looks good! Thanks for addressing my feedback 🙏

PR #700, opened 2022-07-12
[DevDot404] Optimze PRODUCT_LINK_CARDS

ashleemboyer:
## 🗒️ What

- Optimizes generation of `PRODUCT_LINK_CARDS` in `DevDot404`

## 🤷 Why

- The `filter(getIsBetaProduct)` call isn't needed because we can use `__config.dev_dot.beta_product_slugs`
- The `filter((productSlug: ProductSlug) => productSlug !== 'sentinel')` call isn't needed if we replace both this call and the `map` call with a `forEach` on the array of beta product slugs

## 🧪 Testing

- [ ] Go to a page that will 404, such as [/not-a-real-page](https://dev-portal-git-amboptimize-404-product-list-hashicorp.vercel.app/not-a-real-page)
- [ ] The same cards should be listed as listed [in production](https://developer.hashicorp.com/not-a-real-page)


src/views/error-views/dev-dot-404/index.tsx
```diff
@@ -1,37 +1,39 @@
-import React, { ReactElement } from 'react'
-import { ProductSlug } from 'types/products'
+import React, { ReactElement, useMemo } from 'react'
```
zchsh, line 1: ⛏️  Nitpick: Not 100% sure, but doesn't look like `useMemo` is being used here!

```suggestion
import React, { ReactElement } from 'react'
```

zchsh (approved):
✅ This looks good!

📌 💭 For context, rationale for the previous approach was that, while being slightly less efficient, once we're out of "beta" and moving into GA we could then remove the `.filter(getIsBetaProduct)` part. If `beta_product_slugs` is going to be long-lived in config, then this is very much a non-issue.

ashleemboyer:
Discussed in Slack, changes not needed!

PR #702, opened 2022-07-12
Rendering generic nav header on /_error

ashleemboyer:
## 🗒️ What

- Updates the main navigation header to render the non-product header on `/_error` routes (such as 404)

## 🤷 Why

- When we updated `PRODUCT_SLUGS_TO_LOGOS` in #699, it made non-beta product logos start showing up on 404 routes nested under non-beta product slugs (example: https://developer.hashicorp.com/boundary/not-a-page)

## 🛠️ How

- Just updated the logic of when the non-product header is rendered, to include a check for `router.route === '/_error'`

## 🧪 Testing

- [ ] Go to a page under a non-beta product, such as /boundary/not-a-page
- [ ] You should see the same navigation header that you see on the home page

## 💭 Anything else?

This might just be a bandaid until we've released all products in beta / GA. I figured it's better overall to not show non-beta product slugs on 404 pages until we've confirmed a long term plan.

### DIFF
--- a/src/components/navigation-header/index.tsx
+++ b/src/components/navigation-header/index.tsx
@@ -1,3 +1,4 @@
+import { useRouter } from 'next/router'
 import { IconMenu24 } from '@hashicorp/flight-icons/svg-react/menu-24'
 import { IconX24 } from '@hashicorp/flight-icons/svg-react/x-24'
 import { useCurrentProduct, useMobileMenu } from 'contexts'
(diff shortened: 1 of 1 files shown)


BRKalow:
@ashleemboyer ❓ This also now no longer renders the Waypoint and Vault logos when there's a 404 for those products, was that intentional?

ashleemboyer:
> @ashleemboyer ❓ This also now no longer renders the Waypoint and Vault logos when there's a 404 for those products, was that intentional?

@BRKalow yes this was intentional. These changes are intended to be a hot-fix until we can consult with design for a longer term solution (or until GA when it'll be OK to show any product logo).

BRKalow:
@ashleemboyer Gotcha! Was this reported as an issue? It seems like a nice touch to have the product logo placement here, and it's not exactly a secret that we'll be adding additional products to the site 💭

BRKalow (approved):

PR #707, opened 2022-07-14
explore: collapse all tutorials routes to dynamic

kendallstrautman:
## 🔗 Relevant links

- [Asana task](url) 🎟️


src/views/tutorial-view/server.ts
```diff
@@ -137,44 +137,52 @@ const moizeOpts: Options = {
...
+	const allCollections = await cachedGetAllCollections()
+	let paths = []
+	// eventually this will be the whole product list? we need to build from
+	// a product-first standpoint here to ensure the routing is correct as each
+	// page needs to be build from its product-parent
+	__config.dev_dot.beta_product_slugs.map((product) => {
+		// Only build collections where this product is the main 'theme'
+		// @TODO once we implement the `theme` query option, remove the theme filtering
+		// https://app.asana.com/0/1201903760348480/1201932088801131/f
+		// @TODO we may want to adjust this to build based on the collection slug instead of the theme?
```
kendallstrautman, line 150: Need to think this through

src/pages/[product]/tutorials/[collectionSlug].tsx
```diff
@@ -0,0 +1,7 @@
+import CollectionView from 'views/collection-view'
+import { generateStaticFunctions } from 'views/collection-view/server'
+
+const { getStaticProps, getStaticPaths } = generateStaticFunctions()
+
+export { getStaticPaths, getStaticProps }
```
Comment (kendallstrautman) [inline src/pages/[product]/tutorials/[collectionSlug].tsx:6]: I think defining `getStaticProps` and paths would be useful here now. Otherwise this file doesn't really have meaningful data.

src/pages/[product]/tutorials/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+	ProductTutorialsViewProps,
+} from 'views/product-tutorials-view/server'
+import ProductTutorialsView from 'views/product-tutorials-view'
+
+export default ProductTutorialsView
+
+export async function getStaticProps({ params }): Promise<{
+	props: ProductTutorialsViewProps
+}> {
+	const productData = await import(`data/${params.product}.json`)
```
Comment (kendallstrautman) [inline src/pages/[product]/tutorials/index.tsx:13]: Should I use fs.readFile here? is that more performant 💭....

src/views/tutorial-view/server.ts
```diff
@@ -137,44 +137,52 @@ const moizeOpts: Options = {
...
+				const tutorialSlug = splitProductFromFilename(tutorial.slug)
 
-			return {
-				params: {
-					tutorialSlug: [collectionSlug, tutorialSlug] as [string, string],
-				},
-			}
+				return {
+					params: {
+						product,
```
kendallstrautman, line 163: Perhaps we define the product based on the collection slug product.....

src/views/tutorial-view/server.ts
kendallstrautman: This may bite us when we implement cloud, where the cloud theme isn't used.

kendallstrautman: Maybe I should use the directory structure instead. And build paths with the collection slug that have the same product.

src/pages/[product]/tutorials/[collectionSlug].tsx
```diff
@@ -0,0 +1,7 @@
+import CollectionView from 'views/collection-view'
+import { generateStaticFunctions } from 'views/collection-view/server'
+
+const { getStaticProps, getStaticPaths } = generateStaticFunctions()
```
Comment (kendallstrautman) [inline src/pages/[product]/tutorials/[collectionSlug].tsx:4]: I could consider moving the definition of these functions back into this file, might give it more meaning.

ashleemboyer:
Taking a look now! 😊

src/views/tutorial-view/server.ts
```diff
@@ -137,44 +143,47 @@ const moizeOpts: Options = {
...
 
-export async function getTutorialPagePaths(
-	product: ProductOption
-): Promise<TutorialPagePaths[]> {
-	const allCollections = await cachedGetAllCollections({
-		product: { slug: product },
-	})
-	// Only build collections where this product is the main 'theme'
+export async function getTutorialPagePaths(): Promise<TutorialPagePaths[]> {
+	const allCollections = await cachedGetAllCollections()
```
ashleemboyer, line 147: Nothing to do for this comment, but out of curiosity and for the sake of my own learning about the API, would calling `fetchAllCollectionsByProduct` once for each beta product be faster here? I can't remember if `getAllCollections` is one of the endpoints we noted as being slower. Maybe it also becomes less relevant as we onboard each product anyways. 😊

src/pages/[product]/tutorials/[collectionSlug].tsx
Comment (ashleemboyer) [reply src/pages/[product]/tutorials/[collectionSlug].tsx:4]: Good point! I think the `src/views/collection-view/server.ts` is a little long to totally copy/paste over here, but maybe the helpers that the generated `getStaticPaths` and `getStaticProps` functions call could be defined in a separate file? 💭

src/views/collection-view/server.ts
```diff
@@ -112,37 +110,40 @@ export async function getCollectionPageProps(
...
+	const paths = filteredCollections.map((collection) => {
+		// assuming slug structure of :product/:filename
+		const [product, filename] = collection.slug.split('/')
+		return {
+			params: {
+				product,
+				collectionSlug: filename,
+			},
+		}
+	})
```
ashleemboyer, line 130: 📌 Out of scope of this PR since it was already present, but do we need separate `.filter` and `.map` calls here, or could we build a new array and just iterate of `collections` once using `forEach`? 💭

src/views/tutorial-view/server.ts
```diff
@@ -137,44 +143,47 @@ const moizeOpts: Options = {
...
 
 			return {
 				params: {
+					product: productFromCollection,
 					tutorialSlug: [collectionSlug, tutorialSlug] as [string, string],
 				},
 			}
 		})
 	})
+	paths = [...paths, ...currentProductPaths]
```
ashleemboyer, line 171: 📌 Same future-thoughts here, I think we can simplify this with a `allCollections.forEach` that pushes to the `paths` array. 💭

ashleemboyer (commented):
This is going to be an awesome improvement! 🙌🏻

Left a few comments and curious questions, and one other thing I feel strongly about is that we should name the new folder `src/pages/[productSlug]` rather than `src/pages/[product]`, because the `product` variables throughout the codebase represent a `ProductData` object (or some subset of it). I found myself getting confused each time I saw `product` that was intended to be a page param containing a product's slug.

src/views/tutorial-view/utils/get-product-data.ts
```diff
@@ -0,0 +1,23 @@
...
+			)
+		)
+		return productData
+	} catch (e) {
+		console.error(
+			`[Error]: unable to fetch product data for ${product} — ${e.message}`
+		)
+		throw e
+	}
+}
```
BRKalow, line 21: So I don't think this will work during ISR, as the file system isn't kept around. An option could be to use a dynamic import, as that would include all of the json files as JS modules and they would be included in the deployed artifact.

src/views/tutorial-view/utils/get-product-data.ts
```diff
@@ -0,0 +1,23 @@
+import fs from 'fs'
```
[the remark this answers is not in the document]
kendallstrautman: Yes! Good call. That thought came to my mind as I was signing off.

src/views/tutorial-view/server.ts
kendallstrautman: I think it would be preferable to limit network requests, even if `getAllCollections` is a large call. This is where the caching becomes helpful.

src/views/tutorial-view/utils/get-product-data.ts
kendallstrautman: Ah okay I originally had it as a a dynamic import but I've read about that being less performant, perhaps because of what you mentioned with all the json files being included. I'll look into it, forgot about the fs / ISR issues 💀

src/pages/[product]/tutorials/[collectionSlug].tsx
Comment (kendallstrautman) [reply src/pages/[product]/tutorials/[collectionSlug].tsx:4]: Yeah I'm thinking of leaving the helper functions in `/collection-view/server.ts` and just bringing back `getStaticProps` & `getStaticPaths` here.

src/views/collection-view/server.ts
kendallstrautman: I made some updates with your suggestions!

src/lib/get-product-data.ts
```diff
@@ -0,0 +1,23 @@
+import fs from 'fs'
+import path from 'path'
+import moize from 'moize'
+import { LearnProductData, LearnProductSlug } from 'types/products'
+
+function getProductData(product: LearnProductSlug): LearnProductData {
```
kendallstrautman, line 6: This type should be `ProductSlug` and `ProductData` so it can be used anywhere.

BRKalow (approved):
👏

src/views/tutorial-view/utils/get-product-data.ts
kendallstrautman: Breadcrumb per slack. TL:DR vercel may support using FS in ISR now - https://nextjs.org/docs/advanced-features/output-file-tracing

src/views/collection-view/server.ts
```diff
@@ -112,54 +106,24 @@ export async function getCollectionPageProps(
...
-export async function getCollectionPaths(
-	product: ProductOption
-): Promise<CollectionPagePath[]> {
-	const collections = await cachedGetAllCollections({
-		product: { slug: product },
+export async function getCollectionPagePaths(): Promise<CollectionPagePath[]> {
+	const collections = await cachedGetAllCollections()
+	const paths = []
+	collections.forEach((collection) => {
+		// Only build collections where this product is the main 'theme'
```
kendallstrautman, line 113:
```suggestion
		// Only build collections the theme is also a product in 'beta' 
```

kendallstrautman: I should add more context here and a breadcrumb for the future.

BRKalow:
@kendallstrautman looks like we've gotta remove the nomad/tutorials route as well 👀

src/lib/get-product-data.ts
ashleemboyer: Agreed! Is this a note-to-self to resolve in this PR or a fast-follow?

src/views/collection-view/server.ts
ashleemboyer: Oh awesome! Appreciate knocking this out early here, and looks great. 🤩

ashleemboyer (approved):
Woo! This is sweet upgrade 🔥

src/views/tutorial-view/server.ts
ashleemboyer: Makes sense! Thank you for the clarity 😊

kendallstrautman:
Oh yeah! ty @BRKalow that musta creeped in with the merge from main. Will update

zchsh (approved):
🙌 This looks great to me!

PR #709, opened 2022-07-14
Amb.defer collection api call take 2

ashleemboyer:
## 🔗 Relevant links

- [Asana task](url) 🎟️

### DIFF
--- a/src/views/tutorial-view/index.tsx
+++ b/src/views/tutorial-view/index.tsx
@@ -32,10 +32,11 @@ import { getLearnRedirectPath } from 'components/opt-in-out/helpers/get-learn-re
 
 // Local imports
 import {
-	TutorialViewProps,
-	TutorialData,
 	CollectionContext,
+	LayoutContentWrapperProps,
+	TutorialData,
 	TutorialSidebarSidecarProps,
+	TutorialViewProps,
 } from './types'
 import MDX_COMPONENTS from './utils/mdx-components'
 import { formatTutorialToMenuItem, generateCanonicalUrl } from './utils'
@@ -48,27 +49,48 @@ import {
 } from './components'
 import s from './tutorial-view.module.css'
 
+/**
+ * The purpose of this wrapper component is to make it possible to invoke the
+ * `useMobileMenu` hook. It cannot be invoked in `TutorialView`. This is because
+ * it requires being invoked within a `MobileMenuProvider`.
+ *
+ * `MobileMenuProvider` is rendered by `CoreDevDotLayout`, which is indirectly
+ * rendered by `SidebarSidecarLayout`. This means the `useMobileMenu` hook can
+ * only be invoked here by a component rendered within `SidebarSidecarLayout`.
+ */
 const LayoutContentWrapper = ({
 	children,
 	collectionCtx,
 	product,
 	setCollectionViewSidebarSections,
-}) => {
+}: LayoutContentWrapperProps) => {
 	const { mobileMenuIsOpen } = useMobileMenu()
 	const haveLoadedData = useRef(false)
 
-	if (mobileMenuIsOpen && haveLoadedData.current === false) {
+	/**
+	 * Only need to load the data once, on the first open of the mobile menu
+	 */
+	if (haveLoadedData.current === false && mobileMenuIsOpen) {
 		getCollectionViewSidebarSections(product, collectionCtx.current).then(
-			(result) => {
+			(result: CollectionCategorySidebarSection[]) => {
 				haveLoadedData.current = true
 				setCollectionViewSidebarSections(result)
 			}
 		)
 	}
 
-	return children
+	/**
+	 * Wrapping in a fragment to prevent a "return type 'ReactNode' is not a valid
+	 * JSX element" error
+	 */
+	return <>{children}</>
 }
 
+/**
+ * Renders content for tutorial routes.
+ *
+ * /:productSlug/tutorials/:collectionSlug/:tutorialSlug
+ */
 function TutorialView({
 	layoutProps,
 	product,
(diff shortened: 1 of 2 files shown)


src/views/tutorial-view/index.tsx
```diff
@@ -33,64 +28,81 @@ import TutorialsSidebar, {
...
+	 * Only need to load the data once, on the first open of the mobile menu
+	 */
+	if (haveLoadedData.current === false && mobileMenuIsOpen) {
+		getCollectionViewSidebarSections(product, collectionCtx.current).then(
+			(result: CollectionCategorySidebarSection[]) => {
+				haveLoadedData.current = true
+				setCollectionViewSidebarSections(result)
+			}
+		)
+	}
```
BRKalow, line 80: 💭 This should be in a `useEffect` instead of the component body to avoid a side effect during render. We should also add a `.catch` here to ensure we handle a scenario where the fetch fails.

src/views/tutorial-view/index.tsx
```diff
@@ -33,64 +28,81 @@ import TutorialsSidebar, {
...
+ * only be invoked here by a component rendered within `SidebarSidecarLayout`.
+ */
+const LayoutContentWrapper = ({
+	children,
+	collectionCtx,
+	product,
+	setCollectionViewSidebarSections,
+}: LayoutContentWrapperProps) => {
+	const { mobileMenuIsOpen } = useMobileMenu()
+	const haveLoadedData = useRef(false)
```
BRKalow, line 68: 💭 Generally booleans are named using `has` or `is` e.g.: `hasLoadedData`, `isDataLoaded`

I like that you chose to use a ref to avoid any unnecessary state updates 👍

ashleemboyer: Oop! Yes. Didn't have my tenses right when I typed this out, thank you! 😁 Updated in e733359617bc58940421f268e4b7a2cbb307eec6!

src/views/tutorial-view/index.tsx
ashleemboyer: @BRKalow I added `useEffect` and a `catch` in https://github.com/hashicorp/dev-portal/pull/709/commits/642bcf2ca6699944af431f7db9299c17e066931f. What do we want to do if there is an error? If we do nothing, the sidebar where this data shows will not be useful because it'll be just about empty. So I think we _want_ the error to happen.

<details>
  <summary>Screenshot of empty sidebar</summary>
  
</details>

Since this is a UX discussion we probably want to have more widely, can we add a `TODO` in the `catch` block with an Asana task and come back so this PR isn't blocked?

BRKalow: The intent with catching the error here is that we control the error state and don't present a totally unexpected UX to the user. I think in the rare case that the call fails, the empty menu is reasonable and better than the page being broken in some other, unexpected way.

ashleemboyer: TODO added in https://github.com/hashicorp/dev-portal/pull/709/commits/68c2fccc8a46c2d42a6e711ea7c50d151fce36f9!

src/views/tutorial-view/index.tsx
```diff
@@ -33,64 +28,98 @@ import TutorialsSidebar, {
...
+				.catch((error) => {
+					/**
+					 * TODO: revist how we're handling this error
+					 * https://app.asana.com/0/1202097197789424/1202599138117878/f
+					 */
+					const errorMessage = `[TutorialView] error calling \`getCollectionViewSidebarSections\`: ${JSON.stringify(
+						error
+					)}`
+					console.error(errorMessage)
+					throw new Error(errorMessage)
```
BRKalow, line 89: 💭: I don't think we want to re-throw here, otherwise we get an `Uncaught (in promise) Error`:
```
Uncaught (in promise) Error: [TutorialView] error calling `getCollectionViewSidebarSections`:
```

The `console.error` should be adequate!

ashleemboyer: This is the discussion I'm trying to defer with the new ticket linked in the TODO. An error occurring provides information to the user, whereas an empty menu that they expect to be populated does not provide any information to them. I think we need to keep this for now until we have a wider discussion and firmer decision on what to do in the app when frontend errors occur.

BRKalow: The error doesn't do anything other than log to the console, so it's only helpful if a user has the console open, in which case they'll see the message from `console.error`.

Here's what we see in the console now when I block the request:
Screen Shot 2022-07-15 at 12 05 52 PM

And the menu:
Screen Shot 2022-07-15 at 12 07 00 PM

The empty menu is better than some unexpected/broken UI, and we still have the `console.error` which would provide insight if the console was open.

ashleemboyer: When I do the same, I have an error that pops up in the viewport on top of everything else. This is what tells the user that something went wrong.

CleanShot 2022-07-15 at 13 16 39

Since the same message is logged to the console automatically from throwing the error, I can remove the redundant `console.error` for our sake since we read those.

ashleemboyer: Won't our `ErrorBoundary` doesn't catch the error thrown?

ashleemboyer: Ah, I think the `ErrorBoundary` doesn't catch in this case because we're using a `catch`? According to: https://reactjs.org/docs/error-boundaries.html#how-about-event-handlers

BRKalow: Close! Error boundaries don't catch errors thrown in async code, so I don't think we would see the error boundary even without the catch.

Even so, I would argue the error boundary is a much worse experience for the user in this case as the entire page would be gone, whereas the page is still very much usable without these collection links being present in the menu.

ashleemboyer:
Hmm... [weird deploy error](https://vercel.com/hashicorp/dev-portal/3H3zem8CESTbQZqqKMH2K4xuY5AF) 🤔

Redeploying to see if that works!

PR #712, opened 2022-07-14
[spike] Move docs-specific server code to DocsView

ashleemboyer:
## 🗒️ What

- Moves `src/layouts/sidebar-sidecar/server.ts` to `src/views/docs-view/server.ts`
  - Updates the file import in 36 files
- Moves `src/layouts/sidebar-sidecar/utils/product-url-adjusters.ts` to `src/views/docs-view/utils/product-url-adjusters.ts`
- Organizes imports in newly-named `src/views/docs-view/server.ts`

## 🤷 Why

- The server code was only used for docs-specific views. Since `SidebarSidecarLayout` has become used for many views over time, the server code no longer made sense living under the layout.
- The `product-url-adjusters.ts` file was moved because it's only used by the server code.

## 🧪 Testing

We shouldn't need to do much manual testing. None of the code was functionally changed. But if you want to test a few pages, any page that renders `DocsView` will be the one to go to. 😊


BRKalow (approved):
LGTM 👍  Nice work

PR #717, opened 2022-07-18
feat(next-auth): integrate next-auth with HCP Cloud IDP provider

BRKalow:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1202520168081556/1202520168081603/f) 🎟️

## 🗒️ What

Implements initial auth integration following in @dstaley footsteps. Paired with @ashleemboyer to do the first pass integration of `next-auth`.

We still need to validate the production auth client, which has to be done on the developer.hashicorp.com domain (so in production).

Validated using a client in the hcp dev environment:
Screen Shot 2022-07-20 at 4 16 24 PM

Screen Shot 2022-07-20 at 4 17 08 PM

**Note: The auth flow is not expected to work in preview at this point.**

## 🛠️ How

- Following the docs here https://next-auth.js.org/


src/lib/auth/cloud-idp-provider.ts
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * A custom next-auth provider to authenticate via HashiCorp's Cloud IDP service
+ */
+const CloudIdpProvider: Provider = {
+	id: 'cloud-idp',
+	name: 'Cloud IDP',
+	type: 'oauth',
+	wellKnown:
+		__config.dev_dot.auth.idp_url + '/.well-known/openid-configuration',
+	authorization: { params: { scope: 'openid' } },
```
dstaley, line 12: Looks like the IDP team added support for the `profile` and `picture` scopes! I know Learn used an avatar, but I'm not sure if devportal needs one as well.

BRKalow: Yeah! I'm actually able to access the picture and profile attributes right now (see the JWT callback in the API route), so it all seems to be coming through as expected without additional scopes here 💭

dstaley: Ah gotcha! I knew `profile` was a default scope, but I guess `picture` is as well!

src/pages/auth.tsx
```diff
@@ -0,0 +1,23 @@
...
+			</>
+		)
+	}
+	return (
+		<>
+			Not signed in <br />
+			<button onClick={() => signIn('cloud-idp')}>Sign in</button>
+		</>
+	)
+}
```
BRKalow, line 23: This page (hopefully obviously) is just so that we can validate the full auth flow. It'll be removed after the production auth client we had created is validated on developer.hashicorp.com.

src/lib/auth/refresh-access-token.ts
```diff
@@ -0,0 +1,51 @@
...
+// TODO: validate that the refresh token flow works as implemented
+
+/**
+ * Takes a token, and returns a new token with updated
+ * `accessToken` and `accessTokenExpires`. If an error occurs,
+ * returns the old token and an error property
+ *
+ * Adapted from: https://next-auth.js.org/tutorials/refresh-token-rotation#server-side
+ */
+export default async function refreshAccessToken(token: JWT) {
```
BRKalow, line 12: This is not yet proved out, but it felt useful to leave it in as-is. I've got a follow-up ticket to finalize and validate the refresh token flow here: https://app.asana.com/0/1202097197789424/1202640177146339/f

config/production.json
```diff
@@ -9,6 +9,9 @@
 			"applicationId": "4486817b-44b8-4e6c-b6a4-cef83e48cf46",
 			"clientToken": "pub750ceadc5a9c79a6d3da59b534eb5108",
 			"service": "developer.hashicorp.com"
+		},
+		"auth": {
+			"idp_url": "https://auth.idp.hashicorp.com"
```
ashleemboyer, line 14: Question 🙋🏻 Do we need a `session_max_age` here like is being set in `config/base.json`? Or is there a default value being used that we're happy with in prod?

src/pages/_app.tsx
```diff
@@ -55,29 +61,31 @@ export default function App({ Component, pageProps, layoutProps, host }) {
...
-											<Component {...pageProps} />
-										</Layout>
-										<Toaster />
-										{showProductSwitcher ? <PreviewProductSwitcher /> : null}
-									</LazyMotion>
-								</CodeTabsProvider>
-							</CurrentProductProvider>
-						</AllProductDataProvider>
-					</DeviceSizeProvider>
+					<SessionProvider session={session}>
```
ashleemboyer, line 64: Just to confirm: is adding this as a wrapper around most of the existing items here the only change in this area? The changes above are obvious, but GitHub isn't diff-ing this party very nicely 😆

src/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,49 @@
...
+
+			// TODO: validate that the refresh token flow works as implemented
+			// return refreshAccessToken(token)
+
+			return { ...token, error: 'RefreshTokenError' }
+		},
+		/**
+		 * The session callback is called whenever a session is checked. By default, only a subset of the token is returned for increased security.
+		 *
+		 * ref: https://next-auth.js.org/configuration/callbacks#session-callback
```
Comment (ashleemboyer) [inline src/pages/api/auth/[...nextauth].ts:39]: Love these links to docs, thank you! 🙌🏻

src/lib/auth/refresh-access-token.ts
```diff
@@ -0,0 +1,51 @@
...
+				grant_type: 'refresh_token',
+				refresh_token: token.refreshToken as string,
+			})
+
+		const response = await fetch(url, {
+			headers: {
+				'Content-Type': 'application/x-www-form-urlencoded',
+			},
+			method: 'POST',
+		})
```
dstaley, line 29: I know you're following the example from the docs, but I can't recall ever seeing an OAuth `POST` request with parameters in the URL. If this flow doesn't work, that's probably the first thing I'd change.

ashleemboyer (approved):
Checked out this branch locally to test, and all I was missing was `NEXTAUTH_URL` in my `.env.local`! Minor comments & clarifications but this LGTM 👏🏻

src/pages/_app.tsx
BRKalow: Yup! You can hide whitespace changes on a diff which helps in this case: https://github.com/hashicorp/dev-portal/pull/717/files/e849d0e177b466e3bff59dedc320d9a86422a98c?w=1

config/production.json
BRKalow: For now we'll have this inherit from `base.json`, but I expect we'll adjust this for prod at some point!

src/pages/_app.tsx
ashleemboyer: OOF I forgot I turned that off recently when we converted from `SPACE` to `TAB` indentation 🤣

config/production.json
ashleemboyer: Ah, I see! I didn't realize we were merging nested properties of these config files already. 😊

PR #720, opened 2022-07-19
feat(test): test case for remove-non-beta-product-pages

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Adds a short test case for `remove-non-beta-product-slugs` to ensure that any time the list of
removed directories changes we need to address it via a snapshot update, which should give us a signal as to whether or not the change is expected.

## 🤷 Why

We accidentally deleted the new dynamic tutorials routes, which didn't manifest until the production build.

### DIFF
--- a/scripts/remove-non-beta-product-pages.ts
+++ b/scripts/remove-non-beta-product-pages.ts
@@ -38,9 +38,11 @@ async function main() {
 			dir.name !== '[productSlug]'
 		) {
 			console.log(`🧹 removing pages at /${dir.name}`)
-			await fs.promises.rm(path.join(pagesDir, dir.name), {
-				recursive: true,
-			})
+			if (!process.env.DRY_RUN) {
+				await fs.promises.rm(path.join(pagesDir, dir.name), {
+					recursive: true,
+				})
+			}
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


scripts/remove-non-beta-product-pages.ts
```diff
@@ -38,9 +38,11 @@ async function main() {
 			dir.name !== '[productSlug]'
 		) {
 			console.log(`🧹 removing pages at /${dir.name}`)
-			await fs.promises.rm(path.join(pagesDir, dir.name), {
-				recursive: true,
-			})
+			if (!process.env.DRY_RUN) {
```
BRKalow, line 41: Added a `DRY_RUN` env var here so that we can execute the script WITHOUT actually modifying the file system. It allows us to test against the script output.

src/__tests__/remove-non-beta-product-pages.test.ts
```diff
@@ -0,0 +1,35 @@
...
+		🧹 removing pages at /consul
+		🧹 removing pages at /docs
+		🧹 removing pages at /hcp
+		🧹 removing pages at /nomad
+		🧹 removing pages at /packer
+		🧹 removing pages at /sentinel
+		🧹 removing pages at /terraform
+		🧹 removing pages at /vagrant
+		"
+	`)
```
BRKalow, line 33: This snapshot test will fail anytime the output changes, which should be a good signal for us to manually review (we will have to anyways in order to get tests to pass).

ashleemboyer (approved):

thiskevinwang (approved):
Suggested adding 1 comment. Otherwise, nice!

kendallstrautman (approved):

PR #726, opened 2022-07-20
chore: update packer-plugins with fetch retry

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Updates `@hashicorp/platform-packer-plugins` to a newer version which has fetch retries built-in. This should reduce flaky build failures as a result of packer plugin docs failing to fetch.


zchsh (approved):
🙌 💖

PR #729, opened 2022-07-21
Stub in /profile page

ashleemboyer:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1201147515801429/1202621516737909/f) 🎟️

## 🗒️ What

- Updates `signInWrapper` signature based on some chatting with @BRKalow in Slack
- Removes the `Test` key I accidentally added to the `ValidAuthProviderId` `enum` in my last PR 😆 
- Adds `isRequired` option to `useAuthentication` (passed as `required` to `useSession`)
- Adds new `AuthenticatedView` utility component
  - Handles checking if the user is authenticated via the new `isRequired` option
  - Only renders `children` (the gated content) if the user is authenticated
- Adds new `ProfileView` component that uses `AuthenticatedView`
  - _Temporarily_ invokes the `useAuthentication` hook until the "Sign Out" button is present its future home
- Adds new `/profile` page that renders `ProfileView`
- Removes `/auth` page since the new `isRequired` option handles redirects for us

## 🤷 Why

- Since we've verified the successful integration of `next-auth` with Cloud IDP, we're able to get started building gated views.
- The `AuthenticatedView` utility should make it really easy to build additional pages gated by authentication.

## 🧪 Testing

Auth doesn't work in preview URLs yet, so make sure to check out this branch locally to test.

- After a redirect, there is going to be a `HashiHead` error locally. We haven't figured out why yet, but the error is only thrown in local development.

### `ProfilePage`

- [ ] Go to the `/profile` page
- [ ] If you **_are not authenticated_**, you should be redirected to the HCP sign in page that looks like this:
  <details><summary>Expand for screenshot</summary>CleanShot 2022-07-22 at 10 27 18@2x</details>
- [ ] If you **_are authenticated_**, you should see a page like the following:
  <details><summary>Expand for screenshot</summary>CleanShot 2022-07-21 at 14 28 45@2x</details>
- [ ] Clicking "Sign Out" should redirect you back to the sign in page again


src/pages/profile.tsx
```diff
@@ -0,0 +1,12 @@
+import useAuthentication from 'hooks/use-authentication'
+import BaseNewLayout from 'layouts/base-new'
+import ProfileView from 'views/profile-view'
+
+const ProfilePage = () => {
+	const { user } = useAuthentication()
+
+	return <ProfileView user={user} />
+}
```
BRKalow, line 9: Question: Curious why you're passing `user` down here as opposed to using the hook in the nested `ProfileView`?

src/views/authenticated-view/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+	useEffect(() => {
+		if (!isLoading && !isAuthenticated) {
+			router.push(redirectTo)
+		}
+	}, [isAuthenticated, isLoading, redirectTo, router])
+
+	// Show loading message while data is loading
+	if (!user) {
+		return <p>Loading...</p>
+	}
```
BRKalow, line 28: Question: Does this handle the SSR case? So initially we see `Loading...` then once everything resolves on the client in theory it would redirect or render its children?

BRKalow (commented):
Nice! This looks good.

We'll probably want to talk about if it makes sense to leverage the server-side integration to SSR gated pages [per the next-auth docs](https://next-auth.js.org/getting-started/client#require-session), but that's a separate conversation thread! 💭

src/pages/auth.tsx
```diff
@@ -1,20 +1,32 @@
...
-				<button onClick={() => signOut()}>Sign out</button>
-			</>
-		)
+const AuthPage = () => {
+	const router = useRouter()
+	const { isAuthenticated, isLoading, signIn } = useAuthentication()
+
+	// Redirect to the /profile page if the user is already authenticated
+	useEffect(() => {
+		if (!isLoading && isAuthenticated) {
```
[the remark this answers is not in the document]
ashleemboyer: Ah, good catch! Updating

src/pages/profile.tsx
ashleemboyer: I did it to make the view component a little more agnostic. No strong feelings either way; would you like me to update the view to use the hook instead?

src/views/authenticated-view/index.tsx
ashleemboyer: Ooohh I'm going to go check out those docs you linked and see what I find!

ashleemboyer:
@BRKalow made some updates that leverage the `required` property in `useSession`. This allows us to simplify `AuthenticatedView` and makes `/auth` redundant! What do you think?

src/pages/profile.tsx
BRKalow: No action needed, this is fine for now!

src/hooks/use-authentication.ts
```diff
@@ -6,18 +6,30 @@ import { SessionData, UserData, ValidAuthProviderId } from 'types/auth'
...
-const useAuthentication = () => {
-	// Pull data and status from next-auth's hook
-	const { data, status } = useSession()
+const useAuthentication = (options: UseAuthenticationOptions = {}) => {
+	// Get option properties from `options` parameter
+	const { isRequired = false } = options
+
+	// Pull data and status from next-auth's hook, and pass options
+	const { data, status } = useSession({
+		required: isRequired,
```
BRKalow, line 31: blocker: Nice find, and I like its intent! However, in our case we do not want to send folks to the interstitial login page with the list of providers:

Screen Shot 2022-07-22 at 9 10 51 AM
 
We want to route directly to the HCP login page, which is why we have the default `cloud-idp` argument for our `signIn` wrapper.

We could potentially implement our own `isRequired` functionality here based on `next-auth`: https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/react/index.tsx#L100-L109

BRKalow (changes requested):
One comment on how `required` works under the hood that we want to adjust. Everything else looks great though!

src/hooks/use-authentication.ts
ashleemboyer: @BRKalow Updated! I added an `onUnauthenticated` option to the `useAuthentication` hook, and set up a default for the callback. What do you think?

https://github.com/hashicorp/dev-portal/pull/729/commits/d072fc0827a395b9e9ea7e5158f393fe3715906a

BRKalow (approved):
💯

PR #730, opened 2022-07-21
[hotfix] Fixing IconCardLink urls on /docs

ashleemboyer:
## 🗒️ What

Fixes the `IconCardLink` components in the hero of the `/docs` pages

## 🧪 Testing

- [ ] Go to /waypoint/docs
- [ ] Make sure the `IconCardLink` components go to the right links

PR #732, opened 2022-07-22
Add E2E tests for /docs IconCardLink elements

ashleemboyer:
## 🗒️ What

- Adds a new E2E test file for testing `IconCardLink`s in `ProductRootDocsPathLanding`
- Updates the markup in `ProductRootDocsPathLanding` to wrap the `IconTileLogo`, page title, page subtitle, and `IconCardLink`s in a `<header>` element

## 🤷 Why

- These tests are to help us catch bugs like what was fixed in #730.
- The `<header>` element makes it a little easier to query the `IconCardLink` anchor elements during the tests. Using the element here is fine since it conveys that the content in it is the header of the main page content.

## 🛠️ How

I did _a lot_ of Googling and docs reading for this one 😁

- [Writing Tests](https://playwright.dev/docs/writing-tests) - helped me get started
- [Filtering locators](https://playwright.dev/docs/locators#filtering-locators) - helped me figure out how to query for the exact anchor element I wanted the test to invoke `.click()` on
- [Assertions](https://playwright.dev/docs/test-assertions) - helped me find really useful matchers like `toHaveURL`
- [Page](https://playwright.dev/docs/api/class-page) - helped me find the [`waitForURL`](https://playwright.dev/docs/api/class-page#page-wait-for-url) method
- [Parametrize tests](https://playwright.dev/docs/test-parameterize) - inspired me to generate one `test` for each case; so did [Parallelize tests in a single file](https://playwright.dev/docs/test-parallel#parallelize-tests-in-a-single-file)
- [Set timeout for a single test](Set timeout for a single test) - was helpful in allotting more time for tests to run

## 🧪 Testing

Nothing to manually test here. 😊

## 💭 Anything else?

I would expect these tests to fail as each beta product is added to the `beta_product_slugs` array. Do we want to go ahead and import all of the product JSON files now to prevent that?


src/__tests__/e2e/docs-landing-icon-card-links.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+		// There is no IconCardLink for this path, ignore it
+		if (path === 'docs') {
+			return
+		}
+
+		// Create a test for this product + path
+		const testName = `[/${betaProductSlug}/docs] ${name} \`IconCardLink\` goes to /${betaProductSlug}/${path}`
+		test(testName, async ({ baseURL, page }) => {
+			// Triple the available time for the test to ru
+			test.slow()
```
BRKalow, line 35: question: was this to solve a particular issue you were seeing?

BRKalow (commented):
This is great! In general, e2e tests are more expensive to run (read: slower), so it's worthwhile to try and minimize how many assertions we're making.

With that in mind, what do you think about only running this for a single product? We know that the component used for each product is the same, so if it works for one product it should work for all.

src/__tests__/e2e/docs-landing-icon-card-links.spec.ts
ashleemboyer: Yes! The timeout is 30s by default and some tests were sometimes taking longer than that.

ashleemboyer:
@BRKalow I've updated the test to just look at the Waypoint product data, and also simplified some of the test methods used! What do you think?

BRKalow (approved):
👍  nice work!

PR #733, opened 2022-07-22
spike: custom docs content layouts

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️


src/views/docs-view/components/docs-root-landing/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+		}
+
+		if (!heading && child.props.originalType === 'h1') {
+			heading = {
+				title: getElementTextContent(child),
+				id: child.props.id,
+			}
+			return null
+		}
+		if (!subtitle && child.props.originalType === 'p') {
```
zchsh, line 42: It seems like there's potential for some strange edge cases here.

For example, if there's no `h1` in a document, which I think isn't guarded in other ways, we'd end up with an empty title element.

Or, if an `h1` is followed by a non-`p` element (and a `p` element appears later in the document), content would be re-ordered.

I wonder whether this might be a decent case for a `remark` plugin - it seems like we might be able to precisely target `h1 + p` at the start of a document on a given page with slightly more precision. We could replace it with a `dynamic` loaded custom MDX component.

PR #739, opened 2022-07-26
Add enable_auth flag

ashleemboyer:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1202097197789424/1202664135851182/f) 🎟️

## 🗒️ What

- Adds `enable_auth` flag to `base` and `production` configs
- Adds `isEnabled` derived boolean to our `useAuthentication` return value
- Updates `AuthenticationView` to return `ErrorView` if `isEnabled` is `false`

## 🤷 Why

- The `enable_auth` flag is necessary for us to build authenticated user features in isolation without showing these features in production.
- The `isEnabled` boolean from `useAuthentication` will be really nice for conditionally rendering elements such as the Sign In and Sign Out buttons that will live in the navigation header / mobile sidebar.
- Updating `AuthenticationView` to return `ErrorView` if `isEnabled` is `false` is an easy test for checking if the flag is working as expected. And if any views are built on top of `AuthenticatedView`, they'll automatically behave the same without any extra work on the developer.

## 🧪 Testing

We can't test the production environment until this branch is merged into `main` and a new deploy is created, but the local testing steps should help provide confidence that these changes will work in production.

- [ ] Check out this branch locally
- [ ] Run the server with `npm start`
- [ ] Go to the /profile page
- [ ] The page should behave as it was previously, because the `enable_auth` flag is set to `true`
- [ ] Set the `enable_auth` flag in `src/config/development.json` to `false`
- [ ] Go to the `/profile` page
- [ ] You should see the DevDot 404 view

### DIFF
--- a/config/base.json
+++ b/config/base.json
@@ -30,8 +30,9 @@
 		"max_static_paths": 10
 	},
 	"flags": {
-		"enable_product_docs_search": true,
+		"enable_auth": true,
+		"enable_datadog": true,
 		"enable_io_beta_cta": true,
-		"enable_datadog": true
+		"enable_product_docs_search": true
 	}
 }
--- a/config/base.json
+++ b/config/base.json
@@ -30,7 +30,7 @@
 		"max_static_paths": 10
 	},
 	"flags": {
-		"enable_auth": true,
+		"enable_auth": false,
 		"enable_datadog": true,
 		"enable_io_beta_cta": true,
 		"enable_product_docs_search": true
--- a/config/development.json
+++ b/config/development.json
@@ -8,5 +8,8 @@
 	},
 	"learn": {
 		"max_static_paths": 0
+	},
+	"flags": {
+		"enable_auth": true
 	}
 }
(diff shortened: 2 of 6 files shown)


config/base.json
```diff
@@ -30,8 +30,9 @@
 		"max_static_paths": 10
 	},
 	"flags": {
-		"enable_product_docs_search": true,
+		"enable_auth": true,
```
dstaley, line 33: ⛏️ I think the base state should be `false`, at least during this development phase. That way we're enabling auth explicitly in dev/preview as opposed to disabling in prod. Super small nit, so feel free to ignore if you don't vibe with this :)

dstaley (approved):
LGTM!

config/base.json
ashleemboyer: This makes sense to me! Updated in https://github.com/hashicorp/dev-portal/pull/739/commits/bda9e7e1dbfa7f03c3cdcceb7d004b065fbb29e7. 😊

PR #742, opened 2022-07-27
Add mobile Sign In and Sign Up elements

ashleemboyer:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1202097197789424/1202665629707459/f) 🎟️

## 🗒️ What

- Adds `MobileAuthenticationControls` component in `MobileMenuContainer`
- Uses `MobileAuthenticationControls` alongside uses of `MobileMenuContainer`

## 🤷 Why

- I tried not exporting `MobileAuthenticationControls` and directly placing it within `MobileMenuContainer`, but that didn't work because the instances of `MobileMenuContainer` are all styled differently. This approach requires using `MobileAuthenticationControls` everywhere `MobileMenuContainer` is used, but allows us to keep the CSS as simple as possible.

## 🛠️ How

- The `useAuthentication` hook was useful for determining auth state.

## 📸 Design Screenshots

<details>
  <summary>Expand for screenshot</summary>
  <a href="https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=11619%3A114842">Figma link</a>
  
</details>

## 🧪 Testing

- [ ] Check out this code locally (auth doesn't work in preview URLs yet)
- [ ] Make sure you're viewing the app in a desktop viewport
- [ ] Set `auth_enabled` to `false` in `config/development.json`
- [ ] Go to the home page
- [ ] Open the mobile menu
- [ ] You **should not** see the Sign In and Sign Up elements at the top of the mobile menu
- [ ] Go to a page under a product (such as /waypoint)
- [ ] Open the mobile menu
- [ ] You **should not** see the Sign In and Sign Up elements at the top of the mobile menu
- [ ] Set `auth_enabled` to `true` in `config/development.json`
- [ ] Go to the home page
- [ ] Open the mobile menu
- [ ] You **should** see the Sign In and Sign Up elements at the top of the mobile menu
- [ ] Go to a page under a product (such as /waypoint)
- [ ] Open the mobile menu
- [ ] You **should** see the Sign In and Sign Up elements at the top of the mobile menu
- [ ] Click the Sign In button
- [ ] Authenticate as usual (you might not be taken through the flow)
- [ ] Verify you're authenticated by going to the /profile page
- [ ] Go to the home page
- [ ] Open the mobile menu
- [ ] You **should not** see the Sign In and Sign Up elements at the top of the mobile menu
- [ ] Go to a page under a product (such as /waypoint)
- [ ] Open the mobile menu
- [ ] You **should not** see the Sign In and Sign Up elements at the top of the mobile menu

### DIFF
--- a/src/components/mobile-menu-container/index.tsx
+++ b/src/components/mobile-menu-container/index.tsx
@@ -41,20 +41,20 @@ const MobileAuthenticationControls = () => {
 
 	return (
 		<div className={s.mobileAuthenticationControls}>
-			<Button
-				icon={<IconArrowRight16 />}
-				iconPosition="trailing"
-				onClick={() => signIn()}
-				size="small"
-				text="Sign In"
-			/>
 			<ButtonLink
-				color="secondary"
 				href="/sign-up"
 				icon={<IconUserPlus16 />}
 				iconPosition="trailing"
 				size="small"
 				text="Sign Up"
+			/>
+			<Button
+				color="secondary"
+				icon={<IconArrowRight16 />}
+				iconPosition="trailing"
+				onClick={() => signIn()}
+				size="small"
+				text="Sign In"
 			/>
 		</div>
 	)
--- a/src/components/mobile-menu-container/index.tsx
+++ b/src/components/mobile-menu-container/index.tsx
@@ -1,7 +1,12 @@
 import { ForwardedRef, forwardRef } from 'react'
 import classNames from 'classnames'
 import { m, useReducedMotion } from 'framer-motion'
+import { IconUserPlus16 } from '@hashicorp/flight-icons/svg-react/user-plus-16'
+import { IconArrowRight16 } from '@hashicorp/flight-icons/svg-react/arrow-right-16'
 import { useMobileMenu } from 'contexts'
+import useAuthentication from 'hooks/use-authentication'
+import Button from 'components/button'
+import ButtonLink from 'components/button-link'
 import { MobileMenuContainerProps } from './types'
 import s from './mobile-menu-container.module.css'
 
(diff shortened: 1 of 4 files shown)


src/components/mobile-menu-container/index.tsx
```diff
@@ -18,6 +23,43 @@ const MOBILE_MENU_MOTION = {
...
+ * Handles rendering the Sign In and Sign Up UI elements in mobile viewports.
+ * Intended to be used alongside `MobileMenuContainer`.
+ *
+ * @TODO Render user dropdown disclosure for authenticated users
+ * ref: https://app.asana.com/0/1202097197789424/1202665629707469/f
+ */
+const MobileAuthenticationControls = () => {
+	const { isAuthEnabled, isAuthenticated, isLoading, signIn } =
+		useAuthentication()
+	const shouldShowAuthButtons = isAuthEnabled && !isLoading && !isAuthenticated
```
kendallstrautman, line 36: 💭 I wonder if it would make sense to add `showAuthUI` to the return value from the `useAuthentication` hook. Not necessarily for this PR but I imagine this is something we'll be defining in a lot of components.

src/components/mobile-menu-container/mobile-menu-container.module.css
```diff
@@ -6,17 +6,29 @@
...
+	height: calc(100vh - var(--sticky-bars-height));
+	left: -150vw;
+	overflow-y: auto;
+	position: fixed;
+	top: var(--sticky-bars-height);
+	width: 100%;
+	z-index: 1;
+}
+
+.mobileAuthenticationControls {
```
kendallstrautman, line 24: It seems like these buttons are rendering at desktop size also? 
Screen Shot 2022-07-27 at 9 43 49 AM

kendallstrautman (approved):
Looks good! Just had one question about the UI rendering at desktop to checkout.

src/components/mobile-menu-container/mobile-menu-container.module.css
ashleemboyer: Gah! Great catch. I forgot to add the `g-hide-on-desktop` helper class. 😆

ashleemboyer: 🗒️ Fixed in https://github.com/hashicorp/dev-portal/pull/742/commits/5fb1b30cae60c3916881d837b3a57681d70722e5

src/components/mobile-menu-container/index.tsx
ashleemboyer: Ooooh I really like this idea! 🙌🏻

I think we'd need two variables though:

- `showUnauthenticatedUI`
- `showAuthenticatedUI`

Both of these could be derived by `useAuthentication`, and that would mean greatly reducing references to `isAuthEnabled` which is inevitably going away when we release auth features! ✨ I'll explore adding these in one of my PRs where I need to render something different for authenticated users (like the user dropdown disclosure).

PR #743, opened 2022-07-27
Add /sign-up page

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambadd-sign-up-page-hashicorp.vercel.app/sign-up) 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1202661920988683/f) 🎟️

## 🗒️ What

- Adds new `/sign-up` page and `SignUpView`
- Adds new `signUp` function that invokes `signIn` with the default options for signing up
- Updates `signInWrapper` and `signOutWrapper` to accept more options, have defaults set, and then apply the rest of the given options
- Adds a `DEFAULT_SIGN_IN_CALLBACK_URL` set as `/profile` (we can overwrite this on a case-by-case basis)

## 📸 Design Screenshots

[Figma link](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=11365%3A114999)

## 🧪 Testing

NOTES:

1. The SVG image is not the final image we're using, so it's placement is not pixel perfect. There is [a follow-up ticket](https://app.asana.com/0/1202097197789424/1202683836858983/f) to handle it before shipping User Profile. We'll polish its styling there.
2. The actual sign in and sign up functionality can't be tested on preview URLs. To test those, check this branch out locally.

- [ ] Go to [/sign-up](https://dev-portal-git-ambadd-sign-up-page-hashicorp.vercel.app/sign-up)
- [ ] Ensure the page renders correct at desktop views
- [ ] Ensure the page renders correct at tablet views
- [ ] Ensure the page renders correct at mobile views


src/components/button-link/button-link.module.css
```diff
@@ -4,56 +4,57 @@
...
+	composes: hds-font-weight-medium from global;
+
+	/* attributes */
+	align-items: center;
+	border-radius: 5px;
+	border-style: solid;
+	border-width: var(--border-width);
+	display: flex;
+	font-size: var(--font-size);
+	font-family: var(--font-body);
```
ashleemboyer, line 20: 🗒️ Setting `font-family` was the main change I made here. I don't think Prettier formatted CSS files when we converted from SPACE indentation to TABS 😭

src/components/button/button.module.css
```diff
@@ -1,76 +1,77 @@
...
+	/* composition */
+	composes: hds-font-weight-medium from global;
+
+	/* attributes */
+	align-items: center;
+	border-radius: 5px;
+	border-style: solid;
+	border-width: var(--border-width);
+	display: flex;
+	font-family: var(--font-body);
```
ashleemboyer, line 15: 🗒️ Same here -- all I did was add the `font-family` property.

src/views/sign-up-view/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+import s from './sign-up.module.css'
+
+// Text content
+const TITLE = 'Create Your Account'
+const DETAILS = [
+	'Use your account across HashiCorp products',
+	'Track your progress in longer tutorials',
+	'Bookmark tutorials you need to reference regularly',
+]
+const SIGN_UP_BUTTON_TEXT = 'Sign Up'
```
cbfx, line 28: howdy @ashleemboyer ! great work on this sign up page. there was a big push a while back to standardize text casing to sentence case in HCP. just wanted to mention it here. "Sign Up" => "Sign up".

do y'all have a Figma board where i can see more of the designs for the sign up pages?

ashleemboyer: Hi @cbfx! Thanks for calling this out. Here's a link to the Figma for this page: https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=11365%3A114999.

cc-ing @braydenlove and @heitzke for visibility

src/hooks/use-authentication.ts
```diff
@@ -1,17 +1,33 @@
...
  * A minimal wrapper around next-auth/react's `signIn` function. Purpose is to
  * handle invoking the wrapped function with a default value.
  *
  * https://next-auth.js.org/getting-started/client#signin
  */
-const signInWrapper = (provider: ValidAuthProviderId = DEFAULT_PROVIDER_ID) => {
-	return signIn(provider)
+const signInWrapper = (
+	provider: ValidAuthProviderId = DEFAULT_PROVIDER_ID,
+	options: SignInOptions = {}
```
kendallstrautman, line 22:
```suggestion
	provider?: ValidAuthProviderId = DEFAULT_PROVIDER_ID,
	options?: SignInOptions = {}
```
Should these be optional if you're providing a default?

kendallstrautman: Actually im unsure if that's valid ts what I suggested lol but anyways, still a question 😆

src/hooks/use-authentication.ts
```diff
@@ -21,8 +37,31 @@ const signInWrapper = (provider: ValidAuthProviderId = DEFAULT_PROVIDER_ID) => {
  * https://next-auth.js.org/getting-started/client#signout
  */
 const signOutWrapper = (options: SignOutParams = {}) => {
-	const { callbackUrl = '/', redirect = true } = options
-	return signOut({ callbackUrl, redirect })
+	const { callbackUrl = '/', redirect = true, ...restOptions } = options
```
kendallstrautman, line 40: I think we'll want to be able to pass a callback URL here also, similar to the sign-in. Not necessary to add here though, but a note might be nice.

kendallstrautman: Would those `restOptions` override this `callbackUrl` if it was passed?

src/pages/sign-up.tsx
```diff
@@ -0,0 +1,3 @@
+import SignUpView from 'views/sign-up-view'
+
+export default SignUpView
```
kendallstrautman, line 3: 💭 I wonder if it still makes sense to have a 'view' in this case since it's such a standalone page. Not necessary to address but its just a thought.

src/views/sign-up-view/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+	 * @TODO determine loading state UI
+	 */
+	if (isLoading) {
+		return null
+	}
+
+	/**
+	 * Show 404 error if enable_auth is false or the user is already authenticated
+	 */
+	if (!isAuthEnabled || isAuthenticated) {
```
kendallstrautman, line 46: 💭 I wonder if we could just redirect them back to the homepage instead of showing the 404

src/views/sign-up-view/index.tsx
braydenlove: Thanks, will update!

kendallstrautman (approved):
👏

braydenlove:
@ashleemboyer question for you. I am assuming the image isn't scaling (and is being cut off) because we are adding the final asset in later (after we make it for you)? Just want to make sure I am not adding QA items for things you are already aware of :)!

src/pages/sign-up.tsx
zchsh: Loosely held hot take: as routing gets more complex, I feel like it might not be a bad thing to have every page have a `view` component (even for `view` components that are only used on one page). So I'd be happy either way (whether we want to move the view component into this page file, or keep it separate in `views`).

zchsh (approved):
🙌 ✅ Nice work! Preview looks great to me 👍

src/hooks/use-authentication.ts
ashleemboyer: Yep! TypeScript only needs one or the other and doesn't like both.

CleanShot 2022-07-29 at 14 37 36@2x

braydenlove:
**This looks great, caught a few things when we QAed this morning. Here is a [video](https://www.loom.com/share/30f565113b364d1c885fa82e2387cae2) going over all of the items.**

- To help better align the tertiary button to the typography (probably bad css on the p):

```

.sign-up_signInContainer {
gap: 0px;
}
```

```
p.text_root {
margin-top: 2px;
}

```

- As we scale down the viewport we are seeing zoomed in text that is off the grid (starting at 1000px)
- Padding on the right side of the screen doesn't seem to be operating correctly and we are having text going out of frame (starting at ~600px)
- Bullet list check Icons are displaying at different sizes (starting at 348px)

Not sure why I can't upload photos to GitHub right now :/

src/hooks/use-authentication.ts
ashleemboyer: So both `callbackUrl` and `redirect` can be overwritten by providing them in the `options` object, but they are not overwritten by `restOptions`. I've added the extra destructuring line of code in order [to provide default values](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#default_value) for both `callbackUrl` and `redirect`, then pulling `restOptions` so they can be put back into the object sent to `signOut` untouched.

src/pages/sign-up.tsx
ashleemboyer: In this case, I opted for making a view component because of how much CSS was needed and because two SVG assets were needed. This was easier to organize in a view component folder than directly in `src/pages`.

src/views/sign-up-view/index.tsx
ashleemboyer: Good call, I've updated this to have two separate redirects:

- Kept the 404 for `!isAuthEnabled`
- Added a redirect to `/profile` if `isAuthenticated` is true

https://github.com/hashicorp/dev-portal/pull/743/commits/1e7eb3be6e71e29631cd6192f5f84daaed8bf19a

braydenlove:
Last thing I forgot. At mobile we are thinking we just remove the graphic as it will be below all of the actions and the graphic will become very small. Thoughts? [Figma Link](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=11467%3A114715)

ashleemboyer:
@braydenlove just pushed these updates!

- No spacing between "Already have an account?" text and "Sign In" button
- Added `margin-top: 2px` to "Already have an account?" text (this won't withstand dynamic font size units but we can come back to it)
- Check icons no longer shrink on smaller viewports
- No longer showing the aside content on mobile viewports

The other issue you were seeing with the mobile preview looking zoomed in is a quirky thing Chrome is unexpectedly doing. If you zoom out on the page using two fingers on a trackpad in a pinch motion, everything should be good!

braydenlove:
Thanks for the info and how quick you turned those changes around! Confirming that the zoom on the trackpad thing is working to fix the type stuff! I am noticing something weird now where the background gradient layer is splitting off from the bg which is new. I will DM you the image!

braydenlove:
Approved, thanks again!

src/views/sign-up-view/index.tsx
cbfx: @ashleemboyer no problem! nice to meet y'all, btw!

ashleemboyer: Nice to meet you @cbfx!! 👋🏻

PR #746, opened 2022-07-28
Render UserDropdownDisclosure for authenticated users in wide viewports

ashleemboyer:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

- Adds `className` prop to `UserDropdownDisclosure` to enable custom styling in some cases (like the dark navigation header)
- Adds `showAuthenticatedUI` and `showUnauthenticatedUI` booleans to the return object of `useAuthentication` (shout out to @kendallstrautman for the idea in a PR from this week! 🙌🏻)
- Updates `NavigationHeader` to render `UserDropdownDisclosure` for authenticated users on wide viewports (narrow viewports coming soon)

## 📸 Design Screenshots

[Figma link](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=11357%3A114658)

## 🧪 Testing

- [ ] Check out this code locally and run the server (auth isn't working in preview urls yet)
- [ ] Make sure your browser is in desktop width
- [ ] Go to the home page
- [ ] Authenticate yourself using the "Sign In" button in the nav header
- [ ] The page should reload
- [ ] The `UserDropdownDisclosure` should show in place of the "Sign In" and "Sign Up" elements (the element will flash in and we will address that outside of this PR)
- [ ] Your avatar should show in the `UserDropdownDisclosure`
- [ ] Change your browser width to a tablet/mobile viewports
- [ ] The `UserDropdownDisclosure` should no longer show
- [ ] Change your viewport back to a desktop width
- [ ] Open the `UserDropdownDisclosure`
- [ ] You should see the correct label and description for your user
- [ ] You should see three items in the list:
  - [ ] Bookmarks
  - [ ] Account Settings
  - [ ] Sign Out
- [ ] Make sure the Bookmarks link goes to the right place (it's OK that it 404s)
- [ ] Make sure the Account Settings goes to the right place
- [ ] Make sure the Sign Out option works

### DIFF
--- a/src/components/navigation-header/index.tsx
+++ b/src/components/navigation-header/index.tsx
@@ -1,11 +1,22 @@
+// Third-party imports
 import { useRouter } from 'next/router'
+
+// HashiCorp imports
+import { IconBookmark16 } from '@hashicorp/flight-icons/svg-react/bookmark-16'
+import { IconExternalLink16 } from '@hashicorp/flight-icons/svg-react/external-link-16'
 import { IconMenu24 } from '@hashicorp/flight-icons/svg-react/menu-24'
+import { IconSignOut16 } from '@hashicorp/flight-icons/svg-react/sign-out-16'
 import { IconUserPlus16 } from '@hashicorp/flight-icons/svg-react/user-plus-16'
 import { IconX24 } from '@hashicorp/flight-icons/svg-react/x-24'
-import { useCurrentProduct, useMobileMenu } from 'contexts'
+
+// Global imports
 import useAuthentication from 'hooks/use-authentication'
+import { useCurrentProduct, useMobileMenu } from 'contexts'
 import Button from 'components/button'
 import StandaloneLink from 'components/standalone-link'
+import UserDropdownDisclosure from 'components/user-dropdown-disclosure'
+
+// Local imports
 import { NavigationHeaderItem } from './types'
 import {
 	GiveFeedbackButton,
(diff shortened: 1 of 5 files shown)


src/components/navigation-header/index.tsx
```diff
@@ -38,32 +49,58 @@ const MobileMenuButton = () => {
...
+				items={[
+					{
+						href: '/bookmarks',
+						icon: <IconBookmark16 />,
+						label: 'Bookmarks',
+					},
+					{
+						href: 'https://portal.cloud.hashicorp.com/account-settings',
+						icon: <IconExternalLink16 />,
+						label: 'Account Settings',
```
kendallstrautman, line 90: 💭 We probably want this link to open a new tab? If it's going to the external HCP settings page.

src/components/user-dropdown-disclosure/types.ts
```diff
@@ -20,6 +20,7 @@ type UserDropdownDisclosureItem =
 	  }
 
 interface UserDropdownDisclosureProps {
+	className?: DropdownDisclosureProps['className']
```
kendallstrautman, line 23: Total style nit: I think it's neat to have the optional properties at the bottom of interface definitions. Take it or leave it 😸

src/hooks/use-authentication.ts
```diff
@@ -57,8 +57,11 @@ const useAuthentication = (options: UseAuthenticationOptions = {}) => {
 	})
 
 	// Deriving booleans about auth state
+	const isAuthEnabled = AUTH_ENABLED
 	const isLoading = status === 'loading'
 	const isAuthenticated = status === 'authenticated'
+	const showAuthenticatedUI = isAuthenticated
```
kendallstrautman, line 63: This one might feel a bit redundant since it just returns `isAuthenticated`, but do see how it would read better to have these two options. Just thinking outloud.

src/hooks/use-authentication.ts
```diff
@@ -57,8 +57,11 @@ const useAuthentication = (options: UseAuthenticationOptions = {}) => {
...
+	const showUnauthenticatedUI = isAuthEnabled && !isLoading && !isAuthenticated
```
kendallstrautman, line 64: This will be super useful!

kendallstrautman (approved):
LGTM! 🚀

src/components/navigation-header/index.tsx
ashleemboyer: I was wondering about this too. It'll require some refactoring so I'll do it outside of this PR, but I've taken a note to ask design folks about this! 😊

src/components/user-dropdown-disclosure/types.ts
ashleemboyer: Ah, I did't know that was your preference! I tend to alphabetize properties in most cases because it can make them easier to find when there are a lot of them. Might be a style thing to chat about as a group and explore automating. 😎

PR #755, opened 2022-08-01
[WIP] POC of user data fetching

BrandonRomano:


src/pages/api/auth/[...nextauth].ts
```diff
@@ -43,6 +43,7 @@ export default NextAuth({
 				...session,
 				user: { ...session.user, nickname: token.nickname },
 				id: token.sub,
+				access_token: token.cloud_idp_access_token,
```
Comment (BrandonRomano) [inline src/pages/api/auth/[...nextauth].ts:46]: Had to change auth here to expose the `access_token` to the frontend.

We need to pass this entire token so it can be validated on the API side to ensure it was signed by our trusted IDP.

.editorconfig
```diff
@@ -0,0 +1,3 @@
+# Tab indentation
+[*]
+indent_style = tab
```
BrandonRomano, line 3: This addition will come in a separate PR. This slipped in here as I wanted to set this as I was working through this :)

https://github.com/hashicorp/dev-portal/pull/756

src/views/profile-view/components/dynamic-bookmarks-list.tsx
```diff
@@ -0,0 +1,42 @@
...
+					'Content-Type': 'application/json',
+					Authorization: `Bearer ${session.access_token}`,
+				},
+			})
+				.then((res) => res.json())
+				.then((res) => {
+					setBookmarks(res.result)
+				})
+		}
+	}, [])
```
BrandonRomano, line 27: Whatever logic we end up going with, we'll want to pack this into the Learn client.

Just working on a POC here, so this will remain just a simple direct fetch within this PR.

BrandonRomano:
I believe this POC has served its purpose and I'm going to close this out now.

PR #764, opened 2022-08-03
Refactor CardLink anchor element

ashleemboyer:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1202097197789424/1202689706806161/f) 🎟️

## 🗒️ What

- Updates `CardLink` to no longer wrap it's content in an anchor HTML element
- Updates `CardLink` to require the `ariaLabel` prop
- Updates references to `CardLink` that needed an `ariaLabel` added

## 🤷 Why

In a gist, wrapping arbitrary amounts of content in an anchor element provides an inconsistent and often overly verbose screen reader experience. These changes also place a healthy constraint on developers to create/use quality `aria-label`s for usages of `CardLink`.

> Perhaps the worst thing you can do for a block link is to wrap everything in the `<a href>`

[Block Links, Cards, Clickable Regions, Rows, Etc.](https://adrianroselli.com/2020/02/block-links-cards-clickable-regions-etc.html)

## 🧪 Testing

Make sure the `CardLink`s function as you expect on the following pages. Feel free to test the same pages for other products as well.

- [ ] https://dev-portal-git-ambrefactor-cardlink-anchor-hashicorp.vercel.app/
- [ ] https://dev-portal-git-ambrefactor-cardlink-anchor-hashicorp.vercel.app/waypoint
- [ ] https://dev-portal-git-ambrefactor-cardlink-anchor-hashicorp.vercel.app/waypoint/docs
- [ ] https://dev-portal-git-ambrefactor-cardlink-anchor-hashicorp.vercel.app/waypoint/tutorials
- [ ] https://dev-portal-git-ambrefactor-cardlink-anchor-hashicorp.vercel.app/waypoint/downloads


alexcarpenter:
🙋 Do you foresee a use-case when you'd need to style a child element within `CardLink` based on focusing the link? The move to having the link at the end makes this a bit more difficult to grasp, i'd have to look at the underlying implementation to know that the link does not wrap the content.

From my understanding you'd get the same experience read to the user by requiring the aria-label usage while still having the link wrap the content vs the move to absolutely positioning it at the end of the card content.

ashleemboyer:
> 🙋 Do you foresee a use-case when you'd need to style a child element within `CardLink` based on focusing the link? The move to having the link at the end makes this a bit more difficult to grasp, i'd have to look at the underlying implementation to know that the link does not wrap the content and require `focus-within` usage?

I don't forsee that use-case right now, no. I'm not sure that's a capability we have with our current implementation either, because the `className` passed to `CardLink` is currently passed straight to `Card`. [We don't apply any styles to the `<a>` element currently](https://github.com/hashicorp/dev-portal/blob/main/src/components/card-link/index.tsx):

```tsx
const CardLink = ({
  ariaLabel,
  children,
  className,
  href,
  openInNewTab,
}: CardLinkProps): ReactElement => {
  const classes = classNames(s.root, className)
  const target = openInNewTab ? '_blank' : undefined

  return (
    <Link href={href}>
      <a aria-label={ariaLabel} target={target}>
        <Card className={classes}>{children}</Card>
      </a>
    </Link>
  )
}
```

> From my understanding you'd get the same experience read to the user by requiring the aria-label usage while still having the link wrap the content vs the move to absolutely positioning it at the end of the card content.

So the experiences aren't actually the same when all content is wrapped in an anchor. When arrowing through the content -- for example say there's an image, heading, then body text -- each element would be announced as a link. See the second video demo in this section of Adrian Roselli's post: https://adrianroselli.com/2020/02/block-links-cards-clickable-regions-etc.html#Links

src/components/card-link/index.tsx
```diff
@@ -15,12 +16,29 @@ const CardLink = ({
...
 	const target = openInNewTab ? '_blank' : undefined
 
+	if (!ariaLabel) {
+		developmentToast({
+			color: ToastColor.critical,
+			title: 'Error in CardLink',
+			description:
+				'`CardLink` requires the `ariaLabel` prop in order to be announced by screen readers.',
+		})
+	}
```
BRKalow, line 26: Question 🙋: due to `ariaLabel` now being a required prop based on the `CardLinkProps` interface, this block should never be hit, right? Or is this to handle the case where we are getting dynamic data from authorable files?

BRKalow (approved):
I had a question about the prop check based on the type definition, but otherwise this looks good to me! Thanks for providing the resource that led to this change, very helpful!

src/components/card-link/index.tsx
ashleemboyer: Correct it should be hit very seldomly! It can happen really easily if the component is rendered via MDX such as the Swingset page. So this is just a little fallback just in case. 😊

PR #766, opened 2022-08-03
feat: add Dependabot support

dstaley:
## 🗒️ What

This PR adds support for Dependabot to ensure that our internal HashiCorp dependencies are kept up-to-date.

## 🤷 Why

When dependency updates to HashiCorp packages are made by members outside of the Dev Portal team, there's a chance that bug fixes and new features don't flow into Dev Portal.

## 🛠️ How

Adding `.github/dependabot.yml` configured to update once a week.

## 💭 Anything else?

Currently Dependabot is only configured to support `@hashicorp/*` packages. Are there other packages we want to ensure always remain up-to-date?


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
```
BRKalow, line 1: Can we add `hashicorp/web-platform` as a reviewer? https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#reviewers

BRKalow (approved):
Looking forward to seeing if we like this!

PR #779, opened 2022-08-05
chore: empty commit to bypass env var limit

zchsh:
## 🔗 Relevant links

This PR contains a single empty commit, intended to get the latest work on `main` deploying to Vercel as expected.

- 🔗  [Notion doc on why this is necessary](https://www.notion.so/hashicorp/Bypass-intermittent-environment-variable-size-limit-on-Vercel-deploy-14f03b05855042189145f53f999b4671)
- TLDR; the last commit on main has a very long commit message, which fails cause Vercel includes it in env.


thiskevinwang (approved):
👍

PR #797, opened 2022-08-08
ubuntu: pipe to dd when downloading gpg key

shoenig:
This PR swaps the use of `tee` for `dd` when creating the gpg key file
for setting up the apt repository on Debian/Ubuntu systems.

Using `tee` causes binary output to be displayed on the terminal, which
is not desireable (see  https://github.com/hashicorp/nomad/issues/14047)

### DIFF
--- a/src/views/product-downloads-view/helpers.ts
+++ b/src/views/product-downloads-view/helpers.ts
@@ -31,7 +31,7 @@ export const generateDefaultPackageManagers = (
 		{
 			label: 'Ubuntu/Debian',
 			commands: [
-				`wget -O- https://apt.releases.hashicorp.com/gpg | gpg --dearmor | sudo tee /usr/share/keyrings/hashicorp-archive-keyring.gpg`,
+				`wget -O- https://apt.releases.hashicorp.com/gpg | gpg --dearmor | sudo dd of=/usr/share/keyrings/hashicorp-archive-keyring.gpg`,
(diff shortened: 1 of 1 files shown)


src/views/product-downloads-view/helpers.ts
```diff
@@ -31,7 +31,7 @@ export const generateDefaultPackageManagers = (
 		{
 			label: 'Ubuntu/Debian',
 			commands: [
-				`wget -O- https://apt.releases.hashicorp.com/gpg | gpg --dearmor | sudo tee /usr/share/keyrings/hashicorp-archive-keyring.gpg`,
+				`wget -O- https://apt.releases.hashicorp.com/gpg | gpg --dearmor | sudo dd of=/usr/share/keyrings/hashicorp-archive-keyring.gpg`,
```
phinze, line 34:
```suggestion
				`wget -O- https://apt.releases.hashicorp.com/gpg | gpg --dearmor -o /usr/share/keyrings/hashicorp-archive-keyring.gpg`,
```

Hello! I came here also chasing gobbledygook in the install output for Vagrant. I have an alternate fix to suggest though via the [Docker install instructions for linux](https://docs.docker.com/engine/install/ubuntu/). `gpg` supports a `-o` flag so we can just use that directly instead.

phinze (approved):
Would love to see this make it onto the site, but I also have one tweak to suggest!

kendallstrautman:
👋🏻 Hello! checking back to see if we still want to get these updates in? If not, let's close this PR out ☺️

PR #809, opened 2022-08-09
Minimally reproduce static paths bug

BrandonRomano:


src/pages/[productSlug]/integrations/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+import ProductIntegrationsLanding from 'views/product-integrations-landing'
+
+export async function getStaticPaths() {
+	const paths = ['vault', 'packer'].map((productSlug: ProductSlug) => ({
+		params: { productSlug },
+	}))
+	console.log('getStaticPaths', paths)
+	return {
+		paths,
+		fallback: true,
```
Comment (BRKalow) [inline src/pages/[productSlug]/integrations/index.tsx:12]: 💭 https://nextjs.org/docs/api-reference/data-fetching/get-static-paths#fallback-pages

```suggestion
		fallback: false,
```

Comment (BrandonRomano) [reply src/pages/[productSlug]/integrations/index.tsx:12]: :brycefax:

BrandonRomano:
Thanks for the help here @BRKalow 🙏

PR #822, opened 2022-08-11
[refactor] Organizing use-authentication files; storing cloud_idp_access_token

ashleemboyer:
## 🗒️ What

- Pulls `cloud_idp_access_token` from the JWT and names it `accessToken` under `SessionData`
- Nests the `use-authentication` hook in a folder under `src/hooks`
- Pull some code out of the main file into some new ones under the new directory:
  - New `src/hooks/use-authentication/helpers` directory: `signInWrapper`, `signOutWrapper`, and `signUp`
  - New `src/hooks/use-authentication/types.ts` file

## 🤷 Why

- The `cloud_idp_access_token` is needed in order to make authenticated requests to the Learn API.
- The `use-authentication.ts` file was getting pretty long. These changes help make it easier to consume and it's about half the length now.

## 🧪 Testing

We can't test this in preview URLs yet, but the code can be checked out and run locally to make sure the auth functionality still works as expected.


kendallstrautman (approved):
LGTM! I'd wait for Bryce to take a look also though 👍

src/types/auth.ts
```diff
@@ -4,7 +4,8 @@ import { Session } from 'next-auth'
  * Object representing an authentication session.
  */
 type SessionData = Omit<Session, 'user'> & {
-	id?: 'string'
+	accessToken?: string
+	id?: string
```
kendallstrautman, line 8: Just curious why these are optional?

ashleemboyer: I can't remember if we saw issues otherwise or not. They may not need to be optional? Hmmm 🤔

BRKalow: We could try to make them required! I guess _technically_ there's no guarantee that they will be there as the token could not have the data or something, but in practice it might be okay for them to be required!

BRKalow (approved):
LGTM! Nice work

src/types/auth.ts
ashleemboyer: I'll make a mental note to check this out next week and make the update if possible! 😊

PR #827, opened 2022-08-15
fix: terraform sidebar title and overview items

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Fixes duplicative titles and `"Overview"` items in Terraform sidebars.

## 🤷 Why

To match design specs.

## 🛠️ How

- Adds support for `hideSidebarTitle` and `addOverviewItem` properties for `RootDocsPath` items
- Uses these new properties to fix up the Terraform sidebar, and retain existing behaviour for Vault and Waypoint
- Fixes a minor issue where sometimes `isActiveItem` wouldn't match due to trailing slashes (eg `fullPath` is `/terraform/cloud-docs/`, does not match `currentPath` which would be `/terraform/cloud-docs`

## 📸 Design Screenshots

This PR addresses an issue across most Terraform docs categories (all except `/terraform/docs`). An example of the issue in the current "stub", alongside the design spec, is shown below in the context of `/terraform/internals`:

terraform-internals  fix up rendering of “overview” item

The work in this PR yields the following result:

CleanShot 2022-08-15 at 14 39 03@2x

## 🧪 Testing

Visit existing Vault and Waypoint docs categories. These sidebars in each of these docs category contexts should continue to render `"Overview"` items:

- [/vault/api-docs][/vault/api-docs]
- [/vault/docs][/vault/docs]
- [/waypoint/docs][/waypoint/docs]
- [/waypoint/commands][/waypoint/commands]
- [/waypoint/plugins][/waypoint/plugins]

Visit all Terraform docs routes. None of the sidebars in each docs category should contain duplicative visible titles or overview items:

- [/terraform/cdktf][/cdktf]
- [/terraform/cli][/cli]
- [/terraform/cloud-docs][/cloud-docs]
- [/terraform/cloud-docs/agents][/cloud-docs/agents]
- [/terraform/docs][/docs]
    - Note: uses a typical `Overview` item approach
- [/terraform/enterprise][/enterprise]
- [/terraform/internals][/internals]
    - Note: does not have an `Overview` item. Adding one would require more complex logic to inject the `Overview` item after the first node in `nav-data` if the first node is a heading.
    - Intent is to implement this more complex logic in a separate chunk of work: 🎟️  [[terraform] implement automatic overview-item-detection-injection logic](https://app.asana.com/0/0/1202817253755969/f)
- [/terraform/intro][/intro]
- [/terraform/language][/language]
- [/terraform/plugin][/plugin]
- [/terraform/plugin/framework][/plugin/framework]
- [/terraform/plugin/log][/plugin/log]
- [/terraform/plugin/mux][/plugin/mux]
- [/terraform/plugin/sdkv2][/plugin/sdkv2]
- [/terraform/registry][/registry]

## 💭 Anything else?

- Our design spec shows `"Overview"` items for all docs category sidebars. This PR does _not_ try to ensure those are present.
    - Intent is to add these as we make content updates
    - Only specific routes are missing `"Overview"` items. For example, [/terraform/cloud-docs][/cloud-docs] already has an overview item, while [/terraform/internals][/internals] does not. In this case, fixing the content inconsistency seems compatible with the existing `dot-io` setup, so it feels more logical to fix this in content than in implementation.

- A potential alternative to the `hideSidebarTitle` portion of this PR would be to modify all relevant `*-nav-data.json` files in Terraform content. The rationale for not taking the content approach at this time is:
    - Modifying `*-nav-data.json` would affect <https://terraform.io> in unintentional ways (there are multiple sidebar sections currently, so the heading items are important)
    - Modifying `*-nav-data.json` would only fix the issue for the latest version of content.
    - Once we're in GA, and docs on <terraform.io> have been migrated to <https://developer.hashicorp.com>, we'll be able to make the desired content changes without having to worry about effects on <https://terraform.io>

- On an implementation note, I think we can clean up the interface of `docs-view/server`, since it's mostly just passing in `rootDocsPath` at this point. I'm going to handle that in a separate PR, since it's a bit of an optimization. Definitely felt relevant here though, and worth mentioning.
    - 🎟️ [[docs-view] Refactor docs-view/server to reduce duplicative rootDocsPath-related arguments](https://app.asana.com/0/1202097197789424/1202802160011097/f)

[task]: https://app.asana.com/0/1202097197789424/1202785022290137/f
[preview]: https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/
[/cdktf]: https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/cdktf
[/cli]: https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/cli
[/cloud-docs]: https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/cloud-docs
[/cloud-docs/agents]: https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/cloud-docs/agents
[/docs]: https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/docs
[/enterprise]: https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/enterprise
[/internals]: https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/internals
[/intro]: https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/intro
[/language]: https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/language
[/plugin]: https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/plugin
[/plugin/framework]: https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/plugin/framework
[/plugin/log]: https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/plugin/log
[/plugin/mux]: https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/plugin/mux
[/plugin/sdkv2]: https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/plugin/sdkv2
[/registry]: https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/registry
[/vault/api-docs]: https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/vault/api-docs
[/vault/docs]: https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/vault/docs
[/waypoint/docs]: https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/waypoint/docs
[/waypoint/commands]: https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/waypoint/commands
[/waypoint/plugins]: https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/waypoint/plugins


zchsh:
@ashleemboyer @BRKalow This PR makes some changes to fix up the Terraform sidebar to match designs.

Aim here is to leave content changes to be made in `*-nav-data.json` files where possible, and do what we can in the implementation to have the flexibility to meet designs; while not stepping too much further into content-shimming territory. Specifically:
    - `"Overview"` items are no longer strictly added. We don't need to add these for Terraform.
    - Titles can be temporarily hidden. Terraform content already contains headings, so we don't want to add them in (visibly) in those cases. Ultimately I think we'll want to solve this by removing headings from terraform `*-nav-data.json` files. But currently this isn't feasible, as we need the headings for things to display as expected on <terraform.io>.

src/views/docs-view/server.ts
```diff
@@ -260,27 +262,38 @@ export function getStaticGenerationFunctions<
...
+			// If the title is not hidden for this rootDocsPath, include it
+			if (currentRootDocsPath.hideSidebarTitle) {
+				docsSidebarLevel.visuallyHideTitle = true
+			}
+			// If we want to add an "overview" item, add one
+			if (currentRootDocsPath.addOverviewItem) {
+				docsSidebarLevel.overviewItemHref = versionPathPart
+					? `/${product.slug}/${basePath}/${versionPathPart}`
+					: `/${product.slug}/${basePath}`
+			}
```
ashleemboyer, line 288: AFAIK, we want the `overviewItemHref` behavior to be the default from a design standpoint, and removing it to be an exception. Other products depend on this behavior working as well, so this needs to be backwards compatible.

Could we flip this logic with a rename or with an adjustment to the conditional check? Maybe checking that the option has been explicitly turned off will work:

```typescript
// Add "Overview" item, unless explicitly disabled
if (currentRootDocsPath.addOverviewItem !== false) {
  docsSidebarLevel.overviewItemHref = versionPathPart
	  ? `/${product.slug}/${basePath}/${versionPathPart}`
		: `/${product.slug}/${basePath}`
}
```

src/views/docs-view/server.ts
```diff
@@ -260,27 +262,38 @@ export function getStaticGenerationFunctions<
...
+				backToLinkProps: {
+					text: `${product.name} Home`,
+					href: `/${product.slug}`,
+				},
+				levelButtonProps: {
+					levelUpButtonText: `${product.name} Home`,
+				},
+				menuItems: navDataWithFullPaths as EnrichedNavItem[],
+				// If the title is not hidden for this rootDocsPath, include it
+				title: BASE_PATHS_TO_NAMES[basePath] || product.name,
```
ashleemboyer, line 277: Can we retain the `TODO` comment that was present before? Looks like this might be an accidental copy/paste from the new `hideSidebarTitle` block

src/types/products.ts
```diff
@@ -115,6 +115,20 @@ interface RootDocsPath {
 	 * content API should pull from. Defaults to `main`.
 	 */
 	mainBranch?: string
+
+	/**
+	 * An optional property to hide the title of this rootDocsPath
+	 * in the sidebar. Used for Terraform routes where sidebar titles
+	 * are present in nav-data.json.
+	 */
+	hideSidebarTitle?: boolean
```
ashleemboyer, line 124: Should we clarify that this just _visually_ hides the title? Maybe add it to the name, like `visuallyHideSidebarTitle`?

src/components/sidebar/components/sidebar-headings/sidebar-headings.module.css
```diff
@@ -1,10 +1,10 @@
...
-  margin-bottom: 12px;
-  margin-top: 0;
-  padding-left: 8px;
-  padding-right: 8px;
+	color: var(--token-color-foreground-primary);
+	font-size: var(--token-typography-display-100-font-size);
+	font-weight: var(--token-typography-font-weight-semibold);
+	line-height: var(--token-typography-display-100-line-height);
+	margin-bottom: 12px;
+	margin-top: 20px;
```
ashleemboyer, line 7: Is there a particular issue being solved with this setting? It causes the spacing to not match design spec when there is a heading item directly after the filter input. ([Figma link](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=11303%3A116341))

CleanShot 2022-08-15 at 16 48 17@2x

ashleemboyer:
Random VQA find: the nav data for [/terraform/language](https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/language) has an extra divider at the end of it. Could we remove it at the source?

CleanShot 2022-08-15 at 16 49 53@2x

ashleemboyer:
> Random VQA find: the nav data for [/terraform/language](https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/language) has an extra divider at the end of it. Could we remove it at the source?
> 
> CleanShot 2022-08-15 at 16 49 53@2x

Hmm actually looks like this is happening in a couple of places:

- https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/cli
- https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/cloud-docs
- https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/enterprise
- https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/internals
- https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/language

src/components/sidebar/components/sidebar-headings/sidebar-headings.module.css
ashleemboyer: Is this one of the pages that the change was needed? https://dev-portal-git-zsfix-terraform-sidebar-overview-item-hashicorp.vercel.app/terraform/plugin

zchsh: ☝️ yep, exactly - it's where headings appear in the midst of nav-data. Also I am totally face palming right now because I wrote review comments in a text file and then forgot to post them 🤦

zchsh: Update: I've tweaked this to `16px` since we do have a design spec for a heading at the top of nav data, but not for one in middle of nav data. Thanks for catching this & linking the spec 💖

src/components/sidebar/components/sidebar-mobile-controls/index.tsx
```diff
@@ -70,10 +70,19 @@ const LevelDownButton = ({ onClick, text }: ButtonProps) => {
...
 
-interface SidebarMobileControlsProps {
-	levelUpButtonText: ButtonProps['text']
-	levelDownButtonText: ButtonProps['text']
-}
+/**
+ * One or both of `levelUp` and `levelDown`
+ * `ButtonText` properties must be provided.
+ */
+type SidebarMobileControlsProps =
```
zchsh, line 77: I wanted to use the `SidebarProps` type in `docs-view/server`, but needed to make it so either `levelUpButtonText` or `levelDownButtonText` or both could be provided here. This type aims to accomplish that.

src/components/sidebar/helpers/add-nav-item-meta-data.ts
```diff
@@ -56,7 +56,8 @@ export const addNavItemMetaData = (
 			// Found an `EnrichedLinkNavItem` object
 			if (item.hasOwnProperty('fullPath')) {
 				const itemPath = (item as EnrichedLinkNavItem).fullPath
-				const isActive = !foundActiveItem && itemPath === currentPath
+				const isActive =
+					!foundActiveItem && itemPath.replace(/\/$/, '') === currentPath
```
zchsh, line 60: Fixes an issue where items with a trailing slash in their `fullPath` were not matched, even though they're equivalent (eg when comparing `fullPath` of `/terraform/cloud-docs/` with `currentPath` of `/terraform/cloud-docs`)

src/data/terraform.json
```diff
@@ -104,25 +104,29 @@
 			"iconName": "docs",
 			"name": "CDKTF",
 			"path": "cdktf",
-			"productSlugForLoader": "terraform-cdk"
+			"productSlugForLoader": "terraform-cdk",
+			"hideSidebarTitle": true
```
zchsh: Adds newly introduced `hideSidebarTitle` prop for Terraform docs categories where it's needed.

src/data/vault.json
```diff
@@ -151,12 +151,14 @@
 			"iconName": "docs",
 			"name": "General Documentation",
 			"path": "docs",
-			"shortName": "Documentation"
+			"shortName": "Documentation",
+			"addOverviewItem": true
```
zchsh, line 155: I've updated `vault` and `waypoint` data with newly introduced `addOverviewItem` prop. I've gone the "add" route because a lot of Terraform routes would be duplicative otherwise.

I was admittedly on the fence whether to invert this and add `hideOverviewItem` to the Terraform items, though that felt like it might imply that we're hiding an incoming overview item from `nav-data`, which is not the case. The other thing that swayed me was the idea that adding these "Overview" items is something we should do in content, not in implementation. Intent being that we can deprecate `addOverviewItem` route-by-route post-GA by updating content.

Very open to other approaches here 🙇‍♂️

src/data/waypoint.json
```diff
@@ -66,17 +66,20 @@
 			"name": "General Documentation",
 			"path": "docs",
 			"shortName": "Documentation",
+			"addOverviewItem": true,
```
zchsh, line 69: Same as in `vault.json`, making sure to retain existing behaviour with `addOverviewItem` on all `rootDocsPaths`

src/types/products.ts
zchsh: I dig it! ✅ Updated

src/views/docs-view/server.ts
zchsh: @ashleemboyer Definitely agreed that it seems like we want the `overviewItemHref` to be consistent across products. I think to get to that consistency, without code-level shims, I see two broad options:

1. _Remove_ all existing `"Overview"` items, for example from Terraform docs categories' `*-nav-data.json`, and rely on the implementation to shim in the desired content
2. _Add_ `"Overview"` items to all `*-nav-data.json` items where they're missing, and then remove our code that shims in this content

Personally I lean towards the `2.`, so I think we should have the implementation-side shimming of missing `"Overview"` items in sidebar content be the exception rather than the rule. So we should be explicitly setting `addOverviewItem` in cases where we haven't had the chance to make the content update.

That being said, I don't feel too strongly here - it just feels odd to me for us to try to implement a permanent-default fix (rather than temporary fix) for what seems to boil down to a content inconsistency issue. Let me know if that seems reasonable, happy to adjust to do the flip-logic-rename you suggest (say, `doNotAddOverviewItem`) if that still feels preferable 🙇‍♂️

src/views/docs-view/server.ts
zchsh: Absolutely, great catch, this was a slip on my part (exactly as you suspected!). I've added it back 👍

src/views/docs-view/server.ts
ashleemboyer: I think we have different perspectives of what the Overview item is. As it's currently coded, I see it as a feature of the platform, not a shim. If we were injecting this item server-side as we fetched/formatted the sidebar data, I'd see it more like a shim. But regardless of what we call it, isn't it easier for us to code this into the platform and maintain it rather than rely on 9 different teams and repositories to do it?

Our current approach that we've been following for a while is that we place an Overview item in the sidebar. These changes break that approach for other products, so I definitely think they need to be made in a backwards-compatible way. If we want to revisit the Overview item approach, I think it's a larger discussion we need to involve both design and product in.

src/components/sidebar/components/sidebar-headings/sidebar-headings.module.css
ashleemboyer: That should work well! 👍🏻 It'll also continue working if/when dividers are added in since those have a larger margin on them.

src/views/docs-view/server.ts
BRKalow: Can we detect whether or not there is an existing `Overview` item and add it if missing? Considering the constraints we have here:
* We want an `Overview` nav item for every root path
* Some nav data files have `Overview` items already defined
* We don't want to duplicate the nav items on older versions

Doing a conditional "shim" feels like the more robust approach. That also gives us the ability to shift to making this an authoring concern, if that is a route we want to take in the future.

zchsh: @ashleemboyer I think it is easier for us to code the `"Overview"` item into the platform, _if_ we have full author alignment on _not_ including the item. To me getting author alignment either way is necessary, so it seems logical to avoid extra code and instead ask authors to align content by including the `"Overview"` item rather than omitting it.

@BRKalow I'd be on board with a conditional shim. It seems like that'd be a more automatic version of manually specifying `addOverviewItem` in product data. Would we be alright with handling that in a separate PR? (Happy to tackle here in this PR if we'd rather; personally I lean towards a separate ticket and coming back to it after Terraform work is done).

BRKalow (approved):
👍

PR #828, opened 2022-08-15
Add Bookmark fetchers

ashleemboyer:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1202097197789424/1202378757122278/f) 🎟️

## 🗒️ What

- Adds an `ApiBookmark` interface
- Adds a few API functions for bookmarks:
  - `createBookmark`
  - `deleteBookmark`
  - `getBookmark`
  - `getAllBookmarks`
  - `getBookmarksByTutorialIds`
- Updates `post` method so that `bodyJson` is optional

## 🤷 Why

- The new methods are needed to start working on bookmarking features.
- The `post` method doesn't _always_ need `bodyJson` passed.

## 🧪 Testing

There is nothing to manually test here. The new methods are only invoked in tests and the updated `post` method is only invoked in `createBookmark`.


src/lib/learn-client/api/api-types.ts
```diff
@@ -75,6 +75,12 @@ export interface ApiCollectionTutorial
 	tutorial: ApiTutorial
 }
 
+export interface ApiBookmark {
```
kendallstrautman, line 78:
```suggestion
export interface ApiBookmark extends BaseModel {
```
This data will have the `id` & `created_at` etc.

src/lib/learn-client/api/bookmark/delete-bookmark.ts
```diff
@@ -0,0 +1,40 @@
...
+ * https://digital-api-specs.vercel.app/learn#tag/Bookmarks/paths/~1bookmarks~1{tutorial_id}/delete
+ */
+const deleteBookmark = async ({
+	accessToken,
+	tutorialId,
+}: DeleteBookmarkOptions): Promise<DeleteBookmarkResult> => {
+	// Build the API route to make a request to
+	const requestRoute = `${BOOKMARK_API_ROUTE}/${tutorialId}`
+
+	// Make the GET request
```
kendallstrautman, line 26:
```suggestion
	// Make the DELETE request
```

src/lib/learn-client/api/bookmark/get-bookmarks.ts
```diff
@@ -0,0 +1,49 @@
...
+ * https://digital-api-specs.vercel.app/learn#tag/Bookmarks/paths/~1bookmarks/get
+ */
+const getBookmarks = async ({
+	accessToken,
+	tutorialIds,
+}: GetBookmarksOptions): Promise<GetBookmarksResult> => {
+	// Build the API route to make a request to
+	let requestRoute = BOOKMARK_API_ROUTE
+	if (tutorialIds && tutorialIds.length > 0) {
+		const commaSeparatedIds = tutorialIds?.join(',')
```
kendallstrautman, line 30:
```suggestion
		const commaSeparatedIds = tutorialIds.join(',')
```

src/lib/learn-client/api/bookmark/get-bookmarks.ts
```diff
@@ -0,0 +1,49 @@
...
+ * Fetches bookmarks based on given `tutorialIds`. If none are provided, then
+ * all bookmarks will be fetched. If some `tutorialIds` are provided, then they
+ * will be passed in the comma-separated `tutorialIds` query parameter.
+ *
+ * If successful, then an array of bookmark objects is returned. Otherwise
+ * throws an error.
+ *
+ * https://digital-api-specs.vercel.app/learn#tag/Bookmarks/paths/~1bookmarks/get
+ */
+const getBookmarks = async ({
```
kendallstrautman, line 23: 💭 I wonder if it would be nice to split this into two distinct functions, `getBookmarks` where the `tutorialIds` array is required. And then `getAllBookmarks`. To me, that would make it more clear in the codebase when we are calling for a specific set of bookmarks versus all of them. Thoughts?

src/lib/learn-client/api/bookmark/__tests__/create-bookmark.test.ts
```diff
@@ -0,0 +1,68 @@
...
+import { createBookmark } from '../create-bookmark'
+import { BOOKMARK_API_ROUTE } from '..'
+
+const mockBookmark: ApiBookmark = {
+	tutorial_id: 'test-tutorial-1',
+	tutorial: {} as $TSFixMe,
+	user_id: 'test-user',
+}
+
+jest.mock('lib/learn-client', () => {
```
kendallstrautman, line 12: Don't you want to mock just the `createBookmark` function? Or what's the thought behind mocking the entire client?

ashleemboyer: These were set up to do [partial mocks](https://jestjs.io/docs/mock-functions#mocking-partials) of each method in `lib/learn-client`, so only the method used by the bookmark method being tested is what was getting mocked.

I went ahead and changed this in https://github.com/hashicorp/dev-portal/pull/828/commits/c27723485a2696d86fec44eebbcddb9bbe5b0d8a though, so the import is different and we can instead use `spyOn` and `restoreAllMocks`. Both make less code in the end. 😊

src/lib/learn-client/api/api-types.ts
ashleemboyer: Fixed in https://github.com/hashicorp/dev-portal/pull/828/commits/b2a8f3191373ca34985dd519b6ec12e609f5287f!

src/lib/learn-client/api/bookmark/delete-bookmark.ts
ashleemboyer: Fixed in https://github.com/hashicorp/dev-portal/pull/828/commits/5d4fea55d6f0bd69b0f17672f3d0ae8938010452!

src/lib/learn-client/api/bookmark/get-bookmarks.ts
ashleemboyer: Fixed in https://github.com/hashicorp/dev-portal/pull/828/commits/16768a7dfde1e4ef1f3232abbf20a12456405aaa!

src/lib/learn-client/api/bookmark/get-bookmarks.ts
ashleemboyer: Still thinking on this one! I had originally decided to only create one function since there's only one endpoint, to make a 1:1 to mapping.

ashleemboyer: Updated in https://github.com/hashicorp/dev-portal/pull/828/commits/1baea46b14953ddcadf1d06f0bb22bc2a7db69a6! There are now two methods instead of one:

- `getAllBookmarks`
- `getBookmarksByTutorialIds`

src/lib/learn-client/api/bookmark/__tests__/create-bookmark.test.ts
```diff
@@ -0,0 +1,61 @@
...
+const mockBookmark: ApiBookmark = {
+	created_at: null,
+	id: 'test-bookmark-1',
+	tutorial_id: 'test-tutorial-1',
+	tutorial: {} as $TSFixMe,
+	updated_at: null,
+	user_id: 'test-user',
+}
+
+describe('createBookmark', () => {
```
BRKalow, line 15: Nice tests!

src/lib/learn-client/index.ts
```diff
@@ -37,7 +37,7 @@ export function put(path, token, bodyJson) {
 	})
 }
 
-export function post(path, token, bodyJson) {
+export function post(path, token, bodyJson = {}) {
```
BRKalow, line 40: Question 🙋: Curious why you added the default for `bodyJson` here?

BRKalow (approved):
LGTM!

src/lib/learn-client/index.ts
ashleemboyer: To make it optional 😊

PR #829, opened 2022-08-15
🛹 feat: adds bookmark card icon

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksadds-bookmark-card-icon-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1199634971449915/1202791375540706) 🎟️

## 🗒️ What

This PR adds a new `BookmarkButton` component to the tutorial card eyebrow. Production will not render these changes since auth is not enabled. The production tutorial cards should not be affected by these updates. 

## 📸 Design Screenshots

Screen Shot 2022-08-15 at 5 09 13 PM

## 🧪 Testing

- Visit the [tutorial card swingset page](https://dev-portal-git-ksadds-bookmark-card-icon-hashicorp.vercel.app/swingset/components/tutorialcard) and test out the bookmark states. 
- Also check it out on the [tutorial product landing](https://dev-portal-git-ksadds-bookmark-card-icon-hashicorp.vercel.app/vault/tutorials) pages. 

## 💭 Anything else?

This isn't linked up with actual data yet, the click handlers are basically a stub. It's a skateboard for the UI. Eventually this button will need to trigger a dialog render if unauthenticated, and be able to update bookmark state if authenticated.

### DIFF
--- a/src/components/bookmark-button/icons/index.tsx
+++ b/src/components/bookmark-button/icons/index.tsx
@@ -1,56 +1,24 @@
+import { IconBookmark16 } from '@hashicorp/flight-icons/svg-react/bookmark-16'
+import { IconBookmarkAdd16 } from '@hashicorp/flight-icons/svg-react/bookmark-add-16'
+import { IconBookmarkFill16 } from '@hashicorp/flight-icons/svg-react/bookmark-fill-16'
+import { IconBookmarkRemoveFill16 } from '@hashicorp/flight-icons/svg-react/bookmark-remove-fill-16'
+import classNames from 'classnames'
 import s from './icons.module.css'
-
-/**
- * We need these icon to be an inline svg so we can manipulate the colors of the
- * 'plus' & 'minus' sign lines based in the varying bookmark hover states
- *
- * These were made from the flight library:
- * - '@hashicorp/flight-icons/svg-react/bookmark-16'
- * - '@hashicorp/flight-icons/svg-react/bookmark-fill-16'
- * - '@hashicorp/flight-icons/svg-react/bookmark-add-16'
- * - '@hashicorp/flight-icons/svg-react/bookmark-remove-16'
- */
 
 export function AddBookmarkIcon() {
 	return (
-		<svg
-			xmlns="http://www.w3.org/2000/svg"
-			width={16}
-			height={16}
-			fill="none"
-			viewBox="0 0 16 16"
-			aria-hidden={true}
-			className={s.icon}
-		>
-			<path
-				fillRule="evenodd"
-				d="M2 3.25A2.25 2.25 0 014.25 1h7.5A2.25 2.25 0 0114 3.25v10.83a1 1 0 01-1.478.878L8.12 12.564a.25.25 0 00-.238 0l-4.403 2.394A1 1 0 012 14.08V3.25zm2.25-.75a.75.75 0 00-.75.75v9.989l3.664-1.992a1.75 1.75 0 011.672 0l3.664 1.992V3.25a.75.75 0 00-.75-.75h-7.5z"
-				clipRule="evenodd"
-			/>
-			<path
-				id={s.addSymbol}
-				d="M7.75 4a.75.75 0 01.75.75V6h1.25a.75.75 0 010 1.5H8.5v1.25a.75.75 0 01-1.5 0V7.5H5.75a.75.75 0 010-1.5H7V4.75A.75.75 0 017.75 4z"
-			/>
-		</svg>
+		<>
+			<IconBookmark16 className={classNames(s.icon, s.base)} />
+			<IconBookmarkAdd16 className={classNames(s.icon, s.hover)} />
+		</>
 	)
 }
 
 export function RemoveBookmarkIcon() {
 	return (
-		<svg
-			xmlns="http://www.w3.org/2000/svg"
-			width={16}
-			height={16}
-			fill="none"
-			viewBox="0 0 16 16"
-			aria-hidden={true}
-			className={s.icon}
-		>
-			<path d="M4.25 1A2.25 2.25 0 002 3.25v10.83a1 1 0 001.478.878l4.403-2.394a.25.25 0 01.238 0l4.403 2.394A1 1 0 0014 14.08V3.25A2.25 2.25 0 0011.75 1h-7.5z" />
-			<path
-				id={s.removeSymbol}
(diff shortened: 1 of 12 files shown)


src/components/bookmark-button/icons/index.tsx
```diff
@@ -0,0 +1,56 @@
+import s from './icons.module.css'
+
+/**
+ * We need these icon to be an inline svg so we can manipulate the colors of the
+ * 'plus' & 'minus' sign lines based in the varying bookmark hover states
```
ashleemboyer, line 5: This _should_ be possible by setting `color` for the `<button>` the icons are placed in. This is something that works for me with a quick test:

```
<button style={{ color: 'red' }}>
  <IconBookmarkAdd16 />
</button>
```

CleanShot 2022-08-15 at 18 58 25@2x

kendallstrautman: It's not that I need to override the color, it's that I need to conditionally show the 'plus' or 'minus' icon based on the hover state. I'm definitely open to a different way of doing this! lmk if you have ideas.

kendallstrautman: I need to have a way to target just the 'plus' or just the 'minus' icon in order to show / hide them on hover and I'm not sure how to do that with just the flight icon unless I'm targeting elements, which feels prone to error later on if the markup of the icon shifts.

src/components/bookmark-button/index.tsx
```diff
@@ -0,0 +1,35 @@
+import { RemoveBookmarkIcon, AddBookmarkIcon } from './icons'
+import s from './bookmark-button.module.css'
+
+/**
+ * TODO
+ * - create a 'plugged' in HOC that can render modals, update bookmark state etc.
+ * This HOC will be used by this button and the one in tutorial meta
```
kendallstrautman, line 7: Note for follow up work.

src/components/bookmark-button/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+	if (!__config.flags.enable_auth) {
+		return null
+	}
+	const helpText = isBookmarked ? `Remove bookmark` : `Add bookmark`
+	return (
+		<button
+			aria-pressed={!!isBookmarked}
+			onClick={() => {
+				// TODO use the create / destroy methods in the client
+				// or render dialog to prompt auth if not auth'd
```
kendallstrautman, line 25: Another 'stub' function here. Perhaps later on this function will get lifted up and passed down to this component so it's purely visual.

src/components/collection-card/docs.mdx
```diff
@@ -10,6 +10,7 @@ With text heading:
 
 <LiveComponent>{`
 <CollectionCard
+	url="/"
```
kendallstrautman, line 13: This swingset page was erroring since this wasn't present.

src/components/tutorial-collection-cards/components/card-eyebrow/index.tsx
```diff
@@ -1,12 +1,15 @@
...
-		</span>
-	)
+function CardEyebrow({
+	children,
+	className,
+}: {
+	children: ReactElement[] | ReactElement
+	className?: string
+}) {
+	return <span className={classNames(s.eyebrow, className)}>{children}</span>
```
kendallstrautman, line 12: I refactored this to pass children. I think this approach will be helpful, especially as we start to have many eyebrow states with progress etc.

src/components/bookmark-button/icons/index.tsx
ashleemboyer: Ahhh I see now! Maybe rendering two icons based on the state and then hiding one with CSS will work?

```tsx
// If tutorial is not bookmarked
<button className={s.bookmarkButton}>
  <IconBookmark16 className={s.defaultStateIcon} />
  <IconBookmarkAdd16 className={s.hoverStateIcon} />
</button>
```

```tsx
// If tutorial is bookmarked
<button className={s.bookmarkButton}>
  <IconBookmarkFill16 className={s.defaultStateIcon} />
  <IconBookmarkRemoveFill16 className={s.hoverStateIcon} />
</button>
```

```css
.bookmarkButton {
  color: var(--token-color-foreground-faint);

  &:hover, &:focus {
    color: var(--token-color-foreground-strong);

    // Hide in hover state
    & .defaultStateIcon {
      display: none;
    }

    // Show in hover state
    & .hoverStateIcon {
      display: block;
    }
  }
}

// Show in default state
.defaultStateIcon {
  display: block;
}

// Hide in default state
.hoverStateIcon {
  display: none;
}
```

kendallstrautman: That's smart 😄 I'll try that out!

kendallstrautman: Made that update! thanks for the idea 😸

ashleemboyer:
Taking a look now! 😊

src/components/bookmark-button/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+interface BookmarkButtonProps {
+	isBookmarked: boolean
+}
+
+export default function BookmarkButton({ isBookmarked }: BookmarkButtonProps) {
+	const { isAuthEnabled } = useAuthentication()
+	// NOTE! - hiding this component from prod until auth is enabled
+	if (!isAuthEnabled) {
+		return null
+	}
```
ashleemboyer, line 20: Sorry for any confusion on this via Slack! I meant to suggest using the `AUTH_ENABLED` constant from `use-authentication`, which can be imported directly:

```
import { AUTH_ENABLED } from 'hooks/use-authentication'
```

Otherwise, invoking the `useAuthentication` hook will cause some unnecessary JavaScript to run here.

src/components/bookmark-button/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+	const helpText = isBookmarked ? `Remove bookmark` : `Add bookmark`
+	return (
+		<button
+			aria-pressed={!!isBookmarked}
```
ashleemboyer, line 24: Since `isBookmarked` is already a `boolean`, no need to `!!` here!

```suggestion
			aria-pressed={isBookmarked}
```

zchsh (approved):
✅ 🔖 Looks good to me!

ashleemboyer (approved):

PR #837, opened 2022-08-16
feat: add bookmark auth cta

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksadd-bookmark-auth-cta-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202520168081556/1202791375540708) 🎟️

## 🗒️ What

This PR adds the pop up dialog for unauthenticated users, prompting them to sign-in in order to save the bookmark. It also refactors things so that both the tutorial meta and bookmark card icon can use this dialog via an HOC called `withDialog`. We can decide later whether this HOC will also handle the data handling based on user interaction. 

## 📸 Design Screenshots

Screen Shot 2022-08-17 at 3 19 53 PM

## 🧪 Testing
- Visit a page with a [bookmark button](https://dev-portal-git-ksadd-bookmark-auth-cta-hashicorp.vercel.app/terraform/tutorials)
- Hit the button, you should see the sign in prompt dialog. 

To test the full flow, run this locally
- Do the same steps and you should be able to sign in. If you click the bookmark icon when signed in, nothing should happen. 

## 💭 Anything else?

This is just a stub and still needs to be integrated with the work in [#838](https://github.com/hashicorp/dev-portal/pull/838)

PR #838, opened 2022-08-16
Add bookmarks hooks

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambadd-bookmarks-hooks-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1202791375540709/f) 🎟️

## 🗒️ What

- Installs React Query and the Devtools
- Updates `_app.tsx` to work with React Query and the Devtools
- Adds a few new hooks:
  - `useAllBookmarks`
  - `useBookmarkMutations`
  - `useBookmarksByTutorialIds`
  - `useIsBookmarked`
- Updates `Tutorial['id']` to reference `ApiTutorial['id']`
- Updates some of the bookmark API methods to reference `Tutorial['id']` instead of `ApiTutorial['id']`
  - `createBookmark`
  - `deleteBookmark`
  - `getBookmark`
  - `getBookmarksByTutorialIds`
- Updates `ProfileView` to be a test page for rendering all bookmarks and some test tutorial cards
- Removes some no longer needed files for `ProfileView` (and they aren't needed in upcoming designs)

## 🧪 Testing

Throughout the steps below, have the React Query Devtools open and make sure the queries behave how you expect them to.

- [ ] Go to /profile
- [ ] In a card in the Tutorials section of the page, click an "add bookmark" button
- [ ] The Bookmarks list should automatically update
- [ ] The "add bookmark" button you clicked should automatically update


src/components/tutorial-card/types.ts
```diff
@@ -1,6 +1,12 @@
+import { ApiTutorial } from 'lib/learn-client/api/api-types'
 import { ProductOption } from 'lib/learn-client/types'
 
 export interface TutorialCardProps {
+	/**
+	 * The `id` of the Tutorial relevant to the card.
+	 */
+	id: ApiTutorial['id']
```
kendallstrautman, line 8: This would actually be `Tutorial['id]` from this file, https://github.com/hashicorp/dev-portal/blob/main/src/lib/learn-client/types.ts#L11 - since this data is formatted. Just a string so not that it matters much.

src/components/bookmark-button/index.tsx
```diff
@@ -1,33 +1,38 @@
...
 
 interface BookmarkButtonProps {
-	isBookmarked: boolean
+	tutorialId: ApiTutorial['id']
 }
 
-export default function BookmarkButton({ isBookmarked }: BookmarkButtonProps) {
+export default function BookmarkButton({ tutorialId }: BookmarkButtonProps) {
+	const { isBookmarked } = useIsBookmarked({ tutorialId })
+	const { addBookmark, removeBookmark } = useBookmarkMutations()
```
kendallstrautman, line 13: Do you mind if we do this integration in another PR? Just to avoid conflicts with some refactoring work I'm doing [in this PR](https://github.com/hashicorp/dev-portal/pull/837)

kendallstrautman: Once this is merged I can integrate these updates into that PR

ashleemboyer: Sure

src/hooks/bookmarks/use-all-bookmarks.ts
```diff
@@ -0,0 +1,32 @@
+import { useQuery, UseQueryResult } from '@tanstack/react-query'
+import { ApiBookmark } from 'lib/learn-client/api/api-types'
+import { getAllBookmarks } from 'lib/learn-client/api/bookmark'
+import useAuthentication from 'hooks/use-authentication'
+
+interface UseAllBookmarksResult extends Omit<UseQueryResult, 'data'> {
+	bookmarks: undefined | ApiBookmark[]
```
kendallstrautman, line 7: Would the data be 'undefined' if someone isn't authenticated? or what would cause that return

src/hooks/bookmarks/use-is-bookmarked.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+/**
+ * Handles checking if there is a bookmark for the given `tutorialId`.
+ */
+const useIsBookmarked = ({
+	tutorialId,
+}: UseIsBookmarkedOptions): UseIsBookmarkedResult => {
+	// Load up all bookmarks, since we know it primes queries & handles auth
+	const { bookmarks, isFetched: bookmarksAreFetched } = useAllBookmarks()
```
kendallstrautman, line 20: Hmmm I would expect us to just fetch once single tutorial here. Trying to think through this.

src/hooks/bookmarks/use-all-bookmarks.ts
ashleemboyer: The data is `undefined` while the data is being loaded for the first time. So in the case that someone is not authenticated, the query will never execute and the value will be `undefined`.

src/hooks/bookmarks/use-is-bookmarked.ts
ashleemboyer: Based on how this hook is intended to be used (once for each `TutorialCard` or `TutorialMeta`), that would require doing a fetch for each card/meta. So if there are 20 cards on a page, that means 20 fetches. This approach reduces it to one fetch in total.

src/components/tutorial-card/types.ts
ashleemboyer: Updated in https://github.com/hashicorp/dev-portal/pull/838/commits/f6816373e5e945e458540c30519570929b66f3cf. Also replaced `TutorialCardPropsWithId` since it's redundant with these overall changes.

src/components/bookmark-button/index.tsx
ashleemboyer: Actually on second thought, I'm not sure these changes should be undone. They are being used to test the new hooks.

src/hooks/bookmarks/use-bookmarks-by-tutorial-ids.ts
```diff
@@ -0,0 +1,42 @@
...
+
+interface UseBookmarksByTutorialIdsResult extends Omit<UseQueryResult, 'data'> {
+	bookmarks: undefined | ApiBookmark[]
+}
+
+/**
+ * Handles fetching and storing bookmarks with React Query for the given
+ * `tutorialIds`.
+ */
+const useBookmarksByTutorialIds = ({
```
kendallstrautman, line 18: I was considering that the page level view would fetch the specific set of tutorial bookmark data with this function. And then pass that down to the cards, instead of each card calling the hook. 🤔 Not that we have to do that but it was just my thought. 

It does seem a bit odd to me having every tutorial card calling the 'getAllBookmarks' hook. But maybe with the caching that's okay?....

src/components/bookmark-button/index.tsx
kendallstrautman: Once were done testing, we could undo? Or I can just manage the conflicts...

src/components/tutorial-card/types.ts
kendallstrautman: Cool, I think that change would also apply to other places where you're referencing `ApiTutorial` in this PR.

src/hooks/bookmarks/use-is-bookmarked.ts
kendallstrautman: Right, for that use case I figured that the bookmark component itself wouldn't call the `isBookmarked` hook, and that data would be passed down from higher. 

But just based on the naming of this hook, I would expect it to query just a single bookmark.

src/hooks/bookmarks/use-bookmarks-by-tutorial-ids.ts
ashleemboyer: The reason I didn't go with that idea right away is because it requires invoking this hook on every view we have TutorialCards. So far, that means product landing pages, tutorial landing pages, collection views, and install landing pages. The cache is being leveraged correctly here too! Even across several pages, all bookmarks are only being fetched one time. I think your idea is a good optimization to try out in a separate PR since it'll require refactors in a quite a few places and this PR is already doing a lot of refactoring.

src/hooks/bookmarks/use-is-bookmarked.ts
ashleemboyer: While we could pass data down, one of the primary purposes of using the hooks is so we can avoid props drilling. The hook does return the value one single bookmark, it just makes sure the data it needs to do so is already present. And reduces the number of fetches made as a side effect.

kendallstrautman: Well the tutorial meta only cares about one tutorial, not multiple. So the cards are where the multiple fetches come into play. Is there a way to prime a global cache in a way, such that if we make a call for a particular tutorial on say a product landing page, that data would be cached for the tutorial and collection page views? 

Or is this approach what that is intended to do in a way?

src/components/bookmark-button/index.tsx
ashleemboyer: I already had to deal with managing conflicts on this branch after spending many hours on it, so I'd appreciate it ❤️

src/components/tutorial-card/types.ts
ashleemboyer: Which places? Many of them are valid uses of that interface

ashleemboyer: Updated in https://github.com/hashicorp/dev-portal/pull/838/commits/54013f9f393855d971bede11304c57b6b6da75e2!

src/hooks/bookmarks/use-is-bookmarked.ts
ashleemboyer: The query priming approach wasn't working right away and needs some more time to explore, so that's another reason why optimizing this hook has been deferred for now. Once the individual query `bookmark` priming is worked out, then we can step `useIsBookmarked` toward no longer fetching all bookmarks to "prime" itself.

kendallstrautman: In dev mode, I'm seeing that every time a bookmark is created or deleted the 'bookmarks' endpoint is being called with this set-up. Having all bookmarks called when just one needs to be updated doesn't seem like the behavior we want.

What was happening with the individual priming? Should we wait to figure that out before finishing up these hooks?

ashleemboyer: Sorry if I wasn't clear! I want to overall optimize and also resolve the query priming in a separate PR since everything is working as needed here.

src/views/profile-view/index.tsx
```diff
@@ -1,58 +1,47 @@
...
-			<ProfileViewContent user={user} />
+			<ProfileViewContent />
 		</AuthenticatedView>
 	)
 }
 
 /**
  * The content of the ProfileView that is gated behind authentication.
  */
-const ProfileViewContent = ({ user }: ProfileViewProps) => {
```
kendallstrautman, line 24: We will still need the user data for this view. Is there a reason you removed it?

kendallstrautman: Actually let me check the designs again, maybe im getting wires crossed with Learn

kendallstrautman (approved):
Approving with the knowledge that optimization with the way were fetching data based on the threads will occur on follow-up PRs 👌

BRKalow (commented):
Exciting stuff! There are a few things I think we want to adjust here in how we're typing the query hooks, namely relying on the underlying `useQuery` types for inferring the type of `data` instead of overwriting that declaration.

We can also simplify the `useIsBookmarked` hook to select data directly from the result of `useAllBookmarks`, if that is the preferred implementation.

As we start to introduce data fetching, let's also consider how wiring up these components will impact our ability to test and document them. If we add queries to components we render in swingset, we'll likely need to setup the query plumbing there as well. Sometimes having a purely presentational component can be a valuable separation!

src/components/tutorial-card/types.ts
```diff
@@ -35,13 +41,4 @@ export interface TutorialCardProps {
...
 	productsUsed: ProductOption[]
-
-	isBookmarked?: boolean
-}
-
-export interface TutorialCardPropsWithId extends TutorialCardProps {
-	/**
-	 * The tutorial's unique identifier
-	 */
-	id: string
```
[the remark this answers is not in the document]
ashleemboyer: I undid these changes, so marking resolved.

src/components/bookmark-button/index.tsx
```diff
@@ -1,33 +1,38 @@
...
+import { useBookmarkMutations, useIsBookmarked } from 'hooks/bookmarks'
 import { AUTH_ENABLED } from 'hooks/use-authentication'
 import { RemoveBookmarkIcon, AddBookmarkIcon } from './icons'
 import s from './bookmark-button.module.css'
 
-/**
- * TODO
- * - create a 'plugged' in HOC that can render modals, update bookmark state etc.
- * This HOC will be used by this button and the one in tutorial meta
- */
```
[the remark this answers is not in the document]
ashleemboyer: Undid these changes, marking resolved.

src/hooks/bookmarks/use-all-bookmarks.ts
```diff
@@ -0,0 +1,32 @@
+import { useQuery, UseQueryResult } from '@tanstack/react-query'
+import { ApiBookmark } from 'lib/learn-client/api/api-types'
+import { getAllBookmarks } from 'lib/learn-client/api/bookmark'
+import useAuthentication from 'hooks/use-authentication'
+
+interface UseAllBookmarksResult extends Omit<UseQueryResult, 'data'> {
```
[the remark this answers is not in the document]
ashleemboyer: This is mainly just renaming `data` to `bookmarks` so the property has a clearer name that can make the hook consumption a little clearer. I can update it to reference `data` though, rather than remove + set our own.

src/hooks/bookmarks/use-all-bookmarks.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Handles fetching and storing all bookmarks with React Query.
+ */
+const useAllBookmarks = (): UseAllBookmarksResult => {
+	// Get the current user's access token
+	const { session } = useAuthentication()
+	const accessToken = session?.accessToken
+
+	// Fetch all bookmarks with the access token
+	const { data: bookmarks, ...restQueryResult } = useQuery(
```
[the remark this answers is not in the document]
ashleemboyer: It's still being inferred, this just renamed the `data` property to `bookmarks`.

src/hooks/bookmarks/use-bookmark-mutations.ts
```diff
@@ -0,0 +1,101 @@
...
+	 */
+	const onMutationSuccess = (
+		mutationResult: CreateBookmarkResult | DeleteBookmarkResult,
+		mutationVariables: CreateBookmarkOptions | DeleteBookmarkOptions
+	) => {
+		// Pull the modified bookmark's `tutorialId`
+		const mutatedTutorialId = mutationVariables.tutorialId
+
+		// `removeBookmark` returns `undefined`; we can tell which mutation occurred
+		const wasRemoveMutation = mutationResult === undefined
```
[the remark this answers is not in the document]
ashleemboyer: I can make this update.

src/hooks/bookmarks/use-bookmarks-by-tutorial-ids.ts
```diff
@@ -0,0 +1,43 @@
...
+	tutorialIds,
+}: UseBookmarksByTutorialIdsOptions): UseBookmarksByTutorialIdsResult => {
+	// Get the current user's access token
+	const { session } = useAuthentication()
+	const accessToken = session?.accessToken
+
+	// Fetch all bookmarks with the access token
+	const { data: bookmarks, ...restQueryResult } = useQuery(
+		['bookmarks'],
```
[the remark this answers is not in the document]
ashleemboyer: Ah good call, I forgot to come back to this. I paused on thinking about this query key because it's not extremely straightforward. We want to make sure this query gets invalidated on add/remove and I wasn't 100% sure (but will test) if the query you're proposing would get invalidated just based on the `"bookmarks"` portion.

src/hooks/bookmarks/use-is-bookmarked.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+/**
+ * Handles checking if there is a bookmark for the given `tutorialId`.
+ */
+const useIsBookmarked = ({
+	tutorialId,
+}: UseIsBookmarkedOptions): UseIsBookmarkedResult => {
+	// Load up all bookmarks
+	const { bookmarks, isFetched: bookmarksAreFetched } = useAllBookmarks()
```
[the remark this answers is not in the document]
ashleemboyer: I don't think the `select` option is what we want, because it'd cause the `find` to be invoked every time the hook renders. I wrapped the `find` in a `useQuery` so that wouldn't happen, and the value would be cached and accessible anywhere.

But I'll go back to the query priming and see if I can get that working as we want it to.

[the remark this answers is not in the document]
BRKalow: I don't expect that the usage of `find` here will have any meaningful performance impact in practice. If we were noticing performance issues, the selector can also be optimized to ensure it only runs when the underlying data changes, per the linked post!

I can see some benefit to the unique queries per bookmark though, considering we'll need to invalidate them on mutations. 💭

src/hooks/bookmarks/use-all-bookmarks.ts
```diff
@@ -0,0 +1,57 @@
...
+
+	// Get the current user's access token
+	const { session } = useAuthentication()
+	const accessToken = session?.accessToken
+
+	// Set up the `onSuccess` callback
+	const onSuccess = (data: QueryDataType) => {
+		// Prime each individual bookmark query by tutorial id
+		data.forEach((bookmark) => {
+			queryClient.setQueryData(['isBookmarked', bookmark.tutorial_id], true)
```
BRKalow, line 36: Nicee!

BRKalow (approved):
👏

PR #842, opened 2022-08-17
Adding support for src/content/ docs-landing.mdx files

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambadd-docs-landing-mdx-support-hashicorp.vercel.app/waypoint/docs) 🔎
- [Asana task](https://app.asana.com/0/1202702246158644/1202780150813240/f) 🎟️

## 🗒️ What

- Adds support for `docs-landing.mdx` files next to `docs-landing.json` files
- Updates relevant authoring docs
- Removes references to `includeMDXSource` (now programmatically determined on presence of MDX file)

## 🤷 Why

These changes enable authors to write longer form content in `/:productSlug/docs` landing pages, in a file local to this repository (rather than in each product repository). The decision to make this file local is so that authors can make changes that only affect `dev-portal`, rather than also affecting .io sites.

## 🧪 Testing

- [ ] Make sure [/waypoint/docs](https://dev-portal-git-ambadd-docs-landing-mdx-support-hashicorp.vercel.app/waypoint/docs) has not changed [compared to the prod version](https://developer.hashicorp.com/waypoint/docs)
- [ ] Make sure other /docs pages still work as expected, and unchanged


BRKalow:
@ashleemboyer just confirming, based on the ticket we want to support MDX authored here for product landing pages (`ProductLandingView`), but not the root docs landing pages, does that sound right? Root docs paths have authorable MDX via the source repositories.

zchsh:
☝️ +1 to Bryce's question - if this is for docs-landing, I've got 🎟️  [[terraform-docs] support paragraph text on docs landing to match design](https://app.asana.com/0/0/1202785022290112/f). I'm thinking I'll try to get the designs implemented by supporting a new, relatively simple paragraph "block" type (that only supports basic markdown, ie inline links and bold italic underline) in order to get the `/terraform/docs` page implemented:

CleanShot 2022-08-17 at 11 43 49@2x

Very open to input there. Definitely feels like something that, longer-term, we'd probably want to solve with a more robust MDX component system and `layout`-type stuff (maybe specific layouts would expose specific MDX custom components?). For now my intent is to try to go with the simplest next-step solution, but very open to adjusting course if we have any strong opinions on that front.

ashleemboyer:
@BRKalow @zchsh aw shucks! This is my bad. I was in /docs land in my head and wasn't paying close enough attention to the ticket title! 🤣 Will update shortly!

@zchsh re: /terraform/docs, we shouldn't need a new block since we already support MDX for that page. Maybe we just add a boolean on `RootDocsPath` that determines whether the MDX slot goes before or after the marketing content blocks? Let's all chat about that in Slack!

PR #854, opened 2022-08-18
feat: add Registry link to Terraform nav

zchsh:
## 🔗 Relevant links

- [Preview link][/terraform] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

- Adds a `Registry` link to the top nav on Terraform pages
- Adds a `Terraform Registry` link to the `Resources` section in sidebars on Terraform pages

## 🤷 Why

- Parity with <https://www.terraform.io/>, where we have a `Registry` link in the top bar
- Match design specs

## 🛠️ How

- Uses `@custom-media` to make the nav-header breakpoint from mobile-to-desktop more flexible
    - ... and tweaks it to avoid overflow, which is particularly bad for Terraform with the `Registry` item, but was also happening for other products in local dev due to the size of auth elements in the top nav
- Splits out a `get-nav-items` utility to have somewhere to relatively cleanly add the `Registry` item for Terraform
- Adds `data/terraform/additional-sidebar-resources.json` to handle the addition of a `Registry` link in the Resources section of the sidebar

#### Top nav

Frame 1

#### Sidebar resources

CleanShot 2022-08-18 at 21 19 30@2x

## 🧪 Testing

- [ ] Visit a Terraform page, such as [/terraform][/terraform]. The following should apply to _all_ Terraform pages:
    - The top navigation bar should contain a `Registry` item after the `Install` item
    - The top navigation bar should not experience overflow or sqooshed buttons at any breakpoint
    - The Terraform sidebar should contain a `Terraform Registry` item

- [ ] Visit a page under any other product, such as [/waypoint][/waypoint]. The following should apply to _all_ other product pages:
    - The top navigation bar should look and function as it did previously
    - The top navigation bar should not experience overflow or sqooshed buttons at any breakpoint
    - The `Resources` section of the sidebar should look and function as it did previously

## 💭 Anything else?

> **Note**: Pretty soon after I started on this task in detail, I realized it might warrant broader discussion. So the contents of this PR might not have the level of polish we want to merge to main; at this point I felt an early review might be worthwhile to see if there's anything fundamental we want to change about the underlying goal or design spec.

[preview]: https://dev-portal-git-zsterraform-add-registry-nav-link-hashicorp.vercel.app/
[task]: https://app.asana.com/0/1202097197789424/1202785022290103/f
[/terraform]: https://dev-portal-git-zsterraform-add-registry-nav-link-hashicorp.vercel.app/terraform
[/waypoint]: https://dev-portal-git-zsterraform-add-registry-nav-link-hashicorp.vercel.app/waypoint


src/components/navigation-header/components/home-page-content/index.tsx
```diff
@@ -54,7 +54,7 @@ const HomePageHeaderContent = () => {
 					src={require('../../img/logo-white.svg?include')}
 				/>
 			</div>
-			<div className="g-hide-on-mobile g-hide-on-tablet">
+			<div className={sharedNavStyles.leftSideDesktopOnlyContent}>
```
zchsh, line 57: In order to fit the Registry link, I've done a very light refactor allow us to use a more flexible breakpoint for switching between the "mobile" and "desktop" modes of `navigation-header`.

This className change is one part of that. In general, where necessary to achieve this type of flexibility, I've replaced uses of `g-hide-on-*` as those use fixed breakpoints.

---

> **Note**: 💭 This might not feel ideal. I think a key note here is that I think, long term, this type of flexibility will be essential if we want the site to genuinely look and function well at any viewport size (rather than have it be optimized only at specific, somewhat arbitrary "mobile", "tablet", and "desktop" breakpoints).

> 📌 Future thought: Even better, from my perspective, would be to find a way to have "dynamic breakpoints" - this way, menus would collapse when necessary to prevent overflow rather than at specific pixel width breakpoints. However, that type of enhancement feels well outside the scope of this PR.

src/components/navigation-header/components/primary-nav-link/index.tsx
```diff
@@ -1,29 +1,23 @@
...
 import s from './primary-nav-link.module.css'
 
 interface PrimaryNavLinkProps {
 	ariaLabel: string
 	navItem: {
-		id?: string
-		isSubmenu?: boolean
 		label: string
-		pathSuffix?: string
+		url: string
```
zchsh, line 10: We want to link to the Terraform Registry in the top nav (https://registry.terraform.io/). In order to achieve this, I've modified the interface and component slightly here to allow for full URLs, rather than strictly requiring a path suffix only.

The URL construction that previously happened inside this component now happens outside of it. Hopefully this aligns with our goal of separating logic and presentational concerns as well.

src/components/navigation-header/components/product-page-content/index.tsx
```diff
@@ -29,6 +29,7 @@ import {
 } from '..'
 import sharedNavStyles from '../../navigation-header.module.css'
 import s from './product-page-content.module.css'
+import { getNavItems } from './utils/get-nav-items'
```
zchsh, line 32: I've split out nav item generation to the `getNavItems` utility

src/components/navigation-header/components/product-page-content/utils/__tests__/get-nav-items.test.ts
```diff
@@ -0,0 +1,161 @@
+import waypointProductData from 'data/waypoint.json'
+import terraformProductData from 'data/terraform.json'
+import { getNavItems } from '../get-nav-items'
+import { ProductData } from 'types/products'
+
+describe('getNavItems', () => {
```
zchsh, line 6: I've added tests for `getNavItems` with inline snapshots. Happy to adjust the testing approach if we have something else in mind - for me personally, inline snapshots felt like a good balance given the return data.

src/components/navigation-header/components/product-page-content/utils/get-nav-items.ts
```diff
@@ -0,0 +1,57 @@
...
+		{
+			type: 'link',
+			label: 'Install',
+			url: `/${currentProduct.slug}/tutorials`,
+		},
+	]
+	/**
+	 * For Terraform, add a "Registry" item
+	 */
+	if (currentProduct.slug == 'terraform') {
```
zchsh, line 46: It felt a bit odd to push an extra link based on a product slug, rather than for example managing nav data in a JSON file as authorable content. But I think this aligns with our current intent of keeping as much consistency as possible between products.

src/components/navigation-header/navigation-header.module.css
```diff
@@ -1,3 +1,6 @@
+@custom-media --nav-header-is-mobile (max-width: 1200px);
+@custom-media --nav-header-is-desktop (min-width: 1201px);
```
zchsh, line 2: This is where the bulk of the relatively minor refactor for more breakpoint flexibility happens.

The focal point here is create new `@custom-media` and using that where we previously might've used our more set-in-stone `mobile`, `tablet`, and `desktop` breakpoints.

src/content/terraform/additional-sidebar-resources.json
```diff
@@ -0,0 +1,6 @@
+[
+	{
+		"title": "Terraform Registry",
+		"href": "https://registry.terraform.io/"
```
zchsh, line 4: In the design spec, this item is supposed to come _before_ other resource items.

However, given we'd already built this functionality, my intent here is to run this by design and see if it'd be fine to put the extra `Registry` item after other items, to avoid a refactor here.

> **Note**: 💭 If design strictly wants the item first, one approach that feels viable to me would be, instead of having optional `additional-sidebar-resources.json`, have an optional `sidebar-resources.json`. We'd then either use that authored set of resources if present, or fall back to a default set like we're currently generating. But we would not try to mix authored resources with fallback defaults.

zchsh:
@ashleemboyer @BRKalow This PR adds the Registry item to Terraform's top nav, and to Terraform's sidebar resources.

> ⚠️ Broad note here: I've spiked this in, but in requesting review, am not necessarily saying this is the direction we should take. Very open to discussion as to how we should move forward, whether that's trying to finalize this PR, and taking some of the compromises here back to design for further review.

The changes I made to make this work in a way that felt acceptable:
- Made the breakpoint for nav-header going mobile more flexible, so that we can tweak it to accommodate items (see note below)
- Split out `get-nav-items`, which generates top nav items for a product. The change in functionality here is adding a `Registry` item if the product slug is `terraform`. This felt a bit odd, but felt better than trying to mix authored and generated content. Very open to alternate approaches to achieving this (or as mentioned, also open to thinking about whether we want to achieve this).

---

### Note on changed breakpoint

The change in breakpoint for mobile-to-desktop nav-header seems significant enough to warrant some extra context. I'd started trying to meet the design spec to add the `Registry` item, and realized there were complications to responsive behaviour:

terraform-registry-in-nav-issue

Looking into it further, I noticed we'll have these issues anyways on other products once authentication is enabled, for example Waypoint and Vault:

waypoint-top-nav-issue

vault-top-nav-issue

With this in mind, I figured even if we don't want to add `Registry` to the top nav (eg if it feels too inconsistent), we'll very likely want the flexibility to adjust the nav-header mobile-to-desktop breakpoint to something other than our fixed `mobile`, `tablet`, and `desktop` breakpoints. Definitely something to run by design, of course (maybe also points to the benefit of investing more in responsive mockups!)

src/content/terraform/additional-sidebar-resources.json
ashleemboyer: `additional-sidebar-resources.json` was a product-driven decision and. See this Slack thread for context: https://hashicorp.slack.com/archives/C03A8CW77FU/p1659577575329129. We should get design to update the Figma to reflect the decision that was made in that thread.

src/components/navigation-header/components/product-page-content/index.tsx
```diff
@@ -141,34 +142,17 @@ const ProductPageHeaderContent = () => {
...
 								const ariaLabel = `${currentProduct.name} ${label}`
 
 								let ItemContent
-								if (items) {
+								if (type == 'submenu') {
 									ItemContent = PrimaryNavSubmenu
-								} else {
+								} else if (type == 'link') {
 									ItemContent = PrimaryNavLink
 								}
```
ashleemboyer, line 157: The `type` abstraction doesn't feel super necessary here.

src/components/navigation-header/components/product-page-content/utils/get-nav-items.ts
```diff
@@ -0,0 +1,57 @@
...
+		},
+		{
+			type: 'link',
+			label: 'Tutorials',
+			url: `/${currentProduct.slug}/tutorials`,
+		},
+		{
+			type: 'link',
+			label: 'Install',
+			url: `/${currentProduct.slug}/tutorials`,
```
ashleemboyer, line 40: 🚧 This should link to `downloads`

src/components/navigation-header/components/product-page-content/utils/types.ts
```diff
@@ -0,0 +1,17 @@
...
+	label: string
+	url: string
+}
+
+export interface NavItemSubmenu {
+	type: 'submenu'
+	label: string
+	iconColorTheme: ProductSlug
+	items: NavigationHeaderItem[]
+}
```
ashleemboyer, line 15: These types are already defined in `PrimaryNavSubmenu` and `PrimaryNavLink`. I think they should be referenced here as the source of truth.

src/components/navigation-header/components/product-page-content/utils/get-nav-items.ts
```diff
@@ -0,0 +1,57 @@
...
+	]
+	/**
+	 * For Terraform, add a "Registry" item
+	 */
+	if (currentProduct.slug == 'terraform') {
+		items.push({
+			type: 'link',
+			label: 'Registry',
+			url: 'https://registry.terraform.io/',
+		})
```
ashleemboyer, line 51: I think we want to open this link in a new tab since it's external to the site. That's how we're handling the link for this item in the Sidebar Resources section as well.

zchsh: Sweet, can do! ✅ Have updated to support an `openInNewTab` property here, and have used it 👍

src/components/navigation-header/components/product-page-content/utils/types.ts
zchsh: Sounds good - have updated this file to reference those types 👍 Let me know if that works!

src/components/navigation-header/components/product-page-content/utils/get-nav-items.ts
zchsh: 🙌 Great catch, ty! ✅ Updated.

src/components/navigation-header/components/product-page-content/index.tsx
zchsh: Sounds good - have removed 👍 I found I couldn't use `items` directly without a type warning, so have used `navItem.hasOwnProperty('items')`.

zchsh:
@ashleemboyer Thanks for the excellent feedback 💖 Have done my best to address it, let me know what you think 🙇‍♂️

ashleemboyer (approved):

PR #856, opened 2022-08-19
[hotfix] Fixing CardLink <a> order and spacing issue

ashleemboyer:
## What

Fixes a spacing issue at the top of some CardLink elements.

Was introduced when `{children}` was moved after `<Link>` to fix the TAB order of elements. Have created a follow-up ticket to fix the TAB order in a different way: https://app.asana.com/0/1202097197789424/1202822098515463/f.

## Testing

- [ ] Go to /vault
- [ ] Make sure there is no extra spacing at the top of cards

### DIFF
--- a/src/components/card-link/index.tsx
+++ b/src/components/card-link/index.tsx
@@ -30,6 +30,7 @@ const CardLink = ({
 
 	return (
 		<Card className={classes}>
+			{children}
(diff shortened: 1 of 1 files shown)


BRKalow (approved):
👍 makes sense

PR #849, opened 2022-08-17
feat: support custom docs nav items

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Updates the icons, text, and in some cases links for Terraform's `Documentation` nav items.

## 🤷 Why

To match our approved design spec.

## 🛠️ How

- Adds `lib/docs/get-docs-nav-items`
    - Intent is to handle setting up docs nav items for all products
- Re-uses the above utility in navigation-header and sidebar
- Updates `supported-icons` in nav-header to support options for Terraform
- Updates `terraform.json` with text and icons to match designs

### Context

terraform - update text and icons in top nav and sidebar

### Screenshot from this PR's preview

CleanShot 2022-08-17 at 17 48 27@2x

## 🧪 Testing

- [ ] Visit a Terraform page, such as [/terraform][/terraform]
    - Ensure the `Documentation` dropdown looks and functions as expected
- [ ] Visit another product page, such as [/waypoint][/waypoint]
    - Other products' `Documentation` dropdowns should function as previously, without regression

## 💭 Anything else?

> **Note**: the two-column layout for the dropdown is intentionally not included in this PR. It will be handled in a separate chunk of work: 🎟️  [[terraform] implement two-columns for Docs dropdown in top nav](https://app.asana.com/0/1202097197789424/1202765004267093/f)

[task]: https://app.asana.com/0/1202097197789424/1202691941299957/f
[preview]: https://dev-portal-git-zsupdate-docs-category-icons-and-text-hashicorp.vercel.app/
[/terraform]: https://dev-portal-git-zsupdate-docs-category-icons-and-text-hashicorp.vercel.app/terraform
[/waypoint]: https://dev-portal-git-zsupdate-docs-category-icons-and-text-hashicorp.vercel.app/waypoint

PR #857, opened 2022-08-19
fix: hide icon cards on /terraform/docs

zchsh:
## 🔗 Relevant links

- [Preview link][/terraform/docs] 🔎
- [Asana task](url) 🎟️

## 🗒️ What

Hides icon cards on [/terraform/docs][/terraform/docs].

## 🤷 Why

To match design specs.

## 🛠️ How

- Does not render icon cards if `currentProduct.slug == "terraform"`
- Adjusts spacing logic to ensure spacing is acceptable whether or not icon cards are rendered

### Context

terraform-docs  remove icon cards, not shown in design

### Screenshot from this PR

CleanShot 2022-08-19 at 06 28 06@2x

## 🧪 Testing

- [ ] Visit [/terraform/docs][/terraform/docs]. Icon cards should not be visible.
- [ ] Visit any other product docs landing, such as [/vault/docs][/vault/docs]. Icon cards should be visible.

[task]: https://app.asana.com/0/1202097197789424/1202785022290109/f
[/terraform/docs]: https://dev-portal-git-zsterraform-hide-docs-landing-i-1b050c-hashicorp.vercel.app/terraform/docs
[/vault/docs]: https://dev-portal-git-zsterraform-hide-docs-landing-i-1b050c-hashicorp.vercel.app/vault/docs


src/views/product-root-docs-path-landing/components/hero/hero.module.css
```diff
@@ -1,13 +1,11 @@
 .root {
-	margin-bottom: 24px;
-	padding-bottom: 24px;
+	padding-bottom: 48px;
```
zchsh, line 2: I've collected the previous combined margin + padding into one metric.

src/views/product-root-docs-path-landing/components/hero/hero.module.css
```diff
@@ -27,3 +25,8 @@
 .pageSubtitle {
 	color: var(--token-color-foreground-primary);
 }
+
+.iconCardGridLink {
+	margin-top: 32px;
```
zchsh, line 30: Moved `margin-bottom` off of `.iconTextAndWrapper` and onto `.iconCardGridLink`. This way, when `.iconCardGridLink` isn't present, we don't end up creating extra space.

src/views/product-root-docs-path-landing/components/hero/index.tsx
```diff
@@ -34,7 +34,11 @@ const ProductRootDocsPathLandingHero = ({
 					<Text className={s.pageSubtitle}>{pageSubtitle}</Text>
 				</div>
 			</div>
-			<ProductRootDocsPathLandingIconCardLinkGrid />
+			{currentProduct.slug !== 'terraform' ? (
```
zchsh, line 37: Currently the design spec shows no icon cards on the `/terraform/docs` landing view, so that's what I'm aiming for. Chatting with Brayden, it sounds likely that these will get added back in for Terraform in some capacity in the future, but we don't yet have a design spec for which cards to add.

zchsh:
@ashleemboyer @braydenlove @BRKalow This PR removes the "icon cards" from the `/terraform/docs` landing page.

Intent here is to match design specs in the most straightforward way possible. Have added some more context below in case it's helpful.

---

## Extra context on this change

I suspect in the future we'll want to add at least some of these icon cards back in some capacity. That being said, I think finding a good solution here might require relatively significant design involvement.

For example, I considered the possibility of using something the`get-nav-items` from #849 to generate these icon cards, rather than strictly showing icon cards. But can see why this might not be the greatest UX solution when there are many nav items, as we end up with very duplicative navigation elements:

CleanShot 2022-08-18 at 21 42 00@2x

With this in mind, intent here is to stick with the approved mockups for now, even if that means diverging from our system a bit, in order to deliver on our current approved design specs. Happy to adjust the approach here if there are any strong opinions 👍

ashleemboyer (approved):

src/views/product-root-docs-path-landing/components/hero/index.tsx
BRKalow: thought 💭 : Do we want to make this configurable via the `docs-landing.json` content file? I'm fine with either approach, just curious if you considered that!

zchsh: 👆 I think it could be a great option - I hadn't considered that specifically, but did consider maybe using `get-nav-items` like we do in #849 . I ultimately leaned away from it, since it feels like hiding these items is probably a temporary thing, so maybe better to handle it with a one-off conditional rather than introducing a new facet of authorable content. Very open to reconsidering within this PR though, I do like the specificity of putting something in `docs-landing.json`, and I suppose for now we'd do like an empty array to match the design spec? 🤔

BRKalow: @zchsh Yeah I was even thinking just a toggle, like `hideCardLinks: true` or something. Also totally fine with leaving it as you have it if it feels like this will be temporary 😄

BRKalow (approved):

braydenlove (approved):
Thanks @zchsh, this is much cleaner and we still have all the links in dropdown + sidebar. We will circle back later when we have the time to figure out what we actually want to show here.

PR #858, opened 2022-08-19
feat: support two column docs dropdowns

zchsh:
## 🔗 Relevant links

- [Preview link][/terraform] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Updates dropdowns so that Terraform's `Documentation` dropdown renders in two columns.

## 🤷 Why

To match design specs.

## 🛠️ How

Updates our top nav dropdown logic to render in two columns when there are many items:
- Adds an optional CSS class that will to render item groups in two columns
- Conditionally add that class if the item group contains `>= 10` items

### Context

terraform - update text and icons in top nav and sidebar

### Screenshot from this PR

> **Note**: icon and text updates are handled in a separate PR, #849. This PR focuses solely on getting the two-column layout working. As such, the screenshot below intentionally does _not_ have updated icons and text yet. 

CleanShot 2022-08-19 at 06 56 13@2x

## 🧪 Testing

- [ ] Visit a Terraform page, such as [/terraform][/terraform]
    - The `Documentation` dropdown in the top nav should display in two columns
- [ ] Visit any other product page, such as [/vault][/vault]
    - The `Documentation` dropdown in the top nav should be in one column, as it already has been
    - Note: technically the distinguishing factor here is on the # of items in the dropdown, not on the product slug. If we do end up with another product with `>= 10` items in an `itemGroup` in a top nav dropdown, we'll see that dropdown section go to two columns as well.
- [ ] Other dropdowns, such as the product dropdown, should function as expected
    - Note that the product dropdown, with all products in beta, will have 9 items. So should still be in one column
    - If we add a tenth item, the dropdown will go to two columns. See the `Anything else` section below for details

## 💭 Anything else?

🚨 **This change is intentionally _not_ targeted specifically to `Terraform`** 🚨 

As a result, _if_ we end up with `>= 10` items in any item group in a dropdown, or _if_ we were to change the threshold from `10` to some lower integer, we would see two-column dropdowns in other dropdowns, as seen in these speculative examples:

<details>
<summary>Not this PR, but maybe relevant: threshold set to 9, and all products in beta</summary>

CleanShot 2022-08-19 at 06 40 24@2x

</details>

<details>
<summary>Not this PR, but maybe relevant: threshold set to 5, and mixed beta products</summary>

CleanShot 2022-08-19 at 06 41 04@2x

CleanShot 2022-08-19 at 06 41 27@2x

</details>

> **Note**: It is certainly possible to adjust the logic to only target the Terraform docs dropdown, if that is the intent. It felt more consistent and maintainable to not make this product distinction. I think the two-column layout looks great in the speculative examples above, so that definitely influenced this call. Very open to other perspectives, would love any thoughts 🙇‍♂️

[task]: https://app.asana.com/0/1202097197789424/1202765004267093/f
[/terraform]: https://dev-portal-git-zsterraform-two-column-docs-dropdown-hashicorp.vercel.app/terraform
[/vault]: https://dev-portal-git-zsterraform-two-column-docs-dropdown-hashicorp.vercel.app/vault


src/components/navigation-header/components/dropdown-menu/dropdown-menu.module.css
```diff
@@ -1,146 +1,152 @@
...
-  padding: 0;
+	list-style: none;
+	margin: 0;
+	padding: 0;
+
+	&.twoColumns {
+		column-count: 2;
+		width: max-content;
+		gap: 24px;
+	}
```
zchsh, line 80: This is the meaningful change in this file - the styles for our two-column layout option. The rest is formatting.

src/components/navigation-header/components/dropdown-menu/index.tsx
```diff
@@ -243,7 +243,12 @@ const NavigationHeaderDropdownMenu = ({
 										{label}
 									</Text>
 								)}
-								<ul aria-labelledby={itemGroupLabelId} className={s.itemGroup}>
+								<ul
+									aria-labelledby={itemGroupLabelId}
+									className={classNames(s.itemGroup, {
+										[s.twoColumns]: items.length >= 10,
```
zchsh, line 249: We switch to a two-column drop-down layout when there are 10 or more items.

> **Note**: 10 is a somewhat arbitrary number here, I'm going off the fact that Terraform has this many items in the dropdown currently. We can of course tweak this threshold if there's another number that would feel better.

zchsh:
@ashleemboyer @BRKalow @braydenlove This PR implements logic for two-column dropdowns, in order to meet our design spec for Terraform.

🚨 I want to specifically call out the `Anything else` section in the PR description. In brief:
- The logic here is _not_ specific to the Terraform Documentation dropdown
- It _could_ apply to any dropdown with a large number of items in a particular section of the dropdown.

ashleemboyer (approved):

BRKalow:
I think it make sense to have this logic apply to the menu across the entire site, not just Terraform 👍 Thanks for the context and additional background!

braydenlove (approved):
I dig this and appreciate the flexibility! We think these dropdowns could be really impactful spots for future content so I love the fact that we can make these 2 column across the site <3! Next time we chat remind me to show you some future concept work.

PR #862, opened 2022-08-19
Integrate bookmarks UI

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambintegrate-bookmarks-ui-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/0/1202822591039666/f) 🎟️

## 🗒️ What

- Updates bookmark buttons to have `z-index: 1`, since `CardLink` had to be rolled back to render its `Link` after its `children`
- Updates `src/components/mobile-menu-container/mobile-menu-container.module.css` to have `z-index: 2` so it is above the buttons that are now `z-index: 1`
- Adds `options` parameters to `addBookmark` and `removeBookmark` exposed by `usesBookmarkMutations`
- Updates `withDialog`:
  - to accept a `tutorialId` rather than `isBookmarked`, and has it leverage `useIsBookmarked` instead
  - invoke `addBookmark` and `removeBookmark` when appropriate
  - leverage the `onSuccess` callback for add/remove bookmark for showing bookmark toasts
- Updates `TutorialCardBookmarkButton` to accept `tutorialId` (instead of `IsBookmarked)`
- Adds `tutorialId` prop to `TutorialMeta` (so it can pass it to its bookmark button)
  - Updates references in: `TutorialView` and `WellArchitectedFrameworkTutorialView`
- Removes `isBookmarked` reference in `src/components/tutorial-card/docs.mdx`

## 🤷 Why

I thought about leaving `withDialog` as is, and passing a `addOrRemoveBookmark` function. That would have required passing the function everywhere `TutorialCardBookmarkButton` and `TutorialMetaBookmarkButton` are rendered, which would have required abstracting a shareable function with `isBookmarked` and `tutorialId` parameters. That would have required both buttons to accept `isBookmarked` and `tutorialId` props so that they could be passed along.

So instead of adding multiple props to multiple components, I decided to keep them as minimal as possible and only add the `tutorialId` prop which can be used to derive `isBookmarked`. I figure we can try to abstract `addOrRemoveBookmark` later if we find we need to. My major goals here were to make/keep as many things presentational as possible, while also reusing as much code as possible.

## 🛠️ How

Used the sweet, new `ConnectedTutorialCard` and `withDialog` helpers, and also updated `addBookmark` and `removeBookmark` so we can leverage some callbacks React Query exposes for `mutate` calls.

## 🧪 Testing

- [ ] While unauthenticated, try bookmarking a tutorial from a tutorial card
- [ ] You should see the sign in dialog pop up
- [ ] While unauthenticated, try bookmarking a tutorial from a tutorial page
- [ ] You should see the sign in dialog pop up
- [ ] Sign in
- [ ] Make sure you can **add** a bookmark from a tutorial card
- [ ] Make sure you can **remove** a bookmark from a tutorial card
- [ ] Make sure you can **add** a bookmark from a tutorial page
- [ ] Make sure you can **remove** a bookmark from a tutorial page

## 💭 Anything else?

This implementation does not yet leverage `useBookmarksByTutorialIds`, which primes individual `isBookmarked` queries. Kendall and I chatted about this today and we decided to save that for the PR that comes after this one. It will require using that hook at the View level, which means we may need to do some refactoring in those views to get `tutorialIds` to pass to `useBookmarksByTutorialIds`.

### DIFF
--- a/src/components/bookmark-button/bookmark-button.module.css
+++ b/src/components/bookmark-button/bookmark-button.module.css
@@ -3,16 +3,17 @@
 	--hoverIconOpacity: 0;
 	--baseIconOpacity: 1;
 
-	color: var(--token-color-foreground-faint);
-	position: relative;
-	display: flex;
 	align-items: center;
-	cursor: pointer;
 	background: none;
 	border: none;
-	padding: 0;
-	width: var(--buttonSize);
+	color: var(--token-color-foreground-faint);
+	cursor: pointer;
+	display: flex;
 	height: var(--buttonSize);
+	padding: 0;
+	position: relative;
+	width: var(--buttonSize);
+	z-index: 1;
 
 	@media (prefers-reduced-motion: no-preference) {
 		transition: color 0.175s ease;
@@ -28,13 +29,13 @@
 
 	&:focus-visible::before {
 		border-radius: 5px;
-		right: -1px;
-		top: -1px;
 		bottom: -1px;
-		left: -1px;
 		box-shadow: var(--token-focus-ring-action-box-shadow);
 		content: '';
+		left: -1px;
 		position: absolute;
+		right: -1px;
+		top: -1px;
 	}
 
 	&:focus,
--- a/src/components/bookmark-button/helpers/with-dialog.tsx
+++ b/src/components/bookmark-button/helpers/with-dialog.tsx
@@ -1,10 +1,11 @@
 import { useCallback, useState } from 'react'
-import Dialog from 'components/dialog'
+import { Tutorial } from 'lib/learn-client/types'
 import useAuthentication from 'hooks/use-authentication'
+import { useBookmarkMutations, useIsBookmarked } from 'hooks/bookmarks'
+import Dialog from 'components/dialog'
 import { BookmarkButtonProps } from '../types'
 import BookmarkSignInPrompt from '../sign-in-dialog'
-import { useBookmarkMutations, useIsBookmarked } from 'hooks/bookmarks'
-import makeBookmarkToast from 'components/bookmark-button/toast/make-bookmark-toast'
+import makeBookmarkToast from '../toast/make-bookmark-toast'
 
 /**
  * This HOC serves as a 'controller'
@@ -17,7 +18,7 @@ export function withDialog(BookmarkComponent: React.FC<BookmarkButtonProps>) {
 	return function BookmarkComponentWithDialog({
 		tutorialId,
 	}: {
-		tutorialId: BookmarkButtonProps['tutorialId']
+		tutorialId: Tutorial['id']
 	}) {
 		const { isAuthenticated, signIn } = useAuthentication()
 		const { isBookmarked } = useIsBookmarked({ tutorialId })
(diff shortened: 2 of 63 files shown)


src/components/bookmark-button/bookmark-button.module.css
```diff
@@ -3,16 +3,17 @@
...
 	border: none;
+	color: var(--token-color-foreground-faint);
+	cursor: pointer;
+	display: flex;
+	height: var(--buttonSize);
 	padding: 0;
+	position: relative;
 	width: var(--buttonSize);
-	height: var(--buttonSize);
+	z-index: 1;
```
ashleemboyer, line 16: 🗒️ This was needed to place the button on top of `CardLink`'s `a::before` element.

src/components/mobile-menu-container/mobile-menu-container.module.css
```diff
@@ -18,7 +18,7 @@
 	position: fixed;
 	top: var(--sticky-bars-height);
 	width: 100%;
-	z-index: 1;
+	z-index: 2;
```
ashleemboyer, line 21: 🗒️ This was needed to place the mobile menu container on top of bookmark buttons (now `z-index: 1`).

src/components/bookmark-button/helpers/with-dialog.tsx
```diff
@@ -1,66 +1,65 @@
...
 	return function BookmarkComponentWithDialog({
-		isBookmarked,
-	}: Pick<BookmarkButtonProps, 'isBookmarked'>) {
+		tutorialId,
+	}: {
+		tutorialId: Tutorial['id']
+	}) {
 		const { isAuthenticated, signIn } = useAuthentication()
+		const { isBookmarked } = useIsBookmarked({ tutorialId })
+		const { addBookmark, removeBookmark } = useBookmarkMutations()
```
kendallstrautman, line 25: So I think we should lift this up to `ConnectedTutorialCard` and that can serve as the 'controller'. That controller could also pass down the `addOrRemoveBookmark` handler. And we would only have to handle the `tutorialId` at that level instead of passing it down. Thoughts?

kendallstrautman: If we take that approach, this component will have the sole responsibility of _render a dialog or not_. And the `Connected` component will have the responsibility of _fetch the dynamic data and decide what to render in the eyebrow_.

src/components/tutorial-card/connected-tutorial-card.tsx
```diff
@@ -4,9 +4,8 @@ import TutorialCard from '.'
...
 
 export function ConnectedTutorialCard(props: TutorialCardPropsWithId) {
-	/**
-	 * const isBookmarked = useIsBookmarked(props.id)
-	 */
+	const { id: tutorialId } = props
+
 	return (
 		<TutorialCard
 			{...props}
```
kendallstrautman, line 11:
```suggestion
	const { id: tutorialId, ...rest } = props

	return (
		<TutorialCard
			{...rest}
```
Since we don't need to pass the id to the tutorial card 👍

kendallstrautman (approved):
LGTM! I can follow-up [re: this comment](https://github.com/hashicorp/dev-portal/pull/862/files#r950492892) in another PR 😄👏

PR #841, opened 2022-08-17
[ProductDownloadsView] Restricting "Featured Tutorials" section to just tutorials

ashleemboyer:
## 🔗 Relevant links

- [Vault Preview link](https://dev-portal-git-ambrestrict-featured-tutorials-section-hashicorp.vercel.app/vault/downloads) 🔎
- [Waypoint Preview link](https://dev-portal-git-ambrestrict-featured-tutorials-section-hashicorp.vercel.app/waypoint/downloads) 🔎
- [Asana task](https://app.asana.com/0/1202702246158644/1202780278912298/f) 🎟️

## 🗒️ What

- Updates the Feature Tutorials section of the downloads views to only show tutorials
- Updates the Vault and Waypoint sections to have 6 tutorials

## 🤷 Why

- We do not want to mix and match types of cards, especially agter bookmarking and progress are implemented.

## 🧪 Testing

- [ ] Go to [/vault/downloads](https://dev-portal-git-ambrestrict-featured-tutorials-section-hashicorp.vercel.app/vault/downloads)
- [ ] Make sure the Featured Tutorials content is OK
- [ ] Go to [/waypoint/downloads](https://dev-portal-git-ambrestrict-featured-tutorials-section-hashicorp.vercel.app/waypoint/downloads)
- [ ] Make sure the Featured Tutorials content is OK

PR #865, opened 2022-08-19
Leverage useBookmarksByTutorialIds, add new TutorialCardsGridList

ashleemboyer:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/0/1202822098515485/f) 🎟️

## 🗒️ What

- Adds new `TutorialCardsGridList` component that renders tutorial cards using `CardsGridList` and pre-fetches `isBookmarked` state for the tutorials in the list
- Updates all `CardsGridList`s that rendered tutorial cards to leverage `TutorialCardsGridList`

## 🤷 Why

- The logic and code comments for the following was a bit long to put in 4 separate places:
  - pulling tutorial ids from an array of data _and_ building a React element render
  - then invoking `useBookmarksByTutorialId`
  - then rendering a loading state based on `isFetching` and `isRefetching`
  - then rendering the `CardsGridList`
- This lets approach gives us a lower maintenance and consistent rendering state for code that was behaving exactly the same in 4 places.

## 🧪 Testing

- [ ] Make sure you are not authenticated
- [ ] There should be no visual changes on the affected pages:
  - [ ] Product landing ([/waypoint](https://dev-portal-git-ambprime-bookmark-queries-in-views-hashicorp.vercel.app/waypoint))
  - [ ] Install landing ([/waypoint/downloads](https://dev-portal-git-ambprime-bookmark-queries-in-views-hashicorp.vercel.app/waypoint/downloads))
  - [ ] Product tutorials landing ([/waypoint/tutorials](https://dev-portal-git-ambprime-bookmark-queries-in-views-hashicorp.vercel.app/waypoint/tutorials))
  - [ ] Collection view ([/waypoint/tutorials/get-started-docker](https://dev-portal-git-ambprime-bookmark-queries-in-views-hashicorp.vercel.app/waypoint/tutorials/get-started-docker))
- [ ] Sign in
- [ ] There should no longer be a GET request for every bookmark on the following pages, just one single `bookmarks?tutorialIds=...,...` GET request:
  - [ ] Product landing (/waypoint)
  - [ ] Install landing (/waypoint/downloads)
  - [ ] Product tutorials landing (/waypoint/tutorials)
  - [ ] Collection view (/waypoint/tutorials/get-started-docker)


src/hooks/bookmarks/use-all-bookmarks.ts
```diff
@@ -20,7 +20,7 @@ interface UseAllBookmarksResult
  * Handles fetching and storing all bookmarks with React Query.
  */
 const useAllBookmarks = ({
-	enabled,
+	enabled = true,
```
kendallstrautman, line 23: Should this be connected to the authentication state?

src/components/cards-grid-list/components/tutorial-cards-grid-list/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+	 * We also do not want to block this component from rendering content, so that
+	 * is why `isLoading` is not being used.
+	 *
+	 * The bookmark queries are disabled
+	 * when is auth is disabled or no user is authenticated. When queries are
+	 * `disabled`, `isLoading` is `true` forever.
+	 */
+	const isFirstLoad = isFetching && !isRefetching
+	if (isFirstLoad) {
+		return null
```
kendallstrautman, line 53: 💭 Ideally we hook up the loading state to the skeleton in the future.

src/views/product-downloads-view/components/featured-tutorials-section/index.tsx
```diff
@@ -24,24 +22,7 @@ const FeaturedTutorialsSection = ({
 			>
 				Featured Tutorials
 			</Heading>
-			<CardsGridList>
-				{featuredLearnCards.map((cardProps: FeaturedLearnCard) => {
-					const { id, type } = cardProps
-					if (type == 'collection') {
```
kendallstrautman, line 30: I think we want to retain this one because of the collection cards?

src/components/cards-grid-list/components/tutorial-cards-grid-list/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	return <CardsGridList {...restProps}>{cardsGridListItems}</CardsGridList>
```
kendallstrautman, line 56: 💭 Here's a though. It could be cool to have this just wrap the children so the rendering of the list items could remain flexible? 
```
<CardGridList>
    <PrimedTutorialBookmarks tutorials={tutorials}>
        {tutorials.map(....
           <li><ConnectedTutorialCard/></li>
         }   
     </PrimedTutorialBookmarks>
  </ CardGridList>
```
This way, if on a certain page we have a different grid layout or like in the case of `src/views/product-downloads-view/components/featured-tutorials-section/index.tsx`, where we render both collections and tutorials as children, it can be more easily handled. 

Just thinking out-loud. This is also I think a fine thing to package since it is repeated throughout many views.

src/hooks/bookmarks/use-all-bookmarks.ts
ashleemboyer: It is! 😊 The `enabled` option passed to `useQuery` below is set using this new option and `accessToken` retrieved from `useAuthentication`.

src/views/product-downloads-view/components/featured-tutorials-section/index.tsx
ashleemboyer: This section no longer renders collection cards as of #841! 😊

kendallstrautman: Oh okay! It looks like `FeaturedLearnCard` still shows that the[ collection type ](https://github.com/hashicorp/dev-portal/blob/main/src/views/product-downloads-view/types.ts#L40-L42)is acceptable. Maybe we should make a follow up task to update that type?

kendallstrautman (approved):
A big improvement for sure! Perhaps in the future we could abstract further so that we only make one network call for this data per view. 

When we tie this bulk call to the component, then for example on some of the product tutorial landing pages may render multiple of these grid card lists. So if we separated this into a function that just made the call for all the tutorial ids on the page and wasn't responsible for rendering, that could be a future optimization!

ashleemboyer:
> A big improvement for sure! Perhaps in the future we could abstract further so that we only make one network call for this data per view.
> 
> When we tie this bulk call to the component, then for example on some of the product tutorial landing pages may render multiple of these grid card lists. So if we separated this into a function that just made the call for all the tutorial ids on the page and wasn't responsible for rendering, that could be a future optimization!

I'll keep thinking on this! I liked your idea of a `PrimedTutorialBookmarks `. One thing I wanted to watch out for was how many times the data is iterated over. So rather than have one component iterate over data to get tutorial IDs, then another one mapping over the same data to render `<li>` elements, I went with one that does both. But this does only end up applying to these components!

PR #869, opened 2022-08-22
 feat: user profile bookmarks stub

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksuser-profile-stub-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1199634971449915/1202841857338448) 🎟️

## 🗒️ What

This PR stubs out the main content of the user profile bookmarks page. 

It renames `/profile` 👉 `/profile/bookmarks`, since there will eventually be a proper base profile page. All bookmarks are fetched and rendered in a grid without pagination. 

## 🧪 Testing

- Pull down this branch, run things locally.
- Log in, open the user dropdown and navigate to the bookmarks page. 
- The grid of your bookmark data should render. 

Screen Shot 2022-08-22 at 10 23 00 AM

## 💭 Anything else?

- Pressing the bookmark button on the cards now should only trigger a console log. In a follow-up PR I will add the confirmation dialog that actually updates the data. 
- The sidebar layout still needs to be implemented. 
- The base empty state still needs to be implemented.


src/pages/profile/bookmarks.tsx
```diff
@@ -0,0 +1,9 @@
+import BaseNewLayout from 'layouts/base-new'
+import ProfileBookmarksView from 'views/profile/bookmarks-view'
+
+const BookmarksPage = () => {
```
kendallstrautman, line 4: Perhaps I should confirm with design on this routing update. Lmk if I'm missing any context!

ashleemboyer: This routing update seems good to me with the knowledge that there will _probably_ be a `/profile` page in the future! 👍🏻

src/views/profile/bookmarks-view/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * The content of the ProfileView that is gated behind authentication.
+ */
+const ProfileBookmarksViewContent = () => {
+	const { bookmarks, isLoading } = useAllBookmarks({ enabled: true })
+
+	if (isLoading) {
+		return null // TODO return loading skeleton
+	}
```
ashleemboyer, line 30: Can this be updated to check for first load via `isFetching` and `isRefetching` like [`TutorialCardsGridList`](https://github.com/hashicorp/dev-portal/blob/main/src/components/cards-grid-list/components/tutorial-cards-grid-list/index.tsx#L51-L54) checks? I wonder if we should add an `isFirstLoad` boolean in the return object of these hooks. 💭

src/views/profile/bookmarks-view/helpers/render-bookmark-cards.tsx
```diff
@@ -0,0 +1,49 @@
...
+		collectionCtx,
+		productsUsed,
+	} = formatTutorialData(bookmark.tutorial)
+	const duration = getReadableTime(readTime)
+	return (
+		<li key={id}>
+			<TutorialCard
+				heading={name}
+				description={description}
+				url={getTutorialSlug(slug, collectionCtx.default.slug)}
```
ashleemboyer, line 32: What if a user has bookmarked the tutorial in the context of the non-default collection? Does this always link to the tutorial in the default collection?

src/views/profile/bookmarks-view/helpers/render-bookmark-cards.tsx
```diff
@@ -0,0 +1,49 @@
...
+				productsUsed={productsUsed.map((p: ProductUsed) => p.product.slug)}
+				eyebrowSlot={
+					<>
+						<span>{duration}</span>
+						<BookmarkButtonIconOnly
+							isBookmarked={true}
+							handleOnClick={() => console.log('Trigger confirmation dialog!')}
+						/>
+					</>
+				}
```
ashleemboyer, line 45: I worry about this adding unnecessary maintenance complexity because it adds another piece of code where we'll need to make changes any time a tutorial card eyebrow changes. Can we leverage `TutorialCardWithBookmark` here by adding some props to it?

src/views/profile/bookmarks-view/bookmarks-view.module.css
```diff
@@ -0,0 +1,12 @@
+.subheading {
+	margin: 12px 0 30px 0;
+	color: var(--token-color-foreground-primary);
+
+	@media (--medium-up) {
+		margin-bottom: 38px;
+	}
+}
```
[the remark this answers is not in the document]
kendallstrautman: Oh thank you! good catch 👐

src/views/sign-up-view/index.tsx
```diff
@@ -57,7 +57,7 @@ const SignUpView = () => {
 	 * Redirect to profile page if user is already authenticated
 	 */
 	if (isAuthenticated) {
-		router.replace('/profile')
+		router.replace('/profile/bookmarks')
```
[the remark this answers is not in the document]
kendallstrautman: Yes! Can do

src/views/profile/bookmarks-view/helpers/render-bookmark-cards.tsx
kendallstrautman: This button interaction needs to behave much differently, it renders a different dialog and doesn't need to check for auth etc. So this seems like a better approach to me!

src/views/profile/bookmarks-view/helpers/render-bookmark-cards.tsx
kendallstrautman: It does, this is how bookmarks currently behave in learn. There isn't collection associated data with the bookmark model in the API right now so we have no way of knowing the collection context of the tutorial at the time of bookmarking. We could note this as a future improvement.

src/views/profile/bookmarks-view/helpers/render-bookmark-cards.tsx
kendallstrautman: Imo, a case like this is why we have the slot pattern. Because it needs slightly different treatment and this easily allows us to do that!

BRKalow (approved):
LGTM! Nice work

src/views/profile/bookmarks-view/helpers/render-bookmark-cards.tsx
ashleemboyer: Got it. Thank you!

src/views/profile/bookmarks-view/bookmarks-view.module.css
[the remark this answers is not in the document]
ashleemboyer: Oh shoot. I said what I meant to include in the screenshot, but I messed up my tablet screenshot! The `38px` is the spacing between `.subheading` and the "Your bookmarks" heading after it, rather than the spacing to the wrapper around the "Your bookmarks" heading. The tablet and above spacing should be `32px`. Sorry for the confusion!

src/views/profile/bookmarks-view/helpers/render-bookmark-cards.tsx
ashleemboyer: Totally understand where you're coming from! The tradeoff here is that we have duplicate code for the same presentations, so if we have to make updates to the presentation of tutorial cards with bookmark buttons, we'll have to do it in two places. Like when Progress comes into play this week, won't it have to be implemented in two places with these changes? Here and in `TutorialCardWithBookmark`? Since the behavior of the button is the only thing that differs between the two, that's where I figure it would be useful to customize that with a prop passed to `TutorialCardWithBookmark`.

kendallstrautman: I figure we can make abstractions for the progress work when we get there. I could see a future where we refactor all the eyebrows into one place when we're doing progress, which will make these different composed iterations easier to maintain together.

zchsh: 💭  I dig this approach in the context of this PR - agreed we may want to abstract in the future (the "progress will have to go in two places" point seems like an excellent one); personally I don't feel we need to include that abstraction in this PR.

zchsh (approved):
✅ I ran through testing steps locally, this looks great to me! 🙌

PR #872, opened 2022-08-22
Add Checkbox

ashleemboyer:
## 🔗 Relevant links

- Preview urls 🔎 
  - [`CheckboxField`](https://dev-portal-git-ambadd-checkbox-hashicorp.vercel.app/swingset/components/checkboxfield)
  - [`CheckboxGroup`](https://dev-portal-git-ambadd-checkbox-hashicorp.vercel.app/swingset/components/checkboxgroup)
- [Asana task](https://app.asana.com/0/1202097197789424/1202816749416416/f) 🎟️

## 🗒️ What

- Adds new `src/components/form` directory
  - Adds a basic `Form` component, not used anywhere but there and ready for when it will be used
- Adds several subdirectories in `src/components/form`:
  - `components`: for generic atoms that can be used to compose fields/groups
  - `base-controls`: for the base elements used in fields/groups, like `<input>`s with specific `type`s set
  - `field-controls`: for piecing together atoms and base controls
  - `group-controls`: for piecing together several fields, and some usage of atoms
- Adds `Errors` atom: renders an icon and a list of error messages. Can be used at the field or group level, and its styles never change between the two uses.
- Adds `HelperText` atom: renders text that describes a field or group. Its styles never change between the two uses. In fields, it should be referenced as the element that describes the base control via `aria-describedby` on the control and `id` on the `HelperText`.
- Adds `Label` atom: renders a `<label>` element and has `font-weight` customizations for use at the group level when a group does not have a legend.
- Adds `Legend` atom: renders a `<legend>` with reusable (and customizable) styles.
- Adds `CheckboxControl`: an `<input>` of type `checkbox` and has quite a bit of custom styling that's good to reuse
- Adds `CheckboxField` and Swingset page: uses a few atoms and `CheckboxControl` to create a field that can include a label, helper text, and error messages. Handles associating `CheckboxControl` and `Label` with `useId`.
- Adds `CheckboxGroup` and Swingset page: uses `CheckboxField` and some atoms to show at the group level. Does not expose `children` as prop because it handles several things for us with less code and higher quality markup:
  - Not showing helper text at the `CheckboxField` level when `layout=horizontal`
  - Setting the `Label` font weight based on whether or not there is a `Legend`
  - Setting the `name` prop for all rendered `CheckboxField`
  - Not showing errors at the `CheckboxField` level
- Updates Swingset `componentsRoot` blob to enable nesting `docs.mdx` files in subdirectories of `src/components`
- Upgrades the `@hashicorp/design-system-tokens` package so the new form-related tokens are accessible
  - Removes reference to deprecated `--token-color-neutral-border-strong` in `src/swingset-docs/color.mdx`
  - Replaces reference to deprecated `--token-color-neutral-border-strong` with `--token-color-border-strong ` in `src/views/profile/bookmarks-view/components/empty-state/empty-state.module.css`

## 📸 Design Screenshots

- [HDS Checkbox Figma file](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/HDS-Product---Components?node-id=9120%3A23131)
- [HDS Checkbox documentation page](https://design-system-components-hashicorp.vercel.app/components/form/checkbox)

## 🧪 Testing

- Go to the [`CheckboxField` Swingset page](https://dev-portal-git-ambadd-checkbox-hashicorp.vercel.app/swingset/components/checkboxfield)
- Make sure elements in the examples are styled correctly according to docs
- Make sure the elements behave as you expect
- Go to the [`CheckboxGroup` Swingset page](https://dev-portal-git-ambadd-checkbox-hashicorp.vercel.app/swingset/components/checkboxgroup)
- Make sure elements in the examples are styled correctly according to docs
- Make sure the elements behave as you expect

## 💭 Anything else?

Future component ideas:

- `FieldSet`, which would handle setting up the `grid` approach used in `CheckboxField` for example
  - Maybe it would have some validation of `children` and/or automatically place certain children in the grid based on the order of the `children`? This would reduce some CSS responsibility of placing elements in their appropriate `grid-area`.


src/components/form/components/errors/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+const Errors = ({ className, messages = [] }: ErrorsProps) => {
+	const classes = classNames(s.root, className)
+
+	return (
+		<div className={classes}>
+			<IconAlertDiamondFill16 className={s.icon} />
+			<ul className={s.list}>
+				{messages.map((error: string) => (
+					// eslint-disable-next-line react/jsx-key
+					<li className={s.message}>{error}</li>
```
BRKalow, line 15: suggestion: instead of disabling this, could we use the message as a key?

src/components/form/field-controls/checkbox-field/types.ts
```diff
@@ -0,0 +1,13 @@
+import { LabelProps } from 'components/form/components'
+
+interface CheckboxFieldProps {
+	errors?: string[]
+	helperText?: string
+	id?: string
+	initialIsChecked?: boolean
+	label: string
```
BRKalow, line 8: thought: I think ideally we'll want to support an arbitrary React element for `label`, in addition to a string. I know this doesn't align with the spec, but in the case of Tutorial Library specifically we'll also want to be rendering product icons within the label.

src/components/form/group-controls/checkbox-group/docs.mdx
```diff
@@ -0,0 +1,243 @@
...
+			label: 'Label of control #1',
+		},
+		{
+			label: 'Label of control #2',
+		},
+		{
+			label: 'Label of control #3',
+		},
+	]}
+/>
```
BRKalow, line 23: thought: It would be great to see a more compositional approach here:

```jsx
<CheckboxGroup ... >
	<CheckboxField ... />
	<CheckboxField ... />
	<CheckboxField ... />
</CheckboxGroup>
```

BRKalow (commented):
Looks great!

src/components/form/components/errors/index.tsx
ashleemboyer: Oh, oops! I meant to make that the array index, not totally disable 😆

src/components/form/field-controls/checkbox-field/types.ts
ashleemboyer: If you see `CheckboxField` being used there, I can update! From what I saw at a glance, I wasn't sure if it would be used because there won't be things like helper text or error messages.

src/components/form/group-controls/checkbox-group/docs.mdx
ashleemboyer: I don't think we should use a compositional approach here because there are quite a few things unique to `CheckboxField`s when inside of a `CheckboxGroup`:

- Cannot show `helperText` when `CheckboxGroup`'s `layout` is `'horizontal'`
- Label should have `'regular'` font weight when `layout` is `'vertical'`, `'semibold'` otherwise
- The `name` prop is set based on `CheckboxGroup`'s `name` prop (so all checkboxes should have the same `name`)
- Do not want to show errors for individual checkboxes, so `errors` must be omitted

We can handle all of this automatically with less code if we don't take a compositional approach. If you're looking for a compositional approach for styling reasons, I think we could allow a `className` prop in `CheckboxGroupOption`? Otherwise, I plan to make a `FieldSet` component outside of this PR, and that would enable creating a custom `CheckboxGroup` with a compositional approach if absolutely needed.

src/components/form/field-controls/checkbox-field/types.ts
BRKalow: Makes sense! I think we'll want to use it to ensure the base checkbox styles are consistent.

src/components/form/components/errors/index.tsx
ashleemboyer: ✅ Fixed in: https://github.com/hashicorp/dev-portal/pull/872/commits/e7bc40697947acf57081819bc027eb0dbc522220

src/components/form/field-controls/checkbox-field/types.ts
ashleemboyer: Changes:

- Updated `label` to allow `ReactNode` in: https://github.com/hashicorp/dev-portal/pull/872/commits/06faf823f78f10faf5cdf399388090753473c016
- Added an example to the `CheckboxField` Swingset page in: https://github.com/hashicorp/dev-portal/pull/872/commits/c1957d8d1317d48f389bd2301691e2b776686c42

BRKalow (approved):
👏

PR #877, opened 2022-08-23
feat: local and deploy previews with developer UI

BRKalow:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1202097197789424/1202647508675368/f) 🎟️

## 🗒️ What

Introduces support for local and deploy previews in content repositories with the developer UI.

* Refactors the content preview build step into two separate implementations, `IoPreviewBuilder` and `DeveloperPreviewBuilder`
* Adjusts static generation logic to only generate paths for the previewed product

Preview in vault generated from this branch: https://vault-git-brkfeat-developer-preview-hashicorp.vercel.app/vault/docs/what-is-vault

## 🧪 Testing

- [ ] Checkout this branch: https://github.com/hashicorp/vault/tree/brk.feat/developer-preview
- [ ] Run `npm run start` in the website dir
- [ ] Visit `/vault/docs/what-is-vault`
- [ ] See the addition to the page heading based on the PR ("Very secure"), noting it's loading from local content
- [ ] Visit this preview: https://vault-git-brkfeat-developer-preview-hashicorp.vercel.app/vault/docs/what-is-vault
- [ ] See the addition to the page heading based on the PR ("Very secure"), noting it's loading from local content

To test the Docker workflow:
- [ ] Checkout this branch: https://github.com/hashicorp/vault/tree/brk.feat/developer-preview
- [ ] run `make website/build-local` in the website dir
- [ ] run `make website/local`
- [ ] Make changes to any content file
- [ ] Check the browser, validate changes are present


src/pages/packer/plugins/index.tsx
```diff
@@ -36,12 +37,17 @@ async function getStaticProps(ctx) {
...
 			staticProps.props.layoutProps.sidebarNavDataLevels[2].menuItems
-		// Fetch and merge in remote plugins nav data with the partialNavData
-		const rawNavData = await appendRemotePluginsNavData(
-			remotePluginsFile,
-			partialNavData,
-			''
-		)
+
+		let rawNavData = partialNavData
+		if (!isDeployPreview() || isDeployPreview('packer')) {
```
BRKalow, line 42: I'm not happy with some of these new conditionals. An alternative would be to remove the actual page files for unrelated products, like we do for the io previews.

src/views/docs-view/server.ts
```diff
@@ -139,11 +146,60 @@ export function getStaticGenerationFunctions<
...
+						typeof extraOptions.remarkPlugins === 'function'
+							? extraOptions.remarkPlugins(params)
+							: extraOptions.remarkPlugins
+
+					return [
+						...remarkPluginsFromExtraOptions,
+						...remarkPluginsForFileSystemContent,
+					]
+				},
+			})
```
BRKalow, line 189: It might be a good idea to abstract out this logic into a separate method, this file is getting pretty long and complex.

scripts/content-repo-preview/build.ts
```diff
@@ -1,12 +1,20 @@
...
+
+// our CWD
+const cwd = process.cwd()
 
 function checkEnvVars() {
 	// Filter out defined env vars, leaving only the missing ones
-	const missingEnvVars = ['REPO', 'DEV_IO', 'IS_CONTENT_PREVIEW'].filter(
-		(key) => !process.env[key]
-	)
+	const missingEnvVars = ['REPO'].filter((key) => !process.env[key])
```
BRKalow, line 17: This doesn't really make a ton of sense now given that we are deriving most of the environment variables 💭

src/lib/env-checks.js
```diff
@@ -89,8 +89,11 @@ function isContentDeployPreview(productSlug) {
 	return isDeployPreview() && isProxiedProduct(productSlug)
 }
 
-function isDeployPreview() {
-	return process.env.IS_CONTENT_PREVIEW
+function isDeployPreview(productSlug) {
+	const isProductSlugMatching =
+		!productSlug || productSlug === process.env.PREVIEW_FROM_REPO
```
BRKalow, line 94: Adding support for an optional `productSlug` here, which let's us answer the question "is it a deploy preview for a specific product"

src/pages/packer/plugins/index.tsx
zchsh: if i'm reading right, is the unhappiness here also related to `/packer/plugins` (& `/boundary/api-docs` to an extent) not having their `server` stuff as well abstracted as other docs routes?

zchsh (approved):
✅ Tested this locally with both `npm run start` and Docker, and reviewed the deploy preview PR. This is awesome! 🤩 Stoked 💯

For some reason I had in my mind it was gonna be some big refactor to get this working... Brycefax strikes again with the remote/local loader setup 🧑‍🍳😘⚡️🙌

---

Random side note, when running with Docker (but not with `npm run start`), I saw an `npm` version warning but likely a "me" thing, and didn't seem to affect preview working, so dropping it in `<details />` below...

<details>
<summary>npm version warning that didn't seem matter</summary>

CleanShot 2022-08-24 at 13 21 39@2x

</details>

src/pages/packer/plugins/index.tsx
BRKalow: A bit! Maybe it won't be an issue in practice, but anytime we're relying on data from another repo through the GH API, we'll likely need a conditional like this. It also just feels awkward coming upon these conditionals without the right context 💭  There are three states that we realistically should handle for every case:

* Running in dev-portal, always remote
* Running in a content deploy preview, but not the source of the data (remote)
* Running in a content deploy preview, in the source of the data (local)

I think we're safe for now, but it _feels_ like there could be an abstraction to be created here...

BRKalow:
@zchsh Thanks for the review 🙇 Good call on the npm version warning/error, I also noticed that. I'll see if I can update the `Dockerfile` here to fix.

src/pages/packer/plugins/index.tsx
kendallstrautman: ☝️ This feels like really helpful context. Should we document this in a README that explains how docs content is handled in dev dot or add a comment here?

BRKalow: Great idea!

BRKalow: @kendallstrautman I dropped a section in the repo docs: https://github.com/hashicorp/dev-portal/blob/brk.feat/developer-preview/docs/README.md#remote-content--application-context

thiskevinwang:
From `waypoint/website`, I used this branch as the Docker context and it worked great 👍 

```diff
# ./website/Makefile
DOCKER_RUN_FLAGS=-it \
		--publish "3000:3000" \
		--rm \
		--tty \
		--volume "$(PWD)/content:/app/content" \
		--volume "$(PWD)/public:/app/public" \
		--volume "$(PWD)/data:/app/data" \
		--volume "$(PWD)/redirects.js:/app/redirects.js" \
		--volume "next-dir:/app/website-preview/.next" \
		--volume "$(PWD)/.env:/app/.env" \
		-e "REPO=waypoint" \
+		-e "PREVIEW_MODE=developer"



.PHONY: website/build-local
website/build-local:
	@echo "==> Building local Docker image"
+	@docker build https://github.com/hashicorp/dev-portal.git\#brk.feat/developer-preview \
		-t $(DOCKER_IMAGE_LOCAL)
```

kendallstrautman (approved):
💚 Followed the testing steps and it works as expected!

A cool follow-up thing to add for the author-experience could be hot-reloading with next-remote-watch 😈

PR #878, opened 2022-08-23
fix(vault/downloads): changelog url

thiskevinwang:
# Description

This addresses https://github.com/hashicorp/vault/issues/16829

See https://dev-portal-git-kevin-vault-changelog-hashicorp.vercel.app/_proxied-dot-io/vault/downloads


BRKalow (approved):
This makes more sense.

PR #881, opened 2022-08-24
feat: bookmarks sidebar layout

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksbookmarks-sidebar-layout-hashicorp.vercel.app) 🔎
- [Asana task](https://app.asana.com/0/1199634971449915/1202832753754544) 🎟️
- [Design File](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=12138%3A127457)

## 🗒️ What

This PR sets up the bookmarks view to use the sidebar / sidecar layout & creates a custom sidebar to handle the unique needs of this view. 

## 🛠️ How

~~I had to add a prop to the sidebar sidecar layout so the mobile authenticated disclosure dropdown could be opened on initial state. This is prop-drilly, but works.~~ Adjusting things to the mobile menu authentication component forces the dropdown disclosure to be open on all 'profile' routes. 

I chose to create a custom sidebar for this view since it does not behave the same as the other docs/tutorial sidebars. The designs don't call for walking nav levels and are specific about two different sidebar states between mobile and desktop views. This is because the mobile disclosure renders the sidebar items shown on desktop. 

## 📸 Design Screenshots

Screen Shot 2022-08-24 at 5 04 03 PM
Screen Shot 2022-08-24 at 5 03 58 PM

## 🧪 Testing

- pull down this branch
- start the dev server and sign in 
- navigate to /profile/bookmarks
- you should see the proper sidebar layout
- with the correct nav items on deskop

- shrink down to mobile view 
- check the nav items and behavior matches the design spec. 

## Anything Else?
The `Resources` section on the sidebar is generated even though the designs don't render those. I think for consistency this is okay but probably something we should chat about in the long term.

### DIFF
--- a/src/components/mobile-menu-container/components/mobile-user-disclosure/index.tsx
+++ b/src/components/mobile-menu-container/components/mobile-user-disclosure/index.tsx
@@ -50,11 +50,15 @@ const renderItem = (
 	return <li key={index}>{content}</li>
 }
 
-const MobileUserDisclosure = ({ items, user }: MobileUserDisclosureProps) => {
+const MobileUserDisclosure = ({
+	items,
+	user,
+	initialOpen,
+}: MobileUserDisclosureProps) => {
 	const { icon, description } = getUserMeta(user)
 
 	return (
-		<Disclosure containerClassName={s.root}>
+		<Disclosure containerClassName={s.root} initialOpen={initialOpen}>
 			<DisclosureActivator className={s.activator}>
 				<span className={s.iconAndTextWrapper}>
 					<span className={s.icon}>{icon}</span>
--- a/src/components/mobile-menu-container/index.tsx
+++ b/src/components/mobile-menu-container/index.tsx
@@ -37,7 +37,11 @@ const MOBILE_MENU_MOTION = {
  * Handles rendering the Sign In and Sign Up UI elements in mobile viewports.
  * Intended to be used alongside `MobileMenuContainer`.
  */
-const MobileAuthenticationControls = () => {
+const MobileAuthenticationControls = ({
+	initialOpen,
+}: {
+	initialOpen?: boolean
+}) => {
 	const { showAuthenticatedUI, showUnauthenticatedUI, signIn, signOut, user } =
 		useAuthentication()
 
@@ -68,7 +72,11 @@ const MobileAuthenticationControls = () => {
 		)
 	} else if (showAuthenticatedUI) {
 		content = (
-			<MobileUserDisclosure items={getUserMenuItems({ signOut })} user={user} />
+			<MobileUserDisclosure
+				items={getUserMenuItems({ signOut })}
+				user={user}
+				initialOpen={initialOpen}
+			/>
 		)
 	}
 
--- a/src/components/mobile-menu-container/index.tsx
+++ b/src/components/mobile-menu-container/index.tsx
@@ -2,6 +2,7 @@
 import { ForwardedRef, forwardRef } from 'react'
 import classNames from 'classnames'
 import { m, useReducedMotion } from 'framer-motion'
+import { useRouter } from 'next/router'
 
 // HashiCorp imports
 import { IconArrowRight16 } from '@hashicorp/flight-icons/svg-react/arrow-right-16'
@@ -37,11 +38,8 @@ const MOBILE_MENU_MOTION = {
  * Handles rendering the Sign In and Sign Up UI elements in mobile viewports.
  * Intended to be used alongside `MobileMenuContainer`.
  */
-const MobileAuthenticationControls = ({
-	initialOpen,
-}: {
-	initialOpen?: boolean
-}) => {
+const MobileAuthenticationControls = () => {
+	const { asPath } = useRouter()
 	const { showAuthenticatedUI, showUnauthenticatedUI, signIn, signOut, user } =
 		useAuthentication()
 
@@ -75,7 +73,7 @@ const MobileAuthenticationControls = ({
 			<MobileUserDisclosure
 				items={getUserMenuItems({ signOut })}
 				user={user}
-				initialOpen={initialOpen}
+				initialOpen={asPath === '/profile/bookmarks' ? true : false}
 			/>
 		)
 	}
--- a/src/components/mobile-menu-container/index.tsx
+++ b/src/components/mobile-menu-container/index.tsx
@@ -73,7 +73,7 @@ const MobileAuthenticationControls = () => {
 			<MobileUserDisclosure
 				items={getUserMenuItems({ signOut })}
 				user={user}
-				initialOpen={asPath.includes('/profile') ? true : false}
+				initialOpen={asPath.startsWith('/profile')}
 			/>
 		)
 	}
--- a/src/components/mobile-menu-container/index.tsx
+++ b/src/components/mobile-menu-container/index.tsx
@@ -73,7 +73,7 @@ const MobileAuthenticationControls = () => {
 			<MobileUserDisclosure
 				items={getUserMenuItems({ signOut })}
 				user={user}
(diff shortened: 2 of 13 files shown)


src/layouts/sidebar-sidecar/types.ts
```diff
@@ -17,6 +17,7 @@ interface BaseProps {
 	AlternateSidebar?: (props: any) => ReactElement
 	optInOutSlot?: ReactElement
 	versions?: VersionSelectItem[]
+	initialOpenMobileAuthControls?: boolean
```
kendallstrautman, line 20: I don't love this drilling but I couldn't think of a different way to enable this opened menu state. Seemed like the easiest option with the current setup.

src/components/mobile-menu-container/components/mobile-user-disclosure/index.tsx
```diff
@@ -50,11 +50,15 @@ const renderItem = (
...
+const MobileUserDisclosure = ({
+	items,
+	user,
+	initialOpen,
+}: MobileUserDisclosureProps) => {
 	const { icon, description } = getUserMeta(user)
 
 	return (
-		<Disclosure containerClassName={s.root}>
+		<Disclosure containerClassName={s.root} initialOpen={initialOpen}>
```
kendallstrautman, line 61: The designs call for this disclosure to be already open when the mobile menu is opened. The way I have it now, this is drilled down from `SidebarSidecarLayout` 🤷‍♀️  Open to alternative approaches.

src/components/sidebar/components/nav-list/index.tsx
```diff
@@ -0,0 +1,6 @@
+import { ReactNode } from 'react'
+import s from '../../sidebar.module.css'
+
+export default function NavList({ children }: { children: ReactNode }) {
+	return <ul className={s.navList}>{children}</ul>
```
kendallstrautman, line 5: Extracted this into a reusable component since I needed some of the sidebar styles in the custom sidebar for the bookmarks view and this seemed better than importing from the stylesheet.

src/components/sidebar/components/sidebar-nav-menu-item/index.tsx
```diff
@@ -27,6 +27,7 @@ import {
 	SupportedIconName,
 } from './types'
 import s from './sidebar-nav-menu-item.module.css'
+import { IconSignOut16 } from '@hashicorp/flight-icons/svg-react/sign-out-16'
```
kendallstrautman, line 30: TODO: reorder this import.

src/components/sidebar/components/sidebar-nav-menu-item/index.tsx
```diff
@@ -150,6 +151,26 @@ const SidebarNavLinkItem = ({ item }: SidebarNavLinkItemProps) => {
 	}
 }
 
+/**
+ * Handles rendering a button and icon for the sidebar.
+ * Currently used for a 'sign out' action on the profile page
+ */
+export function SidebarNavMenuButton({ item }: SidebarNavMenuButtonProps) {
```
kendallstrautman, line 158: Created a new component here to handle the `Sign out` button in the sidebar. This is created to support the needs of this usecase, I figure we can expand / abstract as further usecases for this crop up.

src/views/profile/bookmarks-view/index.tsx
```diff
@@ -7,6 +7,15 @@ import Heading from 'components/heading'
...
 import s from './bookmarks-view.module.css'
+import SidebarSidecarLayout from 'layouts/sidebar-sidecar'
+import { generateTopLevelSubNavItems } from 'lib/generate-top-level-sub-nav-items'
+import { SidebarNavMenuItem } from 'components/sidebar/components'
+import Sidebar from 'components/sidebar'
+import { IconSignOut16 } from '@hashicorp/flight-icons/svg-react/sign-out-16'
+import { SidebarNavMenuButton } from 'components/sidebar/components/sidebar-nav-menu-item'
+import useAuthentication from 'hooks/use-authentication'
+import { useDeviceSize } from 'contexts'
+import NavList from 'components/sidebar/components/nav-list'
```
kendallstrautman, line 18: TODO: reorder these imports

src/views/profile/bookmarks-view/index.tsx
```diff
@@ -15,11 +24,71 @@ import s from './bookmarks-view.module.css'
...
+				sidebarNavDataLevels={[]}
+				sidecarSlot={null}
+			>
+				<ProfileBookmarksViewContent />
+			</SidebarSidecarLayout>
 		</AuthenticatedView>
 	)
 }
 
+function ProfileBookmarksSidebar() {
```
kendallstrautman, line 47: TODO: refactor and separate into own files.

src/components/mobile-menu-container/components/mobile-user-disclosure/index.tsx
kendallstrautman: 🤔  good idea, yeah that does feel better than this approach. ill update.

src/components/sidebar/components/nav-list/index.tsx
```diff
@@ -0,0 +1,6 @@
+import { ReactNode } from 'react'
+import s from '../../sidebar.module.css'
+
+export default function NavList({ children }: { children: ReactNode }) {
```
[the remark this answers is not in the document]
kendallstrautman: That is better! I'll update 👐

src/components/mobile-menu-container/components/mobile-user-disclosure/index.tsx
kendallstrautman: I updated `MobileAuthenticatedControls` to handle this 👌 Feels better to me!

src/components/mobile-menu-container/index.tsx
```diff
@@ -68,7 +70,11 @@ const MobileAuthenticationControls = () => {
 		)
 	} else if (showAuthenticatedUI) {
 		content = (
-			<MobileUserDisclosure items={getUserMenuItems({ signOut })} user={user} />
+			<MobileUserDisclosure
+				items={getUserMenuItems({ signOut })}
+				user={user}
+				initialOpen={asPath.includes('/profile') ? true : false}
```
ashleemboyer, line 76: Checking for `includes('/profile')` ends up being caught on paths like `/waypoint/docs/runner/profiles`, so what do you think about using `startsWith` instead? Also, `includes` and `startsWith` both return a boolean, so you can simplify this to just pass the return value of the function. 😊

```suggestion
				initialOpen={asPath.startsWith('/profile')}
```

src/components/sidebar/components/sidebar-nav-menu-item/index.tsx
ashleemboyer: Perfect place to put this! 🙌🏻 Totally agree with this approach.

src/components/sidebar/components/sidebar-nav-list/index.tsx
```diff
@@ -0,0 +1,6 @@
+import { ReactNode } from 'react'
+import s from '../../sidebar.module.css'
+
+export default function SidebarNavList({ children }: { children: ReactNode }) {
+	return <ul className={s.navList}>{children}</ul>
+}
```
ashleemboyer, line 6: Great abstraction! ✨

ashleemboyer (approved):
Only one small request in `src/components/mobile-menu-container/index.tsx`, so going to approve! No need to re-request review after that's addressed. Nicely done!! 👏🏻

src/components/mobile-menu-container/index.tsx
kendallstrautman: Ohhhh good call! Will update 👍 ty

PR #882, opened 2022-08-25
Fix tablet width mobile menu

ashleemboyer:
## 🗒️ What

- Fixes the mobile menu on tablet viewports, with some custom logic pulled from `useDeviceSize`, since this is the only place where we have customized a special breakpoint.

## 🤷 Why

- When we updated the nav header breakpoint recently, we did catch that we needed to update the breakpoint used by `MobileMenuProvider` as well.

## 🧪 Testing

- [ ] Check that the mobile menu shows at `1200px` or narrower
- [ ] Make sure the mobile menu closes when:
  - [ ] The menu is open and a page navigation occurs
  - [ ] The menu is open and the viewport is changed to `>= 1201px`

### DIFF
--- a/src/components/navigation-header/navigation-header.module.css
+++ b/src/components/navigation-header/navigation-header.module.css
@@ -23,7 +23,7 @@
 	position: relative;
 	width: 100%;
 
-	@media (--nav-header-is-mobile) {
+	@media (--dev-dot-show-mobile-menu) {
 		padding-bottom: 14px;
 		padding-top: 14px;
 	}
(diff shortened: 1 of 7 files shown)


zchsh (approved):
🥇  Dang, thanks for handling this! 💖 

Found one other spot where we might need to change from `isDesktop`, have opened up #883 which targets this PR to try to fix. Intent is to fix an issue where, between the new `1200px` breakpoint and `isDesktop`, we're still seeing the desktop-style "back to link" menu contents rather than the multi-pane mobile menu contents.

Even without that tweak, this PR definitely fixes the broken behaviour on tablet - so approving regardless 👍

PR #891, opened 2022-08-30
feat: tutorial progress create & update

zchsh:
- feat: stub in dev progress view, but it shows bookmarks
- feat: set up fetching all progress
- feat: demo fetch all progress on test view
- feat: start thinking through tutorial card use
- feat: test out with-auth-elems tutorial card
- fix: isolate progress work to test progress view
- feat: continue react-query hooks implementation
- feat: implement collection progress connection
- feat: spike tutorial progress create-update
- chore: reorder content on test view
- feat: refactor tutorial-progress to use react-query

## 🔗 Relevant links

- [Asana task](url) 🎟️


src/hooks/progress/use-update-tutorial-progress.ts
```diff
@@ -0,0 +1,150 @@
...
+ * if the tutorial does not yet have recorded progress,
+ * the tutorial will be marked as `visited` (0 percent).
+ *
+ * When the startRef element moves out of the viewport,
+ * the tutorial will be marked as `in_progress` (50 percent).
+ *
+ * When the endRef element moves into the viewport,
+ * the tutorial will be marked as `complete` (100 percent).
+ */
+export function useUpdateTutorialProgress({
```
zchsh, line 39: This hook handles tracking progress, simplifying the consuming `TutorialView`'s responsibility to placing the `startRef` and `endRef` on elements. This is very similar to existing functionality in the `hashicorp/learn` repo.

Note: Acknowledging [early feedback in Slack to possibly combine this with `use-tutorial-progress-mutations`](https://hashicorp.slack.com/archives/C02T80JTNS2/p1661377009842339), I've ended up having this separate as I found combining them yielded a pretty huge file. Very open to other approaches though 🙇‍♂️

src/hooks/progress/use-update-tutorial-progress.ts
```diff
@@ -0,0 +1,142 @@
...
+		 * If we have positive progress, or need to record initial progress,
+		 * then we create or update our progress (formatted as percent for the API).
+		 *
+		 * Note that we don't ever downgrade the percent value of progress.
+		 */
+		const existingPercent = progressLabelToPercent(tutorialProgressLabel)
+		const newPercent = progressLabelToPercent(newProgressLabel)
+		const hasPositiveProgress = parseInt(newPercent) > parseInt(existingPercent)
+		const needsInitialProgress = tutorialProgressLabel === null
+		if (needsInitialProgress || hasPositiveProgress) {
```
zchsh, line 98: We don't want to make negative progress updates - if the API says progress is `100` percent, even if we've just loaded a view and can't see `endRef` we don't want to downgrade this progress.

We do need to ensure that if the API progress is `null`, we create a new progress record.

src/lib/learn-client/api/api-types.ts
```diff
@@ -75,6 +76,24 @@ export interface ApiCollectionTutorial
...
+	extends Pick<BaseModel, 'id' | 'created_at' | 'updated_at'> {
+	user_id: string
+	collection_id: uuid
+	tutorial_id: uuid
+	complete_percent: TutorialProgressPercent
+	/**
+	 * Note: so far, we've only needed the tutorial_id from this record type.
+	 * Perhaps we could not include detailed tutorial data in the API response?
+	 */
+	tutorial: ApiTutorialLite
```
zchsh, line 89: Added notes here as I _think_ we'll be able to drop these from the API response and significantly reduce the amount of data we're fetching. But very much outside the scope of this PR, and also I'm not _sure_ we'd want to drop the data since I don't have full context on why it was added.

src/lib/learn-client/api/progress/formatting.ts
```diff
@@ -0,0 +1,40 @@
...
+	TutorialProgressPercent,
+} from 'lib/learn-client/types'
+
+/**
+ * Convert from API's complete_percent to the TutorialProgressLabel enum.
+ *
+ * This is necessary as in the context of Dev Dot, we want the label,
+ * but in the context of the Learn API, we want complete_percent.
+ */
+export function progressPercentToLabel(
```
zchsh, line 12: Formatting utilities to convert between the API format (a limited set of percentage values), and the format we want to display and reason around on Dev Dot (`visited`, `in_progress`, or `completed`).

This is largely informed by existing behaviour in `hashicorp/learn` - while it's technically possible for the API to hold a variety of percentage values, we only ever create or update with `0`, `50`, or `100` (corresponding to `visited`, `in_progress`, and `completed`).

src/lib/learn-client/api/progress/get-tutorial-progress.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+type GetTutorialProgressResult = null | ApiCollectionTutorialProgress
+
+/**
+ * Fetches progress on a single tutorial.
+ *
+ * https://digital-api-specs.vercel.app/learn#tag/Collection-Tutorial-Progress/paths/~1collections~1{collection_id}~1tutorials~1{tutorial_id}~1progress/get
+ */
+const getTutorialProgress = async ({
```
zchsh, line 20: I've added get, update, and create functions for `lib/learn-client/progress`. I expect we'll rarely if ever use these directly, seems more likely we'll always use them through the related `react-query` hooks.

src/lib/learn-client/types.ts
```diff
@@ -10,6 +10,11 @@ export type uuid = string
 export type slug = string
 export type identifier = uuid | slug
 
+export interface TutorialIdCollectionId {
```
zchsh, line 13: Used this type in a few places already, and I think it'll come up even more through the rest of user progress work, so felt like a decent time to elevate it to this file.

zchsh:
@ashleemboyer @kendallstrautman @BRKalow This PR implements the "create and update" part of user progress (on tutorials!).

Have tried to be thorough here with context since there's quite a bit going on (or at least it felt that way to me since i hadn't used `react-query` yet!). I've added an overview video & testing steps video to the PR description. 

src/hooks/progress/use-update-tutorial-progress.ts
```diff
@@ -0,0 +1,143 @@
...
+	const { isAuthenticated } = useAuthentication()
+	// We need to know if progress exists, to know whether to "create" or "update"
+	const { tutorialProgressLabel } = useTutorialProgress({
+		tutorialId,
+		collectionId,
+	})
+	// `react-query` mutations are what we'll use to "create" or "update" progress
+	const tutorialProgressMutations = useTutorialProgressMutations()
+	// We don't want to run multiple mutations at once
+	const [isRunningMutation, setIsRunningMutation] = useState<boolean>(false)
```
BRKalow, line 53: 💭 I think we can determine this from the mutation hook return value: https://tanstack.com/query/v4/docs/reference/useMutation

In particular, `status`:

> Will be:
> `idle` initial status prior to the mutation function executing.
> `loading` if the mutation is currently executing.
> `error` if the last mutation attempt resulted in an error.
> `success` if the last mutation attempt was successful.

src/hooks/progress/use-tutorial-progress.ts
```diff
@@ -0,0 +1,62 @@
...
+						 * to know whether the record exists or not.
+						 */
+						if (data == null) {
+							return null
+						}
+						return progressPercentToLabel(data.complete_percent)
+					}
+				),
+			{ enabled: !!accessToken }
+		)
```
ashleemboyer, line 54: We should store the full, raw data returned from the API in the cache rather than a formatted / pared down version. This enables any part of the app to use the data however it wants on a long term scale, and minimizes the maintenance we have to do in this layer of the data fetching.

src/lib/learn-client/api/api-types.ts
```diff
@@ -75,6 +76,24 @@ export interface ApiCollectionTutorial
 	tutorial: ApiTutorial
 }
 
+export interface ApiCollectionTutorialProgress
+	extends Pick<BaseModel, 'id' | 'created_at' | 'updated_at'> {
```
ashleemboyer, line 80: Since this is `Pick`-ing all the attributes, can we just extend `BaseModel` here?

src/lib/learn-client/api/api-types.ts
ashleemboyer: We definitely should hang onto these until the API response does otherwise. 😊

src/lib/learn-client/api/progress/formatting.ts
ashleemboyer: @kendallstrautman do you have more context on why we only create/update Tutorial:Collection progress with 0, 50, or 100? Doesn't this mean that a user will have to get 50% of the way through a tutorial before their progress is set as "in progress"?

src/lib/learn-client/api/progress/create-tutorial-progress.ts
```diff
@@ -0,0 +1,56 @@
...
+
+	// Make the request
+	const requestResult = await post(requestRoute, accessToken, {
+		complete_percent: completePercent,
+	})
+
+	// Return null if the request status is 404
+	if (requestResult.status === 404) {
+		return null
+	}
```
ashleemboyer, line 42: This can be omitted since this is a POST request and not a GET request.

ashleemboyer (changes requested):
Have a few comments sprinkled in here but looking good!

Would also like to see the changes in `src/hooks/progress/use-update-tutorial-progress.ts` and `src/views/tutorial-view/index.tsx` moved to a separate PR for the ease of reviewing.

src/lib/learn-client/api/progress/formatting.ts
zchsh: @ashleemboyer With reference to both the existing `hashicorp/learn` code and to this PR, the `0`, `50`, and `100` percent values are not currently tied to scroll progress in a strict sense. Instead we use `start` and `end` markers. 
- when a page is loaded, we set `0%` in the DB (`visited` in the front-end)
- when the page is scrolled enough that the `start` marker is out of view, we set `50%` in the DB (`in_progress`)
- when the page is scrolled enough that the `end` marker is in view, we set `100%` in the DB (`complete`)

The `in_progress` state has been tied to `50%` previously, so the intent in this PR is to maintain that mapping.

Let me know if that clarifies. This might also related to the layer of opinion built into the `react-query` layer.

src/hooks/progress/use-tutorial-progress.ts
zchsh: @ashleemboyer I think a layer of opinion in the query layer here _might_ be valuable. But definitely unsure, would love to get others opinions - cc @BRKalow @kendallstrautman .

The rationale is that we seem to want to consistently refer to `visited`, `in_progress`, and `complete` on the front-end, and that we don't ever refer to actual percentages in the UI. So the `react-query` layer is meant to add that layer of opinion, rather than asking the query consumer to have to parse out the `visited`, `in_progress`, and `complete` status every time they query.

We don't currently seem to have any plans to record or use tutorial progress in a more granular way, though definitely take the point that if we ever do want to change to using full percentages (or rather the specific milestones that the API supports), then we'd need to refactor these hooks (and our display components, of course, though I think we'd need to refactor those either way if we change how we show progress).

Let me know if that seems reasonable - happy to refactor as suggested, with the raw set of records coming from the query and each component that queries doing the parsing, if we'd prefer things that way 👍

zchsh: ☝️ At a higher level this will probably also related to imminent `collection` progress work:

In the spike I'd built a similar layer of opinion, where the query layer handled the logic of "count the completed tutorials in this collection" and returned an integer, rather than returning an array of `ApiCollectionTutorialProgress` records. We could instead return the array of `ApiCollectionTutorialProgress` and have the consumer count through the array of records.

src/hooks/progress/use-update-tutorial-progress.ts
zchsh: @BRKalow  I've updated to include an `isReady` boolean as part of the return object from `useTutorialProgressMutations()`. Have removed `isRunningMutation` state. Let me know if that roughly lines up with what you were expecting 🙇

src/lib/learn-client/api/api-types.ts
zchsh: 💖 Ah great call, I'd mindlessly copied from somewhere else 😁 ✅ updated!

src/lib/learn-client/api/api-types.ts
zchsh: For sure! 💯  Intent here was to make a note of this in case anyone wonders why these are included in the payload - as far as I can tell, we don't have a strong reason to include them.

zchsh:
@ashleemboyer @BRKalow thanks for the feedback! Have done my best to address everything, either via code changes or comments.

The "how much opinion should `react-query` add" felt like a potentially broader discussion, so have left that open for now to get more input. Definitely happy to adjust either way, very open to further thoughts, but felt worth pausing before diving into a refactor of removing the opinion from the `react-query` layer 👍

src/hooks/progress/use-tutorial-progress-mutations/index.ts
```diff
@@ -0,0 +1,123 @@
...
+			)
+		},
+		[updateTutorialProgressMutation, accessToken]
+	)
+
+	/**
+	 * Consumers likely only want to perform calculations in preparation to
+	 * run these mutations if status is "idle" or "success". Otherwise,
+	 * attempt to run the mutation may result in a continuous loop.
+	 */
```
ashleemboyer, line 108: Where is this happening that `isReady` is needed?

src/hooks/progress/use-tutorial-progress.ts
BRKalow: This would be a good candidate for a [query selector](https://tanstack.com/query/v4/docs/guides/migrating-to-react-query-3#query-data-selectors). This way, the query cache remains purely a response cache and we get future flexibility without burdening the consumer of the hook.

```suggestion
	// Fetch a progress records by `tutorialId`, then filter by `collectionId`
	const { data: tutorialProgressLabel, ...restQueryResult } =
		useQuery<QueryDataType>(
			[TUTORIAL_PROGRESS_QUERY_ID, tutorialId, collectionId],
			() => getTutorialProgress({ accessToken, tutorialId, collectionId }),
			{
				enabled: !!accessToken,
				select: data => {
					  /**
					  * Data may be null, if a progress record does not exist
					  * for this tutorialId + collectionId combination.
					  *
					  * We pass this on to the query consumer, as they may need
					  * to know whether the record exists or not.
					  */
				        if (data == null) {
					        return null
				        }
					return progressPercentToLabel(data.complete_percent)
				}
			}
		)
```

ashleemboyer: I think the cache should hold 0 opinion about the data.

ashleemboyer (changes requested):
Can we please separate these files' changes from this PR?

- `src/views/tutorial-view/index.tsx`
- `src/hooks/progress/use-update-tutorial-progress.ts`

src/hooks/progress/use-tutorial-progress-mutations/index.ts
zchsh: It's being used in `use-update-tutorial-progress`- we don't want to continually re-calculate progress and re-do mutations if we have in progress.

zchsh:
@ashleemboyer Ah apologies for missing the separate file changes in the last update - I've stashed those changes for now. Definitely harder to test without them, but could see why we might not want to include them alone (even with the auth feature flag).

I've updated the query layer to store queried data in full, using [select](https://tanstack.com/query/v4/docs/guides/migrating-to-react-query-3#query-data-selectors) (thanks for the suggestion @BRKalow !).

src/lib/learn-client/api/progress/formatting.ts
kendallstrautman: ^ zach explained it really well. From the application perspective, we are only concerned with four states: 'no progress' (record doesn't exist yet), 'visited' (0), 'in progress' (50 - but this is an arbitrary number beween 0-100), & 'complete' (100).

The percentages in the API were created that way to make the system more flexible in case we wanted to capture more finite progress metrics later down the line.

src/lib/learn-client/api/api-types.ts
kendallstrautman: I don't think we would update the API to not return this data so that the endpoint aligns with other enrichment patterns in the API but we could remove returning this in the client methods during formatting if its not needed in the application.

src/lib/learn-client/api/progress/get-tutorial-progress.ts
```diff
@@ -0,0 +1,48 @@
...
+	const requestResult = await get(requestRoute, accessToken)
+
+	// Return null if the request status is 404
+	if (requestResult.status === 404) {
+		return null
+	}
+
+	// Return data as JSON if result is OK
+	if (requestResult.ok) {
+		const { result } = await requestResult.json()
```
kendallstrautman, line 38: You could remove the extra 'tutorial' & 'collection' data here if we want to reduce the amount of data and its not being used.

zchsh: It's not being used, but it sounds like we want to keep `react-query` data matchy-matchy with the API response for now. Happy to switch this up if there are strong opinions though 🙇‍♂️

src/lib/learn-client/types.ts
```diff
@@ -260,6 +265,21 @@ export enum CompanyLogoOption {
...
+	complete = 'complete',
+}
+
+export enum TutorialProgressPercent {
+	Zero = '0',
+	TwentyFive = '25',
+	Fifty = '50',
+	SeventyFive = '75',
+	Ninety = '90',
+	OneHundred = '100',
```
kendallstrautman, line 280: Any reason we're using a different casing here?

kendallstrautman (approved):
Looks good to me! Will do further testing in #900

src/lib/learn-client/api/progress/get-tutorial-progress.ts
kendallstrautman: Sounds good!

ashleemboyer (approved):

src/lib/learn-client/types.ts
zchsh: No strong reason here, leaned towards matching learn-api [which uses this casing](https://github.com/hashicorp/learn-api/blob/64be7e42bb8852c7c777d60e6732b908fc3042a5/src/models/collection_tutorial_progress.ts#L11).

Going to merge this as-is, but definitely happy to adjust in #900 if we want. I guess for consistency with `dev-dot` we'd want camel case like `seventyFive` etc? Or maybe snake like `seventy_five`? Seeing both in the above types so on the fence.

PR #899, opened 2022-08-31
fix: do not revalidate SSG'd pages in deploy previews

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Set `revalidate: false` if we are in a deploy preview and reading from the filesystem. This fixes an issue where ISR attempts to read from the filesystem and errors.


mhonnatti (approved):
This makes sense to have for previews!

PR #900, opened 2022-08-31
feat: activate tutorial progress tracking

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Activates progress tracking on tutorial views.

## 🤷 Why

To continue user progress implementation.

## 🛠️ How

- Adds a `use-update-tutorial-progress` hook, which handles the update logic
- Uses the above hook on tutorial view
- Also renames progress `label` names from #891 to `status`

## 🧪 Testing

- One option is to follow testing steps described in #891 
- Another option is to check out #901 locally, and test there
    - #901 is based on this PR, but also includes progress display in tutorial view sidebar. Makes it much easier to debug

## 💭 Anything else?

- ~There's some `console.log` in `use-update-tutorial-progress` that should be removed. But found these _very_ useful for testing, so have left them in for now for testing's sake.~
    - ✅ EDIT: removed

- We could consider putting this behind a feature flag if we're worried about progress being tracked before it's being displayed.
    - Rationale why I didn't do this: authentication features generally are already behind a feature flag (I think?), so I felt an additional flag would be a bit redundant
    - EDIT: confirmed that the auth feature flag carries through to this work - the `react-query` related hooks check for access tokens and aren't enabled unless they're present, and in our effect hook we only try to run progress mutations if we're authenticated, so mutations will never even get called unless someone is authenticated.

[task]: https://app.asana.com/0/1202097197789424/1202900023863567/f 
[preview]: https://dev-portal-git-zstutorial-progress-create-upda-771f06-hashicorp.vercel.app/


src/hooks/progress/use-tutorial-progress-mutations/index.ts
```diff
@@ -5,7 +5,7 @@ import {
 	createTutorialProgress as createTutorialProgressApi,
 	updateTutorialProgress as updateTutorialProgressApi,
 } from 'lib/learn-client/api/progress'
-import { progressPercentToLabel } from 'lib/learn-client/api/progress/formatting'
```
zchsh, line 8: There are many changes in this PR that rename progress `label` to progress `status`. Apologies for the noise!

src/hooks/progress/use-tutorial-progress.ts
```diff
@@ -4,17 +4,17 @@ import { ApiCollectionTutorialProgress } from 'lib/learn-client/api/api-types'
 import {
 	getTutorialProgress,
 	GetTutorialProgressResult,
-	progressPercentToLabel,
+	progressPercentToStatus,
```
zchsh, line 7: Changes in this file are the `label` → `status` rename.

src/lib/learn-client/api/progress/formatting.ts
```diff
@@ -1,38 +1,38 @@
 import {
-	TutorialProgressLabel,
+	TutorialProgressStatus,
```
zchsh, line 2: Changes in this file are also the progress `label` → `status` rename.

src/lib/learn-client/api/progress/__tests__/formatting.test.ts
```diff
@@ -1,44 +1,44 @@
 import {
-	TutorialProgressLabel,
+	TutorialProgressStatus,
```
zchsh, line 2: More `label` → `status` renames.

src/hooks/progress/use-update-tutorial-progress.ts
```diff
@@ -0,0 +1,186 @@
...
+			let requestType: string
+			if (needsInitialProgress) {
+				requestType = 'CREATED'
+				tutorialProgressMutations.createTutorialProgress(mutationArgs)
+			} else {
+				requestType = 'UPDATED'
+				tutorialProgressMutations.updateTutorialProgress(mutationArgs)
+			}
+
+			console.log(`${requestType} progress: ${newProgressStatus}`)
```
zchsh, line 166: I've left in some debug comments for local testing & validation.

src/hooks/progress/use-update-tutorial-progress.ts
```diff
@@ -0,0 +1,186 @@
...
+		 * On initial client-side navigation, this effect may be called
+		 * before startRefs and endRefs are updated. We can detect this
+		 * by comparing data-ref-id attributes we've added to the
+		 * elements startRef and endRef are attached to.
+		 */
+		const startRefId = startEntry?.target.getAttribute('data-ref-id')
+		const endRefId = endEntry?.target.getAttribute('data-ref-id')
+		const compareId = `${tutorialId}_${collectionId}`
+		const isStaticIdsMismatch =
+			startRefId !== compareId || endRefId !== compareId
```
zchsh, line 77: This is a key part of the effect that we'd been trying to debug on the ambient zoom yesterday. This seems to work exactly as expected now 👍

I've tried to add thorough yet brief comments on all the reasons we'd want to bail from this effect, as well as brief rationale for each effect dependency. Have formatted with early `return` so hopefully it's clearer than a single combined conditional.

zchsh:
@ashleemboyer @BRKalow  @kendallstrautman @thiskevinwang tagged y'all here cause I feel like you all have varying yet useful context on this work.

TLDR; is this became a bit of rabbit hole yesterday, I think we've come out the other end though and have something functioning and reasonable.

Testing that progress is working as expected is likely easiest by pulling down #901 as that contains sidebar display as well.

src/hooks/progress/use-update-tutorial-progress.ts
```diff
@@ -0,0 +1,186 @@
...
+export function useUpdateTutorialProgress({
+	tutorialId,
+	collectionId,
+}: TutorialIdCollectionId): UpdateProgressReturnInterface {
+	// We shouldn't try to update progress unless we're authenticated
+	const { isAuthenticated } = useAuthentication()
+	// We need to know if progress exists, to know whether to "create" or "update"
+	const { tutorialProgressStatus } = useTutorialProgress({
+		tutorialId,
+		collectionId,
```
kendallstrautman, line 48: Should we have this be tied into auth with `enabled`? [See the bookmarks](https://github.com/hashicorp/dev-portal/blob/main/src/hooks/bookmarks/use-is-bookmarked.ts#L32). Or maybe that should be in `useTutorialProgress`.

src/hooks/progress/use-update-tutorial-progress.ts
```diff
@@ -0,0 +1,186 @@
...
+			return
+		}
+
+		/**
+		 * Determine the new progress state, which we'll then update
+		 * - 'complete' (100%) if the end ref is visible
+		 * - 'in_progress' (50%) if the start ref is no longer visible
+		 * - 'visited' (0%) in all other cases
+		 */
+		let newProgressStatus: TutorialProgressStatus
```
kendallstrautman, line 124: 💭 It might help with readability to encapsulate this into a function. So one can easily scan all the early returns and then read name like `handleUpdateProgressStatus` to make this useEffect easy to digest. 

Just a thought. It's up to you.

src/views/tutorial-view/index.tsx
```diff
@@ -188,6 +189,18 @@ function TutorialView({
...
 	]
 
+	/**
+	 * Keep track of progress for authenticated users
+	 *
+	 * Note that we attach data-ref-id to avoid some
+	 * client-side-navigation-related progress tracking quirks.
+	 */
+	const progressRefsId = `${id}_${collectionCtx.current.id}`
+	const progressRefs = useUpdateTutorialProgress({
```
kendallstrautman, line 199: Are you planning on creating a separate hook to return the progress value for this tutorial to render the metadata UI? Or should this hook also return the progress status with the refs?

kendallstrautman: 🤔 Perhaps here we could call for the entire collection's progress and just pass this down to the UI (sidebar & metadata). [See example in Learn](https://github.com/hashicorp/learn/blob/master/pages/tutorials/%5B...slug%5D.tsx#L145-L148)

ashleemboyer: The `useUpdateTutorialProgress` hook shouldn't need to return the updated progress value if we leverage React Query like we do for bookmarks. The `useTutorialProgress` hook will automatically return new data if the query key uses is invalidated on successful update. I believe we're already set up correctly for this already because `useTutorialProgress`'s `makeOnMutationSuccess` invokes `setQueryData`.

ashleemboyer: I did a quick test, it works as far as I can tell! So anywhere we use `useTutorialProgress` will automatically get the updated data without any extra work on our part.

src/hooks/progress/use-update-tutorial-progress.ts
```diff
@@ -0,0 +1,186 @@
...
+ * On initial load, regardless of startRef or endRef visibility,
+ * if the tutorial does not yet have recorded progress,
+ * the tutorial will be marked as `visited` (0 percent).
+ *
+ * When the startRef element moves out of the viewport,
+ * the tutorial will be marked as `in_progress` (50 percent).
+ *
+ * When the endRef element moves into the viewport,
+ * the tutorial will be marked as `complete` (100 percent).
+ */
```
ashleemboyer, line 38: Excellent explanation, thank you! 🙌🏻

src/hooks/progress/use-update-tutorial-progress.ts
```diff
@@ -0,0 +1,186 @@
...
+			return
+		}
+
+		/**
+		 * If we're not authenticated, we bail early, there's no
+		 * point in trying to make a mutation.
+		 */
+		if (!isAuthenticated) {
+			return
+		}
```
ashleemboyer, line 97: I think this should be very first in the block to really emphasize that nothing is happening if there is no authenticated user.

ashleemboyer: After more reading, here's my full suggestion based on what you currently have coded (overly condensed just for the sake of pseudocode):

```typescript
useEffect(() => {
  // no computation needed to do the conditional check, var is already available
  if (!isAuthenticated) return
  
  // no computation needed to do the conditional check, var is already available
  if (typeof tutorialProgressStatus == 'undefined') return
  
  // requires computation to do the conditional, but less than the next block
  const isReadyToRunMutation = tutorialProgressMutations.isReady
  if (!isReadyToRunMutation) return
  
  // Requires the most computation to do conditional, so should be listed last in early bails
   const startRefId = startEntry?.target.getAttribute('data-ref-id')
  const endRefId = endEntry?.target.getAttribute('data-ref-id')
  const compareId = `${tutorialId}_${collectionId}`
  const isStaticIdsMismatch = startRefId !== compareId || endRefId !== compareId
  if (isStaticIdsMismatch) return
  
  // ... rest of the stuff
}, [])
```

---

What do you think about adding an additional early bail based on `tutorialProgressStatus`? As far as I can tell, we also don't need to do anything in this hook if `tutorialProgressStatus === 'complete'` since we can't have negative progress.

- If you do think it'd be good to add this, I'd place it second in the early bails list I've suggested, right after the `!isAuthenticated` one.

src/hooks/progress/use-update-tutorial-progress.ts
```diff
@@ -0,0 +1,186 @@
...
+
+		/**
+		 * We don't want to run multiple mutations at once.
+		 * To avoid this, tutorialProgressMutations provided `isReady`,
+		 * which we respect here.
+		 */
+		const isReadyToRunMutation = tutorialProgressMutations.isReady
+		if (!isReadyToRunMutation) {
+			return
+		}
```
ashleemboyer, line 116: I'm not sure if this check is needed since React Query already handles de-deduplicating network requests for us. It's totally cool to leave it for now for the sake of time, but could you add a ticket to our backlog to explore whether or not this is necessary? 🙏🏻

src/hooks/progress/use-update-tutorial-progress.ts
ashleemboyer: I'm in agreement here; this `useEffect` is quite long at over 100 lines. I think we could  quickly break it into a couple of helpers. 😊

- `canUpdateTutorialProgress`: this would contain all of the early bail `if` conditions that just do `return` when the conditions are true.
- `updateTutorialProgress`: this would run the code within the `if (needsInitialProgress || hasPositiveProgress)` block.

Feel free to use any other function names! These are just my quick ideas for naming.

ashleemboyer (changes requested):
This implementation is so neat! Thanks again for breaking the new hook into a separate PR. It really helped me wrap my head around this specific feature, helping me give this code the attention it deserves. 💖

Changes I'd really like to see before merge, in order of highest to lowest preference:

- Creating some helpers to shorten the long `useEffect`
- Adding an `if` to the `useEffect` that bails early if the tutorial's progress status is "complete"
- Reordering the early bails so the lowest computation ones are first

---

Some additional general commentary that didn't fit as individual comments:

- Question: What happens when a user loads a _super_ short tutorial for the first time? If I'm understanding the code correctly, the order of operation would be:
  - Try to fetch progress, get a 404
  - Do a POST request to create new progress entry w/ "visited" status
  - End ref is in view, so a PUT request happens to update status to "complete"
- Follow-up question: if this is what happens, would we be able to just go skip the immediate update and create the progress entry with "complete" status? Definitely out of scope for this PR!! 😊 (Also not necessary to do if we run out of time, but a polish ticket in the backlog would be lovely!)
- General thought: I wasn't expecting `useUpdateTutorialProgress` to be "passive", in that it doesn't return a function that we invoke. Maybe nothing to do here since the name is already long and you wrote an excellent block comment above it, but could a rename help provide more information? I like the name `useUpdateTutorialProgressWithScroll`, but it's SO long! 😆
- Future polish/optimization: there are a lot of re-renders happening for `TutorialView`. I think some of the suggestions I made in my review comments could help? Hard to say right now. But what do you think about a backlog ticket for exploring further optimization once the progress feature is stable and ready for polish/refinement?

src/views/tutorial-view/index.tsx
zchsh: To confirm Ashlee's comments, plan is:
- This PR makes individual queries. For now this means we fire off a bunch of individual requests, which is not optimal, but gives us convenience from a component architecture standpoint
- In a subsequent PR, I'll add in a batch query at the tutorial view level (similar to [useAllBookmarksById in TutorialCardsGridList, as used on eg collection view](https://github.com/hashicorp/dev-portal/blob/bcae5421fc244c091dc5085d8a1218ace4345dd1/src/components/cards-grid-list/components/tutorial-cards-grid-list/index.tsx#L36).
- The batch query (to be added in next PR) will get fired before the individual queries, and make one request. The subsequent individual queries won't end up making a bunch of requests cause the data they need will already be cached (more optimal! ☺️ )

src/hooks/progress/use-update-tutorial-progress.ts
zchsh: ✅ removed

src/hooks/progress/use-update-tutorial-progress.ts
zchsh: ✅ updated to split out `canUpdateTutorialProgress` & `updateTutorialProgress` (names are great, hardest problem solved haha 🙌 )

src/hooks/progress/use-update-tutorial-progress.ts
zchsh: ✅ sweet, this makes a ton of sense to me, as does the additional early bail - have updated accordingly!

src/hooks/progress/use-update-tutorial-progress.ts
zchsh: Absolutely, added a ticket: 🎟 [investigate if react-query automatically de-duplicates requests as we expect](https://app.asana.com/0/1202097197789424/1202915360953452/f)

For context, I added this in as otherwise I was seeing very quickly repeated mutations - I think because when a mutation is called, then the result of `useTutorialProgressMutations` updates in some way, and therefore re-triggers the `useEffect` hook which has a dependency on `tutorialProgressMutations`.

Here's a screenshot just after initial load where I've removed this condition and added a `RUNNING MUTATION` log in the newly split out `updateTutorialProgress` helper:

CleanShot 2022-09-02 at 11 16 50@2x

zchsh:
@ashleemboyer @kendallstrautman Thanks for the reviews here!

I've tried to address all feedback 💖 Specifically, I've made the following changes:
- Split out `canUpdateTutorialProgress` and `updateTutorialProgress` to reduce the line length of the `useEffect` (went with those names, they seem good to me!)
- Added an early bail if `progress == "complete" to `canUpdateTutorialProgress`
- Re-ordered the early bails in `canUpdateTutorialProgress` for lowest-computation first 👍

---

Re: _super_ short tutorial, I believe the order of operations would be:
- Tries to fetch progress, gets a 404
- All `canUpdateTutorialProgress` conditions are met, _and_ `endRef` is in view. So does a POST request with `complete` status.

Have validated this by zooming out a ton on a short tutorial view (Waypoint install) and filtering logs I added in that rnun every time the `updateTutorialProgress` gets called:

https://user-images.githubusercontent.com/4624598/188182235-8f47c00b-cdac-4fe1-8597-944d2df9fde5.mp4

---

Re: renaming `useUpdateTutorialProgress`, have renamed to `useTutorialProgressRefs` (since it returns `progressRefs`!). Let me know if that seems good 👍

---

Re: many `TutorialView` re-renders... My understanding is roughly that `TutorialView` re-render causes this PR's `useEffect` to run, rather than the other way around... so I think the _re-rendering_ of `TutorialView` is likely unrelated to work in this PR, as the only addition to the render body there is the `progressRefs`, and we've been really careful in the effect to never run duplicate mutations (which I think would be the only effect of the `useEffect` that would cause a re-render). I think a backlog ticket is a great idea though 👍 Added 🎟 [investigate reducing re-renders of TutorialView](https://app.asana.com/0/1202097197789424/1202915360953455/f)

ashleemboyer (approved):
Looks great!! ✨ Thanks for making these changes. 💖

- Glad to know that short tutorials already work as expected!!
- LOVE the `useTutorialProgressRefs` name 👏🏻

Nicely done! 🙌🏻

PR #901, opened 2022-08-31
feat: tutorial sidebar progress display

zchsh:
## 🔗 Relevant links

- [Asana task](url) 🎟️

### DIFF
--- a/src/components/sidebar/components/sidebar-nav-menu-item/index.tsx
+++ b/src/components/sidebar/components/sidebar-nav-menu-item/index.tsx
@@ -116,7 +116,7 @@ const SidebarNavLinkItem = ({ item }: SidebarNavLinkItemProps) => {
 						<SidebarNavMenuItemBadge {...(item as $TSFixMe).badge} />
 					) : undefined
 				}
-				icon={isExternal ? <IconExternalLink16 /> : undefined}
+				icon={isExternal ? <IconExternalLink16 /> : item.rightIcon ?? null}
 			/>
 		</>
 	)
--- a/src/components/sidebar/types.ts
+++ b/src/components/sidebar/types.ts
@@ -1,4 +1,4 @@
-import { ReactNode } from 'react'
+import { ReactElement, ReactNode } from 'react'
 import { SidebarBackToLinkProps } from './components/sidebar-back-to-link'
 import { SidebarMobileControlsProps } from './components/sidebar-mobile-controls'
 
@@ -120,6 +120,7 @@ interface MenuItem {
 	title?: string
 	heading?: string
 	isOpen?: boolean
+	rightIcon?: ReactElement
 }
 
 interface SidebarBaseProps {
--- a/src/components/tutorials-sidebar/index.tsx
+++ b/src/components/tutorials-sidebar/index.tsx
@@ -64,13 +64,14 @@ function TutorialViewSidebarContent({ items }: SectionListProps) {
 function SectionList({ items }: SectionListProps) {
 	return (
 		<ul className={s.listRoot}>
-			{items.map(({ text, href, isActive }: ListItemProps) => {
+			{items.map(({ text, href, isActive, rightIcon }: ListItemProps) => {
 				return (
 					<ListItem
 						key={`${text}${href}`}
 						text={text}
 						href={href}
 						isActive={isActive}
+						rightIcon={rightIcon}
 					/>
 				)
 			})}
@@ -78,8 +79,10 @@ function SectionList({ items }: SectionListProps) {
 	)
 }
 
-function ListItem({ href, isActive, text }: ListItemProps) {
-	return <SidebarNavMenuItem item={{ isActive, title: text, href }} />
+function ListItem({ href, isActive, text, rightIcon }: ListItemProps) {
+	return (
+		<SidebarNavMenuItem item={{ isActive, title: text, href, rightIcon }} />
+	)
 }
 
 function SectionTitle({ text }: SectionTitleProps) {
--- a/src/components/tutorials-sidebar/types.ts
+++ b/src/components/tutorials-sidebar/types.ts
@@ -1,3 +1,4 @@
+import { ReactElement } from 'react'
 import { SidebarProps } from 'components/sidebar'
 
 export interface TutorialSidebarProps {
@@ -37,6 +38,7 @@ export interface ListItemProps {
 	href: string
 	isActive?: boolean
 	text: string
+	rightIcon?: ReactElement
 }
 
 export interface SectionListProps {
--- a/src/components/sidebar/components/sidebar-nav-menu-item/types.ts
+++ b/src/components/sidebar/components/sidebar-nav-menu-item/types.ts
@@ -25,6 +25,7 @@ interface SidebarNavMenuItemProps {
 type SupportedIconName = 'home' & ProductSlug
 interface SidebarNavLinkItem extends MenuItem {
 	leadingIconName?: SupportedIconName
+	rightIcon?: ReactElement
 }
 
 interface SidebarNavLinkItemProps {
--- a/src/views/tutorial-view/utils/index.tsx
+++ b/src/views/tutorial-view/utils/index.tsx
@@ -12,10 +12,14 @@ export function formatTutorialToMenuItem(
 	currentPath: string
 ): ListItemProps {
 	const path = getTutorialSlug(tutorial.slug, collectionSlug)
+
+	const rightIcon = <span style={{ color: 'red' }}>Icon</span>
+
 	return {
 		text: tutorial.name,
 		href: path,
 		isActive: path === currentPath,
+		rightIcon,
 	}
 }
 


src/components/sidebar/components/sidebar-nav-menu-item/index.tsx
```diff
@@ -116,7 +116,7 @@ const SidebarNavLinkItem = ({ item }: SidebarNavLinkItemProps) => {
 						<SidebarNavMenuItemBadge {...(item as $TSFixMe).badge} />
 					) : undefined
 				}
-				icon={isExternal ? <IconExternalLink16 /> : undefined}
```
zchsh, line 119: `SidebarNavMenuItem` now supports a `rightIcon` property. This seems necessary to render the progress icon.

src/components/tutorials-sidebar/tutorial-progress-icon/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+	if (status == 'complete') {
+		return <IconCheckCircle16 className={s.iconComplete} />
+	} else if (status == 'in_progress') {
+		return <IconCircleHalf16 className={s.iconInProgress} />
+	} else if (status == 'visited' && IS_DEV) {
+		/**
+		 * TODO: this should be removed before going to production,
+		 * but is useful for dev testing.
+		 */
+		return <IconCircle16 className={s.iconVisited} />
```
zchsh, line 19: 🚧 This has been useful for testing, leaving it for now, but should remove before merging. (Ditto with associated styles).

src/components/tutorials-sidebar/types.ts
```diff
@@ -39,8 +41,13 @@ export interface ListItemProps {
...
 }
 
+export interface TutorialListItemProps extends ListItemProps {
+	tutorialId: uuid
+	collectionId: uuid
+}
+
 export interface SectionListProps {
-	items: ListItemProps[]
+	children: ReactNode
```
zchsh, line 50: Have update this prop to be `children` for flexibility:
- tutorial landing and collection views pass regular `ListItem` for collection sidebar items
- tutorial view passes new `TutorialListItem` for tutorial sidebar items, and these hook into the progress state

src/views/tutorial-view/index.tsx
```diff
@@ -182,7 +182,7 @@ function TutorialView({
 			children: (
 				<TutorialViewSidebarContent
 					items={collectionCtx.current.tutorials.map((t) =>
-						formatTutorialToMenuItem(t, collectionCtx.current.slug, currentPath)
+						formatTutorialToMenuItem(t, collectionCtx.current, currentPath)
```
zchsh, line 185: Updated args here as we need both collection `id` (for tutorial progress, which is specific to collection context) as well as `slug` (as previously, used to build the tutorial URL).

zchsh:
@ashleemboyer @BRKalow @kendallstrautman This PR adds tutorial sidebar progress display, for tutorial items only.

Notes in `Anything else` section are likely relevant - intent in this PR is to focus on the UI part, and do the "batch query" stuff for the page in a subsequent PR.

src/components/sidebar/components/sidebar-nav-menu-item/index.tsx
BRKalow: nit: it might be helpful for readability to pull out `icon` into a variable above the JSX given we've got two conditionals in its assignment now.

src/components/tutorials-sidebar/types.ts
BRKalow: 💭 should this be:
```suggestion
	children: ReactNode[]
```

src/components/tutorials-sidebar/index.tsx
```diff
@@ -49,32 +52,78 @@ function CollectionViewSidebarContent({
...
+	collectionId,
+}: TutorialListItemProps) {
+	/**
+	 * Query for progress, and display the appropriate status icon
+	 */
+	const { tutorialProgressStatus } = useTutorialProgress({
+		tutorialId,
+		collectionId,
+	})
+	const rightIcon = <TutorialProgressIcon status={tutorialProgressStatus} />
```
kendallstrautman, line 123: 💭 perhaps we could make a wrapper component here that is like `TutorialListItemWithProgress`, in case we want to render this component elsewhere and not show the progress data.

kendallstrautman: See an example with the tutorial card here https://github.com/hashicorp/dev-portal/blob/main/src/components/tutorial-card/helpers/with-bookmark.tsx

src/components/tutorials-sidebar/index.tsx
```diff
@@ -49,32 +52,78 @@ function CollectionViewSidebarContent({
...
+	href,
+	isActive,
+	text,
+	tutorialId,
+	collectionId,
+}: TutorialListItemProps) {
+	/**
+	 * Query for progress, and display the appropriate status icon
+	 */
+	const { tutorialProgressStatus } = useTutorialProgress({
```
kendallstrautman, line 119: 🤔 Instead of each tutorial calling for its own progress data here. I think it would make more sense to call for the entire [collection's progress state](https://digital-api-specs.vercel.app/learn#tag/Collection-Tutorial-Progress/paths/~1collections~1{id}~1progress/get). That would make it one single call. And then the progress status would be passed down to these children. Thoughts?

kendallstrautman: [See related comment here ](https://github.com/hashicorp/dev-portal/pull/900/files#r961006560)

src/components/tutorials-sidebar/index.tsx
kendallstrautman: If we go with the collection call route, then disregard my comments above.

src/components/tutorials-sidebar/tutorial-progress-icon/tutorial-progress-icon.module.css
```diff
@@ -0,0 +1,28 @@
+.progressIcon {
+	width: 12px;
+	height: 12px;
+}
```
ashleemboyer, line 4: 🙋🏻 Purely out of my own curiosity, what led you to deciding on this approach? I personally probably would've done something like this, just based on my own habits:

```css
.iconComplete,
.iconInProgress,
.iconVisited {
  height: 12px;
  width: 12px;
}

.iconComplete {
  color: var(--token-color-palette-green-200);
}

.iconInProgress {
  color: var(--token-color-palette-blue-200);
}

.iconVisited {
  color: magenta;
}
```

OR I might have done something like this with variables and named the classes in a way that they could've been applied using the `classnames` package:

```css
.root {
  color: var(--icon-color);
  height: 12px;
  width: 12px;
}

.complete {
  --icon-color: var(--token-color-palette-green-200);
}

.in_progress {
  --icon-color: var(--token-color-palette-blue-200);
}

.visited {
  --icon-color: magenta;
}
```

No action needed here, just riffing 🎸

src/components/tutorials-sidebar/types.ts
```diff
@@ -39,8 +41,13 @@ export interface ListItemProps {
 	text: string
 }
 
+export interface TutorialListItemProps extends ListItemProps {
+	tutorialId: uuid
+	collectionId: uuid
+}
```
ashleemboyer, line 47: Should `Tutorial['id']` and `Collection['id']` be referenced here? Those should probably reference `uuid` as their source of truth if they're not. 💭

ashleemboyer:
Tested the functionality out locally, working great! 🙌🏻

I also looked closely at the network requests being made to make sure we're leveraging React Query as we expect, and that's solid. When the PUT requests are happening to update progress, and `setQueryData` is invoked on successful update mutation, the `useTutorialProgress` is automatically returning the new data to update the icon without doing an extra GET request! 👏🏻👏🏻👏🏻

src/components/sidebar/types.ts
```diff
@@ -120,6 +120,13 @@ interface MenuItem {
 	title?: string
 	heading?: string
 	isOpen?: boolean
+	/**
+	 * Optional icon to display at right of the menu item.
+	 * Note that if this menu item is an external link,
+	 * the `rightIcon` will be ignored and an external link
+	 * icon will be shown instead.
+	 */
+	rightIcon?: ReactElement
```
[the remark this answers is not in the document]
zchsh: oooh nice for sure, I like `trailingIcon` a lot, and of course consistency is key 💖 also dig `trailing / leading` cause it seems great for if (when?) we start worrying more about right-to-left language layouts 🙌

src/components/tutorials-sidebar/types.ts
zchsh: ☝️ I'm not sure! I see `ReactNode` in a ton of other places in our code base... but have never dug super deep into the types here, both seem valid 🤷‍♂️ For consistency I'd lean towards keeping it as-is, but happy to do a deep dive or adjust if there are any strong opinions.

src/components/tutorials-sidebar/index.tsx
zchsh: If we want to render this component without progress data, I think we can use `SidebarNavMenuItem` directly. 

That being said, can definitely see why we might want something like `with-bookmark` or even `connected-bookmark-button` for progress... Had noodled on that a bit in the spike with a higher-order-component, see [connect-tutorial-progress-component](https://github.com/hashicorp/dev-portal/blob/7785c299ab24658922ce28df0e6dce2efc78ee8a/src/components/tutorial-card/components/connect-tutorial-progress-component/index.tsx).

Would it be alright if we defer the refinement here to a future PR? I feel like splitting out something like `connect-tutorial-progress-component` might come more naturally in adding the connected tutorial card.

src/components/tutorials-sidebar/index.tsx
zchsh: @kendallstrautman @ashleemboyer ☝️ yep, exactly - it seems like with `react-query`, compared to direct API requests, rather than have to choose "should we do this individually or in batch" I think we can do both. Writing this out cause I'm still building my understanding of it:
- individual components that need to data always make their individual query. Ensures they have the data they need, and avoids prop drilling... but means a ton of separate requests as a result of those queries if we do nothing else.
- At the level of any view or component that renders many individual query-data-needing components, we can optimize the # of requests by making a batch query. My understanding is this populates the `react-query` cache so that all the lil subsequent individual queries don't result in a ton of separate requests.

src/components/sidebar/types.ts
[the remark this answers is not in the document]
zchsh: ✅ renamed!

src/components/tutorials-sidebar/tutorial-progress-icon/tutorial-progress-icon.module.css
zchsh: I think there are a couple things that make me lean towards `compose` in this case:

- **One rule block per selector**
    - with compose, I can write exactly one rule block for `.iconComplete`, and it forms a complete picture of the styles being applied. The use of `compose` means the connection between `.iconComplete` and `.progressIcon` is clear from looking at that one rule block.
    - Where in the first example, unless you can see the whole file at once, neither of the two rule blocks that apply to `.iconComplete` necessarily indicate that there's another place you need to look in order to get a complete picture of the relevant styles.

That's obviously not a big distinguishing factor in this case, but I dig "one rule block per selector" as a more general rule so that I don't ever have to worry "am i missing some other relevant styles somewhere else in the file". There's a more relevant example in the Notion page [Code Style & Practices > CSS Modules](https://www.notion.so/hashicorp/CSS-Modules-753f2647f7aa4a598e28990d9f0d05a8), under the `Group styles by element, not by state or media type` point under `Recommended Practices`.

- **Cleaner markup**
    - this is a really minor one compared to the above IMO... but in cases where either a single `compose` or the CSS custom props in the second example will work for a use case, I'd lean very slightly towards `compose` to keep the markup and style mapping simpler (exactly one associated `className` to look at).
    - I also find it _slightly_ easier to trace `compose` than CSS custom props, since the former is very explicit about where to look

That being said, I also really dig the use of CSS custom props for anything even just slightly more complex (eg, maybe the `--size` and `--icon-color` varies!)... so would be very on the fence deciding between the second example and `compose` as used in this PR 👍 Also `compose` seems to have rules around using it with nested selectors that I can never quite remember, so generally very on board with the CSS custom props approach 💯

src/components/sidebar/components/sidebar-nav-menu-item/index.tsx
zchsh: sweet, thanks for these suggestions 💖 ✅ updated!

src/components/tutorials-sidebar/types.ts
zchsh: ✅ ah great call, updated!

src/components/tutorials-sidebar/index.tsx
ashleemboyer: @zchsh yep, you're right on target! 🎯

src/components/tutorials-sidebar/tutorial-progress-icon/tutorial-progress-icon.module.css
ashleemboyer: Awesome! 👏🏻 Thank you for all the detail. This makes total sense. 😊

src/components/tutorials-sidebar/types.ts
ashleemboyer: Yeah this one is kinda weird! It feels like `ReactNode[]` would be the way to go because the `map` is sending multiple children to this component, but TypeScript also isn't complaining about `ReactNode` being used so maybe it's fine? 🤔

zchsh: Here's a relatively definitive answer: in `@types/react`, the `PropsWithChildren` type (used for `React.FC`, for example) uses `ReactNode`:

```
type PropsWithChildren<P> = P & { children?: ReactNode | undefined };
```

ref: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/fb66211010980641f01bcfe8ffbc0cc3d71989ca/types/react/index.d.ts#L799

src/components/tutorials-sidebar/tutorial-progress-icon/tutorial-progress-icon.module.css
```diff
@@ -0,0 +1,28 @@
...
+
+.iconInProgress {
+	composes: progressIcon;
+
+	/**
+	TODO: in design spec this is --token-color-foreground-action,
+	which is the same hex value (#1060ff). Very minor nitpick, but need to
+	check whether use of action was intentional, or whether use of palette-blue
+	would be more semantically correct (since this icon is not interactive.
+	*/
```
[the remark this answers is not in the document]
zchsh: Asked about this via Slack DMs, but as this is super minor, I imagine there's higher priority things, so going to leave as-is with this comment for now 👍 May have to merge like this 😬

src/lib/learn-client/types.ts
```diff
@@ -79,7 +79,7 @@ export type CollectionLite = Pick<
 >
 
 export interface Collection {
-	id?: uuid // uuid
+	id: uuid // uuid
```
zchsh, line 82: Collection `id` is required for Tutorial progress to work. I think it's not "optional" from a `learn-client` perspective, as it's always available from the API.

src/components/tutorials-sidebar/tutorial-progress-icon/index.tsx
zchsh: ✅ removed this as I think we're ready for final review & merge 👍

zchsh:
@ashleemboyer @BRKalow @kendallstrautman With #900 merged, this feels ready for final review 👍

I think I've address all feedback from the previous round, but let me know if I missed anything 🙇 Thanks all! 💖

BRKalow (approved):
Code looks good! Pulled it down and all seems to be working! 👏

PR #907, opened 2022-09-01
kevin/type app

thiskevinwang:
- chore(_app): improve types; export
- chore: refactor `boundary/api-docs`

## 🔗 Relevant links

- [Asana task](url) 🎟️

### DIFF
--- a/src/pages/_app.tsx
+++ b/src/pages/_app.tsx
@@ -19,7 +19,9 @@ import {
 	CurrentProductProvider,
 	DeviceSizeProvider,
 } from 'contexts'
-import fetchLayoutProps from 'lib/_proxied-dot-io/fetch-layout-props'
+import fetchLayoutProps, {
+	ComponentMaybeWithQuery,
+} from 'lib/_proxied-dot-io/fetch-layout-props'
 import { isDeployPreview, isPreview } from 'lib/env-checks'
 import { makeDevAnalyticsLogger } from 'lib/analytics'
 import EmptyLayout from 'layouts/empty'
@@ -50,7 +52,7 @@ export default function App({
 	pageProps: { session, ...pageProps },
 	layoutProps,
 	host,
-}) {
+}: CustomAppProps) {
 	useAnchorLinkAnalytics()
 	useEffect(() => makeDevAnalyticsLogger(), [])
 
@@ -80,6 +82,9 @@ export default function App({
 	 */
 	const allLayoutProps = {
 		...pageProps.layoutProps,
+		// @ts-expect-error - layoutProps is inferred from `fetchLayoutProps`,
+		// which current resolves to `unknown | null`.
+		// Spread types may only be created from object types.
 		...layoutProps,
 	}
 
@@ -119,7 +124,7 @@ export default function App({
 	)
 }
 
-App.getInitialProps = async ({ Component, ctx }) => {
+App.getInitialProps = async ({ Component, ctx }: GetInitialPropsParams) => {
 	// Determine the product being served through our rewrites so we can fetch the correct layout data
 	let proxiedProduct
 	if (ctx.pathname.includes('_proxied-dot-io/vault')) {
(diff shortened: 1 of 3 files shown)


src/pages/boundary/api-docs/[[...page]].tsx
```diff
@@ -21,15 +23,21 @@ const targetFile = {
...
-			<OpenApiPageContents
-				info={apiPageProps.info}
-				operationCategory={apiPageProps.operationCategory}
-				renderOperationIntro={apiPageProps.renderOperationIntro}
-			/>
-		</SidebarSidecarLayout>
+		<OpenApiPageContents
+			info={apiPageProps.info}
+			operationCategory={apiPageProps.operationCategory}
+			renderOperationIntro={null}
```
Comment (thiskevinwang) [inline src/pages/boundary/api-docs/[[...page]].tsx:39]: I removed `apiPageProps.renderOperationIntro` since that wasn't present, **and** `renderOperationIntro` expects a function, which isn't serializable (via `getStaticProps`) anyways.

src/pages/boundary/api-docs/[[...page]].tsx
```diff
@@ -135,4 +143,5 @@ export async function getStaticProps({ params }) {
 	}
 }
 
+ApiDocsView.layout = ApiDocsLayout
```
Comment (thiskevinwang) [inline src/pages/boundary/api-docs/[[...page]].tsx:146]: This file serves as a POC use case, leverage the newly created/exported `PageWithLayout` type.

src/pages/boundary/api-docs/[[...page]].tsx
```diff
@@ -21,15 +23,21 @@ const targetFile = {
 // The path to read from when running local preview in the context of the boundary repository
 const targetLocalFile = '../../internal/gen/controller.swagger.json'
 
-function ApiDocsView({ layoutProps, apiPageProps }) {
+import { SidebarSidecarLayoutProps } from 'layouts/sidebar-sidecar'
+import type { PageWithLayout } from 'pages/_app'
+
+function ApiDocsLayout(props: SidebarSidecarLayoutProps) {
+	return <SidebarSidecarLayout {...props} />
+}
```
Comment (BRKalow) [inline src/pages/boundary/api-docs/[[...page]].tsx:31]: Q: Is this extra layer necessary?

src/pages/boundary/api-docs/[[...page]].tsx
```diff
@@ -21,15 +23,21 @@ const targetFile = {
...
+
+type ApiDocsViewProps = InferGetStaticPropsType<typeof getStaticProps>
+const ApiDocsView: PageWithLayout<ApiDocsViewProps> = ({ apiPageProps }) => {
```
Comment (BRKalow) [inline src/pages/boundary/api-docs/[[...page]].tsx:34]: It would be interesting to figure out how to get `layoutProps` typed in `getStaticProps` based on a page's specified layout 💭

src/pages/boundary/api-docs/[[...page]].tsx
Comment (thiskevinwang) [reply src/pages/boundary/api-docs/[[...page]].tsx:31]: Not really, but it'd be nice to keep our "LayoutComponent" consumption consistent — i.e. use `<Page>.layout` rather than as a child.

Somewhat like this https://github.com/hashicorp/dev-portal/blob/06937b9b391539ab4d1970cde04d3f9019d37630/src/views/docs-view/index.tsx#L109-L110

Comment (BRKalow) [reply src/pages/boundary/api-docs/[[...page]].tsx:31]: Agreed, sometimes it needs to be as a child though if we're passing props which aren't serializable to the layout.

This should work 💭 

```ts
ApiDocsView.layout = SidebarSidecarLayout
```

src/pages/_app.tsx
```diff
@@ -152,3 +157,47 @@ App.getInitialProps = async ({ Component, ctx }) => {
...
+/**
+ * This is our custom param type for `App.getInitialProps`
+ */
+type GetInitialPropsParams = Omit<AppContext, 'Component'> & ComponentWithLayout
+
+/**
+ * This is our custom type for our Next.js `App` component
+ */
+type CustomAppProps = AppPropsWithLayout &
+	Awaited<ReturnType<typeof App['getInitialProps']>>
```
ashleemboyer, line 203: To keep this file as short as possible, could we move these to a file under `src/types/`? Maybe `src/types/layouts.ts`?

src/pages/boundary/api-docs/[[...page]].tsx
Comment (thiskevinwang) [reply src/pages/boundary/api-docs/[[...page]].tsx:31]: Oh yea good call, that is actually what I wanted :homer:

src/pages/boundary/api-docs/[[...page]].tsx
Comment (thiskevinwang) [reply src/pages/boundary/api-docs/[[...page]].tsx:34]: 💭 I haven't done any digging, but there's a `satisfies` TS operator coming soon that _might_ be what we need for this

src/pages/_app.tsx
thiskevinwang: Done

BRKalow (approved):
LGTM

PR #914, opened 2022-09-01
chore: adds link to 'HashiCorp Developer' nav icon

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksupdate-nav-hashicorp.vercel.app/well-architected-framework) 🔎
- [Asana task](https://app.asana.com/0/1202702246158644/1202863899279051) 🎟️

## 🗒️ What

To improve the UX for the waf pages, this link needed to be added to the base nav header. 

## 🤷 Why

To allow for ease of navigation between the waf views and the home page. 

## 🧪 Testing

- visit the [waf page](https://dev-portal-git-ksupdate-nav-hashicorp.vercel.app/well-architected-framework), click the link to the home page
- no styles should be affected

### DIFF
--- a/src/components/navigation-header/components/home-page-content/index.tsx
+++ b/src/components/navigation-header/components/home-page-content/index.tsx
@@ -6,6 +6,7 @@ import { NavigationHeaderDropdownMenu } from '..'
 import sharedNavStyles from '../../navigation-header.module.css'
 import s from './home-page-content.module.css'
 import { NavigationHeaderItemGroup } from 'components/navigation-header/types'
+import Link from 'next/link'
 
 const HomePageHeaderContent = () => {
 	const betaProductItems = []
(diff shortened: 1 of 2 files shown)


src/components/navigation-header/components/home-page-content/index.tsx
```diff
@@ -48,12 +49,17 @@ const HomePageHeaderContent = () => {
...
+				<a
+					aria-label="HashiCorp Developer Home"
+					className={sharedNavStyles.contentBeforeNav}
+				>
+					<InlineSvg
+						className={s.siteLogo}
+						src={require('../../img/logo-white.svg?include')}
+					/>
+				</a>
+			</Link>
```
ashleemboyer, line 62: Can we keep this wrapped in a div in case we need to add more stuff in it in the future?

```suggestion
			<div className={sharedNavStyles.contentBeforeNav}>
				<Link href="/">
					<a aria-label="HashiCorp Developer Home">
						<InlineSvg
							className={s.siteLogo}
							src={require('../../img/logo-white.svg?include')}
						/>
					</a>
				</Link>
			</div>
```

zchsh: Personally I don't have a strong opinion here, and would be fine with it either way — if we need to add stuff in the future, we can always add the wrapper `<div />` then 👍

zchsh (approved):
✅ Looks good to me! V convenient 👍

src/components/navigation-header/components/home-page-content/index.tsx
kendallstrautman: I'm not super opinionated so I'll add it back!

PR #922, opened 2022-09-02
fix: tutorial meta margin

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-kstutorial-meta-layout-fix-hashicorp.vercel.app/well-architected-framework/operational-excellence/operational-excellence-automate-infrastructure) 🔎
- [Asana task](url) 🎟️

## 🗒️ What

Fixes an issue of too much margin when there is no interactive lab button. 
Screen Shot 2022-09-02 at 1 00 21 PM

## 🧪 Testing

- [visit this page](https://dev-portal-git-kstutorial-meta-layout-fix-hashicorp.vercel.app/well-architected-framework/operational-excellence/operational-excellence-automate-infrastructure)
- and other tutorial pages with an [interactive lab ](https://dev-portal-git-kstutorial-meta-layout-fix-hashicorp.vercel.app/vault/tutorials/getting-started/getting-started-first-secret)to verify things work as expected.

## 💭 Anything else?

This issue mostly affects prod since the bookmark button renders on the previews.


src/components/tutorial-meta/index.tsx
```diff
@@ -56,15 +56,23 @@ export default function TutorialMeta({
...
-				</span>
+				{/**
+				 * TODO: remove this conditional once auth is released to prod,
+				 * since the bookmark button will always render. If we don't
+				 * conditionally render this and there is no interactive lab button,
+				 * a margin stacking occurs that disrupts the layout
+				 */}
+				{isInteractive || isAuthEnabled ? (
+					<span className={s.buttonGroup}>
+						<InteractiveLabButton />
```
ashleemboyer, line 67: I think we need to wrap this in `isInteractive` now since this block can be triggered if `isInteractive` is false and `isAuthEnabled` is true? 🤔

kendallstrautman: But if `isAuthEnabled` is true then the bookmark button will render which is okay!

kendallstrautman: Basically we don't want this wrapper div to render if there isn't a child button

kendallstrautman: The interactive lab button won't render if there isn't an active lab in the context. So that is handled elsewhere

ashleemboyer: Yes, true! But so will the `InteractiveLabButton` even if `isInteractive` is false.

Here's what I'm meaning to suggest:

```typescript
{isInteractive || isAuthEnabled ? (
	<span className={s.buttonGroup}>
		{isInteractive ? <InteractiveLabButton /> : null}
		{/*// NOTE! - hiding this component from prod until auth is enabled */}
		{isAuthEnabled ? (
			<TutorialMetaBookmarkButton
				tutorial={{ id: tutorialId, name: heading.text }}
			/>
		) : null}
	</span>
) : null}
```

ashleemboyer: Oops! Didn't refresh before sending that. I see what you're saying! `InteractiveLabButton` has this:

```typescript
if (!ctx.labId) {
	return null
}
```

ashleemboyer (approved):
All good to go!! 👍🏻

ashleemboyer:
Updating the branch with `main` in case that fixes the build error 🤔

ashleemboyer:
(Updating again because I merged #923 that was already up-to-date and ready for merge! 🙏🏻)

PR #925, opened 2022-09-02
feat: tutorial progress toasts

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

When a tutorial changes to `"complete"` on a tutorial view, displays a tutorial progress toast notification.

## 🤷 Why

To continue user progress implementation, for feature parity with <https://learn.hashicorp.com/>

## 🛠️ How

- Adds `get-all-progress` `learn-client` utility, to allow fetching batched progress data from the Learn API
- Adds `use-collection-progress` hook, to query collection progress by ID, using `get-all-progress`
- Adds `use-previous` utility function, to track changes in a variable (same as existing `learn` implementation)
- Adds `use-progress-toast` hook
    - Uses `use-previous` to detect when progress status for a tutorial changes
    - Uses `use-collection-progress` to detect when completing a specific tutorial means the collection is complete too

### Design Spec

Screen Shot 2022-08-19 at 6 13 34 PM

### Screenshot from this PR

CleanShot 2022-09-06 at 11 37 10@2x

### Screen recording from this PR

https://user-images.githubusercontent.com/4624598/188677241-e57d0fd1-b27d-4144-af9b-89a2eb876506.mp4

## 🧪 Testing

- [ ] Sign in
- [ ] Visit any tutorial view
- [ ] When you complete the first few tutorials in a collection...
    - [ ] The progress should be indicated in the sidebar, same as in #901
    - [ ] You should _also_ see a `Tutorial complete` toast pop up
- [ ] When you complete the last remaining incomplete tutorial in a collection...
    - [ ] Same as other tutorials, except you should see special green success toast, with a `Collection complete` message

## 💭 Anything else?

- This PR also includes two minor but general toast VQA fixes, which I noticed while working on the progress toasts. Figured they're both one-line CSS fixes, so made sense to include.

[task]: https://app.asana.com/0/1202097197789424/1202831197852602/f
[preview]: https://dev-portal-git-zstutorial-progress-toast-hashicorp.vercel.app/

### DIFF
--- a/src/components/toast/components/toast-display/toast-display.module.css
+++ b/src/components/toast/components/toast-display/toast-display.module.css
@@ -54,7 +54,6 @@
 
 .contentArea {
 	flex-grow: 1;
-	padding-top: 2px;
 }
 
 .title {
--- a/src/components/toast/components/toast-display/toast-display.module.css
+++ b/src/components/toast/components/toast-display/toast-display.module.css
@@ -1,84 +1,85 @@
 .root {
-  border-radius: 6px;
-  border: 1px solid;
-  box-shadow: var(--token-elevation-higher-box-shadow);
-  display: flex;
-  padding: 16px;
+	border-radius: 6px;
+	border: 1px solid;
+	box-shadow: var(--token-elevation-higher-box-shadow);
+	display: flex;
+	padding: 16px;
+	width: 100%;
 
-  &.color-neutral {
-    background: var(--token-color-surface-faint);
-    border-color: var(--token-color-border-strong);
+	&.color-neutral {
+		background: var(--token-color-surface-faint);
+		border-color: var(--token-color-border-strong);
 
-    --title-and-icon-color: var(--token-color-foreground-strong);
-  }
+		--title-and-icon-color: var(--token-color-foreground-strong);
+	}
 
-  &.color-highlight {
-    background: var(--token-color-surface-highlight);
-    border-color: var(--token-color-border-highlight);
+	&.color-highlight {
+		background: var(--token-color-surface-highlight);
+		border-color: var(--token-color-border-highlight);
 
-    --title-and-icon-color: var(--token-color-foreground-highlight-on-surface);
-  }
+		--title-and-icon-color: var(--token-color-foreground-highlight-on-surface);
+	}
 
-  &.color-warning {
-    background: var(--token-color-surface-warning);
-    border-color: var(--token-color-border-warning);
+	&.color-warning {
+		background: var(--token-color-surface-warning);
+		border-color: var(--token-color-border-warning);
 
-    --title-and-icon-color: var(--token-color-foreground-warning-on-surface);
-  }
+		--title-and-icon-color: var(--token-color-foreground-warning-on-surface);
+	}
 
-  &.color-success {
-    background: var(--token-color-surface-success);
-    border-color: var(--token-color-border-success);
+	&.color-success {
+		background: var(--token-color-surface-success);
+		border-color: var(--token-color-border-success);
 
-    --title-and-icon-color: var(--token-color-foreground-success-on-surface);
-  }
+		--title-and-icon-color: var(--token-color-foreground-success-on-surface);
+	}
 
-  &.color-critical {
-    background: var(--token-color-surface-critical);
-    border-color: var(--token-color-border-critical);
+	&.color-critical {
+		background: var(--token-color-surface-critical);
+		border-color: var(--token-color-border-critical);
 
-    --title-and-icon-color: var(--token-color-foreground-critical-on-surface);
-  }
+		--title-and-icon-color: var(--token-color-foreground-critical-on-surface);
+	}
 }
 
 .iconArea {
-  color: var(--title-and-icon-color);
-  flex-shrink: 0;
-  margin-right: 12px;
+	color: var(--title-and-icon-color);
+	flex-shrink: 0;
+	margin-right: 12px;
 
-  & svg {
-    display: block;
-  }
+	& svg {
+		display: block;
+	}
 }
 
 .contentArea {
-  flex-grow: 1;
-  padding-top: 2px;
+	flex-grow: 1;
+	padding-top: 2px;
 }
 
 .title {
-  composes: hds-typography-body-200 from global;
-  color: var(--title-and-icon-color);
-  font-weight: var(--token-typography-font-weight-semibold);
+	composes: hds-typography-body-200 from global;
+	color: var(--title-and-icon-color);
+	font-weight: var(--token-typography-font-weight-semibold);
 }
 
 .description {
-  composes: hds-typography-body-200 from global;
-  color: var(--token-color-foreground-primary);
+	composes: hds-typography-body-200 from global;
+	color: var(--token-color-foreground-primary);
 
-  &:not(:first-child) {
-    margin-top: 4px;
-  }
+	&:not(:first-child) {
+		margin-top: 4px;
+	}
 }
 
 .actions {
-  display: flex;
-  flex-wrap: wrap;
-  gap: 8px;
-  margin-top: 16px;
+	display: flex;
+	flex-wrap: wrap;
+	gap: 8px;
+	margin-top: 16px;
 }
 
 .closeArea {
-  flex-shrink: 0;
-  margin-left: 16px;
+	flex-shrink: 0;
+	margin-left: 16px;
 }
--- a/src/hooks/progress/use-collection-progress.ts
+++ b/src/hooks/progress/use-collection-progress.ts
@@ -1,0 +1,44 @@
+import { useQuery, UseQueryResult } from '@tanstack/react-query'
+import useAuthentication from 'hooks/use-authentication'
+import {
+	getAllProgress,
+	GetAllProgressResult,
+} from 'lib/learn-client/api/progress'
+import { Collection } from 'lib/learn-client/types'
+
+type QueryDataType = GetAllProgressResult
+
+interface UseCollectionProgressOptions {
+	collectionId: Collection['id']
+}
+
+interface UseCollectionProgressResult
+	extends Omit<UseQueryResult<QueryDataType>, 'data'> {
+	data: UseQueryResult<QueryDataType>['data']
+}
+
+/**
+ * Handles checking if there is progress for the given `collectionId`.
+ */
+function useCollectionProgress({
+	collectionId,
+}: UseCollectionProgressOptions): UseCollectionProgressResult {
+	// Get the current user's access token
+	const { session } = useAuthentication()
+	const accessToken = session?.accessToken
+
+	// Fetch progress records by `collectionId`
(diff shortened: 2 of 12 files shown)


src/components/toast/components/toast-display/toast-display.module.css
```diff
@@ -1,84 +1,84 @@
...
-  border: 1px solid;
-  box-shadow: var(--token-elevation-higher-box-shadow);
-  display: flex;
-  padding: 16px;
+	border-radius: 6px;
+	border: 1px solid;
+	box-shadow: var(--token-elevation-higher-box-shadow);
+	display: flex;
+	padding: 16px;
+	width: 100%;
```
zchsh, line 7: The meaningful changes here are:

### Addition of `width: 100%` to `.root`

Fixes 🎟️ [[vqa] fix toast alignment](https://app.asana.com/0/1202097197789424/1202930794848928/f).

CleanShot 2022-09-06 at 11 23 56@2x

### Removal of `padding-top: 2px` from `.contentArea`

Fixes minor alignment slash pixel-pushy issue.

CleanShot 2022-09-06 at 11 15 01@2x

---

The remaining changes are formatting changes. Not sure why this formatting thrash is happening, but feels like something worth investigating post-launch 👍

src/hooks/progress/use-collection-progress.ts
```diff
@@ -0,0 +1,44 @@
...
+
+interface UseCollectionProgressResult
+	extends Omit<UseQueryResult<QueryDataType>, 'data'> {
+	data: UseQueryResult<QueryDataType>['data']
+}
+
+/**
+ * Handles checking if there is progress for the given `collectionId`.
+ */
+function useCollectionProgress({
```
zchsh, line 23: The `use-collection-progress` hook is necessary for tutorial completion toast display because we want to display a special "collection complete" toast when the last remaining incomplete tutorial in a collection is completed.

src/hooks/progress/use-tutorial-progress-mutations/index.ts
```diff
@@ -44,6 +44,9 @@ const useTutorialProgressMutations = (): UseTutorialProgressMutationsResult => {
 				[TUTORIAL_PROGRESS_QUERY_ID, tutorialId, collectionId],
 				data
 			)
+
+			// Invalidate related collection progress
+			queryClient.invalidateQueries(['collectionProgress', collectionId])
```
zchsh, line 49: We want to invalidate any `use-collection-progress` queries for a tutorial progress update that involves a specific collection ID, as this new update affects collection progress.

src/hooks/use-previous.ts
```diff
@@ -0,0 +1,12 @@
+import { useRef, useEffect } from 'react'
+
+/**
+ * Keep track of the previous value of a variable
+ */
+export default function usePrevious<T = never>(value: T) {
```
zchsh, line 6: Swiped this from the existing `hashicorp/learn` codebase - in general the toast logic implementation in this PR is pretty similar. Used to detect when progress status for a tutorial changes.

src/lib/learn-client/api/progress/get-all-progress.ts
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * Accepts optional arrays of `tutorialIds` and `collectionIds`.
+ * When either of these are provided, we return all progress records
+ * that match either:
+ * - any one of the provided `tutorialIds`
+ * - any one of the provided `collectionIds`
+ *
+ * https://digital-api-specs.vercel.app/learn#tag/Progress/paths/~1progress/get
+ */
+const getAllProgress = async ({
```
zchsh, line 29: This does the API data fetching we need for `use-collection-progress`.

It'll also be used later for batch fetching tutorial progress to prime individual queries.

> **Note**: We could use the [collection-specific endpoint](https://digital-api-specs.vercel.app/learn#tag/Collection-Tutorial-Progress/paths/~1collections~1{id}~1progress/get) instead... but with `collectionId` passed, and used as a query param, this endpoint does the exact same thing, and will definitely be used for things like batch fetching tutorial progress for the sidebar. So figured it made sense to set up `get-all-progress` in `learn-client` first (heck, we may not even need a separate `get-collection-progress`? since `use-collection-progress` can always use `get-all-progress` under the hood).

src/views/tutorial-view/index.tsx
```diff
@@ -201,6 +202,17 @@ function TutorialView({
 		collectionId: collectionCtx.current.id,
 	})
 
+	/**
+	 * Display toast when progress changes to complete.
+	 */
+	useProgressToast({
```
zchsh, line 208: `useProgressToast` does the work, but we do of course need to hook it up in the view.

src/views/tutorial-view/utils/make-progress-toast/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+	} else {
+		return {
+			title: 'Tutorial complete!',
+			description: `You have ${remainingTutorialsCount} tutorials left in this collection.`,
+			icon: <IconCheckCircle24 className={s.tutorialCompleteIcon} />,
+		}
+	}
+}
+
+export default function makeProgressToast(remainingTutorialsCount: number) {
```
zchsh, line 33: Makes the progress toast! This felt straightforward, props to the toast system 👍

src/views/tutorial-view/utils/use-progress-toast.ts
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * Given a tutorialId, collectionId, and count of tutorials in the collection,
+ *
+ * Show progress toast when the specified tutorial changes progress,
+ * from some non-"complete" status to "complete".
+ *
+ * This toast will give information related to collection progress,
+ * which is why we need the array of tutorial IDs in the collection.
+ */
+export function useProgressToast({
```
zchsh, line 16: This is similar in many ways to the previous implementation in `hashicorp/learn`. I've leaned into comments here to explain the early returns from the effect and so on, but happy to dive in further if needed 🤿

zchsh:
@ashleemboyer @kendallstrautman @BRKalow This PR implements tutorial progress toast.

Mostly I just worked off the existing implementation - felt much more straightforward than the initial work of getting tutorial progress tracking working as expected 😅 Let me know what y'all think! 🙇‍♂️

src/views/tutorial-view/utils/use-progress-toast.ts
BRKalow, line 16: No action needed immediately here, just something to noodle on, but I would expect that the implementation of the toasts _might_ be simpler if we triggered them in our mutation `onSuccess` callbacks. 💭 In the effect here we're tracking the transition of the status value to determine when progress was updated, where the `onSuccess` callback would be the "definitive" point at which we know the state changes.

BRKalow (approved):
Pulled it down and it seems to be working well! Nice work

src/views/tutorial-view/utils/use-progress-toast.ts
zchsh: Sweet, I dig this 👍 I guess we'd need to access collection progress in the mutation `onSuccess` as well though, right? 🤔 Will put my noodle cap on during optimization mode... made a ticket for later: 🎟️ [[progress toast] maybe simplify implementation by triggering toasts from mutation onSuccess callback](https://app.asana.com/0/1202097197789424/1202938785223946/f)

BRKalow: Ah that is true 🤔 Now I'm wondering if the API could just return that information for us as part of the `POST` response 💭

Anyways, totally agree we can revisit at a later date!

src/lib/learn-client/api/progress/get-all-progress.ts
```diff
@@ -0,0 +1,56 @@
...
+}: GetAllProgressOptions): Promise<GetAllProgressResult> => {
+	// Add query params to the URL, if applicable
+	let url = PROGRESS_API_ROUTE
+	const qs = queryString.stringify({ tutorialIds, collectionIds })
+	if (qs !== '') {
+		url += `?${qs}`
+	}
+
+	// Make the GET request
+	const requestResult = await get(url, accessToken)
```
ashleemboyer, line 42: 🙋🏻 Do we need to handle a 404 status?

src/views/tutorial-view/utils/use-progress-toast.ts
ashleemboyer: I think we can and should make the tutorial complete toasts render `onSuccess` of mutations. We show some toasts for bookmark interactions in this way.

Collection complete will be a little more challenging since we're not mutating that data and only fetching it. For that, I'd suggest updating `useCollectionProgress` to accept an `onSuccess` option that gets passed as an option to its `useQuery` and rendering the collection complete toast from there based on the query's data. This approach _should_ remove the need for checking several different pieces of state.

Nothing to do here though! This works and we can come back when there's time. ✨

ashleemboyer (approved):
Tested locally and only noticed one thing! It can be resolved in another PR though.

- On really short tutorials that are marked complete immediately on load, the toast isn't showing up for me. The tutorial I observed this on was: `/nomad/tutorials/transport-security/security-concepts`

src/lib/learn-client/api/progress/get-all-progress.ts
zchsh: I don't think so - unlike with individual requests, we're always hitting the same route (`/progress`), so a 404 would be very unexpected. Even if a user has no progress records, we expect back `data` that's an empty array.

src/views/tutorial-view/utils/use-progress-toast.ts
zchsh: @ashleemboyer Sounds good!

I think there are definitely some challenges here... I think even if we used `useCollectionProgress` `onSuccess` to trigger the `"Collection complete"` toast, we'd still need to know collection progress wherever we trigger the other `"Tutorial complete, ? tutorials left"` (as we need the count, and we don't want to render a `"0 tutorials left"` toast).

Very on board with investigating further later, in 🎟️ [[progress toast] maybe simplify implementation by triggering toasts from mutation onSuccess callback](https://app.asana.com/0/1202097197789424/1202938785223946/f) 👍

zchsh:
@ashleemboyer ah great catch re: really short tutorials with complete-on-load. Have fixed that now ✅

Context on the fix for posterity: Issue was because we were bailing if `!prevStatus`, even though `null` is a useful `prevStatus`. Now we bail only if `typeof prevStatus == "undefined"`. This way, when status changes from `null` to `"complete"`, toast still happens:

### Client-side nav

https://user-images.githubusercontent.com/4624598/188970054-987384d2-6fee-4217-b72d-08b4ade678f5.mp4

### Full reload

https://user-images.githubusercontent.com/4624598/188970043-3dfb2052-7ad0-4273-8f6b-efb10e01a0d1.mp4

PR #938, opened 2022-09-07
Use `useErrorPageAnalytics` hook on 404 page

nandereck:
🎟️ [Asana Task](https://app.asana.com/0/1202791227659279/1202865459293217/f)

---
## 🗒️ What

This PR leverages the new `useErrorPageAnalytics` hook (see [this PR](https://github.com/hashicorp/web-platform-packages/pull/86)) for pushing 404-related data to Segment. This PR replaces the `window.analytics.track` call on the 404 page with said hook.

## 🤷 Why

It was recently discovered that, across our various web properties, we make the following call:

`window.analytics.track(window.location.href, { ... } )`

The first parameter of the `track` method is used downstream in our various data warehouses to create a new table. Since we're using `window.location.href`, we're creating a table for _every_ URL that 404s. This is not ideal for analytics.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
 				"@hashicorp/localstorage-polyfill": "^1.0.14",
 				"@hashicorp/mktg-global-styles": "^4.2.3",
 				"@hashicorp/mktg-logos": "^1.3.1",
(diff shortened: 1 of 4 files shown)


BRKalow (approved):
👍 👍

PR #942, opened 2022-09-07
Adding currentCommand to CommandBarContext

ashleemboyer:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/0/1202932738893664/f) 🎟️

## 🗒️ What

- Renames `setIsOpen` to `toggleIsOpen` and abstracts the toggling logic
- Adds `currentCommand` property to `CommandBarContext`'s state
- Exposes `setCurrentCommand` function from `CommandBarContext`
- Breaks `CommandBarState` `interface` into two new ones:
  - `CommandBarContextState`
  - `CommandBarContextValue`
- Renders a small demo in the `CommandBarDialog` body for testing `setCurrentCommand`

## 🤷 Why

This approach does some of the groundwork for supporting additional commands, with our current default being `"search"` since we don't have other commands yet.

## 🧪 Testing

- [ ] Go to the preview URL
- [ ] Make sure the dialog still opens with the button in the header
- [ ] Make sure the dialog still opens/closes with cmd/ctrl+k
- [ ] When the dialog opens the header should show "search" as the current command
- [ ] The body should have a button for toggling the current command
- [ ] Click the button to toggle the command
- [ ] The header text should update to show the correct command

### DIFF
--- a/src/components/command-bar/index.tsx
+++ b/src/components/command-bar/index.tsx
@@ -1,4 +1,10 @@
-import { createContext, useContext, useEffect, useState } from 'react'
+import {
+	createContext,
+	useCallback,
+	useContext,
+	useEffect,
+	useState,
+} from 'react'
 import {
 	CommandBarActivator,
 	CommandBarDialog,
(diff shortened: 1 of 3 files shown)


src/components/command-bar/index.tsx
```diff
@@ -45,21 +79,37 @@ const CommandBarProvider = ({ children }: CommandBarProviderProps) => {
...
 			document.removeEventListener('keydown', handleKeyDown)
 		}
-	}, [])
+	}, [toggleIsOpen])
 
 	return (
-		<CommandBarContext.Provider value={{ isOpen, setIsOpen }}>
+		<CommandBarContext.Provider
+			value={{ ...state, setCurrentCommand, toggleIsOpen }}
+		>
```
BRKalow, line 87: future 💭: we'll want to `useMemo` the context value here to minimize re-renders on components which access the context value.

BRKalow (approved):
Nice!

src/components/command-bar/index.tsx
ashleemboyer: Good call! I can totally make that update in [Make CommandBarDialog track current tags in its state](https://app.asana.com/0/1201147515801429/1202932738893665/f)

PR #943, opened 2022-09-07
[CommandBarActivator] Add keyboard shortcut Badges

ashleemboyer:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1201147515801429/1202932738893647/f) 🎟️

## 🗒️ What

- Updates `CommandBarActivator` to render two `Badge`s for the `CommandBarDialog` keyboard shortcut
  - First `Badge` is always the `command-16` icon
  - Second `Badge` is always "K"

## 📸 Design Screenshots

[Figma link](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=13182%3A168690)

image

## 🧪 Testing

- [ ] Go to the preview url
- [ ] Make sure the new badges are styled according to the Figma

## 💭 Anything else?

There are some hydration challenges with rendering the first `Badge` based on the user's OS because the only way to detect the OS is with JavaScript. For now I've set this up to just render the `command-16` icon and we'll have to revisit the idea of swapping the icons.


src/components/command-bar/components/activator/index.tsx
```diff
@@ -8,19 +11,37 @@ const CommandBarActivator = ({
 	visualLabel,
 }: CommandBarActivatorProps) => {
 	const { setIsOpen } = useCommandBar()
+	const isMacOs = getIsUserAgentMacOs()
```
BRKalow, line 14: 🤔 I believe this has the potential to cause hydration errors, as the initial markup will always be rendered with this as `false`

ashleemboyer: Want to render 2 icons and do our usual CSS magic to show/hide one of them?

BRKalow: I think that would be ideal! If it's possible to detect windows via CSS? Otherwise maybe we could render something that has both icons?

Screen Shot 2022-09-07 at 3 20 34 PM

ashleemboyer: Updated! The CSS magic didn't work because JavaScript was still needed to detect the OS. I also tried the idea in your screenshot, but the `/` isn't vertically aligning like regular characters. Went with just the `command-16` icon and have added a note to the PR description.

BRKalow (approved):

PR #945, opened 2022-09-07
feat: Update Docs Version Selector

pbortnick:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-pbnew-version-selector-hashicorp.vercel.app/vault/docs/secrets/kv) 🔎
- [Preview - product docs landing page](https://dev-portal-git-pbnew-version-selector-hashicorp.vercel.app/waypoint/docs) 🔍 
- [Preview - docs with long title](https://dev-portal-git-pbnew-version-selector-hashicorp.vercel.app/waypoint/docs/extending-waypoint/passing-values) 🔍 
- [Preview - doc with landing hero](https://dev-portal-git-pbnew-version-selector-hashicorp.vercel.app/vault/api-docs) 🔍 
- [Asana task](https://app.asana.com/0/1202097197789424/1202707964020042/f) 🎟️

## 🗒️ What

Update `<VersionContextSwitcher />` to use `<DropdownDisclosure />` component

## 🛠️ How

- Add `versions` to `DocsViewProps`
- Render in DocsView if versions is defined
- Do the same in the product docs landing page


src/components/version-context-switcher/version-context-switcher.module.css
```diff
@@ -1,50 +1,18 @@
...
-}
+.dropdownList {
+	& ul {
+		max-height: 234px;
+		overflow: scroll;
 
-.option {
-  /* composition */
-  composes: hds-typography-body-200 from global;
+		& li:first-child span {
```
pbortnick, line 6: As per design, 
> Ideally we would truncate in a way where we keep (latest) (beta) etc visible at all times.

src/components/version-context-switcher/version-context-switcher.module.css
```diff
@@ -1,50 +1,18 @@
...
+			display: -webkit-box;
+			-webkit-line-clamp: 2;
+			-webkit-box-orient: vertical;
+			overflow: hidden;
+		}
+	}
```
pbortnick, line 12: As per design, 
> Ideally we would truncate in a way where we keep (latest) (beta) etc visible at all times.

src/components/version-context-switcher/version-context-switcher.module.css
```diff
@@ -1,50 +1,28 @@
...
+		overflow: scroll;
 
-.option {
-  /* composition */
-  composes: hds-typography-body-200 from global;
+		& li:first-child span {
+			display: -webkit-box;
+			-webkit-line-clamp: 2;
+			-webkit-box-orient: vertical;
+			overflow: hidden;
```
pbortnick, line 20: Doubt we will run into this often but as per design:
> Ideally we would truncate in a way where we keep (latest) (beta) etc visible at all times.

src/views/product-downloads-view/components/downloads-section/downloads-section.module.css
```diff
@@ -20,6 +20,10 @@
 	/* Flex-grow to fill space when in separate row */
 	flex-grow: 1;
 	width: max-content;
+
+	@media (max-width: 402px) {
+		width: 100%;
+	}
```
pbortnick, line 26: Solves this issue on smaller screens

File

src/components/version-context-switcher/types.ts
```diff
@@ -14,14 +12,14 @@ export interface VersionContextSwitcherProps {
...
-	initialValue?: ContextSwitcherOption['value']
+	initialValue?: ContextSwitcherOption
 
 	/**
 	 * (optional) A function invoked when the value of the switcher has changed.
 	 * It is invoked after the component has run its internal code for handling
 	 * changes to the switcher.
 	 */
-	onChange?: ChangeEventHandler<HTMLSelectElement>
+	onChange?: (x: string) => void
```
kendallstrautman, line 22:
```suggestion
	onChange?: (version: string) => void
```
💭 Not sure if this is the right suggestion, but a more descriptive parameter would be good!

kendallstrautman:
@pbortnick I added the 'do not merge' label given your description note at the top, feel free to update as you need.

pbortnick:
> @pbortnick I added the 'do not merge' label given your description note at the top, feel free to update as you need.

Woops! Meant to delete that. Thanks tho 👍

BRKalow:
@pbortnick I think the intent here was to update the the version selector rendered on the bottom left of docs pages to use the new component, but not necessarily update the selector on the install pages (though we could align both of these with this new component!)

kendallstrautman (approved):
Looks great to me! 👏  Pending, others feedback 🙇‍♀️

src/components/docs-version-switcher/docs-version-switcher.module.css
```diff
@@ -1,114 +1,36 @@
...
+
+		& ul {
+			width: 100%;
+			max-width: 100%;
+		}
+	}
+
+	& ul {
+		max-height: 234px;
+		overflow: scroll;
```
BRKalow, line 22: suggestion: can we use `overflow: auto` here so that the scrollbars will only show up if necessary?
```suggestion
		overflow: auto;
```

Notice the unneeded horizontal scrollbar here:
Screen Shot 2022-09-09 at 1 04 04 PM

src/components/docs-version-switcher/docs-version-switcher.module.css
```diff
@@ -1,114 +1,36 @@
...
-    color: var(--token-color-foreground-strong);
-    text-decoration: none;
-  }
-
-  &:focus-visible {
-    color: var(--token-color-foreground-strong);
-    z-index: 1;
-  }
+.docsVersionSwitcher {
+	@media (--small) {
```
BRKalow, line 2: suggestion: let's use a custom media query from those defined [here](https://github.com/hashicorp/dev-portal/blob/brk.feat/a11y-announce-results/src/styles/custom-media.css#L9) to ensure consistent breakpoints!

src/views/docs-view/docs-view.module.css
```diff
@@ -1,3 +1,22 @@
...
 	margin-top: 48px;
 }
+
+.contentWithVersions {
+	position: relative;
+
+	& .versionSwitcherWrapper {
+		width: 100%;
+
+		@media (--medium-up) {
```
BRKalow, line 11: suggestion: let's use a custom media query from those defined [here](https://github.com/hashicorp/dev-portal/blob/brk.feat/a11y-announce-results/src/styles/custom-media.css#L9) to ensure consistent breakpoints!

src/components/docs-version-switcher/docs-version-switcher.module.css
pbortnick: [Done here](https://github.com/hashicorp/dev-portal/pull/945/commits/f199daa4cffe74cea11b9e341d590a679d7b9efd#diff-d6b6fc60cb7fc4bc24f617ad293058ae0b7f76ced0f7605fa3d4db08dbbc69a0L22-R22) 😄

src/components/docs-version-switcher/docs-version-switcher.module.css
pbortnick: [done](https://github.com/hashicorp/dev-portal/pull/945/commits/f199daa4cffe74cea11b9e341d590a679d7b9efd#diff-d6b6fc60cb7fc4bc24f617ad293058ae0b7f76ced0f7605fa3d4db08dbbc69a0R2)

src/views/docs-view/docs-view.module.css
pbortnick: [and done](https://github.com/hashicorp/dev-portal/pull/945/commits/f199daa4cffe74cea11b9e341d590a679d7b9efd#diff-4d5c3e4fe15d4318a0f3ae5ea5bdc0b6cfbf15494fd7ac1a26c63a1cac2fbbdfR11)

BRKalow (commented):
@pbortnick So I think we might need to adjust the server code a bit, as we have some additional logic which determines if we pass down `versions` to the layout:

https://github.com/hashicorp/dev-portal/blob/main/src/views/docs-view/server.ts#L331-L340

This is causing the version selector to appear on some sections which are technically "unversioned": https://dev-portal-git-pbnew-version-selector-hashicorp.vercel.app/terraform/registry/providers/docs

src/components/docs-version-switcher/docs-version-switcher.module.css
```diff
@@ -1,114 +1,36 @@
...
+		}
+
+		& > div {
+			width: 100%;
+		}
+
+		& ul {
+			width: 100%;
+			max-width: 100%;
+		}
```
BRKalow, line 17: suggestion / stylistic: I think having explicit classnames here can be helpful just to more clearly correlate with the underlying elements, e.g. when I go look at the component source I can see immediately where the styles are intended to be applied.

src/components/docs-version-switcher/docs-version-switcher.module.css
```diff
@@ -1,114 +1,36 @@
...
+	}
+
+	& ul {
+		max-height: 234px;
+		overflow: auto;
+
+		& li:first-child span {
+			display: -webkit-box;
+			-webkit-line-clamp: 2;
+			-webkit-box-orient: vertical;
```
BRKalow, line 27: 💭 Is this feasible without the `-webkit` prefixed properties / values? That might help browser compat, and would ensure we're relying on standard properties

src/components/docs-version-switcher/index.tsx
```diff
@@ -57,66 +52,46 @@ const DocsVersionSwitcher = ({ options }: DocsVersionSwitcherProps) => {
...
-			</NavigationDisclosure>
+				<DropdownDisclosureLabelItem>
+					{currentProduct.name}
+				</DropdownDisclosureLabelItem>
+				{options
+					// Hide currently selected version from dropdown list
+					.filter(
+						(option: VersionSelectItem) =>
+							option.version !== selectedOption.version
+					)
```
BRKalow, line 72: Not something actionable now, as I know this is in the design spec, but just an observation...it feels a little disorienting to have the contents of the dropdown change depending on the selected version. 💭

src/components/docs-version-switcher/index.tsx
```diff
@@ -57,66 +52,46 @@ const DocsVersionSwitcher = ({ options }: DocsVersionSwitcherProps) => {
...
+						} else {
+							href = getTargetPath({
+								basePath: `${currentProduct.slug}/${currentRootDocsPath.path}`,
+								asPath: currentPath,
+								version: option.version,
+							})
+						}
+						return (
+							<DropdownDisclosureLinkItem
+								aria-label={`${currentProduct.name} ${option.label}`}
```
BRKalow, line 86: question: Curious why you opted to add an aria label here? I would think that the link text would be adequate in this context.

src/components/landing-hero/landing-hero.module.css
```diff
@@ -6,6 +6,10 @@
 	&.hasSubtitle {
 		align-items: flex-start;
 	}
+
+	@media (--dev-dot-tablet-up) {
+		margin-right: 25%;
+	}
```
BRKalow, line 12: thought: I wonder how we could get this one to apply only when versions are present, like in `docs-view` via `.contentWithVersions`.

💭 Do you think we could add a non CSS modules class and reference it with `:global`

```css
&:global(.contentWithVersions) {
	@media (--dev-dot-tablet-up) {
		margin-right: 25%;
	}
}
```

BRKalow (commented):
Looking good! I left a minor comments, but overall I think we're really close here!

src/components/docs-version-switcher/index.tsx
pbortnick: Ahh I think that was leftover when I switched from `<DropdownDisclosureButton />`. Deleting!

src/components/docs-version-switcher/docs-version-switcher.module.css
pbortnick: Agreed, however the styles here are editing another component (`<DropdownDisclosure />`). We could add comments to the css to address this concern. Alternatively, we could add `className` props to the main component, but this may be overkill. What do you think?

src/components/landing-hero/landing-hero.module.css
pbortnick: @BRKalow is this what you had in mind?
https://github.com/hashicorp/dev-portal/pull/945/commits/65b44bb7d3331cbcd2b5995f05720e42c6da0beb

src/views/docs-view/docs-view.module.css
```diff
@@ -1,3 +1,33 @@
...
+		@media (--dev-dot-mobile) {
+			margin-top: 0;
+		}
+	}
+
+	@media (--dev-dot-tablet-up) {
+		& h1:first-child,
+		& :global(.landing-hero) {
+			margin-right: 25%;
+		}
```
BRKalow, line 31: Nice! Can we add a comment above this block just explaining what it's for and why we're doing it?

src/components/docs-version-switcher/docs-version-switcher.module.css
BRKalow: Ooh awesome! Thanks for linking that resource 😄 Makes sense!

src/components/docs-version-switcher/docs-version-switcher.module.css
BRKalow: Ah word, this makes sense! Maybe we can add a comment here explaining what it's for (making the dropdown full width on mobile)? It feels like this might be something we want to bake into the component at some point with a prop 💭

pbortnick: Definitely!

src/views/docs-view/docs-view.module.css
pbortnick: [Added here](https://github.com/hashicorp/dev-portal/pull/945/files#diff-4d5c3e4fe15d4318a0f3ae5ea5bdc0b6cfbf15494fd7ac1a26c63a1cac2fbbdfR32-R33) 😄

src/components/docs-version-switcher/docs-version-switcher.module.css
pbortnick: [Added](https://github.com/hashicorp/dev-portal/pull/945/files#diff-d6b6fc60cb7fc4bc24f617ad293058ae0b7f76ced0f7605fa3d4db08dbbc69a0R3)!

BRKalow (approved):
Nice work!

PR #951, opened 2022-09-08
[CommandBarContext] Update to track current tags

ashleemboyer:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1202097197789424/1202932738893665/f) 🎟️


BRKalow (approved):
LGTM! Agreed regarding exploring use of a reducer. I think eventually it could end up simplifying the state management as things get more complex.

PR #953, opened 2022-09-08
Add ContentTypeContext

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambcontent-type-context-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1201147515801429/1202932738893673/f) 🎟️

## 🗒️ What

- Adds new context in `src/contexts/current-content-type.tsx`
- Replaces `PageWithLayout` type with `CustomAppComponent`
- Adds `contentType` property to `CustomAppComponent`
- Renders `CurrentContentTypeProvider` in `_app.tsx`, getting the value from `Component.contentType`
- Sets docs and tutorials `contentType` on many views

## 🤷 Why

- We need to know the content type being viewed for the search results shown in the command bar. When in a tutorials context, the "Tutorials" tab with search results is opened on load, for example.
- This approach is more stable long-term than looking at the page URL to determine what content type is being viewed.

## 🧪 Testing

- [ ] Go to the preview URL
- [ ] In various pages, open the command bar
- [ ] Check that the "Current content type" shown in the dialog body is what you expect


src/types/_app.ts
```diff
@@ -1,20 +1,32 @@
...
+import { CurrentContentType } from 'contexts'
 
-import type { PageWithLayout } from './layouts'
+type CustomAppComponent<
+	PageProps = Record<string, unknown>,
+	InitialProps = PageProps
+> = NextPage<PageProps, InitialProps> & {
+	contentType?: CurrentContentType
+	layout?: FC
+}
```
BRKalow, line 12: 🤔 I don't think `CustomAppComponent` makes sense as a name in the context of its usage, which is a type for any of our page components. What do you think about `CustomPageComponent`? or `DeveloperPageComponent`?

src/contexts/current-content-type.tsx
```diff
@@ -0,0 +1,35 @@
...
+type CurrentContentType = 'global' | 'docs' | 'tutorials'
+
+const CurrentContentTypeContext = createContext<CurrentContentType>(undefined)
+
+interface CurrentContentTypeProviderProps {
+	children: ReactNode
+	currentContentType: CurrentContentType
+}
+
+const CurrentContentTypeProvider = ({
```
BRKalow, line 12: suggestion: A brief doc string here explaining what this provider is for would be very valuable 🙏

BRKalow (commented):
Looking great! I left a few suggestions

src/types/_app.ts
ashleemboyer: Ahhh you're right! I'll think on this a little more

src/contexts/current-content-type.tsx
ashleemboyer: Can do!

src/types/_app.ts
ashleemboyer: Updated to `CustomPageComponent` in https://github.com/hashicorp/dev-portal/pull/953/commits/0f071d57e59642b1135acb79dde4097f79b533d1 ✅

src/contexts/current-content-type.tsx
ashleemboyer: Added a block comment in https://github.com/hashicorp/dev-portal/pull/953/commits/efe66407da5f8fd65c48200b1205ee2cf853b454 ✅

src/contexts/current-content-type.tsx
```diff
@@ -0,0 +1,47 @@
...
+ * Stores the current content type being viewed in a page. This information is
+ * useful when we want to show a different UI based on the type of content a
+ * user is actively viewing.
+ *
+ * Example: Global Search will automatically show tutorial search results before
+ * docs search when `currentContentType` is "tutorials".
+ *
+ * To change the value of `currentContentType`, the `contentType` property can
+ * be added to a component that is rendered as a page. See `CustomPageComponent`
+ * for reference.
```
BRKalow, line 22: perfect!

PR #954, opened 2022-09-08
chore: sitemap

thiskevinwang:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

This introduces `next-sitemap` for sitemap generation at build

This will generate
- `/public/robots.txt`
- `/public/sitemap-0.xml`
- `/public/sitemap.xml`

## 🧪 Testing

https://developers.google.com/search/docs/advanced/sitemaps/build-sitemap?hl=en#autositemap


next-sitemap.config.js
```diff
@@ -0,0 +1,7 @@
+/** @type {import('next-sitemap').IConfig} */
+module.exports = {
+	siteUrl: process.env.SITE_URL || 'https://developer.hashicorp.com',
+	generateRobotsTxt: true, // (optional)
+	exclude: ['/swingset/*'],
```
BRKalow, line 5: I think we should also exclude `/onboarding/`, right @kendallstrautman ?

alexcarpenter: Noticed `/swingset` is still being included too.

thiskevinwang: Updated the matcher to `/swingset*`

next-sitemap.config.js
```diff
@@ -0,0 +1,7 @@
+/** @type {import('next-sitemap').IConfig} */
+module.exports = {
+	siteUrl: process.env.SITE_URL || 'https://developer.hashicorp.com',
+	generateRobotsTxt: true, // (optional)
+	exclude: ['/swingset*'],
```
BRKalow, line 5:
```suggestion
	exclude: ['/swingset*', '/onboarding/*'],
```

BRKalow:
@thiskevinwang It looks like this only picks up pages which are statically generated, is that accurate?

thiskevinwang:
> @thiskevinwang It looks like this only picks up pages which are statically generated, is that accurate?

@BRKalow I think that is the case according to this: https://github.com/iamvishnusankar/next-sitemap/issues/31... 

Do you think we should build it (with all pages) and commit the sitemap file, and remove it from the site build/deploy process?

BRKalow:
I think that would potentially duplicate a lot of the static generation step if we decoupled it from the build. I wonder if our SEO consultant could give us recommendations / insight into what the sitemap should include if we're not generating the full site statically.

thiskevinwang:
@BRKalow So it sounds like generating a _partially complete_ sitemap is **not** detrimental to our SEO.

Should we roll with this?

BRKalow (approved):
LGTM! We can check the production sitemap after this rolls out to see if there are any more tweaks we want to make.

PR #958, opened 2022-09-08
[CommandBar] Add DialogBody command property; Render Product Tag for search command when currentProduct exists

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambcommand-bar-product-tag-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1201147515801429/1202932738893658/f) 🎟️

## 🗒️ What

- Moves the header, body, and footer subcomponents of `CommandBarDialog` to their own files
- Renames `Command` type to `CommandBarCommand`
- Adds `DialogBody` property to `CommandBarCommand`
- Adds `useSetUpAndCleanUpCommandState` utility hook
- Adds a `currentCommand` toggle button to `CommandBarDialogBody` for demoing `useSetUpAndCleanUpCommandState`
- Adds `DialogBody` to search and settings command configurations
  - the `DialogBody` for settings is very basic for now
  - the `DialogBody` for search handles adding a `Tag` for `currentProduct` if it exists, and also handles cleaning up that state on unmount using the new `useSetUpAndCleanUpCommandState` utility hook
- Renders `currentCommand`'s `DialogBody` in `CommandBarDialogBody`
- Updates `addTag` to prevent duplicate tags from being added

## 🧪 Testing

- [ ] Go to the preview URL's home page
- [ ] Open the command bar
- [ ] There should not be a product Tag present
- [ ] Go to a product's page
- [ ] Open the command bar
- [ ] There should be a Tag present for the product
- [ ] Remove the Tag
- [ ] Close and reopen the command bar
- [ ] The Tag should be present again
- [ ] Close and reopen the command bar
- [ ] Only one Tag should be present
- [ ] Toggle between commands multiple times
- [ ] The dialog should change between the two commands how you expect it to
- [ ] There should be no state or React hooks issues when toggling


src/components/command-bar/components/dialog/header.tsx
```diff
@@ -0,0 +1,51 @@
...
+import Badge from 'components/badge'
+import { useCommandBar, CommandBarTag } from 'components/command-bar'
+import Tag from 'components/tag'
+import s from './command-bar-dialog.module.css'
+
+const CommandBarDialogHeader = () => {
+	const { currentCommand, currentTags, removeTag } = useCommandBar()
+
+	const { useOnInitialLoad = () => null } = currentCommand
+	useOnInitialLoad()
```
ashleemboyer, line 11: 🗒️ This is new!

src/components/command-bar/components/dialog/index.tsx
```diff
@@ -1,135 +1,32 @@
...
-	children,
 	isOpen = false,
 	onDismiss = () => null,
 }: CommandBarDialogProps) => {
+	// Get the current command from Command Bar state
+	const { currentCommand } = useCommandBar()
+
+	// Invoke currnet command's initial loading hook
+	const { useOnInitialLoad = () => null } = currentCommand
+	useOnInitialLoad()
```
BRKalow, line 18: 💭  This _might_ break the rules of hooks if/when someone switches to a command with a different `useOnInitialLoad` hook which itself has different calls to `useState`/`useEffect`/etc.

This feels like a potential justification for having commands support a custom `render` method, or custom body component, which would allow the command to run any arbitrary hooks / effects on render, without expanding the top-level `Command` interface.

ashleemboyer: Is there a good way to test this?

How do you envision the `render` method working, and where do you see it being invoked? We need something that will allow all parts of the dialog to be prepared for UI, which is why I've placed this here.

BRKalow: It could be tested by switching commands to something that has a different `useOnInitialLoad` hook.

I haven't given it much thought, but `render` would likely get called in `CommandBarDialogBody`:

```ts
  const { currentCommand } = useCommandBar()

	return (
		<div className={s.body}>
            {currentCommand.render()}
			<div>Current content type: {currentContentType}</div>
		</div>
	)
```

or...

```ts
  const { currentCommand } = useCommandBar()

	return (
		<div className={s.body}>
            {<currentCommand.Render />}
			<div>Current content type: {currentContentType}</div>
		</div>
	)
```

I don't think this would have any functional difference to what you're doing now. Effects don't block rendering in any way, so the sequencing would be the same.

```ts
const searchCommand = {
  // ...
  Render: () => {
    useOnInitialLoad()
    return (
      // search body
    )
  }
}
```

ashleemboyer:
@BRKalow made some updates! Added a button to the dialog body for testing. I think this is working better!

src/components/command-bar/components/dialog/header.tsx
ashleemboyer: Irrelevant now, so resolving 😁

src/components/command-bar/commands/search.tsx
```diff
@@ -1,12 +1,59 @@
...
+	const cleanUpCommandState = useCallback(() => {
+		if (currentProductTag) {
+			removeTag(currentProductTag.id)
+		}
+	}, [currentProductTag, removeTag])
+
+	/**
+	 * Leveraging the set up + clean up hook exposed by CommandBarDialog.
+	 */
+	useSetUpAndCleanUpCommandState(setUpCommandState, cleanUpCommandState)
```
BRKalow, line 45: neat! ✨

src/components/command-bar/commands/settings.tsx
```diff
@@ -1,12 +1,17 @@
 import { IconSettings24 } from '@hashicorp/flight-icons/svg-react/settings-24'
-import { SupportedCommand } from '../types'
+import { CommandBarCommand, SupportedCommand } from '../types'
 
-const settingsCommand = {
+const DialogBody = () => {
+	return <h1>Sutff for Settings command</h1>
```
BRKalow, line 5: Sutff 😄 
```suggestion
	return <h1>Stuff for Settings command</h1>
```

BRKalow (approved):
Looks good! I like this pattern of delegating rendering to the command, feels clean 💅

src/components/command-bar/commands/settings.tsx
ashleemboyer: HA good _sutff_ 🤣

PR #964, opened 2022-09-08
feat: adds custom icon card grid item support

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksadds-custom-docs-landing-icon-tiles-hashicorp.vercel.app/hcp/docs) 🔎
- [Asana task](https://app.asana.com/0/1202702246158644/1202932567024650) 🎟️

## 🗒️ What

This PR introduces a new `iconCardGridItems` property to the docs landing template. If this is passed in the content file, it will overrides the root docs paths as the source for generating the icon tile grid. 

This was used in the HCP designs and I do think it adds good context in terms of UX for hcp. Definitely open to alternative approaches here and a discussion as to whether we want to use this for other products, as I know it's been asked for. 

## 📸 Design Screenshots

Screen Shot 2022-09-08 at 4 16 28 PM

## 🧪 Testing

- [Visit this link](https://dev-portal-git-ksadds-custom-docs-landing-icon-tiles-hashicorp.vercel.app/hcp/docs)
- Verify the custom links render and populate the tiles
- Visit another [product docs landing](https://dev-portal-git-ksadds-custom-docs-landing-icon-tiles-hashicorp.vercel.app/nomad/docs), verify the root docs paths still generate the tiles. 

## 💭 Anything else?

The designs show a 2-up grid, but imo that isn't worth the effort here.


src/content/hcp/docs-landing.json
```diff
@@ -1,5 +1,27 @@
 {
 	"pageSubtitle": "Learn and develop your knowledge of HashiCorp Cloud Platform with these tutorials and code resources.",
+	"iconCardGridItems": [
```
kendallstrautman, line 3: Made this authorable...open to suggestions though if we'd prefer not to do this.

BRKalow (commented):
I think having this authorable is reasonable. The risk being that we could lose consistency across products, though we would need a conditional for HCP either way. 💭

zchsh (approved):
✅ 🙌 💯 This seems like a great way to achieve what we need for HCP!

📌  Now I'm looking at the `currentProduct.slug !== 'terraform'` conditional and thinking we could probably use like `iconCardGridItems: []` in `terraform/docs-landing.json` there instead... IMO definitely a future thing tho and definitely outside the scope of this PR 👍

PR #966, opened 2022-09-09
chore: improve getTutorials return typing

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Adds [function overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads) to the `getTutorials` type definition to refine the return type based on the arguments.

Currently, the return type is always `Promise<Tutorial[] | Omit<Tutorial, 'content'>[]>`, even though we can statically determine this based on the second argument, `withContent`. Function overloads allow us to declare different types for the function based on varying arguments, so we can say `if withContent is false, omit the 'content' property` and make that available to consumers.

Before:
Screen Shot 2022-09-09 at 9 33 55 AM

After:
Screen Shot 2022-09-09 at 9 33 30 AM

### DIFF
--- a/src/lib/learn-client/api/tutorial/index.ts
+++ b/src/lib/learn-client/api/tutorial/index.ts
@@ -38,7 +38,13 @@ export async function getTutorial(
 	throw error
 }
 
-// getTutorials
+/**
+ * Fetches tutorials based on provided identifiers. The following function overloads ensure the return type is accurate based on how the
+ * `withContent` argument is passed in.
+ *
+ * @param idsOrSlugs - the tutorial identifiers used to fetch
+ * @param withContent - determines whether or not content is included in the returned tutorial records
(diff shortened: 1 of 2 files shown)


thiskevinwang (approved):
Nice

src/lib/learn-client/api/tutorial/index.ts
```diff
@@ -39,6 +39,17 @@ export async function getTutorial(
 }
 
 // getTutorials
```
EnMod, line 41: ❓ : I like this for narrowing down types 💯 Would it make sense to change this comment to read `// getTutorials type safeguards` or similar, to denote that this series of functions is specifically for narrowing down `getTutorials`'s return type?

PR #969, opened 2022-09-09
feat: add `SelfReferentialCanonicalTag`

thiskevinwang:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-kevin-self-canonical-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/0/1202872751843252/f) 🎟️

## 🗒️ What

This updates `head-metadata` to generate a self-referential canonical `link`

## 🧪 Testing

Checkout this branch locally (or use the deploy preview) and use `cURL` to verify that a correct `link` tag is rendered on the server response

```
curl https://dev-portal-git-kevin-self-canonical-hashicorp.vercel.app/ | grep -i canonical
curl http://localhost:3000/boundary/api-docs | grep -i canonical
```


src/components/head-metadata/index.tsx
```diff
@@ -44,23 +45,26 @@ export default function HeadMetadata(props: HeadMetadataProps) {
...
-		>
-			<link rel="icon" href="/favicon.ico" sizes="any" />
-			<link rel="icon" href="/icon.svg" type="image/svg+xml" />
-			<meta name="viewport" content="width=device-width, initial-scale=1.0" />
-			<meta
-				name="google-site-verification"
-				content="zRQZqfAsOX-ypXfU0mzAIzb5rUvj5fA4Zw2jWJRN-JI"
-			/>
+		<>
+			<SelfReferentialCanonicalTag />
```
kendallstrautman, line 49: One note is that the tutorial view pages already have canonicals set because the content is duplicated across collections. Can we adjust this to be optional or make sure the tutorial views don't get this self-referential tag?

BRKalow: FWIW, I _think_ this might already be handled by `next/head`, it should dedupe tags and prioritize those which are rendered deeper in the react tree.

src/components/self-referential-canonical-tag/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+const useFullURL = (base: string = 'https://developer.hashicorp.com') => {
+	const { asPath } = useRouter()
+	return new URL(asPath, base).toString()
+}
+
+const SelfReferentialCanonicalTag = () => {
+	const canonicalURL = useFullURL()
+	return (
+		<Head key={canonicalURL}>
```
BRKalow, line 12: I don't think we need a `key` prop here.

src/components/self-referential-canonical-tag/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+			<link rel="canonical" href={canonicalURL} />
```
BRKalow, line 13: I think we can add a `key` prop here to ensure canonicals rendered in page components take priority. We might also have to add the same key to existing canonical tags rendered.

```suggestion
			<link rel="canonical" href={canonicalURL} key="canonical" />
```

ref: https://nextjs.org/docs/api-reference/next/head

thiskevinwang: Added: https://github.com/hashicorp/dev-portal/pull/969/commits/fef495a8670901cd6b0790b291fcf1b469b49b90

src/components/head-metadata/index.tsx
thiskevinwang: This should allow for deduping - https://github.com/hashicorp/dev-portal/pull/969/commits/fef495a8670901cd6b0790b291fcf1b469b49b90

BRKalow (approved):
LGTM!

Validated that the nested view canonical tags are taking precedence here: https://dev-portal-git-kevin-self-canonical-hashicorp.vercel.app/boundary/tutorials/hcp-getting-started

PR #971, opened 2022-09-09
feat: tutorial card progress display

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Implements tutorial card progress display.

## 🤷 Why

To match existing <https://www.learn.hashicorp.com> user progress functionality.

## 🛠️ How

- Updates `with-bookmark` `TutorialCard` helper to show progress too, renames to `with-auth-elements`
- Uses `with-auth-elements` in `TutorialCardsGridList`, which updates all views (except `/profile/bookmarks`)
- Many smaller updates to mesh with the above two updates, see self-review comments for details.

### Design specs

🎨 [Figma link][figma]

Screen Shot 2022-08-19 at 6 05 36 PM

### Screenshots from this PR

CleanShot 2022-09-12 at 11 26 47@2x

CleanShot 2022-09-12 at 11 33 11@2x

## 🧪 Testing

- [ ] Sign in
- [ ] Make some progress on some tutorials
    - You should continue to see progress reflected in the sidebar
- [ ] Visit any view with tutorial cards you've made progress on
    - You should see a progress icon & label displayed at the top of any tutorial card that is `in_progress` or `completed`
    - Note that `visited` progress is intentionally not displayed on tutorial cards
    - 🚧 ... except on `/profile/bookmarks`, which will be handled separately  🎟️ [feat: progress display on /profile/bookmarks](https://app.asana.com/0/0/1202964504180144/f)

## 💭 Anything else?

- 📌 🚧 The Tutorial Cards on `/profile/bookmarks` will be handled in a separate PR.
    - I've set up a follow-up ticket, 🎟️ [feat: progress display on /profile/bookmarks](https://app.asana.com/0/0/1202964504180144/f), to update the display logic on that view.
    - Once this PR is finalized, intent is to use the approach finalized in this PR for that view as well. But including that additional piece in this PR felt like it might balloon scope a little bit.

- 📌 We could consider having `TutorialCard` always include the auth element _logic_ wrapper, rather than having the separate `with-auth-elements` helper component.
    - Note this doesn't mean we'd always show auth elements - just that whenever we render a tutorial card, we'd always check for authentication, and if the flag is enabled and a user is logged in, then we'd show auth elements on the card.
    - I think this could make sense, as I can't think of a situation where we'd want to show a tutorial card without auth elements for an authenticated user.
    - However, this would increase the breadth of the card refactor in this PR, so felt well outside the scope of this PR. Wanted to flag in case this merits a follow-up ticket.

[task]: https://app.asana.com/0/1202097197789424/1202831197852590/f
[preview]: https://dev-portal-git-zstutorial-card-progress-display-hashicorp.vercel.app/
[figma]: https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=12357%3A140992


src/components/cards-grid-list/components/tutorial-cards-grid-list/index.tsx
```diff
@@ -24,7 +24,7 @@ const TutorialCardsGridList = ({ tutorials, ...restProps }) => {
 		tutorialIds.push(tutorial.id)
 		cardsGridListItems.push(
 			<li key={tutorial.id}>
-				<TutorialCardWithBookmark {...tutorial} />
+				<TutorialCardWithAuthElements {...tutorial} />
```
zchsh, line 27: I've modified `TutorialCardWithBookmark` to include progress as well, and renamed to `TutorialCardWithAuthElements`.

src/components/tutorial-card/components/progress-icon-and-label/index.tsx
```diff
@@ -0,0 +1,15 @@
+import TutorialProgressIcon from 'components/tutorial-progress-icon'
+import { progressStatusToLabel } from 'lib/learn-client/api/progress'
+import { TutorialProgressStatus } from 'lib/learn-client/types'
+import s from './progress-icon-and-label.module.css'
+
+function ProgressIconAndLabel({ status }: { status: TutorialProgressStatus }) {
```
zchsh, line 6: This is the lil display component to show the progress icon and label.

src/components/tutorial-card/helpers/build-aria-label.ts
```diff
@@ -12,21 +12,25 @@ const PRODUCT_LABEL_MAP: Record<ProductOption, string> = {
...
+	return `${speakableDuration} tutorial.`
+}
+
 export function buildAriaLabel({
 	heading,
-	duration,
 	productsUsed,
 	hasVideo,
 	hasInteractiveLab,
+	eyebrowSlotAriaLabel,
```
zchsh, line 27: I needed a way to have the `aria-label` reflect progress. With our `eyebrowSlot` flexibility on the side of things we can render visibly, it felt right to reflect that flexibility in the `aria-label`. I've gone with `eyebrowSlotAriaLabel` to reflect this intent.

Note that we continue to default this to speakable duration in cases where `eyebrowSlot` renders `duration` - see `tutorial-card` for that default logic.

src/components/tutorial-card/helpers/format-tutorial-card.ts
```diff
@@ -1,32 +1,40 @@
...
-	 * If `collectionSlug` is provided, it overrides that default
-	 * collection context. Note that `collectionSlug` is not validated;
+	 * which we use if `collectionContext` is not provided.
+	 *
+	 * If `collectionContext` is provided, it overrides that default
+	 * collection context. Note that `collectionContext` is not validated;
 	 * please ensure it corresponds to a valid collection.
 	 */
-	collectionSlug?: string
+	collectionContext?: Pick<CollectionLite, 'id' | 'slug'>
```
zchsh, line 26: We continue to require `CollectionLite` data on the `tutorial` argument for `format-tutorial-card`. This is necessary to display progress, as we need the collection `id` to query progress.

I've modified the second argument, which allows for overriding the default collection context, in order to also require `{ slug, id }` for the "alternate" collection context (rather than just `{ slug }` as this was previously).

src/components/tutorial-card/helpers/with-auth-elements.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { TutorialCardBookmarkButton } from 'components/bookmark-button'
+import TutorialCard, { TutorialCardPropsWithId } from '..'
+import ProgressIconAndLabel from '../components/progress-icon-and-label'
+import { getSpeakableDuration } from './build-aria-label'
+
+/**
+ * Displays a TutorialCard, which shows additional user-data-specific elements
+ * for authenticated users.
+ */
+export function TutorialCardWithAuthElements(props: TutorialCardPropsWithId) {
```
zchsh, line 17: This builds off and renames the `with-bookmark` helper we'd set up previously.

src/components/tutorial-card/index.tsx
```diff
@@ -22,6 +26,7 @@ function TutorialCard({
 	hasVideo,
 	hasInteractiveLab,
 	eyebrowSlot,
+	eyebrowSlotAriaLabel = getSpeakableDuration(duration),
```
zchsh, line 29: We now support an `eyebrowSlotAriaLabel` prop. This allows consumers who are passing custom content into `eyebrowSlot` to hook into `build-aria-label` for the card, to ensure custom content is reflected in the `aria-label`.

This felt necessary to support including progress information in the `aria-label`.

We default to `getSpeakableDuration` if `eyebrowSlotAriaLabel` isn't provided, so that at minimum, we'll always have an `aria-label` equivalent to what we were generating previously.

src/components/tutorial-progress-icon/index.tsx
```diff
@@ -1,13 +1,25 @@
+import classNames from 'classnames'
 import { TutorialProgressStatus } from 'lib/learn-client/types'
 import { IconCheckCircle16 } from '@hashicorp/flight-icons/svg-react/check-circle-16'
 import { IconCircleHalf16 } from '@hashicorp/flight-icons/svg-react/circle-half-16'
 import s from './tutorial-progress-icon.module.css'
 
-function TutorialProgressIcon({ status }: { status: TutorialProgressStatus }) {
+function TutorialProgressIcon({
```
zchsh, line 7: Have lifted `TutorialProgressIcon` out of `sidebar`, as we also now use it in `TutorialCard`.

Have also added support for the slightly different `size` needed for use in the `tutorial-card` eyebrow.

src/components/tutorial-progress-icon/tutorial-progress-icon.module.css
```diff
@@ -0,0 +1,21 @@
+.progressIcon {
+	&.status_in_progress {
+		--color: var(--token-color-palette-blue-200);
```
zchsh, line 3: Have confirmed with design that we want to use `--token-color-palette-blue-200`.

Note: still need to implement the subtle colour variation on hover here. I'm leaving this for a separate task as I think it'll be enough to warrant a separate PR, since we'll need to pass in `active` and `hover` states somehow (perhaps through CSS custom properties? Or through props?) 🎟 [[progress-icon] add hover and active color states](https://app.asana.com/0/1202097197789424/1202964504180132/f)

src/hooks/progress/use-tutorial-progress.ts
```diff
@@ -14,9 +14,11 @@ import {
 import { PROGRESS_BATCH_QUERY_ID, TUTORIAL_PROGRESS_SINGLE_QUERY_ID } from './'
 import { useHasWaitedForQuery } from 'hooks/use-has-waited-for-query'
 
+type QueryResult = TutorialProgressStatus | null
```
zchsh, line 17: Realized this type was incorrect, as the queried result can be `null` if there is not progress record. I think I had this at some previous point, but reverted it at some point during the batch work.

src/lib/learn-client/api/progress/formatting.ts
```diff
@@ -38,3 +38,41 @@ export function progressStatusToPercent(
...
+	visited: 'Tutorial is visited.',
+	in_progress: 'Tutorial is in progress.',
+	complete: 'Tutorial is complete.',
+}
+
+/**
+ * Given a TutorialProgressStatus,
+ * Returns text representing the status suitable for screen reader use.
+ */
+export function progressStatusToAriaLabel(
```
zchsh, line 74: I needed utilities to get from a progress status to separate labels, one suitable for display and one suitable for screen readers. Felt natural to include them here in `formatting`.

src/views/collection-view/index.tsx
```diff
@@ -69,7 +69,7 @@ function CollectionView({
 			<CollectionTutorialList
 				isOrdered={ordered}
 				tutorials={tutorials.map((t: ClientTutorialLite) =>
-					formatTutorialCard(t, slug)
+					formatTutorialCard(t, collection)
```
zchsh, line 72: `formatTutorialCard` now needs `{ id, slug }`, not just `slug`. Passing in the `collection` felt like the clearest update here.

src/views/onboarding/collection-view/index.tsx
```diff
@@ -53,6 +53,7 @@ export default function OnboardingCollectionView({
 					isOrdered={ordered}
 					tutorials={tutorials.map((t: ClientTutorialLite) => ({
 						id: t.id,
+						collectionId: id,
```
zchsh, line 56: `collectionId` is now needed for `TutorialCard` (or rather `TutorialCardWithAuth` as used in `TutorialCardGrid`), in order to display progress.

src/views/product-downloads-view/helpers.ts
```diff
@@ -359,14 +360,42 @@ export const generateFeaturedTutorialsCards = async (
...
 		const defaultContext = tutorialData.collectionCtx.default
 		const tutorialLiteCompat = { ...tutorialData, defaultContext }
 
-		const formattedTutorialCard = formatTutorialCard(tutorialLiteCompat)
-		if (collectionSlug) {
-			formattedTutorialCard.url = `/${productSlug}/tutorials/${collectionSlug}/${tutorialSlug}`
+		/**
+		 * We have a collection context to use other than the default.
+		 */
+		let collectionContext = tutorialData.collectionCtx.default
```
zchsh, line 380: Updated the helper here to work with the revised `formatTutorialCard`.

The refactor here didn't feel the cleanest, so tried to be thorough with comments, and have also added this task (also linked to in code comments): 🎟 [[types] consider refactoring TutorialLite type](https://app.asana.com/0/1202097197789424/1202964504180140/f).

src/views/well-architected-framework/collection-view/index.tsx
```diff
@@ -44,6 +44,7 @@ export default function WellArchitectedFrameworkCollectionView({
 				isOrdered={ordered}
 				tutorials={tutorials.map((t: ClientTutorialLite) => ({
 					id: t.id,
+					collectionId: id,
```
zchsh, line 47: `collectionId` is now needed for `TutorialCard` (or rather `TutorialCardWithAuth` as used in `TutorialCardGrid`), in order to display progress.

zchsh:
@ashleemboyer @BRKalow @kendallstrautman This PR implements tutorial card progress display.

This represents a relatively significant refactor, so have tried to be thorough in self-review comments, and in inline comments where it felt like it made sense. Let me know what y'all think 🙇

src/components/tutorial-card/helpers/format-tutorial-card.ts
BRKalow: Question: Can a tutorial have different progress states across collections? 💭

src/components/tutorial-card/helpers/with-auth-elements.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+	/**
+	 * Determine whether to render progress status.
+	 * We only show progress for at a minimum state of `in_progress`.
+	 */
+	const isProgressLoaded = typeof tutorialProgressStatus !== 'undefined'
+	const meetsProgressThreshold =
+		parseInt(progressStatusToPercent(tutorialProgressStatus)) >=
+		parseInt(progressStatusToPercent(TutorialProgressStatus.in_progress))
+	const shouldRenderProgress = isProgressLoaded && meetsProgressThreshold
```
BRKalow, line 37: nit suggestion: what do you think about this? It felt clearer to me looking at this to use the status enum instead of coercing it to a number. We could also pull out the "display progress statuses" into a shared constant if we need to check elsewhere.

```suggestion
  const meetsProgressThreshold = [TutorialProgressStatus.in_progress, TutorialProgressStatus.complete].includes(tutorialProgressStatus)
	const shouldRenderProgress = isProgressLoaded && meetsProgressThreshold
```

BRKalow (approved):
This looks good! I pulled it down and tested it out, all seems to be working 👀 

Great job!

---

> 📌 We could consider having TutorialCard always include the auth element logic wrapper, rather than having the separate with-auth-elements helper component.

I think this could make sense. The only thing potentially worth calling out is that we're coupling the presentational components to auth & data fetching. Which is okay, but might make it hard to test and/or use in other contexts down the road.

src/components/tutorial-card/helpers/with-auth-elements.tsx
zchsh: I dig it! 👍

I was very on the fence... part of me is wondering whether it might make sense to keep the progress data in `percent` for longer, so that progress comparison things are more intuitive... Then we'd only convert the `progressPercentToStatus` right before we display it 🤔 The enums made the most sense to me at first since they were easiest to map to the implementation... but that definitely  seems a candidate for some refactoring 👍

zchsh:
@BRKalow thanks for the quick review and suggestions! 💖 

---

Re: always using auth wrapper, great call re: keeping the presentation layer 👍 I suppose we could still keep the layers separate (eg keep `TutorialCardWithAuthElements` and `TutorialCard`)? Maybe the intent would be less to collapse the layers, and more to generally avoid using the purely presentational `TutorialCard`, perhaps not exposing it as the default export of `components/tutorial-card` (instead using `TutorialCardWithAuthElements` as the default export). Does that seem like it'd get us the best of both worlds in a sense?

PR #980, opened 2022-09-12
feat: show progress on /profile/bookmarks tutorial cards

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Updates `/profile/bookmarks` view to display progress on tutorial cards shown there.

## 🤷 Why

For fully consistent display of progress on tutorial cards.

## 🛠️ How

Refactors `/profile/bookmarks` card rendering to use `TutorialCardsGridList`.

### Design specs

🎨 [Figma link][figma]

Screen Shot 2022-08-19 at 6 05 36 PM

### Screenshots from this PR

CleanShot 2022-09-12 at 11 33 11@2x

## 🧪 Testing

- [ ] Sign in
- [ ] Make some progress on some tutorials, and bookmark them
    - You should continue to see progress reflected in the sidebar
- [ ] Visit `/profile/bookmarks`
    - You should see a progress icon & label displayed at the top of any tutorial card that is `in_progress` or `completed`
    - Note that `visited` progress is intentionally not displayed on tutorial cards

## 💭 Anything else?

- 📌 EDIT: updated with more flexible `BookmarkButtonComponent` prop approach (thanks @BRKalow!). ~I don't think this is necessarily the ideal approach, it feels like adding options to `TutorialCard`.~
    - At the same time, I think we might want to consolidate auth elements into `TutorialCard` as defaults rather than things that need to be composed in.
    - So very on the fence about approach here, and open to alternate suggestions to achieve the aim of consistent progress display on tutorial cards across the experience 🙇 

[task]: https://app.asana.com/0/1202097197789424/1202964504180144/f
[preview]: https://dev-portal-git-zstutorial-card-progree-on-prof-abc762-hashicorp.vercel.app
[figma]: https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=12357%3A140992


src/components/tutorial-card/helpers/with-auth-elements.tsx
```diff
@@ -15,7 +16,12 @@ import { getSpeakableDuration } from './build-aria-label'
  * for authenticated users.
  */
 export function TutorialCardWithAuthElements(props: TutorialCardPropsWithId) {
-	const { id: tutorialId, collectionId, ...rest } = props
+	const {
+		id: tutorialId,
+		collectionId,
+		renderBookmarkConfirmationDialog,
```
zchsh, line 22: `TutorialCardWithAuth` now accepts an optional prop `renderBookmarkConfirmationDialog. This allows `/profile/bookmarks` to use the same component as we do elsewhere to render tutorial cards.

This doesn't feel ideal to me, but feels like it might be slightly preferable to doing the inverse (which felt like it would be recreating the logic of `TutorialCardWithAuthElements` in `render-bookmark-cards`).

src/components/tutorial-card/helpers/with-auth-elements.tsx
```diff
@@ -36,6 +42,14 @@ export function TutorialCardWithAuthElements(props: TutorialCardPropsWithId) {
 		parseInt(progressStatusToPercent(TutorialProgressStatus.in_progress))
 	const shouldRenderProgress = isProgressLoaded && meetsProgressThreshold
 
+	/**
+	 * In certain contexts, we want the bookmark button to render
+	 * with a confirmation dialog box.
+	 */
+	const BookmarkButton = renderBookmarkConfirmationDialog
```
zchsh, line 49: If `renderBookmarkConfirmationDialog` is truthy, we use `BookmarkButtonWithRemoveDialog`.

Note: could probably make sense to move `BookmarkButtonWithRemoveDialog` into `components/bookmark-button`. Left it in place for now for cleaner diff in initial review, happy to move if this approach feels acceptable overall.

src/views/profile/bookmarks-view/index.tsx
```diff
@@ -86,9 +88,21 @@ const ProfileBookmarksViewContent = () => {
 							</DropdownDisclosureButtonItem>
 						</DropdownDisclosure>
 					</span>
-					<CardsGridList fixedColumns={2}>
-						{bookmarks.sort(sortBy.sort).map(renderBookmarkCard)}
-					</CardsGridList>
+					<TutorialCardsGridList
```
zchsh, line 91: Bookmarks view now uses `TutorialCardsGridList`.

zchsh:
@ashleemboyer @BRKalow @kendallstrautman This PR gets tutorial card progress display from #971 into the `/profile/bookmarks` view.

As mentioned in review notes, I'd initially thought maintaining a strict composition-y approach would be best, and still kinda feel that way but doing so didn't end up feeling feasible given bandwidth. Figured I'd put this out there in case it feels reasonable. Very open to exploring alternate approaches to the `renderBookmarkConfirmationDialog` props used in this PR, if y'all have any suggestions 🙇

src/components/tutorial-card/helpers/with-auth-elements.tsx
BRKalow: Thought / alternative approach: this component could accept an optional prop, `BookmarkButtonComponent`:

```tsx
<TutorialCardWithAuthElements
	// ...
	BookmarkButtonComponent={BookmarkButtonWithRemoveDialog}
/>
```

6 one way, half dozen the other 🤷

BRKalow (commented):
Approach seems reasonable!

src/components/tutorial-card/helpers/with-auth-elements.tsx
zchsh: Note: have refactored to accept a `BookmarkButtonComponent` prop instead 👍

src/components/tutorial-card/helpers/with-auth-elements.tsx
zchsh: Nice, I like this! 💖 Updated to accept a `BookmarkButtonComponent` prop, rather than `renderBookmarkConfirmationDialog` 👍

zchsh:
@BRKalow I've updated here with your `BookmarkButtonComponent` suggestion - feels better to me! 💯

kendallstrautman (approved):
Went through the testing steps. Seems like a pragmatic approach! I think we'll land on the right combination once we get auth rendering on all the cards 👌

PR #981, opened 2022-09-12
feat(BrandedCallout): open ext url in new tab

thiskevinwang:
## 🔗 Relevant links

- [Dev Dot - Design VQA - Product Views](https://docs.google.com/document/d/1oGOdwsy1wktMTA4UCb2rwtVjKDxLlQiA7Irkl98cMZE/edit#heading=h.w9perc2u6b5a) 

## 🗒️ What

This updates our `BrandedCallout` to detect _external_ URLs, and open them in new tabs.

## 🧪 Testing

Visit `/boundary/tutorials` at the deploy preview link <TODO> and verify that the hero-link opens in a new tab.


src/views/product-tutorials-view/components/product-view-content/components/branded-callout/index.tsx
```diff
@@ -45,6 +45,12 @@ function BrandedCallout({
 				'--gradient-stop': `var(--token-color-${product}-gradient-faint-stop)`,
 		  }
 		: gradientDefault
+
+	const isExternal =
+		typeof location !== 'undefined' &&
+		cta.url.startsWith('http') &&
+		!cta.url.startsWith(location.origin)
```
ashleemboyer, line 52: 🙋🏻 Is it possible to use our `isAbsoluteUrl` helper here?

thiskevinwang: `isAbsoluteUrl` doesn't appear to differentiate between an _internal/external_ link, so it isn't particularly helpful in this situation

kendallstrautman (approved):
👍 thank you!

src/views/product-tutorials-view/components/product-view-content/components/branded-callout/index.tsx
ashleemboyer: We use it that way in several places throughout the app, did you find a particular case where it doesn't work in that way?

thiskevinwang: In this situation, I'd need to open:
- https://developer.hashicorp.com in the same tab
- https://www.hashicorp.com in a new tab

`isAbsoluteUrl` returns `true` for both of these.

---

If you're suggesting/asking if I can replace the following:
```diff
- cta.url.startsWith('http')
+ isAbsoluteUrl(cta.url)
```
that's certainly an option

ashleemboyer: Got it! This component can receive input that comes from a remote source too, so that makes sense why we can't say:

- absolute URL = external
- non-absolute URL = internal

Other uses like sidebar have made it possible to make these assumptions.

---

💭 What do you think about moving the `isExternal` check into its own file in `src/lib/`? I can see us benefitting from either updating `isAbsoluteUrl` to use it, or replace references to `isAbsoluteUrl` with the new helper!

thiskevinwang: For a reusable `isExternal` helper, I'd like to leverage the `host` value that we're grabbing in `_app` `.getInitialProps`, but I suspect that will be a more involved effort, so my thinking is to do that in a follow up PR

https://github.com/hashicorp/dev-portal/blob/cc3161acb52f08b35135664e166a0334093cfd01/src/pages/_app.tsx#L63-L64

PR #993, opened 2022-09-13
feat: updates vagrant landing

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksupdates-vagrant-landing-hashicorp.vercel.app/vagrant) 🔎
- [Asana task](url) 🎟️

## 🗒️ What

This PR adds a new 'callout' block to the product landing view to accommodate for the extra cta cards in the vagrant designs. It renames 'GetStartedBlock' --> 'CalloutBlock' to make it more generic. I retained the `get_started` property in the landing page schema for now, as that still seems like a relevant authoring helper. 

## 🧪 Testing

- [Visit this page](https://dev-portal-git-ksupdates-vagrant-landing-hashicorp.vercel.app/vagrant)
- Reference it against the [design spec. 
](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=13122%3A171671)

## 💭 Anything else?
Screen Shot 2022-09-12 at 5 44 13 PM

Right now the spacing is setup where headings have the large top-margin. Thinking about how to handle this best since the 'callout' card doesn't need to be paired with a heading. 

Also note that in the design spec, the intro text is show written out, which isn't included in this update.


src/content/vagrant/product-landing.json
```diff
@@ -23,6 +23,24 @@
 		]
 	},
 	"blocks": [
+		{
+			"type": "callout",
+			"heading": "Vagrant vs. Other Software",
+			"body": "Vagrant is not the only tool to manage virtual machines and development environments. ",
+			"iconCardLinks": [
+				{ "icon": "tools", "text": "CLI Tools", "url": "/todo" },
```
kendallstrautman, line 31: Not sure where to link these to.

src/views/product-landing/components/product-landing-blocks/index.tsx
```diff
@@ -14,6 +15,18 @@ function ProductLandingBlocks({ blocks }: ProductLandingBlocksProps) {
 			{blocks.map((block: ProductLandingBlock, idx: number) => {
 				const { type } = block
 				switch (type) {
+					case 'callout':
+						return (
+							<div key={idx} className={s.cardsMargin}>
```
kendallstrautman, line 20: Thoughts on how to best handle margin since these cards aren't paired with a heading? See the description for my full thoughts.

src/components/callout-card/index.tsx
```diff
@@ -6,17 +6,17 @@ import Card from 'components/card'
 import { developmentToast, ToastColor } from 'components/toast'
 import IconCardLink from 'components/icon-card-link'
 import StandaloneLink from 'components/standalone-link'
-import { GetStartedCardProps } from './types'
-import s from './get-started-card.module.css'
+import { CalloutCardProps } from './types'
+import s from './callout-card.module.css'
 
-function GetStartedCard({
+function CalloutCard({
```
zchsh, line 12: 👍 I dig the rename

src/views/product-landing/schema.ts
```diff
@@ -187,6 +162,60 @@ const ProductLandingGetStartedSchema = Joi.alternatives().try(
...
+	| CollectionCardsContentBlock
+	| LinkedCardsContentBlock
+	| CalloutContentBlock
+const ProductLandingContentBlockSchema = Joi.object({
+	type: Joi.string()
+		.required()
+		.valid('heading', 'tutorial_cards', 'linked_cards'),
+})
+	.when(Joi.object({ type: 'callout' }).unknown(), {
+		then: Joi.object({
```
zchsh, line 184: ⛏️ Nit: if we want, we could declare another schema for this, and keep things grouped by block type (types & schemas for each block type together, with the broader schema assembling a bunch of smaller ones rather than declaring things inline).

src/views/product-landing/components/product-landing-blocks/index.tsx
zchsh: Maybe we'd have a new `calloutCardMargin` class here? Based on the design spec, looks like that could be something like:

```css
.calloutCardMargin {
  margin-top: 48px 0;
}
```

zchsh (approved):
✅ Looks good to me so far!

Found it a bit hard to evaluate spacing relative to the design spec, since the content doesn't seem fully up-to-date yet (lots of cards in between `Vagrant vs Other Software` and the `Hacktoberfest` callout), but seems like something we could tweak as a follow up so personally would be fine with merging without getting into spacing nitty-gritty (yet) 👍 💯

src/views/product-landing/schema.ts
kendallstrautman: Good call, will update

PR #995, opened 2022-09-13
feat(notifications): implement welcome toast notifications

pbortnick:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-pbimplement-notifications-hashicorp.vercel.app) 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1202939992537941/f) 🎟️

## 🗒️ What

Implement welcome notifications on all pages except the homepage.

## 🛠️ How

- Create `makeWelcomeToast` function and run in `useEffect` in _app.tsx
- Steps in function:
1. Return if toast has been permanently dismissed or if session is active
2. Check if welcome toast was ever shown AND if the cookie for it has expired. If so, permanently dismiss the notification.
3. If cookies have never been set, set a cookie that expires in 3 months and another that expires later, which demonstrates that the notification has been viewed.
4. Activate session cookie
5. Display `toast`
- For permanent dismissal send `permanentlyDismiss` as `onDismissCallback` that sets dismissal cookie and removes all other cookies.

## 🧪 Testing

- [ ] Check that toast does not appear on [homepage](https://dev-portal-git-pbimplement-notifications-hashicorp.vercel.app)
- [ ] Navigate to another page and check that toast appears
- [ ] Stay on same page for 15 seconds and check that toast auto-dismisses
- [ ] Refresh page and check that toast does not appear

To check that dismissal works properly: 
- [ ] Comment out `inSession` check & run locally
- [ ] Click on the X in the toast
- [ ] Refresh page and make sure toast does not appear


src/lib/make-welcome-notification.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const WELCOME_COOKIE = 'dev-dot-first-welcome-notification'
+const INITIALIZATION_COOKIE = 'dev-dot-welcome-notification-initialized'
+const DISMISS_COOKIE = 'dev-dot-dismiss-welcome'
+const SESSION_COOKIE = 'dev-dot-welcome-session'
+
+function permanentlyDismiss() {
+	// Set dismiss cookie that expires after a year assuming we won't need these notifications in a year after GA
+	Cookies.set(DISMISS_COOKIE, true, {
+		expires: 365,
```
kendallstrautman, line 12: 💭 if we want this to be 1 year from GA, perhaps we pass a date object instead?

kendallstrautman (commented):
Did some testing and it works as expected! 👏 Just had one note on whether we want to set a hard date for this functionality. I could see it as something we forget to cleanup. 

Do we know if product wants us to release this notification immediately or after the products are launched?

src/lib/make-welcome-notification.ts
```diff
@@ -0,0 +1,61 @@
...
+	if (toastInitialized && !welcomeToastCookie) {
+		permanentlyDismiss()
+		return
+	}
+
+	// If no cookies have been set, set cookie that expires after 3 months
+	// Also set cookie that shows initialization of notification
+	if (!welcomeToastCookie && !toastInitialized) {
+		Cookies.set(INITIALIZATION_COOKIE, true, {
+			expires: 365,
```
kendallstrautman, line 43: Could be easier to read if we store some of these numbers in variables. Esp the 15000 one.

src/lib/make-welcome-notification.ts
pbortnick: Yeah sounds good! We can also add a check to see if the current date is later than the date we want to remove these on. You can see that [here](https://github.com/hashicorp/dev-portal/pull/995/commits/95791b152223a33dacd60f704d0c168caf0b3a93#diff-b043befe629def842d164cb756567813f1e0dcacc5ea16d005ae2a99d8b7188dR38). I am confirming now that we won't want these in ~a year and the other question you asked regarding release 👍

src/components/opt-in-out/helpers/make-beta-welcome-toast.tsx
```diff
@@ -1,15 +0,0 @@
...
-
-export function makeBetaWelcomeToast(optInPlatform: OptInPlatformOption) {
-	// Get the product name
-	const platformName = PlatformOptionTitles[optInPlatform]
-	// Make toast
-	toast({
-		color: ToastColor.highlight,
-		title: `Welcome to the ${platformName} Website Beta!`,
-		description:
-			'You can leave the Beta at anytime by clicking the Leave Beta button near the start of the main content area.',
```
kendallstrautman, line 12: 🤔 are they really sure they want to remove this toast while we still have beta opt in flows in effect? Seems odd to remove.

kendallstrautman: I won't fight it but it just seems weird to me lol

src/lib/make-welcome-notification.ts
```diff
@@ -0,0 +1,52 @@
+import Cookies from 'js-cookie'
+import { toast, ToastColor } from 'components/toast'
+
+const WELCOME_COOKIE = 'dev-dot-first-welcome-notification'
+
+const DISMISS_COOKIE = 'dev-dot-dismiss-welcome'
+const SESSION_COOKIE = 'dev-dot-welcome-session'
+const AUTO_DISMISS = 15000 // 15 seconds
+const MAX_DATE = new Date('1/31/2023')
```
kendallstrautman, line 9: nice 😄

src/lib/make-welcome-notification.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export function makeWelcomeToast() {
+	// Return if current date is later than the max date
+	if (new Date() > MAX_DATE) {
+		return
+	}
+
+	// Don't show toast if it has been dismissed or session is still active
+	const dismissWelcomeToast = Cookies.get(DISMISS_COOKIE)
```
kendallstrautman, line 27:
```suggestion
	const hasDismissedWelcomeToast = Cookies.get(DISMISS_COOKIE)
```
naming nit

src/lib/make-welcome-notification.ts
```diff
@@ -0,0 +1,52 @@
...
+		return
+	}
+
+	// If no cookies have been set, set cookie that expires after 3 months from GA
+	const welcomeToastCookie = Cookies.get(WELCOME_COOKIE)
+	if (!welcomeToastCookie) {
+		Cookies.set(WELCOME_COOKIE, true, {
+			expires: MAX_DATE,
+		})
+	}
```
kendallstrautman, line 39: 🤔 we aren't currently doing anything with this welcomeToastCookie. Do we still need it if we just always want to render one toast until the session cookie is set?

src/lib/make-welcome-notification.ts
```diff
@@ -0,0 +1,52 @@
...
+	// Set cookie that expires at end of session to detect whether or not to display
+	// We don't need a condition here since we will have already returned if session is active
+	Cookies.set(SESSION_COOKIE, true)
+
+	toast({
+		color: ToastColor.highlight,
+		title: `Welcome to HashiCorp Developer!`,
+		description: 'Your destination for documentation and tutorials.',
+		autoDismiss: AUTO_DISMISS,
+		onDismissCallback: permanentlyDismiss,
```
kendallstrautman, line 50: Interestingly, if you navigate to another page while the toast is displayed, the `onDismissCallback` fires and sets the perma dismiss cookie. This might not be an issue, but wanted to note it.

pbortnick: Interesting. I do think that is an issue

pbortnick: I think this can be resolved by adding `dismissOnRouteChange: false` to the toast options 👍 . Good catch!

src/components/opt-in-out/helpers/make-beta-welcome-toast.tsx
pbortnick: 🤷‍♀️

src/lib/make-welcome-notification.ts
pbortnick: Yeah you're right we don't need it if we are adding the date check. Alternatively we can keep it with the proper expiration, remove the date check, and then check if the `welcomeToastCookie` exists before rendering. What do you think?
https://github.com/hashicorp/dev-portal/pull/995/commits/ef580a5838b31ec49a7645c21f190ca0dd1f8dd4#diff-b043befe629def842d164cb756567813f1e0dcacc5ea16d005ae2a99d8b7188dR42

kendallstrautman: I think retaining the date check and just removing this cookie would be reasonable! That limits the amount of cookies we're storing 😄

kendallstrautman (approved):
Approving to unblock! But we should address that small welcome cookie discussion before merging 💯 Great work

src/lib/make-welcome-notification.ts
kendallstrautman: 🕵️ 👏

src/lib/make-welcome-notification.ts
pbortnick: I'm fine with that too since we are still involving 2 other cookies in the function 👍

src/lib/make-welcome-notification.ts
```diff
@@ -0,0 +1,43 @@
...
+	// Set dismiss cookie that expires after a year assuming we won't need these notifications in a year after GA
+	Cookies.set(DISMISS_COOKIE, true, {
+		expires: MAX_DATE,
+	})
+	Cookies.remove(SESSION_COOKIE)
+}
+
+export function makeWelcomeToast() {
+	// If the current date is later than the last day we want to show the notifications, return
+	if (new Date() > MAX_DATE) {
```
pbortnick, line 19: @kendallstrautman changed it back here 📆 😸

PR #998, opened 2022-09-13
fix(version switcher): add margin on docs landing page heading

pbortnick:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Add margin right to landing page heading when versions are defined

### Before
Screen Shot 2022-09-13 at 3 55 36 PM

### After
Screen Shot 2022-09-13 at 3 55 30 PM

### DIFF
--- a/src/views/product-root-docs-path-landing/index.tsx
+++ b/src/views/product-root-docs-path-landing/index.tsx
@@ -34,7 +34,7 @@ const ProductRootDocsPathLanding = ({
 	}
 
 	return (
-		<div className={s.root}>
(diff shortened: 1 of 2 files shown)


src/views/product-root-docs-path-landing/index.tsx
```diff
@@ -34,7 +34,7 @@ const ProductRootDocsPathLanding = ({
 	}
 
 	return (
-		<div className={s.root}>
```
pbortnick, line 37: This was only added for the version selector so a class is not needed here otherwise

BRKalow (approved):
LGTM!

PR #1001, opened 2022-09-13
fix: Updating handleTutorialLink to handle multiple query string params

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambfix-handle-tutorial-link-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1202958274929544/f) 🎟️

## 🗒️ What

- Updates `handleTutorialLink` to use [the `URL` API](https://developer.mozilla.org/en-US/docs/Web/API/URL) to analyze its given `nodePath`.
- Fixes `handleTutorialLink` to handle when there are multiple query string parameters, and preserve all params other than the `in` one.

## 🤷 Why

- The `URL` API helps make parsing complex urls easier.
- The method was previously assuming that when there is a query string in the given `nodePath`, that there was only one and that it was the `in` param (used for specifying a non-default collection slug for the tutorial)

## 🧪 Testing

- [ ] Go to [/consul/docs/intro#next-steps](https://dev-portal-git-ambfix-handle-tutorial-link-hashicorp.vercel.app/consul/docs/intro#next-steps)
- [ ] The "Get Started" link in the Next Steps section should no longer have `undefined` as the collection slug portion of the `href`


src/lib/remark-plugins/rewrite-tutorial-links/__tests__/rewrite-tutorial-links.test.ts
```diff
@@ -336,7 +336,6 @@ describe('rewriteTutorialLinks remark plugin', () => {
 		const contents = await remark()
```
kendallstrautman, line 336: It'd be good to add a test case to this file! Not totally necessary for this PR though, could be a follow up if we want to get this patch in fast.

src/lib/remark-plugins/rewrite-tutorial-links/utils/handle-learn-links.ts
```diff
@@ -26,40 +26,70 @@ export function handleTutorialLink(
...
-	if (hasQueryParam) {
-		// isolate the collection name from the query
-		const [, collectionSlug] = queryParam.split('/')
-		finalSlug = `/${product}/tutorials/${collectionSlug}/${filename}`
+	/**
+	 * Construct the full string of query parameters, ignoring the `in` param that
+	 * is not needed on this platform.
+	 *
+	 * NOTE: searchParams.toString() is not used for this because it outputs
+	 * encoded characters (such as the `/` character).
```
BRKalow, line 66: Question: Was this causing an issue? The docs explicitly state that the output of `toString()` is safe for use in a URL 💭 : https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/toString

src/lib/remark-plugins/rewrite-tutorial-links/__tests__/rewrite-tutorial-links.test.ts
ashleemboyer: Do the new test cases in `src/views/product-tutorials-view/helpers/__tests__/detect-and-reformat-learn-url.test.ts` suffice?

src/lib/remark-plugins/rewrite-tutorial-links/utils/handle-learn-links.ts
ashleemboyer: No issues, the `/` character is just more human readable than `%2F`.

src/lib/remark-plugins/rewrite-tutorial-links/utils/handle-learn-links.ts
```diff
@@ -26,40 +26,70 @@ export function handleTutorialLink(
...
+	 */
+	let queryString = ''
+	searchParams.forEach((value: string, key: string) => {
```
kendallstrautman, line 69: 💭 I think you could just remove `in` from the searchParams directly. 

`searchParams.delete(in)`?

ashleemboyer: We totally could, but I figure since `forEach` is being used anyway it's just the same.

kendallstrautman: Then you could do searchParams.toString or something so you wouldn't have to manually append the query stuff, I think.

kendallstrautman (approved):
👏 Ty for fixing this 🙇‍♀️  A few minor comments but if we want to get this patch in and follow up, that is fine!

src/lib/remark-plugins/rewrite-tutorial-links/utils/handle-learn-links.ts
ashleemboyer: I chose not to use `searchParams.toString()` on purpose, for a more human readable URL. Noted this in the block comment as well. 💖

src/lib/remark-plugins/rewrite-tutorial-links/__tests__/rewrite-tutorial-links.test.ts
kendallstrautman: I think it would be best to have test cases in the rewrite plugin itself! But for this PR those updated test cases work 👌

src/lib/remark-plugins/rewrite-tutorial-links/utils/handle-learn-links.ts
BRKalow: Looks like `URL.search` gives us what we want:

```
> const u = new URL('https://developer.hashicorp.com?in=collection/foo/bar&utm_param=docs')
> u.searchParams.delete('in')
> u.search
< '?utm_param=docs'
```

ashleemboyer: To clarify here's what I'm trying to explain with `searchParams.toString()`, regarding the `/` characters:

```
> const u = new URL('https://developer.hashicorp.com?in=collection/foo/bar&param_with_slashes=this/param/has/slashes')
> u.searchParams.toString()
< 'in=collection%2Ffoo%2Fbar&param_with_slashes=this%2Fparam%2Fhas%2Fslashes'
```

BRKalow: Yep, understood! The `.search` property doesn't appear to encode special characters in its output.

PR #1005, opened 2022-09-13
[handleTutorialLink] Using URL.search property for query string

ashleemboyer:
## 🗒️ What

This is a follow up to #1001. After it was merged, @BRKalow found `URL.search` would be useful for the query string.

## 🧪 Testing

There should be no functional or visual changes since #1001.


BRKalow (approved):
👏

PR #1014, opened 2022-09-14
[rewriteTutorialsLink] Handle links with complex query params

ashleemboyer:
## 🗒️ What

- Primarily updates the checks in `rewriteTutorialsLink` to be stronger
- Adds an error check (that doesn't break builds) for links that meet multiple link-type conditions (only one should be met)

## 🤷 Why

- In cases where a product's .io hostname is within a value for a query string param, the link is being detected as both a learn link _and_ a docs link. Because the `isDocsPath` check was being handled first, the link was always being handled as a docs link even if it was actually a learn link.

## 🧪 Testing

- [ ] Go to [/consul/docs/security/encryption#encryption](https://dev-portal-git-ambcomplex-params-in-tutorials-links-hashicorp.vercel.app/consul/docs/security/encryption#encryption)
- [ ] Click the "enable gossip encryption" link
- [ ] The link should not 404

### DIFF
--- a/src/lib/remark-plugins/rewrite-tutorial-links/index.ts
+++ b/src/lib/remark-plugins/rewrite-tutorial-links/index.ts
@@ -129,33 +129,22 @@ export function rewriteTutorialsLink(
 		}
 
 		if (isBetaProduct || isValidSection) {
+			// General variables
 			const urlObject = new URL(url, 'https://learn.hashicorp.com/')
 			const { origin, pathname } = urlObject
 			const urlWithoutOrigin = urlObject.toString().replace(origin, '')
-
-			const collectionPathRegex = new RegExp('^/collections')
-			const isCollectionPath = collectionPathRegex.test(pathname)
-			if (isCollectionPath) {
-				newUrl = handleCollectionLink(urlWithoutOrigin)
-			}
-
-			const tutorialPathRegex = new RegExp('^/tutorials')
-			const isTutorialPath = tutorialPathRegex.test(pathname)
-			if (isTutorialPath) {
-				newUrl = handleTutorialLink(urlWithoutOrigin, tutorialMap)
-			}
-
-			const productHubPathRegex = new RegExp(`^/${product}/?$`)
-			const isProductHubPath = productHubPathRegex.test(pathname)
-			if (isProductHubPath) {
-				newUrl = `/${product}/tutorials`
-			}
-
 			const productIOHostName = productSlugsToHostNames[product]
+
+			// Regexes for each path type
+			const collectionPathRegex = new RegExp('^/collections')
+			const tutorialPathRegex = new RegExp('^/tutorials')
+			const productHubPathRegex = new RegExp(`^/${product}/?$`)
+
+			// Derived path type booleans
+			const isCollectionPath = collectionPathRegex.test(pathname)
+			const isTutorialPath = tutorialPathRegex.test(pathname)
+			const isProductHubPath = productHubPathRegex.test(pathname)
 			const isDocsPath = origin.includes(productIOHostName)
-			if (isDocsPath) {
-				newUrl = handleDocsLink(urlWithoutOrigin, product as ProductSlug)
-			}
 
 			/**
 			 * Check if multiple conditions above were true. Only one should be true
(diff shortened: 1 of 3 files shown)


src/views/product-tutorials-view/helpers/__tests__/detect-and-reformat-learn-url.test.ts
```diff
@@ -166,6 +168,12 @@ describe('detectAndReformatLearnUrl', () => {
 				input: '/tutorials/waypoint/aws-ecs',
 				expected: '/waypoint/tutorials/deploy-aws/aws-ecs',
 			},
+			{
+				input:
+					'/tutorials/consul/gossip-encryption-secure?utm_source=consul.io&utm_medium=docs',
```
kendallstrautman, line 173: 💭 I wonder if they should remove these query params from the docs source anyways since they aren't necessarily relevant to devdot anymore?

kendallstrautman: Not saying we shouldn't support this case, just thinking outloud for content

src/lib/remark-plugins/rewrite-tutorial-links/index.ts
```diff
@@ -114,45 +129,63 @@ export function rewriteTutorialsLink(
...
-			const isProductHubPath = learnProductHub.test(nodePath)
-			const isDocsPath = nodePath.includes(productSlugsToHostNames[product])
-
-			// if its an external link, isolate the pathname
-			if (isExternalLearnLink || isDocsPath) {
-				const fullUrl = new URL(nodePath)
-				// removing the origin from the href instead of only using
-				// 'pathname' so that anchor links are included
-				nodePath = fullUrl.href.replace(fullUrl.origin, '')
+			const urlObject = new URL(url, 'https://learn.hashicorp.com/')
```
kendallstrautman, line 132: Definitely better to use `URL`! 💖 👏

src/lib/remark-plugins/rewrite-tutorial-links/index.ts
```diff
@@ -114,45 +129,63 @@ export function rewriteTutorialsLink(
...
+			 * at a time.
+			 */
+			const isUrlAmbiguous = getIsUrlAmbiguous({
+				isCollectionPath,
+				isTutorialPath,
+				isProductHubPath,
+				isDocsPath,
+			})
+			if (isUrlAmbiguous) {
+				throw new Error(`[rewriteTutorialsLink] found an ambiguous url: ${url}`)
```
kendallstrautman, line 171: If it is ambiguous, do we want to adjust what is returned, for example we may not want to rewrite it at all? Or do we allow for it to be rewritten from whichever condition last caught it?

ashleemboyer: I updated the function to return `newUrl || url`, so the original `url` is what should be returned in this case.

kendallstrautman: 🤔 well if the `newUrl` was set in one of those conditionals, then `url` wouldn't return right?

ashleemboyer: Ah right, since this check is happening after the conditionals. Let me do a reorder of things real quick so that doesn't happen.

kendallstrautman: It could be good to add a test case for an ambiguous url perhaps?

ashleemboyer: Just pushed the re-ordering update in https://github.com/hashicorp/dev-portal/pull/1014/commits/ae6f214b84ab8ac7487aa9f372545ad5145b2c5b. I'm updating docs sidebar to handle external Learn links tomorrow, and I'm going to add more tests and refactor a few things here as I work on it. It will make it a lot easier to come up with the ambiguous url test case. With that fast-follow in mind, are you okay with merging this without the ambiguous url test case?

kendallstrautman (approved):
LGTM! Thank you for these updates, big improvement 💯

PR #1015, opened 2022-09-14
feat: fix up packer plugins loading ref

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What
This PR adjusts logic for Packer plugins to ensure related assets are being loaded from the intended `hashicorp/packer` repo refs.

Specifically, this PR is concerned with the loading of the following assets:
- `plugins-manifest.json`
- `plugins-nav-data.json`
- The `.mdx` file for the `/plugins` root docs path
    - Note: all paths under `/plugins/*` are loaded from separate plugin repositories, not `hashicorp/packer`

This PR sets things up so that both `packer.io` and `dev-dot` load from the `stable-website` branch of `hashicorp/packer`.

## 🤷 Why

To ensure Packer plugins data is being resolved in a predictable and correct way.

## 🛠️ How

- Ensures Packer plugin utils accept and respect a `mainBranch` argument, which allows setting a consistent ref from which assets will be loaded
- Updates `pages/_proxied-dot-io/packer/plugins` files (for `dot-io`) and `pages/packer/plugins` files (for `dev-dot`) to set `mainBranch` to `stable-website`

## 🧪 Testing

- [ ] Preview should build and function as expected for https://developer.hashicorp.com/packer/plugins
    - Preview link: [/packer/plugins][preview]
- [ ] Preview should build and function as expected for https://www.packer.io/plugins
    - Use the product selector in the bottom left to switch to `Packer` to verify this

## 💭 Anything else?

> **Note**: have also added support for and used an `editBranch` argument, which in theory would affect the Edit on GitHub links of any local Packer plugins. In practice, all our plugins under `/plugins/*` are sourced from remote repos, so this logic will likely never be used... but felt right to add this for completeness & hopefully clarity.

[task]: https://app.asana.com/0/1202097197789424/1202979702108666/f
[preview]: https://dev-portal-git-zsuse-latest-version-ref-for-pa-b746e4-hashicorp.vercel.app/packer/plugins


src/components/_proxied-dot-io/packer/remote-plugin-docs/server.ts
```diff
@@ -25,9 +25,14 @@ import rehypePrism from '@mapbox/rehype-prism'
 // which we need to shim cause of how we're fetching remote content here
 import shimRemoteIncludes from 'lib/shim-remote-includes'
 
-async function generateStaticPaths({ navDataFile, remotePluginsFile }) {
+async function generateStaticPaths({
+	navDataFile,
+	remotePluginsFile,
+	mainBranch = 'main',
```
zchsh, line 31: I've ensured `remote-plugin-docs` utils consistently accept `mainBranch` as an option.

This way, at the page file level, we can set distinct settings for `dot-io` and `dev-dot`.

src/lib/get-packer-repo-ref-for-plugins.ts
```diff
@@ -0,0 +1,18 @@
+import { cachedGetProductData } from 'lib/get-product-data'
+import { RootDocsPath } from 'types/products'
+import { getBetaLatestVersionRef } from 'views/docs-view/server'
+
+/**
+ * Resolves the `hashicorp/packer` branch
+ * from which `plugins-manifest.json`, `plugins-nav-data.json`,
+ * and Packer `/plugins` root path content will be loaded.
+ */
+export function getPackerRepoRefForPlugins() {
```
zchsh, line 10: Set this up to try to make the logic very clear for the ref we use for `dev-dot` Packer plugins.

src/pages/_proxied-dot-io/packer/plugins/[...page].tsx
```diff
@@ -25,7 +25,7 @@ const basePath = 'plugins'
 const navDataFile = `data/${basePath}-nav-data.json`
 const localContentDir = `../content/${basePath}`
 const additionalComponents = { Badge, BadgesHeader, PluginBadge, Checklist }
-const mainBranch = 'master'
+const mainBranch = 'stable-website'
```
Comment (zchsh) [inline src/pages/_proxied-dot-io/packer/plugins/[...page].tsx:28]: Have switched from `master` to `stable-website` for the dot-io site, as I believe that's more correct.

src/views/docs-view/server.ts
```diff
@@ -46,13 +46,14 @@ import { getCustomLayout } from './utils/get-custom-layout'
...
  */
-function getBetaLatestVersionRef(slug: string): string | undefined {
+export function getBetaLatestVersionRef(slug: string): string | undefined {
 	const hasDevPortalBranch = [
 		'vault',
 		'waypoint',
 		'boundary',
 		'consul',
 		'nomad',
+		'packer',
```
zchsh, line 56: Have added this to signal that Packer has a `dev-portal` branch which we'll pull from.

src/data/packer.json
```diff
@@ -61,25 +61,21 @@
 		{
 			"iconName": "docs",
 			"name": "General Documentation",
-			"mainBranch": "master",
```
zchsh, line 64: Packer has switched from `master` to `main`, so this is no longer necessary, and is in fact detrimental.

BRKalow (commented):
I don't think this will have any practical impact on this change, but we are moving away from the `dev-portal` branches in each repository and just sourcing from the "normal" latest ref.

@zchsh Does this also impact the `Edit this page` link? We would want that to continue pointing at `main`, as folks shouldn't be targeting `stable-website` with changes directly.

zchsh:
@BRKalow I've updated things here, realized we don't need the `stable-website` business in `packer.json`. Have instead made some minor modifications to the Packer plugin page files. Hoping that's more clear, they are still kind of separate (from the content API, & how other docs are loaded) so felt better not to mix the two configurations.

---

Re: Edit links, I think it's a non-issue as due to our page file setup I think we only ever render two kinds of edit links:
- Links defined by `remoteFile.sourceUrl` for external plugin files
- Links defined our content API (including for the `/packer/plugins` landing page)
In theory though there is a case in the `remote-plugin-docs/server` file where we might need to worry about the `editBranch`... but I believe this is an artifact of a previous setup where plugin docs and other docs lived under the same docs route. Now that we have the separate `/plugins` route, I don't think we ever hit that case (but have updated it just in case!).

BRKalow (approved):
Makes sense!

PR #1016, opened 2022-09-14
feat: Report Jest coverage on PRs

thiskevinwang:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

This adds jest coverage reporting + sticky PR commenting.

Note: There is a bug in `jest-coverage-report-action` that ignores `threshold` values, and defaults to `60`, so our reports will always show up as 🔴 
- https://github.com/ArtiomTr/jest-coverage-report-action/blob/main/src/stages/createReport.ts#L37


thiskevinwang:
It looks like as long as `undefined` is being hardcoded here in the coverage reporting action, `threshold` will always default to `60`

https://github.com/ArtiomTr/jest-coverage-report-action/blob/main/src/stages/createReport.ts#L37

jest.config.js
```diff
@@ -1,13 +1,23 @@
...
+/** @type {import('jest').Config} */
+const config = {
 	setupFilesAfterEnv: ['<rootDir>/.test/setup-jest.js'],
 	roots: ['config', 'src', 'build-libs', 'stylelint-rules'],
 	collectCoverageFrom: [
 		'**/*.{js,jsx,ts,tsx}',
 		'!**/*.d.ts',
 		'!**/node_modules/**',
 	],
+	// https://jestjs.io/docs/configuration#coveragethreshold-object
```
thiskevinwang, line 12:
```suggestion
	// https://jestjs.io/docs/configuration#coveragethreshold-object
	// Note: There is a bug in jest-coverage-report-action that 
	// causes this coverageThreshold to be ignored, and defaulting to 60.
	// - https://github.com/ArtiomTr/jest-coverage-report-action/blob/main/src/stages/createReport.ts#L37
```

mhonnatti:
@BRKalow it looks like @thiskevinwang 's research has landed on the 🔴 being an issue with the way the coverage reporter GH action handles thresholds and our low coverage level. I'm okay okay with this "bug" for the time being. your thoughts?

thiskevinwang:
> @BRKalow it looks like @thiskevinwang 's research has landed on the 🔴 being an issue with the way the coverage reporter GH action handles thresholds and our low coverage level. I'm okay okay with this "bug" for the time being. your thoughts?

@mhonnatti — (@BRKalow highlighted this bug last week)

mhonnatti (approved):

BRKalow (changes requested):
Thanks for taking a stab at this! I think if we can't figure out how to not make it red every time, we should explore another action or at least wait until we're through with our crunch to ship for GA. Otherwise I think it'll just be extra noise and a distraction that's not actionable.

PR #1035, opened 2022-09-17
feat: implement full product on-demand ISR

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Introduces an initial pass at on-demand ISR support. Currently, it supports passing a `product` in the request body, which will revalidate all of a product's docs paths as they are defined in its nav data.


src/components/image/index.tsx
```diff
@@ -97,7 +97,7 @@ function Image({
 					title={title}
 					width={dimensions.width}
 					height={dimensions.height}
-					layout="raw"
+					layout="intrinsic"
```
BRKalow, line 100: TODO: is this equivalent?

src/lib/api-validate-token.ts
```diff
@@ -0,0 +1,50 @@
+import { NextApiHandler, NextApiRequest, NextApiResponse } from 'next/types'
```
BRKalow, line 1: copied from our content workflows repo 💡

src/middleware.ts
```diff
@@ -9,8 +9,9 @@ const OPT_IN_MAX_AGE = 60 * 60 * 24 * 180 // 180 days
```
BRKalow, line 9: updates here to comply with the new, stable middleware API

src/types/_app.ts
```diff
@@ -27,6 +27,6 @@ type CustomAppContext<T = Record<string, unknown>> = AppContextNoComponent &
 /**
  * This is our custom type for our Next.js `App` component
  */
-type CustomAppProps = AppProps & ContextWithLayout
+type CustomAppProps = AppProps<$TSFixMe> & ContextWithLayout
```
BRKalow, line 30: TODO: figure out a better type here

next.config.js
```diff
@@ -104,10 +104,5 @@ module.exports = withSwingset({
 			dangerouslyAllowSVG: true,
 			contentSecurityPolicy: "default-src 'self'; script-src 'none'; sandbox;",
 		},
-		experimental: {
-			images: {
-				layoutRaw: true,
-			},
-		},
```
thiskevinwang, line 111: No longer required for next `12.3.0` correct?

BRKalow: Correct, also now that `next/future/image` is stable!

package.json
```diff
@@ -47,7 +47,7 @@
 	"devDependencies": {
 		"@axe-core/react": "^4.3.1",
 		"@datadog/datadog-api-client": "^1.0.0-beta.7",
-		"@hashicorp/platform-cli": "^2.1.0",
+		"@hashicorp/platform-cli": "^2.4.0-canary-2022816231816",
```
thiskevinwang, line 50: What's the TL;DR for `2.4.0` changes?

BRKalow: Bumped `eslint-config-next` to be aware of the middleware updates

thiskevinwang (approved):
Looks reasonable to me! I haven't personally tested this, but trust that it works!

thiskevinwang:
A plain `GET https://dev-portal-kehosug41-hashicorp.vercel.app/api/revalidate` results in 

`400`

```json
{
  "error": "No product provided."
}
```

thiskevinwang:
Got a successful `200` that took about 2s, with
```json
{
    "product": "vault"
}
```

src/pages/api/revalidate.ts
```diff
@@ -0,0 +1,79 @@
...
+ * specified in its latest nav data.
+ */
+export default validateToken(
+	async function handler(request: NextApiRequest, response: NextApiResponse) {
+		if (request.method !== 'POST') {
+			response.status(StatusCodes.NOT_FOUND)
+			return
+		}
+
+		console.log(request.body)
```
dstaley, line 17: 🙋 do you want to keep this?

src/pages/api/revalidate.ts
```diff
@@ -0,0 +1,79 @@
+import { NextApiRequest, NextApiResponse } from 'next/types'
+import { StatusCodes } from 'http-status-codes'
+import { validateToken } from 'lib/api-validate-token'
+import { cachedGetProductData } from 'lib/get-product-data'
+
+/**
+ * Accepts a POST request with a product slug, triggers revalidation for all of a product's paths
+ * specified in its latest nav data.
+ */
+export default validateToken(
```
dstaley, line 10: ⛏️I think this would be a smidge clearer if you defined the handler, and then exported the wrapped version. I spent a minute looking for where you were validating the token before I realized what was going on.

```ts
async function handler() {}

export default validateToken(handler, options)
```

src/pages/api/revalidate.ts
```diff
@@ -0,0 +1,79 @@
...
+	async function handler(request: NextApiRequest, response: NextApiResponse) {
+		if (request.method !== 'POST') {
+			response.status(StatusCodes.NOT_FOUND)
```
dstaley, line 13: ⛏️I was expecting `METHOD_NOT_ALLOWED`, but I also can understand using 404 if you don't want to expose that there _is_ a route at this path.

src/pages/api/revalidate.ts
```diff
@@ -0,0 +1,79 @@
...
+
+		const navDataPrefixes = productData.rootDocsPaths.map(
+			({ navDataPrefix, path }) => {
+				return { navDataPrefix, path }
+			}
+		)
+
+		// fetch the latest nav data, which will be used to construct paths to revalidate
+		const navDataFiles = await Promise.all(
+			navDataPrefixes.map(({ navDataPrefix, path }) => {
```
dstaley, line 38: ⛏️I think an `async function` would read a bit clearer here.

src/pages/api/revalidate.ts
```diff
@@ -0,0 +1,79 @@
...
+			})
+		}
+
+		// iterate over each nav node and revalidate its path
+		navDataFiles.forEach(([basePath, navData]) => {
+			revalidateNavNodes(basePath, navData)
+		})
+
+		// wait for everything to get revalidated
+		await Promise.allSettled(revalidatePromises)
```
dstaley, line 74: 🙋does this need to be wrapped in a `try {} catch {}` so that we always return a response even if this Promise gets rejected? In local development I'm not getting a response from the server for this endpoint.

dstaley (approved):
LGTM! I added a few minor nitpicks. I think this is a great start to adding revalidation! Excited to see where we go from here.

src/pages/api/revalidate.ts
BRKalow: Ooh good call! Yeah I like 404 here for that reason. Per the spec, we'd want to pass along the proper method in the `Allow` header in our response with the 405, which in turn exposes it. Not that it would be a major negative to implement it this way, but I don't think it provides any benefit for our uses.

src/pages/api/revalidate.ts
BRKalow: That's a great question. I think with `Promise.allSettled` we would have to iterate through the results here and check for a `rejected` status, as it doesn't reject itself. So the `await` on its own _should_ be okay here.

I'm going to add a TODO to add retries and/or additional resiliency here though. Given the volume of requests being sent out from this, it seems likely that we'll see a network hiccup or two.

PR #1040, opened 2022-09-19
feat(tutorials): feedback form

pbortnick:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-pbtutorial-feedback-hashicorp.vercel.app/vault/tutorials/getting-started/getting-started-intro) 🔎
- [Swingset](https://dev-portal-git-pbtutorial-feedback-hashicorp.vercel.app/swingset/components/feedbackform)
- [Asana task](https://app.asana.com/0/1202097197789424/1202638765065372/f) 🎟️

## 🗒️ What

Add tutorial feedback form at the bottom of tutorial pages.

## 🛠️ How

1. Copy over the [`<TutorialForm />` react component](https://github.com/hashicorp/react-components/tree/main/packages/feedback-form) and make the following adjustments:
- Add optional `labelIcon` and `labelSecondary` props to take handle the updated labels in the new design.
- Add `optional` prop to text question, allowing the user to move on if that question is not answered (as shown in designs)
- Use flight icons in the choice buttons
- Update the test file to use `tsx` and resolve any issues that came up from that
- Add a `<Finished />` component to have more control over those styles

2. Copy over the [`<FeedbackPanel />` component](https://github.com/hashicorp/learn/blob/master/components/feedback-panel/index.jsx) from learn.

## 🧪 Testing

- [ ] Submit a feedback and check [this sheet](https://docs.google.com/spreadsheets/d/11UI9p2wJH5rYJOIdZBJGOT0vsxUcHIZiwFm6XLsTOVM/edit#gid=0) to see if your submission was populated.


src/components/feedback-form/index.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+	},
+	{
+		id: 'myQuestion',
+		type: 'text',
+		label: 'This is a question',
+		buttonText: 'Submit',
+	},
+]
+
+describe('FeedbackForm', () => {
```
kendallstrautman, line 40: 👏 ty for the tests!

src/views/tutorial-view/components/feedback-panel/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	try {
+		const payload = {
+			...body,
+			page: document.location.pathname,
+		}
+
+		const apiRequest = async () => {
+			const res = await fetch('/api/feedback', {
```
kendallstrautman, line 31: We don't have an `api/feedback` endpoint yet on dev dot right? We should probably create a new form for this data, or I suppose we could use the old tutorials form 🤔

src/components/feedback-form/docs.mdx
```diff
@@ -0,0 +1,67 @@
+---
+componentName: 'FeedbackForm'
+---
+
+The FeedbackForm component is used at the end of tutorials to collect user feedback.
+
+<LiveComponent>
+{`			<FeedbackForm
```
kendallstrautman, line 8: i'm curious, why did we want to copy over this component instead of just updating and using it from react-components?

kendallstrautman: Was it a styles thing?

src/components/feedback-form/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+	new Promise((resolve) => setTimeout(resolve, delay))
+
+const FeedbackFormContext = createContext<FeedbackFormContextType>({})
+
+interface QuestionProps {
+	question: FeedbackQuestion
+	animate?: boolean
+}
+
+const Question: React.FC<QuestionProps> = ({
```
kendallstrautman, line 40: nit: can we separate this out into its own file?

src/components/feedback-form/docs.mdx
pbortnick: We could do that but it would affect other projects that are using the component if dependencies are updated

pbortnick: Oh I think I misunderstood this question. Yeah the styles are different than in react-components

thiskevinwang (changes requested):
Gave it a quick E2E test, which looked good!

We should remove bugsnag as we're not using that anymore

kellyschummer (approved):
updates look good on my end- Thank you @pbortnick!

src/pages/api/tutorial-feedback.ts
```diff
@@ -0,0 +1,153 @@
...
+}
+
+interface Row {
+	sessionId: string
+	page: string
+	timestamp: Date
+	browser: string
+	os: string
+	platform: string
+	helpful?: string
```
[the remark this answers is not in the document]
kendallstrautman: No worries if we want to keep it 1-1 with the original learn implementation for now though.

src/pages/api/tutorial-feedback.ts
```diff
@@ -0,0 +1,153 @@
...
+			os: `${os.name} ${os.version}`,
+			platform: platform.type,
+		}
+
+		const updatedRow = await findAndUpdate(sheet, newRow)
+
+		if (updatedRow) {
+			await updatedRow.save()
+		} else {
+			//  eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
kendallstrautman, line 138: It'd be nice to fix up this typing but not totally necessary if we want to just get it in.

kendallstrautman (approved):
Did some testing and it seems to work well! 👏 nice job 💯  Let's follow up after merging to make sure it's all good on prod

src/views/tutorial-view/components/feedback-panel/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+			const res = await fetch('/api/tutorial-feedback', {
+				method: 'POST',
+				headers: {
+					'Content-Type': 'application/json',
+				},
+				body: JSON.stringify(payload),
+			})
+
+			if (res.status === 204) {
+				console.log('done')
```
thiskevinwang, line 40: I think we should clean up client-side log statements. What do you think?

src/pages/api/tutorial-feedback.ts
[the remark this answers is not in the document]
pbortnick: I actually don't think this is needed. I just cleaned up the types a bit [here](https://github.com/hashicorp/dev-portal/pull/1040/commits/1fa65bb6385f54c8919530327c5adeb63dd5935d)

src/pages/api/tutorial-feedback.ts
```diff
@@ -0,0 +1,153 @@
...
+			await sheet.addRow(newRow)
+		}
+
+		res.status(204).end()
+	} catch (error) {
+		console.error('unhandled error')
+		console.error(error)
+		res.status(error.status || 500).json({
+			body: { error: 'An unhandled error occurred.' },
+		})
```
thiskevinwang, line 146: A small nit:

```suggestion
		console.error('Unexpected error')
		console.error(error)
		res.status(error.status || 500).json({
			body: { error: 'An unexpected error occurred.' },
		})
```

The error here is being _handled_, but likely just isn't an expected one.

thiskevinwang: Ah, I missed that we're returning specific status codes (like `400` and `403`)

src/pages/api/tutorial-feedback.ts
```diff
@@ -0,0 +1,150 @@
...
+			await sheet.addRow(newRow)
+		}
+
+		res.status(204).end()
+	} catch (error) {
+		console.error('Unexpected error')
+		console.error(error)
+		res.status(error.status || 500).json({
+			body: { error: 'An unexpected error occurred.' },
+		})
```
thiskevinwang, line 146: Just throwing a though out there: Some portion of these caught errors are _expected_ — the `400` and `403` that we are returning from above. So it might be nice to return different body text for those.

But this doesn't seem important as we're not doing anything with the response body on the client.

thiskevinwang (approved):
LGTM

PR #1043, opened 2022-09-19
fix: opt-in redirects for .io sites

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Updates our opt-in redirect generation to generate from our `beta_product_slugs` list and a product's `basePaths`. This means when we rollout a new product we don't need to update the redirects config at all, it should _just work_.

### DIFF
--- a/build-libs/__tests__/build-beta-opt-in-redirect.test.ts
+++ b/build-libs/__tests__/build-beta-opt-in-redirect.test.ts
@@ -2,25 +2,121 @@ import buildBetaProductOptInRedirect from '../build-beta-opt-in-redirect'
 
 describe('buildBetaOptInRedirect', () => {
 	test('builds a redirect definition for beta opt-in', () => {
-		expect(
-			buildBetaProductOptInRedirect('waypoint', ['docs', 'commands', 'plugins'])
-		).toMatchInlineSnapshot(`
-      Object {
-        "destination": "https://developer.hashicorp.com/waypoint/:base/:path*",
-        "has": Array [
-          Object {
-            "key": "waypoint-io-beta-opt-in",
-            "type": "cookie",
-            "value": "true",
-          },
-          Object {
-            "type": "host",
-            "value": "(www\\\\.waypointproject\\\\.io|test-wp\\\\.hashi-mktg\\\\.com)",
-          },
-        ],
-        "permanent": false,
-        "source": "/:base(docs|commands|plugins)/:path*",
-      }
-    `)
+		expect(buildBetaProductOptInRedirect()).toMatchInlineSnapshot(`
+		Array [
+		  Object {
+		    "destination": "https://developer.hashicorp.com/boundary/:base/:path*",
+		    "has": Array [
+		      Object {
+		        "key": "boundary-io-beta-opt-in",
+		        "type": "cookie",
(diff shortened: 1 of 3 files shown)


build-libs/build-beta-opt-in-redirect.js
```diff
@@ -1,33 +1,63 @@
...
-			{
-				type: 'host',
-				value: proxySettings[product].host,
-			},
-		],
-	}
+function buildBetaProductOptInRedirect() {
+	return __config.dev_dot.beta_product_slugs
+		.map((product) => {
+			if (product === 'hcp' || product === 'terraform') {
```
BRKalow, line 39: `hcp` and `terraform` are not served from this repo, so we don't need to worry about the .io specific redirects

build-libs/build-beta-opt-in-redirect.js
```diff
@@ -1,33 +1,63 @@
...
-				value: 'true',
-			},
-			{
-				type: 'host',
-				value: proxySettings[product].host,
-			},
-		],
-	}
+function buildBetaProductOptInRedirect() {
+	return __config.dev_dot.beta_product_slugs
```
BRKalow, line 37: Deriving this from our configs instead of manually calling the function for each product. This should make it future-proof and remove one thing that we need to do as part of the rollout

build-libs/__tests__/build-beta-opt-in-redirect.test.ts
```diff
@@ -2,25 +2,121 @@ import buildBetaProductOptInRedirect from '../build-beta-opt-in-redirect'
...
-          Object {
-            "type": "host",
-            "value": "(www\\\\.waypointproject\\\\.io|test-wp\\\\.hashi-mktg\\\\.com)",
-          },
-        ],
-        "permanent": false,
-        "source": "/:base(docs|commands|plugins)/:path*",
-      }
-    `)
+		expect(buildBetaProductOptInRedirect()).toMatchInlineSnapshot(`
```
BRKalow, line 5: This snapshot contains the redirect definitions for all products, as all the products are present in the development config.

There is no risk to shipping these redirect rules sooner, as it would require someone to manually set the cookie if we don't have the banner showing yet.

zchsh (approved):
✅  Approving from looking at the code for a bit, this seems way better 👍

If there's any testing I can do to further validate, happy to help with that too 🙇

build-libs/redirects.js
```diff
@@ -339,8 +339,7 @@ async function redirectsConfig() {
 	const dotIoRedirects = await buildDotIoRedirects()
 	const devPortalRedirects = await buildDevPortalRedirects()
 	const { simpleRedirects, globRedirects } = splitRedirectsByType([
-		buildBetaProductOptInRedirect('waypoint', ['docs', 'commands', 'plugins']),
-		buildBetaProductOptInRedirect('vault', ['docs', 'api-docs']),
+		...buildBetaProductOptInRedirects(),
```
ashleemboyer, line 342: Oh wow! I didn't even know this was something we should have been updating previously. Great call to automate as much as possible 👍🏻

ashleemboyer (approved):

PR #1044, opened 2022-09-19
fix: collection meta markup

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksfix-collection-meta-html-hashicorp.vercel.app/well-architected-framework/operational-excellence) 🔎
- [Asana task](https://app.asana.com/0/inbox/1199634971443477/1202976762714360/1202976762714366) 🎟️

## 🗒️ What

See the ticket for full context. Some collection descriptions use raw html, this helps to support that.


zchsh (approved):
✅ Love me a `composes` 👍

If we were so inclined, I think since `components/text` does a whole `...rest` thing, we could do eg:

```jsx
<Text className={s.description} dangerouslySetInnerHTML={{ __html: description }} />
```

(personally I am 100% happy with either, tbh i kinda like what feels to me like the relatively simplicity of use `<p />` directly with `compose`'d classNames. Butwanted to mention just in case we have any opinions either way!)

kendallstrautman:
@zchsh ahh good call. I think I do prefer the `composes`, it seems simpler to read instead of having context on how `Text` functions 🤷

PR #1054, opened 2022-09-21
chore: bump memory limit for CDK's serverless functions

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

TF CDK has some large pages, which are exceeding the memory limits of our serverless functions and causing rendering to fail. I'm bumping up the memory limit here for those functions to see if we can get those pages rendering.

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -1,5 +1,10 @@
 {
 	"github": {
 		"silent": false
+	},
+	"functions": {
+		"src/pages/terraform/cdktf/**": {
+			"memory": 2048
+		}
 	}
 }


vercel.json
```diff
@@ -1,5 +1,10 @@
 {
 	"github": {
 		"silent": false
+	},
+	"functions": {
+		"src/pages/terraform/cdktf/**": {
+			"memory": 2048
+		}
```
thiskevinwang, line 8: Docs for posterity
https://vercel.com/docs/project-configuration#project-configuration/functions/value-definition

thiskevinwang (approved):
👍

ansgarm:
@mutahhir just noticed this today 😁 Can we get this merged? Especially now that we have the note on our CDKTF docs pointing to the new portal? 😇

mutahhir:
It seems like the issue was resolved in some other way as I remember previously viewing the CDKTF API reference would not load. It's pretty spiffy now.

BRKalow:
@mutahhir Yeah! We switched to rendering these pages at build time so we should be all set here 🤞

PR #1055, opened 2022-09-21
chore: update token refresh fn

thiskevinwang:
## 🔗 Relevant links

- [Asana task](url) 🎟️


src/lib/auth/refresh-token-set.ts
```diff
@@ -0,0 +1,46 @@
+// The response shape from
+// POST {IDENTITY_PROVIDER}/oauth2/token
+interface TokenSet {
+	access_token: string
+	refresh_token: string
+	scope: string
+	expires_in: number
+	id_token: string
+	token_type: 'bearer'
+}
```
ashleemboyer, line 10: 🙋🏻 Can this be moved to `src/types/auth.ts`?

src/lib/auth/refresh-token-set.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Takes a refresh token, and returns a refreshed tokenSet.
+ *
+ * Adapted from: https://next-auth.js.org/tutorials/refresh-token-rotation#server-side
+ */
+export default async function refreshTokenSet(
```
ashleemboyer, line 17: 💭 Are we able to abstract this a bit to work for any auth provider? Maybe we can pass the current provider's configuration object alongside the `refreshToken`?

src/pages/api/auth/[...nextauth].ts
```diff
@@ -12,26 +17,53 @@ export default NextAuth({
...
+					const { access_token, refresh_token } = await refreshTokenSet(
+						token.refresh_token as string
+					)
+					isDev && console.log('successfully refreshed token set')
+
+					token.access_token = access_token
+					token.refresh_token = refresh_token
+				} catch (err) {
+					return { ...token, error: AuthErrors.RefreshAccessTokenError }
+				}
```
Comment (ashleemboyer) [inline src/pages/api/auth/[...nextauth].ts:56]: 💭 Would it be good to `console.error` here as well? I'm not sure if datadog catches those logs yet or not

src/types/auth.ts
```diff
@@ -6,6 +6,7 @@ import { Session } from 'next-auth'
 type SessionData = Omit<Session, 'user'> & {
 	accessToken?: string
 	id?: string
+	error?: string
```
ashleemboyer, line 9: 🙋🏻 Should this reference the new `AuthErrors` enum instead of `string`?

src/pages/api/auth/[...nextauth].ts
```diff
@@ -41,9 +73,10 @@ export default NextAuth({
 		async session({ session, token }) {
 			return {
 				...session,
-				accessToken: token.cloud_idp_access_token,
+				accessToken: token.access_token,
```
Comment (ashleemboyer) [inline src/pages/api/auth/[...nextauth].ts:76]: Really nice that this is generic now! 🙌🏻

src/views/authenticated-view/index.tsx
```diff
@@ -1,16 +1,25 @@
...
-	})
+	const { isAuthenticated, isAuthEnabled, session, signOut } =
+		useAuthentication({
+			isRequired: true,
+		})
+
+	if (session?.error === AuthErrors.RefreshAccessTokenError) {
+		signOut()
+		return null
+	}
```
ashleemboyer, line 22: 🤚🏻 Should this be handled in the Context since there are other views that use authentication for conditional rendering, but don't all use this view?

ashleemboyer:
The testing worked for me locally! 🎉

src/types/auth.ts
thiskevinwang: Yes! Good catch

src/pages/api/auth/[...nextauth].ts
Comment (thiskevinwang) [reply src/pages/api/auth/[...nextauth].ts:56]: I'm on board!

Comment (thiskevinwang) [reply src/pages/api/auth/[...nextauth].ts:56]: We should have Datadog log drains already set up — see https://vercel.com/hashicorp/dev-portal/settings/integrations

src/views/authenticated-view/index.tsx
```diff
@@ -22,7 +31,12 @@ const AuthenticatedView = ({ children }: AuthenticatedViewProps) => {
 		return null
 	}
 
-	return <>{children}</>
+	return (
```
thiskevinwang, line 34:
```suggestion
	return <>{children}</>
```

src/views/authenticated-view/index.tsx
```diff
@@ -23,6 +32,11 @@ const AuthenticatedView = ({ children }: AuthenticatedViewProps) => {
 	}
 
 	return <>{children}</>
+		<div>
+			<h1>test</h1>
+			{children}
+		</div>
+	)
```
thiskevinwang, line 39:
```suggestion
```

src/views/authenticated-view/index.tsx
thiskevinwang: Which context are you referring to?

ashleemboyer: Oooh shucks! I mistakenly thought `useAuthentication` came from a Context. So scratch that specific idea!

Rather: can we move this to be handled inside of `useAuthentication` so it happens for all views that use the hook?

src/lib/auth/refresh-token-set.ts
thiskevinwang: After looking at this for a bit, it feels like a premature abstraction — going to hold on making this change

src/views/authenticated-view/index.tsx
thiskevinwang: I've moved the `signOut` call into `useAuthentication` and added a basic check to dedupe multiple requests to `signOut` (i.e. multiple rendered components on a page all concurrently attempting to `signOut`

src/lib/auth/refresh-token-set.ts
ashleemboyer: Right on! Thanks for trying it out anyways 😊

src/hooks/use-authentication/index.ts
```diff
@@ -1,7 +1,13 @@
...
-import { SessionData, UserData, ValidAuthProviderId } from 'types/auth'
+import {
+	AuthErrors,
+	SessionData,
+	UserData,
+	ValidAuthProviderId,
+} from 'types/auth'
 import { UseAuthenticationOptions, UseAuthenticationResult } from './types'
 import { signInWrapper, signOutWrapper, signUp } from './helpers'
+import { useEffect } from 'react'
```
ashleemboyer, line 10: ⛏️ Can this be moved to the top of the import list?

src/hooks/use-authentication/index.ts
```diff
@@ -26,6 +32,13 @@ const useAuthentication = (
 		onUnauthenticated,
 	})
 
+	// Logout user if token refresh fails
+	useEffect(() => {
+		if (data?.error === AuthErrors.RefreshAccessTokenError) {
+			signOutWrapper()
+		}
+	}, [data?.error])
```
ashleemboyer, line 40: 🙌🏻

src/views/authenticated-view/index.tsx
```diff
@@ -1,6 +1,9 @@
 import useAuthentication from 'hooks/use-authentication'
```
ashleemboyer, line 1: Looks like these changes aren't needed anymore!

ashleemboyer (approved):
Couple of small things left, but approving assuming that those'll be resolved ezpz. So great to have this in! 🎉

src/hooks/use-authentication/index.ts
thiskevinwang: Done!

src/views/authenticated-view/index.tsx
thiskevinwang: Cleaned up!

PR #1057, opened 2022-09-21
style: refine corner radius on product-nav

braydenlove:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-fork-braydenlove-patch-1-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/0/1203018211842978/f) 🎟️

## 🗒️ What

I am refining the corner radius of the product-nav.

## 🤷 Why

Currently we are using 14px corner radius which looks odd on the far right and left cards visually. It is also not a common radius in HDS, we are moving forward with 6px which is the common radius across HDS.

## 🛠️ How

Initially we wanted it to be a bit more 'round' in appearance but we went too far. 6px feels better aligned to the rest of the system and site.

## 📸 Design Screenshots

Still can't upload images for some reason...

### DIFF
--- a/src/views/homepage/components/product-nav/product-nav.module.css
+++ b/src/views/homepage/components/product-nav/product-nav.module.css
@@ -44,7 +44,7 @@
   display: grid;
   grid-template-columns: repeat(2, 1fr);
   gap: var(--homepage-grid-gap);
-  border-radius: 14px;
+  border-radius: 6px;
 
   @media (--dev-dot-tablet-up) {
     grid-template-columns: repeat(3, 1fr);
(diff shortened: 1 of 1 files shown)


src/views/homepage/components/product-nav/product-nav.module.css
```diff
@@ -44,7 +44,7 @@
   display: grid;
   grid-template-columns: repeat(2, 1fr);
   gap: var(--homepage-grid-gap);
-  border-radius: 14px;
+  border-radius: 6px;
```
kendallstrautman, line 47: 💭 A future improvement here would be to use a css variable. Not necessary for this PR, just thinking outloud.

kendallstrautman (approved):

PR #1065, opened 2022-09-22
[rewriteTutorialLinksPlugin] Fixing tutorial lookup and section rewriting

ashleemboyer:
## 🗒️ What

- Fixes the part of the tutorial link rewriting where the `TUTORIAL_MAP` lookup is happening for `hcp` tutorials. The map stores `cloud`, but `hcp` was being used.
- Fixes valid `SectionOption`s not being rewritten, and updates the `get-is-beta-product` mock to be more restrictive so the tests were no longer false positives.


kendallstrautman:
Awesome, the deploy logs look normal again! 💯

PR #1068, opened 2022-09-22
fix: TFE release notes page

thiskevinwang:


src/constants/version-path.ts
```diff
@@ -1,5 +1,5 @@
 /** This REGEX is used to parse a product version from a URL */
-export const VERSION_IN_PATH_REGEX = /v\d+\.\d+\.(\d+|\w+)/g
+export const VERSION_IN_PATH_REGEX = /v\d+\.\d+\.(\d+|\w+)/i
```
thiskevinwang, line 2: This was the one bug that broke non-TFE versioned pages — swapping `g` for `i` results in non-TFE versioned pages functioning as they did before

BRKalow: ref: https://stackoverflow.com/questions/1520800/why-does-a-regexp-with-global-flag-give-wrong-results/1520853#1520853

BRKalow (approved):

PR #1074, opened 2022-09-22
Add a `clean` command

topfunky:
## 🗒️ What

Authors and contributors often ask how to reset their local directory. This command automates removing auto-generated directories `.next` and `node_modules` with the `npm run clean` command.

It is cross platform with the `rimraf` Node module.

## 🤷 Why

Authors and contributors often need to remove directories created by Next or Node so they can start fresh. This command automates it.

## 🛠️ How

Uses the [rimraf](https://www.npmjs.com/package/rimraf) command for cross platform compatibility.

## Anything Else

Does `package-lock.json` also need to be committed?


BRKalow (commented):
Great idea! Could you also document this in the local development section of https://github.com/hashicorp/dev-portal/blob/main/docs/README.md 🙏

package.json
```diff
@@ -5,6 +5,7 @@
 		"build:deploy-preview": "npx --package @hashicorp/platform-tools@0.5.0 hc-tools --resolve-paths false ./scripts/content-repo-preview/build.ts",
 		"build:webpack-only": "hc-tools next-build-webpack-only",
 		"build": "next build",
+		"clean": "rimraf .next node_modules",
```
kendallstrautman, line 8: 💭 You should be able to just use `rm -rf` I think? [See reference here ](https://github.com/hashicorp/learn/blob/master/package.json#L153). Doesn't matter a ton though 😸

package.json
```diff
@@ -76,6 +77,7 @@
 		"postcss-normalize": "^10.0.1",
 		"postcss-preset-env": "^6.7.0",
 		"react-test-renderer": "^18.2.0",
+		"rimraf": "^3.0.2",
```
kendallstrautman, line 80: If we do use this dep, we should also commit the package-lock updates.

package.json
kendallstrautman: Oh I see your note in the description on why you used `rimraf`, feel free to disregard comment above.

kendallstrautman (approved):
thank you!

PR #1077, opened 2022-09-23
[Tabs] Fix group syncing

ashleemboyer:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1202097197789424/1203017450775750/f) 🎟️

## 🗒️ What

- Updates tab group syncing to occur in `Tabs`, only after `setActiveTabIndex` is called (aka not on first load)
- Removes now unused `useSyncedTabGroups`
- Updates `useTabGroups` to throw an error if `TabProvider` is not rendered
- Updates tests and `CoreDevDotLayout` to render `TabProvider`

## 🤷 Why

- On some pages like [/terraform/tutorials/azure-get-started/install-cli](https://developer.hashicorp.com/terraform/tutorials/azure-get-started/install-cli#install-terraform), where multiple `Tab`s are being rendered, and some of their `Tab` components have `group`s but not all, there was an issue with group syncing that would reset the active tab incorrectly after a tab was clicked.
- `TabProvider` was moved up in the component tree so we don't have to manually render it in several places.

## 🧪 Testing

- [ ] Go to [/terraform/tutorials/azure-get-started/install-cli](https://dev-portal-git-ambfix-tabs-syncing-hashicorp.vercel.app/terraform/tutorials/azure-get-started/install-cli)
- [ ] Scroll through the page
- [ ] Make sure all visible `Tab`s have their first `Tab` selected
- [ ] Go to the [Install Terraform section](https://dev-portal-git-ambfix-tabs-syncing-hashicorp.vercel.app/terraform/tutorials/azure-get-started/install-cli#install-terraform)
- [ ] Click "Homebrew on OS X" tab
- [ ] The tab should be successfully selected
- [ ] Click "Chocolatey on Windows" tab
- [ ] The tab should be successfully selected
- [ ] Click "Linux" tab
- [ ] The tab should be successfully selected


zchsh (commented):
💯  This seems reasonable to me! Tested locally and working as expected 👍

src/components/tabs/index.tsx
```diff
@@ -47,16 +48,29 @@ const Tabs = ({
...
+	 * Handle syncing `activeTabIndex` and `activeTabGroup`.
 	 *
 	 * Note: only works where tabs have groups, eg <Tab group="some-string" />.
 	 */
-	const setSyncedActiveTabIndex = useSyncedTabGroups({
-		activeTabIndex,
-		setActiveTabIndex,
-		tabItems,
-	})
+	const { activeTabGroup, setActiveTabGroup } = useTabGroups()
```
zchsh, line 55: I think this is causing errors where it previously didn't, possibly because it was previously nested within another hook? 🤔 Seems like we either need to add `<TabProvider />` to a bunch of places (`/boundary/api-docs`, mostly, i just skimmed the Vercel logs)... or maybe warn rather than throw in `useTabGroups` 🤔

ashleemboyer: I updated the `TabProvider` rendering in https://github.com/hashicorp/dev-portal/pull/1077/commits/5ad955d20e827e33f60323c8ca175664c7b1475c! It's now rendered in `CoreDevDotLayout`. I think we want to throw an error in the context hook like we do in other context hooks, and I don't think moving the rendering of the provider up in the tree should have issues. Do you forsee any? 🤔

BRKalow: FWIW, I think it's valid that we support rendering `Tabs` outside of the `TabGroupsProvider`. It seems like a "progressive enhancement" sort of thing, not necessarily a hard requirement for using tabs.

That being said, there shouldn't be any practical issue with lifting the provider up, it just seems a little heavy handed that we can't use `<Tabs>` without also rendering the provider.

src/components/tabs/index.tsx
```diff
@@ -47,16 +48,29 @@ const Tabs = ({
 	const tabItems = useTabItems({ children, activeTabIndex, initialActiveIndex })
 
 	/**
-	 * useSyncedTabGroups hooks into TabProvider,
-	 * and keeps activeTabIndex & activeTabGroup in sync.
+	 * Handle syncing `activeTabIndex` and `activeTabGroup`.
 	 *
 	 * Note: only works where tabs have groups, eg <Tab group="some-string" />.
 	 */
-	const setSyncedActiveTabIndex = useSyncedTabGroups({
```
BRKalow, line 55: ❓ question: I'm curious why you opted to remove `useSyncedTabGroups`? It felt like a nice encapsulation of this logic

src/components/tabs/index.tsx
ashleemboyer: Updated in https://github.com/hashicorp/dev-portal/pull/1077/commits/952c344ea4ef3a8d2e3c7438a757506919f736ad

src/components/tabs/index.tsx
ashleemboyer: That code wasn't worked as expected, so when I tried out something else in the body of the `Tabs` component I left code there and decided not to move it into `useSyncedTabGroups`. Updated to move the new code into that file in https://github.com/hashicorp/dev-portal/pull/1077/commits/a14ed10d49c74ce9be546d426fb696dda113529a.

BRKalow (approved):
👏

PR #1085, opened 2022-09-26
fix: react keys

thiskevinwang:
## 🔗 Relevant links

- [Asana task](url) 🎟️


src/views/product-root-docs-path-landing/components/marketing-content/index.tsx
```diff
@@ -137,9 +137,10 @@ const ProductRootDocsPathLandingMarketingContent = ({ blocks }) => {
 
 	return (
 		<div className={s.root}>
-			{blocks.map((block) => {
+			{blocks.map((block, i) => {
+				const key = `${block.headingId}-${i}`
```
ashleemboyer, line 141: Not all of these blocks will have a `headingId`, so i think we'll be safe to just use the index in this case.

```suggestion
				const key = i

```

kendallstrautman:
Closing as this fix is already in main

PR #1089, opened 2022-09-26
feat(docs headings): add permalink component

pbortnick:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-pbheading-permalink-component-hashicorp.vercel.app/consul/docs/discovery/services) 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1201532549545236/f) 🎟️

## 🗒️ What

Add permalink component and use in docs headings.

## 📸 Design Screenshots

🎨 [Design](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer)

Screen Shot 2022-09-26 at 3 32 40 PM

## 🧪 Testing

- [ ] Hover over heading
- [ ] Hover directly over link icon
- [ ] Focus on link with keyboard
- [ ] Click on links
- [ ] Check that url is updated
- [ ] Check that link is vopied to clipboard


src/components/dev-dot-content/mdx-components/mdx-headings/index.tsx
```diff
@@ -48,14 +50,24 @@ export function makeMdxHeadingElement(level: HeadingProps['level']) {
...
-			/>
+			>
+				{/* Temporary permalink render until we update our remark plugin */}
+				{Children.map(children, (child) =>
+					child?.props?.className === '__permalink-h' ? (
+						<MdxHeadingPermalink {...child.props} />
+					) : (
+						child
+					)
+				)}
```
BRKalow, line 69: For others: Pam and I paired on this briefly, and came to the conclusion that this was a reasonable initial implementation. For a few reasons:

* We control the generated element, via our current remark plugin, that adds the anchor with the `__permalink-h` class, so it should be reliable to detect it in this way
* We know we eventually want the logic to live in the heading component for inserting the permalink anchor, so it felt best to put it here now as opposed to, say, our `MdxA` component.
* Once we do update the remark plugin, this would presumably break, and we can make the necessary adjustments in the implementation.

src/components/dev-dot-content/mdx-components/mdx-headings/index.tsx
```diff
@@ -48,14 +50,24 @@ export function makeMdxHeadingElement(level: HeadingProps['level']) {
...
 			<Heading
-				{...props}
+				{...rest}
 				level={level}
 				className={fixedClassName}
 				size={size}
 				weight={weight}
-			/>
+			>
+				{/* Temporary permalink render until we update our remark plugin */}
```
BRKalow, line 62:
```suggestion
				{/* This is a temporary solution to render an updated permalink element while our remark plugin continues to inject the current one. The full implementation will eventually live in this component, but it will require a larger update to the anchorLinks remark plugin.  We detect the injected permalink by the `__permalink-h` class which the remark plugin applies to the element. */}
```

src/components/dev-dot-content/mdx-components/mdx-headings/mdx-headings.module.css
```diff
@@ -27,6 +27,13 @@
 		font-size: 0.8em;
 		font-weight: 500;
 	}
+
+	&:hover {
+		& :global(.__permalink-h) {
+			opacity: 1;
+			transition: opacity 0.2s ease;
```
BRKalow, line 34:
```suggestion
			@media (prefers-reduced-motion: no-preference) {
				transition: opacity 0.2s ease;
			}
```

src/components/dev-dot-content/mdx-components/mdx-heading-permalink/index.tsx
```diff
@@ -0,0 +1,23 @@
+import classNames from 'classnames'
+import { IconLink16 } from '@hashicorp/flight-icons/svg-react/link-16'
+import s from './mdx-heading-permalink.module.css'
+
+export default function MdxHeadingPermalink(props) {
+	const { className, ...rest } = props
+
+	function copyLinkToClipboard(e) {
+		navigator.clipboard.writeText(e.target.closest('a').href)
```
BRKalow, line 9: We should catch any errors that might get thrown from calling `navigator.clipboard.writeText`, just in case! https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText#return_value

src/components/dev-dot-content/mdx-components/mdx-heading-permalink/mdx-heading-permalink.module.css
```diff
@@ -0,0 +1,33 @@
...
+		display: flex;
+		align-items: center;
+		justify-content: center;
+	}
+
+	&:hover,
+	&:focus {
+		background: var(--token-color-palette-neutral-200);
+		opacity: 1;
+		transition: opacity 0.2s ease;
```
BRKalow, line 31: Is this necessary given the above `transition` declaration?

src/components/dev-dot-content/mdx-components/mdx-heading-permalink/mdx-heading-permalink.module.css
```diff
@@ -0,0 +1,33 @@
+.root {
+	composes: g-focus-ring-from-box-shadow from global;
+	display: inline-block;
+	opacity: 0;
+	transition: opacity 0.2s ease;
```
BRKalow, line 5: Let's wrap this in `@media (prefers-reduced-motion: no-preference) { }`

src/components/dev-dot-content/mdx-components/mdx-heading-permalink/index.tsx
pbortnick: @BRKalow does this look good? Open to suggestions of course!
https://github.com/hashicorp/dev-portal/pull/1089/commits/3f38ba8ce610f4d1853024a8559fb8c05446b141

src/components/dev-dot-content/mdx-components/mdx-heading-permalink/mdx-heading-permalink.module.css
pbortnick: Nope! Good catch that was redundant. deleted here:
https://github.com/hashicorp/dev-portal/pull/1089/commits/e42832beb39aeacda8770d588b4713480f1233b4

src/components/dev-dot-content/mdx-components/mdx-heading-permalink/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+	}
```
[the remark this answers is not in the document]
pbortnick: Agreed. Added :)

src/components/dev-dot-content/mdx-components/mdx-headings/mdx-headings.module.css
```diff
@@ -27,6 +28,16 @@
...
 		font-weight: 500;
 	}
+
+	&:hover {
+		& :global(.__permalink-h) {
+			opacity: 1;
+
+			@media (prefers-reduced-motion: no-preference) {
+				transition: opacity 0.2s ease;
+			}
```
ashleemboyer, line 38: Is this needed here since it's declared up in `MdxHeadingPermalink`'s `.root` class?

src/components/dev-dot-content/mdx-components/mdx-heading-permalink/index.tsx
```diff
@@ -0,0 +1,15 @@
...
+import s from './mdx-heading-permalink.module.css'
+
+export default function MdxHeadingPermalink(props) {
+	const { className, level, ...rest } = props
+
+	return (
+		<a className={classNames(s.root, className, s[`h${level}`])} {...rest}>
+			<div className={s.icon}>
+				<IconLink16 color="var(--token-color-foreground-action)" />
+			</div>
```
ashleemboyer, line 12: Is the wrapping `<div>` needed since we can pass a `className` directly to `IconLink16`? I think we also want to not se `color` here in the JSX, but in the CSS. Is there some further context behind using the `color` prop on `IconLink16`?

src/components/dev-dot-content/mdx-components/mdx-heading-permalink/mdx-heading-permalink.module.css
```diff
@@ -0,0 +1,42 @@
...
+	@media (prefers-reduced-motion: no-preference) {
+		transition: opacity 0.2s ease;
+	}
+
+	& .icon {
+		height: 16px;
+		display: flex;
+		align-items: center;
+		justify-content: center;
+	}
```
ashleemboyer, line 19: ⛏️ Could we pull this out of the body of `.root` to align with our [Group styles by element, not by state or media type](https://www.notion.so/hashicorp/CSS-Modules-753f2647f7aa4a598e28990d9f0d05a8#839a408c41ca4a609b5cdde0942fa3af) practice?

src/components/dev-dot-content/mdx-components/mdx-heading-permalink/mdx-heading-permalink.module.css
```diff
@@ -0,0 +1,42 @@
...
+
+	&:focus {
+		opacity: 1;
+		background: var(--token-color-surface-faint);
+
+		& svg * {
+			fill: var(--token-color-foreground-action-active);
+		}
+	}
```
ashleemboyer, line 37: I _think_ we should be able to set the colors of the `svg`s without the `*` character like this:

```css
& svg {
  color: <color>;
}
```

I'm pretty sure it's how we do it in most other places. Have you already tried this approach?

src/components/dev-dot-content/mdx-components/mdx-heading-permalink/index.tsx
pbortnick: I just put the color prop in there because that is what the default color is supposed to be before any action is taken place but we can move it to the css for more consistency. Also, we don't need that `<div>` any more. We can just add `display: flex;` on a class directly on the icon. Good call!

src/components/dev-dot-content/mdx-components/mdx-heading-permalink/mdx-heading-permalink.module.css
pbortnick: You're right! Nope I didn't think of that.

src/components/dev-dot-content/mdx-components/mdx-headings/mdx-headings.module.css
pbortnick: Nope. Good call!

pbortnick:
@ashleemboyer thanks for the review! All of the items should have been addressed in this [latest commit](https://github.com/hashicorp/dev-portal/pull/1089/commits/004a8e48b4f918038f03904bb248469c5ac0716d) but please let me know if I missed something

ashleemboyer (approved):
Looks great!! ✨ 💖

PR #1091, opened 2022-09-26
feat: tutorial scroll progress

zchsh:
## 🔗 Relevant links

- [Asana task](url) 🎟️

### DIFF
--- a/src/layouts/sidebar-sidecar/components/scroll-progress-bar/index.tsx
+++ b/src/layouts/sidebar-sidecar/components/scroll-progress-bar/index.tsx
@@ -1,0 +1,13 @@
+import s from './scroll-progress-bar.module.css'
+
+function ScrollProgressBar() {
+	return (
+		<>
+			<pre className={s.root}>
+				<code>TODO: Scroll progress</code>
+			</pre>
+		</>
+	)
+}
+
+export { ScrollProgressBar }
--- a/src/layouts/sidebar-sidecar/components/scroll-progress-bar/scroll-progress-bar.module.css
+++ b/src/layouts/sidebar-sidecar/components/scroll-progress-bar/scroll-progress-bar.module.css
@@ -1,0 +1,6 @@
+.root {
+	width: 100%;
+	border: 2px solid magenta;
+	background: rgba(255, 0, 0, 0.3);
+	margin: 0;
+}
--- a/src/layouts/sidebar-sidecar/index.tsx
+++ b/src/layouts/sidebar-sidecar/index.tsx
@@ -28,6 +28,7 @@ import {
 	useSidebarNavData,
 } from './contexts/sidebar-nav-data'
 import s from './sidebar-sidecar-layout.module.css'
+import { ScrollProgressBar } from './components/scroll-progress-bar'
 
 const SidebarSidecarLayout = (props: SidebarSidecarLayoutProps) => {
 	const navDataLevels = props.sidebarNavDataLevels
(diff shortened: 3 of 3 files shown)


src/hooks/use-scroll-percentage/helpers/is-browser.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Return `true` if `window` is defined,
+ * `false` otherwise.
+ */
+export function isBrowser(): boolean {
```
zchsh, line 5: A tiny helper that felt worth splitting out, maybe later worth elevating if it feels like it makes sense.

src/hooks/use-scroll-percentage/helpers/round.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Given a number, and the number of decimals to include,
+ * Return the number rounded to the provided number of decimals
+ */
+export function round(number: number, decimalCount: number) {
```
zchsh, line 5: Another tiny helper that felt useful to split out.

src/hooks/use-scroll-percentage/index.ts
```diff
@@ -0,0 +1,109 @@
...
+ * observer to that element and monitor the element and its subtree for
+ * changes. This ensures better accuracy of the `documentScrollHeight` we
+ * use to calculate scroll percentage, as in many views, document height
+ * may change significantly through user interactions that update the DOM.
+ *
+ * For example, on the Tutorial view, the contents in the `main` element
+ * frequently change height from interactions such as changing tabbed content.
+ * For that view, we pass `main` as the `mutationTargetSelector`.
+ */
+export default function useScrollPercentage({
```
zchsh, line 30: This hook is used by the new `ScrollProgress` component in `SidebarSidecarLayout`. One interesting facet is accepting a `mutationTargetSelector` here, there's more detail on why this feels necessary in comments, and in `use-document-scroll-height`.

Have tried to build this in a way that might be re-usable, for example within `use-scroll-progress-analytics`. But leaving that abstraction & re-use step for a separate PR.

src/layouts/sidebar-sidecar/components/scroll-progress-bar/index.tsx
```diff
@@ -0,0 +1,22 @@
+import ProgressBar from 'components/progress-bar'
+import useScrollPercentage from 'hooks/use-scroll-percentage'
+import s from './scroll-progress-bar.module.css'
+
+/**
+ * Displays a scroll progress bar at the bottom of SidebarSidecarLayout.
+ */
+function ScrollProgressBar() {
```
zchsh, line 8: Helper component for `SidebarSidecarLayout`.

src/views/tutorial-view/index.tsx
```diff
@@ -258,6 +258,7 @@ function TutorialView({
 					 * deferring for a follow-up PR since this is functional for the time being.
 					 */
 					sidebarNavDataLevels={sidebarNavDataLevels as any}
+					showScrollProgress={true}
```
zchsh, line 261: Tutorial view uses the newly introduced `showScrollProgress` prop on `SidebarSidecarLayout`.

zchsh:
@BRKalow @kendallstrautman This PR implements scroll progress on tutorial views.

Initially I'd modeled this off of `use-scroll-percentage-analytics`, but I ran into a bit of complexity here with changes to `scrollHeight` due to things like `Tab` content changing or the window resizing. Have done my best to quickly address those so that scroll progress appears accurately, even as `Tab` contents change, the window is resized, or client-side navigation happens.

> **Note**: I think we might have similar issues in `use-scroll-percentage-analytics` to the issues that I've tried to address in this PR via `use-document-scroll-height` etc...  though perhaps not sure they're as much of a concern, since we're only concerned with rough milestones in scroll percentage. That being said, I think there could be situations where it could matter - for example on a tutorial document with long default tabbed content, if a visitor switches to shorter tabs, and `scrollHeight` doesn't get re-calculated, it might be impossible to reach the `100%` scrolled event. So, I've tried to structure this work in a way where we could potentially re-use `use-scroll-percentage` within `use-scroll-percentage-analytics` in the future, if those issues do seem relevant in the latter hook.

alexcarpenter:
@zchsh have we considered using Framer Motion [useScroll hook](https://www.framer.com/docs/use-scroll/) here? 

https://user-images.githubusercontent.com/825855/192844510-e44e7887-663f-4ae9-8a94-c07cb0c85185.mp4

zchsh:
Update here: I've merged in #1116 which simplifies things significantly by offloading a lot of the work to `@motionone/dom` (discovered via `framer-motion`, thanks @alexcarpenter 💖 ). Updating the PR description to reflect 👍

src/components/progress-bar/progress-bar.module.css
```diff
@@ -18,5 +18,7 @@
 	left: 0;
 	position: absolute;
 	top: 0;
-	width: var(--percent-done);
+	width: 100%;
+	transform: scaleX(var(--portion-done));
```
BrandonRomano, line 22: 👨‍🍳 😗 👌 

Transform + width 100%, smart!

BrandonRomano (approved):
Really clean 👌

src/components/progress-bar/progress-bar.module.css
zchsh: Props to @alexcarpenter 💖

kendallstrautman (approved):
👏 this is so cool 💯

PR #1095, opened 2022-09-27
[Global Search] Keyboard Navigation

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambglobal-search-keyboard-nav-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1202948006171576/f) 🎟️

## 🗒️ What

- Updates the hover/focus states for list items
- Removes `tabindex` setting on `role="tabpanel"` element in `Tabs`
- Removes the `tabindex` settings on the `CommandBarListItem` button and link elements
- Removes the arrow key badges from the footer instructions

## 🤷 Why

The arrow key keyboard interactions is too complicated with `Tabs` in the mix, so just using `TAB` and `SHIFT+TAB` simplifies both the implementation and the user experience.

## 🧪 Testing

- [ ] Go to the preview
- [ ] Open the command bar
- [ ] Make sure you can interact with the dialog with keys as expected:
  - Text input is one TAB stop
  - `Tabs` are one TAB stop
    - `Tabs` can be changed with left and right arrow keys
  - Each item in the body is one TAB stop
  - The footer has one TAB stop


src/components/command-bar/components/list/index.tsx
```diff
@@ -27,7 +27,13 @@ const CommandBarList = ({
 
 	return (
 		<div className={s.root}>
-			<Text asElement="p" className={s.label} size={100} weight="semibold">
+			<Text
+				asElement="p"
+				className={s.label}
+				id={labelId}
```
ashleemboyer, line 33: This `id` was missing! 😱

BRKalow (approved):
Working well! I Tested in a few different sections, across browsers, and with VoiceOver.

One note about Firefox, not specific to search: by default Firefox does not tab to links, only inputs and buttons. This can be tweaked by (1) adjusting accessibility settings in MacOS or (2) [adjusting a Firefox config value](https://support.mozilla.org/en-US/questions/1278793).

I would expect that someone who is reliant on keyboard navigation would have this sorted out and configured properly if they were using Firefox and this wouldn't be an issue. Just sharing what I learned!

PR #1114, opened 2022-09-28
[WIP] [Global Search] Using local storage for recent searches

ashleemboyer:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1202097197789424/1202932738893677/f) 🎟️

### DIFF
--- a/src/components/command-bar/commands/search/components/dialog-body/index.tsx
+++ b/src/components/command-bar/commands/search/components/dialog-body/index.tsx
@@ -1,12 +1,18 @@
-import { useCallback, useMemo } from 'react'
+import { useCallback, useEffect, useMemo, useState } from 'react'
 import algoliasearch from 'algoliasearch'
 import { Configure, InstantSearch } from 'react-instantsearch-hooks-web'
 import { IconDocs16 } from '@hashicorp/flight-icons/svg-react/docs-16'
+import { IconHistory16 } from '@hashicorp/flight-icons/svg-react/history-16'
 import { IconLearn16 } from '@hashicorp/flight-icons/svg-react/learn-16'
 import { ProductSlug } from 'types/products'
 import { useCurrentContentType, useCurrentProduct } from 'contexts'
 import { CommandBarTag, useCommandBar } from 'components/command-bar'
 import { useSetUpAndCleanUpCommandState } from 'components/command-bar/hooks'
+import {
+	CommandBarButtonListItem,
+	CommandBarDivider,
+	CommandBarList,
+} from 'components/command-bar/components'
 import Tabs, { Tab } from 'components/tabs'
 import {
 	generateSuggestedPages,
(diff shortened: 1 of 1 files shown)


src/components/command-bar/commands/search/components/recent-searches/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+	CommandBarList,
+} from 'components/command-bar/components'
+import { RecentSearch, RecentSearchesProps } from './types'
+
+const RecentSearches = ({ recentSearches }: RecentSearchesProps) => {
+	const { inputRef, setCurrentInputValue } = useCommandBar()
+
+	if (!recentSearches || recentSearches.length === 0) {
+		return null
+	}
```
BRKalow, line 15: ⛏️ nit: Along with the other suggestions, this could be simplified if we always return the same type/
```suggestion
	if (recentSearches.length === 0) {
		return null
	}
```

src/components/command-bar/commands/search/hooks/use-recent-searches.ts
```diff
@@ -0,0 +1,71 @@
...
+
+		if (!storageValue) {
+			window.localStorage.setItem(RECENT_SEARCHES_KEY, JSON.stringify([]))
+			return []
+		} else {
+			return JSON.parse(storageValue)
+		}
+	} catch (e) {
+		// window may be 'undefined'
+		// localStorage may be disabled
```
BRKalow, line 21: ⛏️ nit: We could `return []` here and simplify the check in the recent searches component.

```suggestion
		// localStorage may be disabled
		return []
```

src/components/command-bar/commands/search/hooks/use-recent-searches.ts
```diff
@@ -0,0 +1,71 @@
...
+				if (JSON.stringify(newList) !== JSON.stringify(recentSearches)) {
+					window.localStorage.setItem(
+						RECENT_SEARCHES_KEY,
+						JSON.stringify(newList)
+					)
+					setRecentSearches(newList)
+				}
+			} catch (e) {
+				// window may be 'undefined'
+				// localStorage may be disabled
```
BRKalow, line 62: ⛏️ nit: We could `return []` here and simplify the check in the recent searches component.

```suggestion
				// localStorage may be disabled
				return []
```

src/components/command-bar/commands/search/components/recent-searches/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+			<CommandBarList label="Recent Searches">
+				{recentSearches.map((recentSearch: RecentSearch) => {
+					return (
+						<CommandBarButtonListItem
+							key={recentSearch}
+							icon={<IconHistory16 />}
+							onClick={() => {
+								setCurrentInputValue(recentSearch)
+								inputRef.current.focus()
+							}}
```
BRKalow, line 28: future 💭 (no action needed): when we have multiple commands, do you think we would need to explicitly set the active command to "search" as part of this too?

BRKalow (approved):
It works great! ✨ left a non-blocking nit suggestion, take it or leave it.

src/components/command-bar/commands/search/components/recent-searches/index.tsx
ashleemboyer: Hmmm I'm not sure. At the moment I would think not, because I don't think clearing the text input should have any effect on what command is active. I would expect that the user has to manually change the command themselves in some way.

PR #1116, opened 2022-09-28
feat: use useScroll like framer-motion

zchsh:
> **Note**: targets #1091, largely replaces it.

This PR replaces custom-build scroll progress update logic with a `use-scroll` hook that leverages [`@motionone/dom`](https://motion.dev/dom/scroll), in a very similar way to [how `framer-motion` uses `@motionone/dom`](https://github.com/framer/motion/blob/main/packages/framer-motion/src/value/use-scroll.ts).

## 🧪 Testing

- [ ] Visit a tutorial view, such as [/waypoint/tutorials/get-started-docker/get-started-intro][/waypoint/tutorials/get-started-docker/get-started-intro]
    - Scroll progress should be visible at the bottom of the page, and should function as expected
    - Scroll progress should remain accurate when the viewport is resized
- [ ] Visit a tutorial view with tabbed content, such as [/waypoint/tutorials/get-started-docker/get-started-install][/waypoint/tutorials/get-started-docker/get-started-install]
    - Scroll progress should function as expected, updating when tabbed content on the page changes the height of the page content
- [ ] Visit another view that uses `SidebarSidecarLayout`, such as [/waypoint/docs/intro][/waypoint/docs/intro]
    - Scroll progress should not be visible on these views. It should only be visible on tutorial views.

[/waypoint/tutorials/get-started-docker/get-started-install]: https://dev-portal-git-zstutorial-scroll-progress-fram-0dd93d-hashicorp.vercel.app/waypoint/tutorials/get-started-docker/get-started-install
[/waypoint/tutorials/get-started-docker/get-started-intro]: https://dev-portal-git-zstutorial-scroll-progress-fram-0dd93d-hashicorp.vercel.app/waypoint/tutorials/get-started-docker/get-started-intro
[/waypoint/docs/intro]: https://dev-portal-git-zstutorial-scroll-progress-fram-0dd93d-hashicorp.vercel.app/waypoint/docs/intro


src/hooks/use-scroll.ts
```diff
@@ -0,0 +1,54 @@
...
+}: UseScrollOptions = {}) {
+	const [scrollValues, setScrollValues] = useState<ScrollValues>({
+		scrollX: 0,
+		scrollXProgress: 0,
+		scrollY: 0,
+		scrollYProgress: 0,
+	})
+
+	useIsomorphicLayoutEffect(() => {
+		return scroll(
```
zchsh, line 36: Taking a very similar approach to what `framer-motion` does internally, using `@motionone`. But this way, we don't need to require the use of `m.div` to use the return values of `useScroll`.

src/layouts/sidebar-sidecar/components/scroll-progress-bar/index.tsx
```diff
@@ -1,20 +1,13 @@
 import ProgressBar from 'components/progress-bar'
-import useScrollPercentage from 'hooks/use-scroll-percentage'
 import s from './scroll-progress-bar.module.css'
 
 /**
  * Displays a scroll progress bar at the bottom of SidebarSidecarLayout.
  */
-function ScrollProgressBar() {
-	const rawPercent = useScrollPercentage({
```
zchsh, line 9: `useScrollPercentage` is baleeted 🎉

src/layouts/sidebar-sidecar/index.tsx
```diff
@@ -96,7 +102,7 @@ const SidebarSidecarLayoutContent = ({
 					{sidebarContent}
 				</div>
 			</MobileMenuContainer>
-			<div className={s.contentWrapper}>
+			<div className={s.contentWrapper} ref={contentRef}>
```
zchsh, line 105: We now track scroll progress through `contentWrapper`, rather than through the whole page.

src/hooks/use-isomorphic-layout-effect.ts
```diff
@@ -0,0 +1,4 @@
+import { useEffect, useLayoutEffect } from 'react'
+import { isBrowser } from './is-browser'
+
+export const useIsomorphicLayoutEffect = isBrowser ? useLayoutEffect : useEffect
```
zchsh, line 4: Added this and `isBrowser` since they'd already seemed like useful things to add, and this felt validated by `framer-motion` having the same utils.

src/components/progress-bar/progress-bar.module.css
```diff
@@ -18,5 +18,7 @@
 	left: 0;
 	position: absolute;
 	top: 0;
-	width: var(--percent-done);
+	width: 100%;
+	transform: scaleX(var(--portion-done));
+	transform-origin: left;
```
zchsh, line 23: Switched to `transform` here, not `width`.

src/components/progress-bar/index.tsx
```diff
@@ -18,13 +18,15 @@ function ProgressBar({
...
+	const portionDone = clamp(percentDone, 0, 100) / 100
+
 	return (
 		<span
 			className={s.root}
 			style={
 				{
 					'--border-radius': rounded ? '2px' : '0px',
-					'--percent-done': `${clamp(percentDone, 0, 100)}%`,
+					'--portion-done': portionDone,
```
zchsh: Switched to `--portion-done` so that the use of `scaleX` is more intuitive in CSS.

Have left the props API the same; could make sense to change in a follow-up PR, but that would require changes to eg. collection progress bars etc, so have left it out of this PR.

package.json
```diff
@@ -134,6 +134,7 @@
 		"@hashicorp/react-vertical-text-block-list": "^7.0.0",
 		"@hashicorp/remark-plugins": "^4.0.0",
 		"@hashicorp/sentinel-embedded": "^0.0.14",
+		"@motionone/dom": "^10.14.2",
```
zchsh, line 137: Added `@motionone/dom`, this replaces a lot of the complexity of #1091 .

alexcarpenter: Framer motion is already included in the bundle. Is there a benefit to adding motionone here too that we don't get from framer motion?

zchsh:
@alexcarpenter @BRKalow @kendallstrautman This PR targets #1091, and uses the same approach as `framer-motion` to reduce the custom implementation complexity of the previous `use-scroll-progress` hook.

I started off using `framer-motion`'s `use-scroll` directly (thanks for this suggestion @alexcarpenter!), but felt a bit locked in to having to use `m.div` from `framer-motion` when I really felt like I didn't need it, so have used `@motionone/dom` here directly. Intent being to maintain the re-usability of `ProgressBar`, and maybe open the door to re-using `use-scroll` in `use-scroll-progress-analytics`, for example.

src/hooks/use-isomorphic-layout-effect.ts
alexcarpenter: Can we leverage `useSafeLayoutEffect` instead? I think they are achieving the same functionality. `src/hooks/use-safe-layout-effect.ts`

src/layouts/sidebar-sidecar/index.tsx
```diff
@@ -61,6 +62,11 @@ const SidebarSidecarLayoutContent = ({
 	const currentlyViewedVersion = getVersionFromPath(currentPath)
 	const sidebarProps = sidebarNavDataLevels[currentLevel]
 	const sidebarIsVisible = !isMobileMenuRendered || mobileMenuIsOpen
+	const contentRef = useRef(null)
+
+	const { scrollYProgress } = useScroll({
+		target: contentRef,
```
alexcarpenter, line 68:
```suggestion
		target: contentRef,
		offset: ["-68px start", "end end"]
```

Sticky elements are not registered during scroll, so the animation doesn't kick in until you've scrolled the distance of the header element which I think is 68px. Is it possible to get that header height value in some fashion?

offset is available as an option in framer-motion useScroll.

zchsh: Ah good call! We can definitely hard-code the `68px` value as a start 💯 We have that value available as a CSS variable, I'll try that out.

The `useScroll` in this PR takes the exact same options as `framer-motion`, since `framer-motion` passes everything through to `@motionone/dom` (at least as far as I can tell!).

src/hooks/use-isomorphic-layout-effect.ts
zchsh: Oh oops! Definitely 💯

package.json
zchsh: I think so! The perceived benefit is that by using `@motionone/dom`, we can set up a `useScroll` hook that returns `number` values for `{ scrollX, scrollXProgress, scrollY, scrollYProgress }` (where `framer-motion`'s `useScroll` returns `MotionValue<number>`). This gives us flexibility to use `useScroll`:
- We can re-use `ProgressBar`, rather than having to rewrite styles or update that component to use `m.div` & accept `MotionValue<number>` rather than a simpler `number`
- We might be able to re-use `useScroll` for things like `use-scroll-percentage-analytics`
- If we need a `MotionValue`, we can wrap any `number` from `useScroll` in `useMotionValue()`. (Admittedly we can technically go the other way, if `scrollProgressY` was a `MotionValue` I guess we could do `scrollProgressY.get()`, but when I tried this it yielded a value that seemed to update much more infrequently than expected, resulting in laggy-looking scroll progress).

Let me know if that makes sense, happy to adjust and dive into a refactor of `ProgressBar` if there are strong opinions that we should use `framer-motion` `useScroll` directly 🙇

src/layouts/sidebar-sidecar/index.tsx
zchsh: Sweet okay we had a function for getting the sticky bar height, which is kinda neat - elevated that function to `lib` and am now using it 💯 thanks for mentioning that! 💖

src/layouts/sidebar-sidecar/components/table-of-contents/use-active-section.ts
```diff
@@ -2,24 +2,7 @@ import { useEffect, useRef, useState } from 'react'
...
- * TODO: this may need to be refactored when we address the brittleness of our
- * header height.
- */
-const getFullNavHeaderHeight = () => {
-	return getCSSVariableFromDocument('--navigation-header-height', {
-		asNumber: true,
-		fallback: NAVIGATION_HEADER_HEIGHT_FALLBACK,
-	}) as number
-}
+import getFullNavHeaderHeight from 'lib/get-full-nav-header-height'
```
zchsh, line 5: Pulled this out to `lib` as we have a second use for it, as part of using `use-scroll`.

package.json
alexcarpenter: Got it, I will leave it up to @BRKalow and team to make the decision on pulling in motionone here vs using Framer Motion.

We seem to be recreating functionality around a library already in use, which feels more brittle in my opinion. Now we're maintaining our own useScroll hook and missing out on some of the benefits you get from motion values. 

- Reduced react rerenders
- subscriptions to values (which seems to be what we'd need to trigger the complete toast on scroll complete?)
- chaining with useTransform hook

> value that seemed to update much more infrequently than expected, resulting in laggy-looking scroll progress

Guessing this by design and is for performance reasons on frequency of updates? This could also be smoothed out with transition easings.

Void of the decision to go the framer or motion one route, this looks good!

alexcarpenter (approved):
👍🏼 Approving based on Bryce's comment in slack.

PR #1118, opened 2022-09-28
Installing @types/shortid

ashleemboyer:
## 🗒️ What

Installs `@types/shortid`

## 🤷 Why

To fix the following TS error:

```
src/components/feedback-form/index.tsx:2:21 - error TS2307: Cannot find module 'shortid' or its corresponding type declarations.
```


BRKalow (approved):
weeeee

PR #1130, opened 2022-09-29
Refactor tutorial feedback api

pbortnick:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-pb-feedback-api-updates-hashicorp.vercel.app/packer/tutorials/docker-get-started/docker-get-started-variables) 🔎
- [Asana task](url) 🎟️

## 🗒️ What

- Update types that were missing
- Create better error handling

## 🧪 Testing

Make sure submissions still work properly by submitting feedback and checking [this sheet](https://docs.google.com/spreadsheets/d/11UI9p2wJH5rYJOIdZBJGOT0vsxUcHIZiwFm6XLsTOVM/edit#gid=0)

### DIFF
--- a/src/pages/api/tutorial-feedback.ts
+++ b/src/pages/api/tutorial-feedback.ts
@@ -7,8 +7,7 @@ import {
 } from 'google-spreadsheet'
 import Bowser from 'bowser'
 
-// const FEEDBACK_SHEET_ID = process.env.TUTORIAL_FEEDBACK_SHEET_ID
-const FEEDBACK_SHEET_ID = '273y81274u'
+const FEEDBACK_SHEET_ID = process.env.TUTORIAL_FEEDBACK_SHEET_ID
 const FEEDBACK_SERVICE_EMAIL = process.env.FEEDBACK_SERVICE_EMAIL
 const FEEDBACK_PRIVATE_KEY = process.env.FEEDBACK_PRIVATE_KEY
 const HASHI_ENV = process.env.HASHI_ENV
--- a/src/pages/api/tutorial-feedback.ts
+++ b/src/pages/api/tutorial-feedback.ts
@@ -20,14 +20,14 @@ interface SurveyResponse {
 interface RequestBody {
 	sessionId: string
 	page: string
-	timestamp: Date
+	timestamp: string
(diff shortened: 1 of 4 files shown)


src/pages/api/tutorial-feedback.ts
```diff
@@ -13,21 +13,22 @@ const FEEDBACK_PRIVATE_KEY = process.env.FEEDBACK_PRIVATE_KEY
...
 	page: string
-	timestamp: Date
-	responses: SurveyResponse[]
+	timestamp: string
 }
 
-interface Row {
-	sessionId: string
-	page: string
-	timestamp: Date
```
pbortnick, line 30: type `Date` was incompatible with `addRow` argument type

src/pages/api/tutorial-feedback.ts
```diff
@@ -132,17 +135,25 @@ const submitFeedback = async (
...
+
+		let errorMessage = 'An unexpected error occurred.'
+
+		if (error.response?.status === 404) {
+			errorMessage = 'Requested entity was not found.'
+		}
+
+		if (error.response?.status === 400) {
+			errorMessage = 'Invalid grant: account not found.'
+		}
```
pbortnick, line 153: Tested these errors and these messages were returned in the actual error

kendallstrautman: 💭 If they are already coming back in the error.message, do we need to specify them here? Or perhaps we just concat with the original message `An unexpected error  occurred: ${error.message}`

pbortnick: I like the idea but the `error.message` for 404 comes back as `'error:0909006C:PEM routines:get_name:no start line' }`. I didn't think that was very clear but if we think otherwise then 👍 . Or maybe there is another common field in the error object that we can use?

pbortnick: I suppose that _is_ the actual error message associdate tho. So I'm fine with this!

kendallstrautman: I'll leave it to you to decide 👍

kendallstrautman (approved):

PR #1133, opened 2022-09-29
[Global Search] Check for invalid hit results

ashleemboyer:
## 🗒️ What

- Adds checks in `DocumentationHit` and `TutorialHit` that prevent rendering the component if a `hit`'s `_highlightResult` is falsy, or the needed properties in the object are not defined

## 🤷 Why

- There was an error happening under `/vault` that caused the site to blow up

## 🧪 Testing

- [ ] Go to /vault
- [ ] Open the command bar
- [ ] Type the letter `t`
- [ ] There should no longer be an error


BRKalow (approved):
LGTM!

PR #1134, opened 2022-09-30
[Badge] Use g-screen-reader-only class for ariaLabel

ashleemboyer:
## 🗒️ What

- Removes two properties from our `Badge` component
  - `ariaDescribedBy`
  - `ariaLabelledBy`
- Adds a `g-screen-reader-only` `<span>` next to the regular `Badge` span that handles labelling `Badge` for screen reader users

## 🤷 Why

- `<span>` doesn't actually accept `aria-label`, so it's not presented to users
- before these changes, the elements were announced as "<aria label>, group"

## 🧪 Testing

- [ ] Go to [the `Badge` Swingset page](https://dev-portal-git-ambbadge-aria-label-hashicorp.vercel.app/swingset/components/badge) on the preview
- [ ] In the Playground, add an `ariaLabel` on the `Badge`: "This is an aria label test"
- [ ] Turn on your device's screen reader
- [ ] Bring your focus to the Playground heading
- [ ] Navigate to the `Badge` with arrow keys (not `TAB`)
  - In VoiceOver, this is VO + Arrow Right
- [ ] The `Badge` should be presented as "This is an arial label test"

No-audio demonstration:

https://user-images.githubusercontent.com/43934258/193285422-26fd7309-1bd3-49aa-b4fc-0f15db53c28a.mov

### DIFF
--- a/src/components/badge/index.tsx
+++ b/src/components/badge/index.tsx
@@ -3,9 +3,7 @@ import { BadgeProps } from './types'
 import s from './badge.module.css'
 
 const Badge = ({
-	ariaDescribedBy,
 	ariaLabel,
-	ariaLabelledBy,
 	className,
 	color = 'neutral',
 	icon,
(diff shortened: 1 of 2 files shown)


BRKalow (approved):
Good catch 😄

ashleemboyer:
> Good catch 😄

@BRKalow Melanie caught this one when checking out global search! 😊

PR #1135, opened 2022-09-30
[Dialog] Require label, update CommandBarDialog label, set CommandBarDialog ariaDescribedBy

ashleemboyer:
## 🗒️ What

- In `CommandBarDialogFooter`:
  - Removes the arrow up and down keys from the instructions (coming in another PR)
  - Makes the `ariaLabels` more explicit
  - Adds `instructionsElementId` prop
- In `CommandBarDialog`:
  - Sets `ariaDescribedBy` prop
  - Sets `label` prop
  - Passes `instructionsElementId` prop to `CommandBarDialogFooter`
- In `Dialog`:
  - Updates `label` prop to be required again
  - Adds `ariaDescribedBy` prop
  - Sets `aria-describedby` on `DialogContent`

## 🤷 Why

- The `label` prop is re-required because it _is_ necessary to have on dialogs
- The `instructionsElementId` was added because in order for `aria-describedby` to work when the dialog comes into focus, the describing element has to be inside of the dialog (aka the footer)

## 🧪 Testing

- [ ] Go the preview
- [ ] Turn on your device's screen reader
- [ ] Open the command bar
- [ ] It should be presented like:

image


BRKalow (approved):
👍 LGTM!

PR #1138, opened 2022-09-30
fix: use framer-motion to fix issue w use-scroll

zchsh:
This PR fixes an issue in #1091 where the custom `use-scroll` hook causes flicker in elements in the `sidecar`.

This PR fixes the issue by using `framer-motion`'s `use-scroll` instead.


zchsh:
@BRKalow I was very wrong about `use-scroll`, the custom hook doesn't work as expected (causes flicker).

cc @alexcarpenter you were right, better to just use `framer-motion` 💖

BRKalow (approved):

PR #1139, opened 2022-09-30
fix: rm font-weight shift from sidecar

zchsh:
Removes the `font-weight` shift for active items from the sidecar `TableOfContents`.

Intent here is prevent layout jank. Have confirmed using `400` as the consistent weight via Slack DM.

### Before

https://user-images.githubusercontent.com/4624598/193311975-7a3f43df-1629-4762-87c8-7368f6da8673.mp4

### After

https://user-images.githubusercontent.com/4624598/193311998-ca04630d-a4d7-49b6-a6b0-903d879d90a2.mp4


BRKalow (approved):
LGTM, assuming design is also 👍

braydenlove (approved):
Great job!

PR #1141, opened 2022-09-30
chore: tf local preview

thiskevinwang:
This enables local preview via Terraform-related repositories

This looks for the `LOCAL_CONTENT_DIR` env var to be present
- ex. `LOCAL_CONTENT_DIR="../docs"`

### DIFF
--- a/src/components/dev-dot-content/mdx-components/mdx-heading-permalink/index.tsx
+++ b/src/components/dev-dot-content/mdx-components/mdx-heading-permalink/index.tsx
@@ -3,10 +3,14 @@ import { IconLink16 } from '@hashicorp/flight-icons/svg-react/link-16'
 import s from './mdx-heading-permalink.module.css'
 
 export default function MdxHeadingPermalink(props) {
-	const { className, level, ...rest } = props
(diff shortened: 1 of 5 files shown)


BRKalow:
@thiskevinwang do you think we should drive this via an env var defined in the scripts that are placed in each repo? 💭

thiskevinwang:
> @thiskevinwang do you think we should drive this via an env var defined in the scripts that are placed in each repo? 💭

@BRKalow given that we currently house some other closely related configurations internally (like `navDataPrefix`), it felt appropriate to **not** use another env var

BRKalow:
Alright, fair enough! I think the distinction here for me is that everything else is needed to correctly grab the remote content, whereas this value is only needed locally when running in the context of said repository.

thiskevinwang:
> I think the distinction here for me is that everything else is needed to correctly grab the remote content, whereas this value is only needed locally when running in the context of said repository.

Hmm, this is a good distinction. Maybe env var **is** the approach then, to avoid cross-contaminating config responsibilities

BRKalow (approved):

PR #1127, opened 2022-09-29
feat: adds support for proxied redirect definitions

BRKalow:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-BRANCH_NAME-hashicorp.vercel.app/PATH_TO_VIEW) 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1203057754313219/f) 🎟️

## 🗒️ What

Support redirect definitions for proxied domains. This is done in a "no-config" way by deriving the proxied domain from the filename.

- `/proxied-redirects/www.waypointproject.io.redirects.js` will apply redirects to `www.waypointproject.io`
- `/proxied-redirects/www.consul.io` will apply redirects to `www.consul.io`
- etc...

PR #1144, opened 2022-09-30
feat: remote product redirects

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Updates our generation of redirects from product repositories to handle the GA scenario. Specifically, these redirects should apply to the developer domain after a product is in GA. Once a product is removed from `beta_product_slugs`, its redirects will apply to `developer.hashicorp.com` instead of the product's .io domain.

Redirects for the .io domains will be handled by the functionality in #1127.


build-libs/redirects.js
```diff
@@ -56,6 +59,26 @@ function addHostCondition(redirects, productSlug) {
 		if (productSlug == PROXIED_PRODUCT) {
 			return redirect
 		}
+
+		// If the productSlug is NOT a beta product, it is GA, so handle the redirect appropriately
+		if (!config['dev_dot.beta_product_slugs'].includes(productSlug)) {
```
thiskevinwang, line 64: 🤔 What's enabling us to access `dev_dot.beta_product_slugs` on config?

BRKalow: It's defined here: https://github.com/hashicorp/dev-portal/pull/1144/files#diff-00a4f13d760b82b998bcae3d31b5fbd8d2e9edf400c352dfa980ecb82fdf8024R18

thiskevinwang: Ah, `flat` 🎉

build-libs/redirects.js
```diff
@@ -155,85 +151,21 @@ async function buildDotIoRedirects() {
...
-	// TODO: intent is to do this after all products have been migrated
-	const consulIoRedirects = [...consulAuthorRedirects]
-	// TODO ... consolidate redirects for other products
 	return [
 		...devPortalToDotIoRedirects,
-		...addHostCondition(boundaryIoRedirects, 'boundary'),
-		...addHostCondition(nomadIoRedirects, 'nomad'),
+		...(await getRedirectsForProduct('boundary')),
+		...(await getRedirectsForProduct('nomad')),
+		...(await getRedirectsForProduct('vault')),
```
thiskevinwang, line 158: 🤔 
```suggestion
		...(await getRedirectsForProduct('vault', 'main')),
```

BRKalow: Hmm, we should probably try and derive the latest ref from the content API 💭

BRKalow: ...we should just make the redirects available via the content API

thiskevinwang: > ...we should just make the redirects available via the content API

Fast-follow, post-GA

thiskevinwang: > Hmm, we should probably try and derive the latest ref from the content API 💭

```bash
curl -s https://content.hashicorp.com/api/content/vault/version-metadata\?partial\=true \
   | jq -r '.result[0].ref'
```

build-libs/__tests__/redirects.test.ts
```diff
@@ -180,4 +194,94 @@ describe('groupSimpleRedirects', () => {
...
+		      },
+		    ],
+		    "permanent": true,
+		    "source": "/vault/docs/foo",
+		  },
+		]
+	`)
+	})
+})
+
```
thiskevinwang, line 286: Nice

build-libs/redirects.js
```diff
@@ -11,9 +11,14 @@ const {
 const fetchGithubFile = require('./fetch-github-file')
 const { isContentDeployPreview } = require('../src/lib/env-checks')
 const buildBetaProductOptInRedirects = require('./build-beta-opt-in-redirect')
+const { loadHashiConfigForEnvironment } = require('../config')
+
+require('isomorphic-unfetch')
```
thiskevinwang, line 16: What's the need for this `require`?

BRKalow: It makes `fetch` available, which isn't defined by default in node. I was running into some type errors trying to import fetch, so just did this...

thiskevinwang: _cough_ NODE 18 _cough_ 😄

BRKalow: Screen Shot 2022-10-04 at 11 39 46 AM

🥲

thiskevinwang: >It makes fetch available, which isn't defined by default in node. I was running into some type errors trying to import fetch, so just did this...

The Next.js runtime polyfills `fetch` right?
But `build-libs/redirects.js` runs outside of this runtime, in the standard `node` runtime?

BRKalow: Correct 👍

build-libs/redirects.js
```diff
@@ -155,85 +175,25 @@ async function buildDotIoRedirects() {
...
-		...addHostCondition(vagrantIoRedirects, 'vagrant'),
-		...addHostCondition(packerIoRedirects, 'packer'),
-		...addHostCondition(consulIoRedirects, 'consul'),
+		...(await getRedirectsForProduct('boundary')),
+		...(await getRedirectsForProduct('nomad')),
+		...(await getRedirectsForProduct('vault')),
+		...(await getRedirectsForProduct('waypoint')),
+		...(await getRedirectsForProduct('vagrant')),
+		...(await getRedirectsForProduct('packer')),
+		...(await getRedirectsForProduct('consul')),
```
thiskevinwang, line 186: Thoughts for parallelizing these async fns?

It could shave off a 10-100 ms off build time.

thiskevinwang (approved):
This seems reasonable to me!

build-libs/redirects.js
BRKalow: Sure sure, updated!

PR #1149, opened 2022-10-03
fix: package-manager code block copy

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Fixes an issue where install commands on the download views were very difficult, if not impossible, to copy. Separate issues affected the ability to copy by selecting text and the ability to copy with the code-block "Copy" button.

## 🤷 Why

To ensure practitioners can easily copy and use the install commands on all downloads views.

## 🛠️ How

- Bumps to new version of code-block that supports multiple single-line commands in a single `shell-snippet`
	- See https://github.com/hashicorp/react-components/pull/736 for details
- Refactors install-view `packageManagers` code to ensure the code is highlighted, by moving some work previously done on the client to the server.
	- This is necessary to fix the select-copy-paste functionality (in the way supported by the `code-block` release above). We use CSS to target the syntax-highlighted `shell-symbol` elements (namely `$`); so highlighting must be applied for the fix to work.

## 🧪 Testing

- [ ] Visit a downloads view, such as [/terraform/downloads](https://dev-portal-git-zsfix-install-view-code-block-copy-hashicorp.vercel.app/terraform/downloads)
	- For any of the "install command" code blocks for any operating system...
	- Select the code, copy it, and paste into a plain text editor. The install commands should have pasted with the correct newlines, and without leading `$` characters.
		- Note that there may be leading spaces with this method; tweaking this would further refinement outside the scope of this PR.
	- Activate the "Copy" button, and paste into a plain text editor. The install commands should have pasted with the correct newlines, and without leading `$` characters or leading spaces.

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/1202097197789424/1203067508838442/f
[preview]: https://dev-portal-git-zsfix-install-view-code-block-copy-hashicorp.vercel.app/terraform/downloads


package.json
```diff
@@ -104,7 +104,7 @@
 		"@hashicorp/react-button": "^6.2.1",
 		"@hashicorp/react-call-to-action": "^4.1.1",
 		"@hashicorp/react-checkbox-input": "^5.0.2",
-		"@hashicorp/react-code-block": "^6.0.1",
+		"@hashicorp/react-code-block": "^6.2.0-canary-20221003171818",
```
zchsh, line 107: Bumps to `code-block` with fixes from hashicorp/react-components#736

src/views/product-downloads-view/index.tsx
```diff
@@ -75,23 +71,6 @@ const ProductDownloadsViewContent = ({
 			isEnterpriseMode
 		),
 	]
-	const packageManagers = useMemo(() => {
```
zchsh, line 78: We build `packageManagers` in `getStaticProps`, so this client-side work is no longer needed.

zchsh:
@BRKalow @BrandonRomano This PR bumps code-block to pull in the "copy multiple single-line commands from a single snippet" fixes from hashicorp/react-components#736 (which isn't published yet, currently running a pre-release as extra confirmation that the fix in that PR works!).

It also moves the install-command-snippet generation for package managers to `getStaticProps` for downloads views, in order to enable code highlighting on those snippets.

src/views/product-downloads-view/helpers.ts
```diff
@@ -142,31 +142,6 @@ export function generateEnterprisePackageManagers(
 	]
 }
 
-export const generatePackageManagers = ({
```
zchsh, line 145: Edit: moved this to a separate file. Based on the bundle analysis, which showed a `200 kB+` increase for all `/downloads` routes, it seemed like `@hashicorp/platform-code-highlighting` was getting bundled if I kept it in this helper file (which is used on the client).

Moved this to `server-helpers`, and updated the implementation there to add highlighting, to see if that prevents the bundle size issue.

BRKalow (approved):
works great!

PR #1158, opened 2022-10-03
fix: boundary stable-website deleted, fix ref

zchsh:
🚨 EDIT: nvm `stable-website` was not intentionally deleted 👍 

## 🗒️ What

Fixes an issue where `dev-dot` builds fail because the `stable-website` ref no longer exists in `hashicorp/boundary`.

Updates to use `main` instead - this means the following pages will pull API docs from `main`:
- https://www.boundaryproject.io/api-docs
- https://developer.hashicorp.com/boundary/api-docs


BRKalow:
@zchsh I just restored the branch, I don't think it should have been deleted 😅

PR #1161, opened 2022-10-04
fix: duplicate canonical tags

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Updates how we render the canonical tag to remove duplicate elements. Also added an e2e test to validate that we only render a single canonical tag.

To reproduce:
* Visit https://developer.hashicorp.com/waypoint
* In the console: `document.querySelectorAll('link[rel="canonical"]').length`
* Observe it is 2, when it should be 1

It seems that leveraging our existing logic in `HashiHead` addresses this. It seems like it could potentially be a bug in Next, but for the sake of time I'm not going to go down the rabbit hole.

Running the new e2e test assertion against production results in:
```
Running 15 tests using 7 workers
  1) [chromium] › seo.spec.ts:35:1 › product landing page should render the metadata ===============

    Error: expect(received).toEqual(expected) // deep equality

    Expected: 1
    Received: 2

      61 |      ).toContain('/og-image/waypoint.jpg')
      62 |
    > 63 |      expect(await page.locator('head link[rel="canonical"]').count()).toEqual(1)
         |                                                                       ^
      64 |
      65 |      expect(
      66 |              await page.locator('head link[rel="canonical"]').getAttribute('href')
```

## 🧪 Testing

- Visit the preview: /waypoint
- Validate that a single canonical tag is present with: `document.querySelectorAll('link[rel="canonical"]').length`
- Navigate to another page
- Validate that the canonical url is correct with: `$('link[rel="canonical"]')?.href`


src/components/head-metadata/index.tsx
```diff
@@ -46,13 +52,13 @@ export default function HeadMetadata(props: HeadMetadataProps) {
 	return (
 		// TODO: OpenGraph image to be passed as the image prop here
 		<>
-			<SelfReferentialCanonicalTag />
```
BRKalow, line 49: Some strange interaction with the usage of a nested `<Head>` tag here, pulling it out and passing the URL directly to `HashiHead` fixes the duplicate tag 🤷‍♂️

thiskevinwang (approved):
Weird! Looks good to me!

I `cURL`d and only 1 canonical `link`
and ran `document.querySelectorAll('link[rel="canonical"]').length` in the browser. Also 1 `link`

PR #1166, opened 2022-10-04
feat: limit static generation for .io sites after GA

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Once a product is removed from the beta array and activated as "GA," there is no practical reason for us to statically generate all of the product's .io site docs pages. This change updates our static paths logic to set a limit of 0 if a product is not in the beta array, which going forward will mean that the product is in GA. This should give us some flexibility to increase static generation for developer without increasing build times too much.


zchsh (approved):
✅ seems like a great call 👍 the build speed caretaker 💖 💯

PR #1177, opened 2022-10-05
feat: update cache warmer to use the revalidate endpoint for developer docs paths

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

This change updates our cache warming script to also warm the cache for our developer docs routes, using the revalidate endpoint.

For each product slug, this will make a call to `developer.hashicorp.com/api/revalidate`, which in turn will ISR every docs route present in our nav data files.

## 🤷 Why

This should ensure that users don't experience slow load times immediately after a deployment when visiting a page which was not statically generated at build time.

### DIFF
--- a/scripts/warm-cache.ts
+++ b/scripts/warm-cache.ts
@@ -28,7 +28,8 @@ interface StaticPathsResponse {
 
 const DEV_PORTAL_URL = config.dev_dot.canonical_base_url
 
-const fetch = createFetch()
+// the socket timeout here matches Vercel's serverless function execution timeout (900s) as defined here: https://vercel.com/docs/concepts/limits/overview#general-limits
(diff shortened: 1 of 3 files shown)


scripts/warm-cache.ts
```diff
@@ -72,6 +72,38 @@ async function getUrlsToCache(product: string): Promise<string[]> {
 	})
 }
 
+/**
+ * Uses the /api/revalidate endpoint to statically generate all docs pages for each product
+ */
+async function warmDeveloperDocsCache() {
```
kendallstrautman, line 78: 🥳

kendallstrautman (approved):
stoked!

PR #1179, opened 2022-10-05
[Tabs] Moving overflowTargetRef placement

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambfix-tabs-controls-rendering-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1203100487715425/f) 🎟️

## 🗒️ What

- Moves the `overflowTargetRef ` from the `div` that contains both tab controls and panels to the `div` containing the tab controls.

## 🤷 Why

- Overflow was being detected due to the content of a tab panel, but we really only care about overflow in the element where the controls live.

## 🧪 Testing

- [ ] Go to [/consul/downloads](https://dev-portal-git-ambfix-tabs-controls-rendering-hashicorp.vercel.app/consul/downloads)
- [ ] Set your viewport width to `1280px` (`1px` before the sidecar goes away)
- [ ] Click the "Linux" tab
- [ ] The `Tabs` should not turn into a `select`

### DIFF
--- a/src/components/tabs/index.tsx
+++ b/src/components/tabs/index.tsx
@@ -68,13 +68,14 @@ const Tabs = ({
 
 	return (
 		<TabNestingProvider>
-			<div ref={overflowTargetRef}>
+			<div>
 				<div
 					className={classNames(s.tabControls, s[`variant--${variant}`], {
 						[s.isCheckingOverflow]: hasOverflow === null,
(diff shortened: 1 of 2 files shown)


BRKalow (approved):
Brilliant! Validated that it happens on the live site and that it does NOT happen on the preview. Nice detective work.

PR #1186, opened 2022-10-06
fix(downloads): ensure downloads pages can ISR

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Updates logic around the path generation to read the install content file. This ensures that [output file tracing](https://nextjs.org/docs/advanced-features/output-file-tracing) will pick u p the path and include it in the build artifact. Currently, ISR fails because the file can't be found.


thiskevinwang (approved):
makes sense to me!

PR #1191, opened 2022-10-06
[addNavItemMetaData] Replace <sup> tags with Badges

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambreplace-sups-with-badge-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1203115842976140/f) 🎟️

## 🗒️ What

- Creates `badge` properties for nav items that hav `<sup>` tags in their `titles`
- Removes the `<sup>` tags from the nav item titles if they're present

## 🤷 Why

- When the `dev-portal` branches were present, we handled this manually in the JSON files. Since we don't have those branches anymore, and some of the manual changes had to be undone, the changes proposed in this PR enable either the old or new format to work.
- This should also enable the new format that renders `Badge`s to work in old versions of docs.

## 🧪 Testing

Make sure the sidebars render as expected on these pages:

- [ ] [/vault/docs](https://dev-portal-git-ambreplace-sups-with-badge-hashicorp.vercel.app/vault/docs)
- [ ] [/waypoint/docs](https://dev-portal-git-ambreplace-sups-with-badge-hashicorp.vercel.app/waypoint/docs)
- [ ] [/hcp/docs](https://dev-portal-git-ambreplace-sups-with-badge-hashicorp.vercel.app/hcp/docs)

### DIFF
--- a/src/components/sidebar/helpers/add-nav-item-meta-data.ts
+++ b/src/components/sidebar/helpers/add-nav-item-meta-data.ts
@@ -13,6 +13,27 @@ interface AddNavItemMetaDataResult {
 }
 
 /**
+ * Returns an object for rendering a Badge in a sidebar item if a <sup> tag is
+ * founnd in its title.
+ */
+const getBadgeFromTitle = (title: string) => {
+	let badge = null
+
+	const regex = new RegExp(/<sup>(.*)<\/sup>$/)
+	const matches = title.match(regex)
+	if (matches && matches.length > 0) {
+		const badgeText = matches[1]
+		badge = {
+			text: badgeText,
+			color: 'neutral',
+			type: 'outlined',
+		}
+	}
+
+	return badge
+}
+
+/**
  * Handles adding meta data to `Sidebar` `EnrichedNavItem` objects.
  *  - For `EnrichedLinkNavItem` objects, an `isActive` property will be added.
  *  - For `EnrichedSubmenuNavItem` objects:
@@ -36,6 +57,29 @@ export const addNavItemMetaData = (
 
 	const itemsWithMetadata = items.map(
 		(item: EnrichedNavItem): NavItemWithMetaData => {
+			let itemCopy = { ...item }
+
+			/**
+			 * If a `badge` object can be determined from a `title` with `<sup>`,
+			 * create the `badge` object and remove the `<sup>` tags from the title.
+			 */
+			if (item.hasOwnProperty('title')) {
+				const itemWithTitle = item as
+					| EnrichedSubmenuNavItem
+					| EnrichedLinkNavItem
+				const badge =
+					itemWithTitle.badge ?? getBadgeFromTitle(itemWithTitle.title)
+				const title = badge
+					? itemWithTitle.title.replace(`<sup>${badge.text}</sup>`, '').trim()
+					: itemWithTitle.title
+
+				itemCopy = {
+					...itemCopy,
+					badge: badge,
+					title: title,
+				}
+			}
+
 			// Found an `EnrichedSubmenuNavItem` object
 			if (item.hasOwnProperty('routes')) {
 				const result = addNavItemMetaData(
@@ -47,7 +91,7 @@ export const addNavItemMetaData = (
 				foundActiveItem = hasActiveChild || foundActiveItem
 
 				return {
-					...item,
+					...itemCopy,
 					routes: result.itemsWithMetadata,
 					hasActiveChild,
 				} as SubmenuNavItemWithMetaData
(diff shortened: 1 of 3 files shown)


src/components/sidebar/helpers/add-nav-item-meta-data.ts
```diff
@@ -12,6 +12,27 @@ interface AddNavItemMetaDataResult {
 	itemsWithMetadata: NavItemWithMetaData[]
 }
 
+/**
+ * Returns an object for rendering a Badge in a sidebar item if a <sup> tag is
+ * founnd in its title.
+ */
+const getBadgeFromTitle = (title: string) => {
```
kendallstrautman, line 19: Is the plan to eventually update this in the source content? If so, maybe we leave some note breadcrumbs about this being temporary 🤔

src/components/sidebar/helpers/add-nav-item-meta-data.ts
```diff
@@ -12,6 +12,27 @@ interface AddNavItemMetaDataResult {
 	itemsWithMetadata: NavItemWithMetaData[]
 }
 
+/**
+ * Returns an object for rendering a Badge in a sidebar item if a <sup> tag is
+ * founnd in its title.
```
kendallstrautman, line 17:
```suggestion
 * found in its title.
```
⛏️

src/components/sidebar/helpers/add-nav-item-meta-data.ts
```diff
@@ -12,6 +12,27 @@ interface AddNavItemMetaDataResult {
...
+ */
+const getBadgeFromTitle = (title: string) => {
+	let badge = null
+
+	const regex = new RegExp(/<sup>(.*)<\/sup>$/)
```
kendallstrautman, line 22:
```suggestion
	const regex = new RegExp(/^<sup>(.*)<\/sup>$/)
```
Should the use always start with `<sup>`?

src/components/sidebar/helpers/add-nav-item-meta-data.ts
ashleemboyer: I don't think we've discussed a solid plan yet, but I'd love for this to be temporary! I think there's some work that'll need to be done first:

- Sunset the docs pages on .io sites
- Get content conformance checks in place, so we can maintain the rule of "no `<sup>` tags in nav data titles"
- Update all content to conform to the check (and IMO past versions of docs too, not just the latest)

Regardless, I'll add a comment so we know why it's here 😊

src/components/sidebar/helpers/add-nav-item-meta-data.ts
ashleemboyer: Nope, the titles don't need to start with `<sup>`! Here's an example of the usage (and I'll add one to the comment I'm writing): https://github.com/hashicorp/cloud.hashicorp.com/blob/33c65ea6659a234b13ce70ce8fc30d82a3e53383/data/docs-nav-data.json#L158

zchsh (approved):
✅ Ran through the testing steps, looks good to me!

As an aside very much outside the contents of this PR, does the badge placement feel a little inconsistent to anyone else? 🤔 Definitely a hard thing to design for... especially as content can be so variable.

<details>
<summary>Example of where the badge placement feels slightly inconsistent and a little off to me (with constructed example of a longer title)</summary>

CleanShot 2022-10-07 at 10 56 55@2x


</details>

ashleemboyer:
> ✅ Ran through the testing steps, looks good to me!
> 
> As an aside very much outside the contents of this PR, does the badge placement feel a little inconsistent to anyone else? 🤔 Definitely a hard thing to design for... especially as content can be so variable.
> 
> Example of where the badge placement feels slightly inconsistent and a little off to me (with constructed example of a longer title)

@zchsh I think I see what you're saying... is it that the `Badge`s aren't right-aligning with each other due to the chevron icons on submenu nav items?

zchsh:
@ashleemboyer yeah exactly! was kind of curious whether anyone else has opinions on whether it'd seem worthwhile to explore other placements (eg maybe as an inline item, on a line above the sidebar item text?) 

heck though i should probably use the feedback form for thoughts like that 😅 will go do that!

ashleemboyer:
> @ashleemboyer yeah exactly! was kind of curious whether anyone else has opinions on whether it'd seem worthwhile to explore other placements (eg maybe as an inline item, on a line above the sidebar item text?)
> 
> heck though i should probably use the feedback form for thoughts like that 😅 will go do that!

@zchsh my guess is maybe we just haven't come across this case yet! It'd be a good thing to put in a board somewhere for post-GA exploration. 😊

PR #1194, opened 2022-10-06
fix: progress-icon hover and active colors

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task]  🎟️

## 🗒️ What

Updates the tutorial sidebar `TutorialListItem` to display the correct progress icon colors on `:hover`, `:focus`, and "active" states.

> **Note**: "active" in this context means "the tutorial link links to the current page", not "the tutorial link is `:active`". We could potentially consider renaming this state to `isCurrent` for clarity, but that seems outside the scope of this PR.

## 🤷 Why

To match design spec.

## 🛠️ How

- Uses `@nest` selector in `ProgressIcon`. This allows `ProgressIcon` to change color based on the `:hover` & `:focus` state of the enclosing `<a />`.
- Passes in an `isActive` prop so that `ProgressIcon` can change color based on the active state of the parent `TutorialListItem`

## 📸 Design Screenshots

🎨 [Figma][figma]

### Design Spec

CleanShot 2022-10-06 at 20 58 11@2x

CleanShot 2022-10-06 at 20 58 32@2x

### Screenshot from this PR

CleanShot 2022-10-07 at 10 45 40

## 🧪 Testing

- [ ] Visit [the preview link][preview], and log in so that user progress features are visible
- [ ] In the sidebar on any tutorial view, such as [/waypoint/tutorials/get-started-nomad/get-started-nomad][/waypoint/tutorials/get-started-nomad/get-started-nomad]...
    - Tutorials that are `completed` and `in_progress` have icons next to them.
    - These icons should show the correct `:hover` and `:focus` icon colors in the sidebar, which are slightly darker than the default state (the difference is subtle!).
    - As well, when a `completed` or `in_progress` tutorial is the active page, the icon color should darken slightly further.

## 💭 Anything else?

🤔  Use of `@nest` CSS selector here is intended to allow us to keep the `ProgressIcon` component less intertwined with related components. One caveat is that we only target `a` > `ProgressIcon` for hover states. In the future, we may also need to target for example `button` > `ProgressIcon`, but for this PR, that felt like a premature addition. This felt efficient, but a little unorthodox, so felt worth calling out.

💭  An alternative approach might be to add a ref to the `a` element we need to know hover and focus state for, by adding this ref in `SidebarNavLinkItem` (using `useHover`, and a new `useFocus` hook, for example). We'd then also need to change the `trailingIcon` prop in `SidebarNavLinkItem` to a render prop, that passes props through to the icon, perhaps with something like `renderTrailingIcon({ isHovered, isFocused, isActive })`. Then, instead of `@nest`, we'd use additional classNames for all the color difference states (much like `.isActive` is already set up).

> **Note**: I'm on the fence about the `@nest` approach, but it felt easier to propose as a starting point. Happy to dive into the `useHover` / `useFocus` ref approach - it does seem like it might be a little more robust in some way 🤔 

[preview]: https://dev-portal-git-zsrefine-progress-icon-color-hashicorp.vercel.app
[/waypoint/tutorials/get-started-nomad/get-started-nomad]: https://dev-portal-git-zsrefine-progress-icon-color-hashicorp.vercel.app/waypoint/tutorials/get-started-nomad/get-started-nomad
[task]: https://app.asana.com/0/1202097197789424/1202964504180132/f
[figma]: https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=11687%3A117718


zchsh:
@BRKalow @braydenlove This PR fixes up the progress icons in tutorial sidebars so that they have the correct coloration on `:hover`, `:focus`, and when the tutorial item represents the current page (aka `isActive`).

kendallstrautman (approved):
Good catch! definitely a v subtle color different 💅

PR #1200, opened 2022-10-07
[Tutorials Library] Handling unsupported refinements in CurrentFilterItem

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambtl-invalid-param-values-hashicorp.vercel.app/tutorials/library) 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1203115842976154/f) 🎟️

## 🗒️ What

- Prevents client-side errors and `console.error`s in DEV environments when an unsupported refinement is found in `CurrentFilterItem`

## 🧪 Testing

- [ ] Go to [/tutorials/library?product=nope](https://dev-portal-git-ambtl-invalid-param-values-hashicorp.vercel.app/tutorials/library?product=nope)
- [ ] The page should not error
- [ ] The page should not show an empty tag
- [ ] Go to [/tutorials/library?edition=nope](https://dev-portal-git-ambtl-invalid-param-values-hashicorp.vercel.app/tutorials/library?edition=nope)
- [ ] The page should not error
- [ ] The page should not show an empty tag
- [ ] Go to [/tutorials/library?randomParamdom=nope](https://dev-portal-git-ambtl-invalid-param-values-hashicorp.vercel.app/tutorials/library?randomParamdom=nope)
- [ ] The page should not error
- [ ] The page should not show an empty tag

### DIFF
--- a/src/views/tutorial-library/components/current-filters/index.tsx
+++ b/src/views/tutorial-library/components/current-filters/index.tsx
@@ -48,6 +48,7 @@ function CurrentFilterItem({ refinement, refine }: CurrentFilterItemProps) {
 	 */
 	if (!labelText) {
 		if (IS_DEV) {
+			// TODO - render something users can see?
 			console.error(
 				'[CurrentFilterItem] Found an unsupported refinement:',
 				JSON.stringify(refinement)
--- a/src/views/tutorial-library/components/current-filters/index.tsx
+++ b/src/views/tutorial-library/components/current-filters/index.tsx
@@ -17,10 +17,13 @@ interface CurrentFilterItemProps {
 	refine: (refinement: Refinement) => void
 }
 
+const IS_DEV = process.env.NODE_ENV !== 'production'
+
 function CurrentFilterItem({ refinement, refine }: CurrentFilterItemProps) {
 	const { label, type, attribute } = refinement
 
-	let labelText = label
+	let labelText
+
 	// This is a "Resource" filter
 	if (type === 'disjunctive') {
 		const resource = RESOURCES.find(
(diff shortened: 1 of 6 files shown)


src/views/tutorial-library/components/current-filters/index.tsx
```diff
@@ -17,10 +17,13 @@ interface CurrentFilterItemProps {
 	refine: (refinement: Refinement) => void
 }
 
+const IS_DEV = process.env.NODE_ENV !== 'production'
+
 function CurrentFilterItem({ refinement, refine }: CurrentFilterItemProps) {
 	const { label, type, attribute } = refinement
 
-	let labelText = label
+	let labelText
```
kendallstrautman, line 25: Whats the context on not wanting to default to `label`?

kendallstrautman: Oh I see, so that we can check if its undefined

kendallstrautman: Are all of the attributes or label possibilities handled in these conditionals below? I just want to make sure we're not removing a default that is actually needed.

kendallstrautman (commented):
🤔 My expectation would be that if there weren't any valid filters applied that the default tutorial grid would render, not 'no results'. Because there are no items to 'clear', thoughts?

Perhaps that fix goes beyond this patch though.

src/views/tutorial-library/components/current-filters/index.tsx
ashleemboyer: > Are all of the attributes or label possibilities handled in these conditionals below? I just want to make sure we're not removing a default that is actually needed.

Yep, I'm pretty sure! I'm hoping the last test case gives us the answer here. Algolia won't do anything with a parameter it doesn't accept, and these ones listed here are all of the ones we accept AFAIK. @BRKalow can you confirm? 😊

ashleemboyer:
> 🤔 My expectation would be that if there weren't any valid filters applied that the default tutorial grid would render, not 'no results'. Because there are no items to 'clear', thoughts?
> 
> Perhaps that fix goes beyond this patch though.

@kendallstrautman I think that'd be a good fix beyond this patch.

Algolia responds like this to query param keys it doesn't accept, but it doesn't check the value of the query param. So it seems like it assumes that whatever query param key it's given has a valid value. But I would think we could add a check somewhere that prevents a search from happening based on the value.

src/views/tutorial-library/components/current-filters/index.tsx
BRKalow: Yep, that's correct!

kendallstrautman (approved):

kendallstrautman:
Nice catch!

PR #1203, opened 2022-10-07
feat: use static paths endpoint for docs static paths

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Uses our analytics data to return the most visited paths from `getStaticPaths` for our docs sections and tutorials sections.

- Introduces a new function, `getStaticPathsFromAnalytics`, which handles fetching from the `static_paths` endpoint and filtering the result against a provided list of `validPaths`.

```ts
const paths = await getStaticPathsFromAnalytics({
	limit: __config.dev_dot.max_static_paths ?? 0,
	pathPrefix: `/${product.slug}/${basePath}`,
	validPaths: pathsFromNavData,
})
```

## 🤷 Why

This ensures we are rendering the most important pages at build time, with the rest being deferred to post-build ISR.


config/development.json
```diff
@@ -1,13 +1,13 @@
...
-		"max_static_paths": 0
+		"max_static_paths": 1
 	},
 	"dev_dot": {
-		"max_static_paths": 0
+		"max_static_paths": 1
 	},
 	"learn": {
-		"max_static_paths": 0
+		"max_static_paths": 1
```
BRKalow, line 10: Just updated these values to validate in development that everything was working as expected. There should be no functional difference as the static paths are not used in development. 👀

src/pages/[productSlug]/tutorials/[...tutorialSlug]/index.tsx
```diff
@@ -5,20 +5,53 @@ import {
...
+
+						return result
+					})
+				})
+			)
+		).flat()
+	} catch {
+		// In the case of an error, fallback to using the base list of generated paths to ensure we do _some_ form of static generation
+		paths = validPaths.slice(0, __config.learn.max_static_paths ?? 0)
+	}
```
Comment (BRKalow) [inline src/pages/[productSlug]/tutorials/[...tutorialSlug]/index.tsx:51]: This is a little funky, but due to the dynamic `productSlug` path portion we've gotta do some trickery. The `validPaths` filtering still works because we ignore the `productSlug` path segment.

In the future, I could see us expanding the `static_paths` endpoint to support multiple levels of params, but this works for now!

kendallstrautman (approved):
👏

PR #1209, opened 2022-10-11
fix: code-block in alert styles

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Fixes an issue where code blocks within alerts did not display correctly.

## 🛠️ How

Adds specificity to styles intended for inline `code`, to avoid them leaking into code-block's `pre > code` element.

### Before

before

### After

after

## 🧪 Testing

- [ ] Visit [the preview][preview]
    - The code block in the alert should be legible. It should not be taking on leaky styles intended for inline code.

## 💭 Anything else?

I think longer term we likely want to switch from our custom ".alert" syntax (`->`, `!>`, `~>`, `=>`) to a custom MDX component, like `<Callout />` or `<Aside />` or something. This would allow us to better scope our CSS, and avoid (recurring) issues such as this one. And it would let us build more interesting callout or aside components (for example, adding support for icons, which has been a feature request in the past). Related:
- 🎟 [Alert aka Callout aka Aside as MDX components, with support for existing custom syntax via Remark](https://app.asana.com/0/1100423001970639/1200034293471289/f)
- 🤔 [Old spike to prove out a remark plugin that would ease the transition](https://github.com/hashicorp/learn/pull/3173)

[task]: https://app.asana.com/0/1202097197789424/1203142623924248/f
[preview]: https://dev-portal-git-zsfix-code-block-in-alert-hashicorp.vercel.app/vault/docs/enterprise/fips/fips1402?optInFrom=vault-io#validating-fips-140-2-inside


BRKalow (approved):
👏

PR #1224, opened 2022-10-13
feat: data-heap-track on toc-list-item

zchsh:
## 🔗 Relevant links 

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Adds `data-heap-track` attributes to table-of-contents anchor links.

## 🤷 Why

To ensure table-of-contents interactions can be analyzed via Heap.

## 🧪 Testing

- [ ] Visit any tutorial or documentation page in [the preview][preview] (or any page with a table of contents in the sidecar area!)
    - Each anchor link in the table of contents should have a `data-heap-track="toc-list-item-index-<index>"` attribute.

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/0/1203161988284273/f
[preview]: https://dev-portal-git-zstoc-heap-track-hashicorp.vercel.app


zchsh:
@BRKalow @topfunky This PR adds a `data-heap-track="toc-list-item-index-<index>"` attribute to table-of-contents `<a />` tags, to make sure those elements can be consistently tracked in Heap.

src/layouts/sidebar-sidecar/components/table-of-contents/index.tsx
```diff
@@ -53,7 +64,11 @@ const TableOfContentsListItem = ({ isActive, slug, title }): ReactElement => {
 
 	return (
 		<li className={className} key={slug}>
-			<a className={s.tableOfContentsListItemAnchor} href={`#${slug}`}>
+			<a
+				className={s.tableOfContentsListItemAnchor}
+				href={`#${slug}`}
+				data-heap-track={`toc-list-item-index-${index}`}
```
BRKalow, line 70: Is there any reason we might want the `slug` here? 💭

zchsh: I had it previously, and then realized it seemed very redundant with `href` 🤔

For example, in Heap I think if we wanted to select an element with a specific slug, we could use the CSS selector `a[data-heap-track*=toc-list-item][href=#<slug>]` for example. But could definitely add it back in if we have a concrete use case for it 👍

BRKalow: Ah that makes sense, thanks for the insight!

BRKalow (approved):

topfunky (approved):
👍🏻

PR #1233, opened 2022-10-14
feat: generic heap-track for cta components

zchsh:
## 🔗 Relevant links

- [Asana task](url) 🎟️


ashleemboyer (commented):
Question 🙋🏻 Should `ButtonLink` and `StandaloneLink` also accept `data-heap-track` and prefix the final value with their names (`button-link` and `standalone-link`) like `button` is doing?

src/components/button-link/index.tsx
```diff
@@ -58,7 +58,7 @@ const ButtonLink = ({
 				rel={openInNewTab ? 'noreferrer noopener' : undefined}
 				target={openInNewTab ? '_blank' : '_self'}
 				onClick={onClick}
-				data-heap-track={dataHeapTrack}
+				data-heap-track={`button-link ${dataHeapTrack}`}
```
ashleemboyer, line 61: 💭 Maybe `dataHeapTrack` should be defaulted to `''` in the props destructuring for the cases where it's not passed in?

zchsh: ah true! one sec...

zchsh: ✅ should be fixed now!

PR #1240, opened 2022-10-14
Adding displayName to Contexts for devtools debugging

ashleemboyer:
## 🗒️ What

Sets the `displayName` property for all of our `Context`s.

## 🤷 Why

It's helpful when debugging with React devtools: https://reactjs.org/docs/context.html#contextdisplayname.

## 🧪 Testing

- [ ] Go to the preview url
- [ ] Open the React devtools
- [ ] Make sure `Context`s have the right display name
  - Note: some `Context`s do not have a display name because they are rendered by third-party code


kendallstrautman (approved):
🧼 💯

PR #1266, opened 2022-10-19
feat(redirects): load redirects for terraform and HCP

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Load redirects from remote repositories for terraform and HCP.


build-libs/redirects.js
```diff
@@ -179,6 +190,9 @@ async function buildProductRedirects() {
 	// Fetch author-oriented redirects from product repos,
 	// and merge those with dev-oriented redirects from
 	// within this repository
+	if (process.env.SKIP_BUILD_PRODUCT_REDIRECTS) {
+		return []
+	}
```
BRKalow, line 195: I added this so that our bundle analysis workflow doesn't attempt to fetch redirects from the GH API and as a result require a GitHub PAT.

PR #1269, opened 2022-10-20
fix: client-side error on error page

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

It appears that when navigating to the error page via a client side navigation (e.g. a 404) this bit of code is erroring as `currentProduct` becomes undefined. I'm not sure if this is the right fix, but it's something to try.

### DIFF
--- a/src/layouts/docs-view-layout/index.tsx
+++ b/src/layouts/docs-view-layout/index.tsx
@@ -13,10 +13,10 @@ import getIsBetaProduct from 'lib/get-is-beta-product'
 const DocsViewLayout = (props: SidebarSidecarLayoutProps) => {
 	const currentProduct = useCurrentProduct()
 	// Products will be removed from the beta array as we shift them to GA, which will hide the opt out button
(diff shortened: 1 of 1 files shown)


zchsh (approved):
👍 approving purely based on the fact this seems like a safe thing to do since `currentProduct` isn't guaranteed to be defined. +1 that it seems like it maybe be related to the missing-versioned-doc issue.

zchsh (commented):
I had another thought here, which is that we probably have `currentProduct` equivalent or some form of `currentProduct.slug` coming through from `getStaticProps`. If that's true, maybe using that could potentially fix the issue without a flash-of-`undefined`-product we might see otherwise?

BRKalow:
@zchsh yeah, so I think the issue here is that when we return `{ notFound: true }` from `getStaticProps`, we lose the `currentProduct` data, along with any other props we might pass. So we need to handle the `notFound` case at a higher level to ensure we aren't trying to render the page without any data.

Screen Shot 2022-10-20 at 10 02 43 AM

PR #1271, opened 2022-10-20
fix(hcp api docs): update packer api spec location

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

It appears that the HCP Packer API specs have moved to a new repository. I've updated our code to pull from the new location.


zchsh (approved):
✅ 🔗  good call - I can see the spec file in that repo, so going off of that (and assuming things are broken currently). But can definitely double-check the preview build once it's done if that's helpful 👍

PR #1278, opened 2022-10-21
feat: Packer GA

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

This PR removes `"packer"` from `beta_product_slugs`, as part of launching Packer into GA. 

[task]: https://app.asana.com/0/0/1203184381598140/f
[preview]: https://dev-portal-git-zsrm-packer-from-beta-hashicorp.vercel.app


BRKalow:
@zchsh just to save another PR, want to update the .io site nav links to point to Developer as part of this? They should be in `src/layouts/_proxied-dot-io/packer` 💭

BRKalow:
OH it looks like you did it here 😄 https://github.com/hashicorp/dev-portal/pull/1282

zchsh:
@BRKalow I can merge #1282 into this PR if we want! I was on the fence whether to roll them together. Leaned slightly towards keeping them separate in the unlikely off-chance we need to roll back.

BRKalow:
@zchsh Fair enough! I think combining them makes it easier to rollback all of the related changes if we do need to do that.

zchsh:
@BRKalow ah good point, that makes sense! I've merged https://github.com/hashicorp/dev-portal/pull/1282 in here, so this PR now has both the `beta` and nav-link URL changes ✅

BRKalow (approved):
👍

PR #1281, opened 2022-10-21
refactor: use happykit for auth flag

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-kshappykit-auth-flag-hashicorp.vercel.app/) 🔎
- [Asana task](url) 🎟️


src/hooks/use-authentication/index.ts
```diff
@@ -39,7 +40,7 @@ const useAuthentication = (
 	}, [data?.error])
 
 	// Deriving booleans about auth state
-	const isAuthEnabled = AUTH_ENABLED
+	const isAuthEnabled = flags?.enableAuth
```
kendallstrautman, line 43: Kyle wanted to use happykit so we could try out percentage based auth rollouts. One thing I don't like about this as opposed to the previous auth flag boolean is the delay in the UI showing since the default is false — feels jumpy and weird.

BRKalow: Yeah, this is the big downfall of runtime feature flagging, especially so for global UI elements. I think for the auth components it's a bit less pronounced as they pop in anyways due to auth needing to resolve, but it delays that process.

kendallstrautman: Word, it does seem to be a similar behavior with or without happykit. I think the best thing would be to follow up and try to minimize this jump / remove the feature flag once we know things are stable.

zchsh (approved):

ashleemboyer (approved):

PR #1301, opened 2022-10-26
[NextAuth] Adding signOut event to sign out of provider

ashleemboyer:
## 🗒️ What

- Configures a `signOut` function in NextAuth's `events` configuration object
- Persists the `id_token` on `account` in `jwt()` for the new `signOut` event function

## 🤷 Why

- According to https://github.com/nextauthjs/next-auth/discussions/3938, NextAuth does not log out of an auth provider entirely for us. So we need to do an extra step on `signOut` in order to completely log out of Cloud IDP.

## 🛠️ How

- [`NextAuth.options.events.signOut`](https://next-auth.js.org/configuration/options#events)

## 🧪 Testing

- [ ] Go to the home page on the preview
- [ ] Sign in
- [ ] Sign out
- [ ] Go to /sign-up
- [ ] Click the sign up button
- [ ] You should be taken to the cloud idp sign up page:
  image
- [ ] Go to the home page
- [ ] Click the sign in button
- [ ] You should be taken to the cloud idp sign in page:
  image

### DIFF
--- a/src/pages/api/auth/[...nextauth].ts
+++ b/src/pages/api/auth/[...nextauth].ts
@@ -1,5 +1,6 @@
 import createFetch from '@vercel/fetch'
 import NextAuth from 'next-auth'
+import { URL } from 'url'
 import CloudIdpProvider from 'lib/auth/cloud-idp-provider'
 import refreshTokenSet from 'lib/auth/refresh-token-set'
 import isJwtExpired from 'lib/auth/is-jwt-expired'
@@ -26,7 +27,11 @@ export default NextAuth({
 					await fetch(CloudIdpProvider.wellKnown)
 				).json()
 				const endSessionEndpoint = wellKnownConfiguration.end_session_endpoint
-				await fetch(`${endSessionEndpoint}?id_token_hint=${token.id_token}`)
+
+				const endSessionUrl = new URL(endSessionEndpoint)
+				endSessionUrl.searchParams.set('id_token_hint', token.id_token)
+
+				await fetch(endSessionUrl.toString())
 			} catch (e) {
 				console.error(
 					'[NextAuth] There was an error in the `signOut` event:',
--- a/src/pages/api/auth/[...nextauth].ts
+++ b/src/pages/api/auth/[...nextauth].ts
@@ -16,7 +16,7 @@ export default NextAuth({
 	// https://next-auth.js.org/configuration/options#events
 	events: {
 		/**
-		 * NOTE: NextAuth log out of auth providers, so we have to handle doing that
+		 * NOTE: NextAuth does not log out of auth providers, so we have to handle doing that
 		 * ourselves in this signOut event.
 		 * https://github.com/nextauthjs/next-auth/discussions/3938
 		 *
--- a/src/pages/api/auth/[...nextauth].ts
+++ b/src/pages/api/auth/[...nextauth].ts
@@ -1,14 +1,33 @@
+import createFetch from '@vercel/fetch'
 import NextAuth from 'next-auth'
 import CloudIdpProvider from 'lib/auth/cloud-idp-provider'
 import refreshTokenSet from 'lib/auth/refresh-token-set'
 import isJwtExpired from 'lib/auth/is-jwt-expired'
 import { AuthErrors } from 'types/auth'
 
+const fetch = createFetch()
+
 const isDev = process.env.NODE_ENV === 'development'
 
 export default NextAuth({
 	session: {
 		maxAge: 2592000, // 30 days
+	},
+	// https://next-auth.js.org/configuration/options#events
+	events: {
+		/**
+		 * NOTE: NextAuth log out of auth providers, so we have to handle doing that
+		 * ourselves in this signOut event.
+		 * https://github.com/nextauthjs/next-auth/discussions/3938
+		 *
+		 * cloud-idp and Ory Hydra docs:
+		 * https://hashicorp.atlassian.net/wiki/spaces/CLOUD/pages/2321449597/cloud-idp+and+Ory+Hydra
+		 */
+		async signOut({ token }) {
+			await fetch(
+				`${__config.dev_dot.auth.idp_url}/oauth2/sessions/logout?id_token_hint=${token.id_token}`
+			)
+		},
 	},
 	providers: [CloudIdpProvider],
 	callbacks: {
@@ -24,9 +43,11 @@ export default NextAuth({
 
 			// initial call during sign in
 			if (isInitial) {
-				// persist access_token and refresh_token on the session token
+				// persist access_token, refresh_token, and id_token on the session token
+				// id_token is needed for the signOut event above
 				token.access_token = account.access_token
 				token.refresh_token = account.refresh_token
+				token.id_token = account.id_token
 
 				// Picture is passed to session.user.image
 				token.picture = profile.picture as string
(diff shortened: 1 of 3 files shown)


kendallstrautman (approved):

src/pages/api/auth/[...nextauth].ts
```diff
@@ -1,15 +1,34 @@
...
+		 * NOTE: NextAuth log out of auth providers, so we have to handle doing that
+		 * ourselves in this signOut event.
+		 * https://github.com/nextauthjs/next-auth/discussions/3938
+		 *
+		 * cloud-idp and Ory Hydra docs:
+		 * https://hashicorp.atlassian.net/wiki/spaces/CLOUD/pages/2321449597/cloud-idp+and+Ory+Hydra
+		 */
+		async signOut({ token }) {
+			await fetch(
+				`${__config.dev_dot.auth.idp_url}/oauth2/sessions/logout?id_token_hint=${token.id_token}`
```
Comment (kendallstrautman) [inline src/pages/api/auth/[...nextauth].ts:28]: My only future thought as an improvement is that we could construct this url with the URL API, but this totally works!

src/pages/api/auth/[...nextauth].ts
```diff
@@ -1,15 +1,34 @@
...
 const isDev = process.env.NODE_ENV === 'development'
 
 export default NextAuth({
 	session: {
 		maxAge: 2592000, // 30 days
 	},
+	// https://next-auth.js.org/configuration/options#events
+	events: {
+		/**
+		 * NOTE: NextAuth log out of auth providers, so we have to handle doing that
```
Comment (BRKalow) [inline src/pages/api/auth/[...nextauth].ts:19]: 💭 are we missing a word in this comment? `...NextAuth (does not?) log out of auth providers...`?

Comment (ashleemboyer) [reply src/pages/api/auth/[...nextauth].ts:19]: Haha yes! Oops 🤣

Comment (ashleemboyer) [inline src/pages/api/auth/[...nextauth].ts:19]:
```suggestion
		 * NOTE: NextAuth does not log out of auth providers, so we have to handle doing that
```

src/pages/api/auth/[...nextauth].ts
```diff
@@ -1,15 +1,34 @@
...
+		 * ourselves in this signOut event.
+		 * https://github.com/nextauthjs/next-auth/discussions/3938
+		 *
+		 * cloud-idp and Ory Hydra docs:
+		 * https://hashicorp.atlassian.net/wiki/spaces/CLOUD/pages/2321449597/cloud-idp+and+Ory+Hydra
```
Comment (BRKalow) [inline src/pages/api/auth/[...nextauth].ts:24]: As this repo is open source, it might be best if we don't link to internal references like this. Maybe we can explain that this is the `end_session_endpoint` from `.well-known` file?

Comment (ashleemboyer) [reply src/pages/api/auth/[...nextauth].ts]: Updated here: https://github.com/hashicorp/dev-portal/pull/1301/commits/aadc63ce12859d807812f58876295ea0fe200dc4

src/pages/api/auth/[...nextauth].ts
```diff
@@ -1,15 +1,34 @@
...
+		 */
+		async signOut({ token }) {
+			await fetch(
+				`${__config.dev_dot.auth.idp_url}/oauth2/sessions/logout?id_token_hint=${token.id_token}`
+			)
+		},
```
Comment (BRKalow) [inline src/pages/api/auth/[...nextauth].ts:27]: nit: Should we wrap this in a `try/catch`? Or is there not really a meaningful way to handle the error here?

If we do not end up doing anything with the result here, we can just return the fetch call:
```suggestion
		signOut({ token }) {
			return fetch(
				`${__config.dev_dot.auth.idp_url}/oauth2/sessions/logout?id_token_hint=${token.id_token}`
			)
		},
```

src/pages/api/auth/[...nextauth].ts
Comment (ashleemboyer) [reply src/pages/api/auth/[...nextauth].ts]: Fixed here: https://github.com/hashicorp/dev-portal/pull/1301/commits/71cb1bbedb4900fd4ce9a3ea4bdd9e734cef1046

src/pages/api/auth/[...nextauth].ts
Comment (ashleemboyer) [reply src/pages/api/auth/[...nextauth].ts:27]: I tried returning the `await fetch` originally, but there's a TS issue with that:

```
Type '({ token }: { session: Session; token: JWT; }) => Promise<Response>' is not assignable to type '(message: { session: Session; token: JWT; }) => Awaitable<void>'.
  Type 'Promise<Response>' is not assignable to type 'Awaitable<void>'.
    Type 'Promise<Response>' is not assignable to type 'PromiseLike<void>'.
      Types of property 'then' are incompatible.
        Type '<TResult1 = Response, TResult2 = never>(onfulfilled?: (value: Response) => TResult1 | PromiseLike<TResult1>, onrejected?: (reason: any) => TResult2 | PromiseLike<...>) => Promise<...>' is not assignable to type '<TResult1 = void, TResult2 = never>(onfulfilled?: (value: void) => TResult1 | PromiseLike<TResult1>, onrejected?: (reason: any) => TResult2 | PromiseLike<TResult2>) => PromiseLike<...>'.
          Types of parameters 'onfulfilled' and 'onfulfilled' are incompatible.
            Types of parameters 'value' and 'value' are incompatible.
              Type 'Response' is not assignable to type 'void'.ts(2322)
```

Comment (ashleemboyer) [reply src/pages/api/auth/[...nextauth].ts]: Wrapped in a `try/catch` here! I think we'll be able to catch these errors in datadog now, right? https://github.com/hashicorp/dev-portal/pull/1301/commits/d85fd5623db7ff2137f3e2fff7b2e3983eb24647

Comment (BRKalow) [reply src/pages/api/auth/[...nextauth].ts]: @ashleemboyer yeah, datadog will get these logs!

src/pages/api/auth/[...nextauth].ts
```diff
@@ -1,15 +1,38 @@
...
+	events: {
+		/**
+		 * NOTE: NextAuth does not log out of auth providers, so we have to handle doing that
+		 * ourselves in this signOut event.
+		 * https://github.com/nextauthjs/next-auth/discussions/3938
+		 */
+		async signOut({ token }) {
+			try {
+				await fetch(
+					`${__config.dev_dot.auth.idp_url}/oauth2/sessions/logout?id_token_hint=${token.id_token}`
```
Comment (squaresurf) [inline src/pages/api/auth/[...nextauth].ts:26]: It would be best to get this url from the well known configuration (i.e. `${__config.dev_dot.auth.idp_url}/.well-known/openid-configuration`).

src/pages/api/auth/[...nextauth].ts
```diff
@@ -1,15 +1,38 @@
...
+	// https://next-auth.js.org/configuration/options#events
+	events: {
+		/**
+		 * NOTE: NextAuth does not log out of auth providers, so we have to handle doing that
+		 * ourselves in this signOut event.
+		 * https://github.com/nextauthjs/next-auth/discussions/3938
+		 */
+		async signOut({ token }) {
+			try {
+				await fetch(
```
Comment (squaresurf) [inline src/pages/api/auth/[...nextauth].ts:25]: I don't see this fetch when logging out with the vercel app. Do you know why that may be?

Comment (dstaley) [reply src/pages/api/auth/[...nextauth].ts:25]: If you mean in your browser's network inspector, I believe this network request is server-side.

Comment (squaresurf) [reply src/pages/api/auth/[...nextauth].ts:25]: Ah! I didn't realize that this was server side. Thank you for clarifying.

src/pages/api/auth/[...nextauth].ts
Comment (ashleemboyer) [reply src/pages/api/auth/[...nextauth].ts:26]: That sounds like a great idea! I'll make that update

src/pages/api/auth/[...nextauth].ts
Comment (ashleemboyer) [reply src/pages/api/auth/[...nextauth].ts:25]: Can confirm! This request happens server-side

squaresurf (approved):

PR #1304, opened 2022-10-27
chore: ignore unprefixed product redirects

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Filters redirects incoming from product repositories, to only allow those redirects that start with `/${product}`.

## 🤷 Why

Redirects that are not prefixed will not work as expected, so we want to ignore them, and provide feedback to authors.

[task]: https://app.asana.com/0/1202097197789424/1203186532172074/f
[preview]:  https://dev-portal-git-zsignore-unprefixed-redirects-hashicorp.vercel.app/


build-libs/redirects.js
```diff
@@ -167,20 +168,72 @@ async function getRedirectsForProduct(
 
 	let repo = product
 
-	const rawRedirects = isDeployPreview(product)
```
zchsh, line 170: This was flagged by linting as a nested ternary. Have refactored to an `if` block while I'm here 👍

build-libs/redirects.js
```diff
@@ -167,20 +168,72 @@ async function getRedirectsForProduct(
...
+	 *
+	 * We also print out a warning. Note that this warning may not be visible
+	 * to authors, unless they view logs for the preview build they're working on.
+	 */
+
+	/** @type {Redirect[]} */
+	const invalidRedirects = []
+	const validRedirects = parsedRedirects.filter((entry) => {
+		// Redirects must be prefixed with the product slug.
+		const isPrefixed = entry.source.startsWith(`/${product}`)
```
zchsh, line 216: For now `isPrefixed` seems like the only validity concern we have. I figure if others come up, probably worth splitting out a `filterValidRedirects` function or something 🤔 But for now this felt alright.

build-libs/redirects.js
```diff
@@ -167,20 +168,72 @@ async function getRedirectsForProduct(
...
+		if (!isPrefixed) {
+			invalidRedirects.push(entry)
+		}
+		return isPrefixed
+	})
+
+	/**
+	 * Log a warning for any invalid authored redirects
+	 */
+	if (invalidRedirects.length > 0) {
```
zchsh, line 227: Aimed for a single warning message, listing all invalid redirects. Happy to rethink this if there's another formatting that might seem preferable 🙇

BRKalow: I think this is the right approach, it makes it easier to see all of the issues at a glance!

build-libs/redirects.js
```diff
@@ -167,20 +168,72 @@ async function getRedirectsForProduct(
...
+		let message = `Found invalid redirects. Invalid redirects are ignored.`
+		message += ` Please ensure all redirects start with "/${product}".`
+		message += ` The following redirects must be updated to start with "/${product}":`
+		message += `\n${JSON.stringify(invalidRedirects, null, 2)}`
+		console.log(message)
+	}
```
BRKalow, line 233: 💭 What do you think about abstracting the filtering / validating into a separate method? This will make it easier to refactor the code away when/if we shift this to our future content conformance setup.

zchsh: I dig it, that makes a lot of sense to me! Will go do that now ... 🏃

zchsh: ✅ split out as separate fn, and added tests!

BRKalow (approved):
👏

zchsh:
Note before mergin: I'm seeing warnings in build logs here. Going to detail each of the warnings I'm seeing:
- For Boundary, backporting looks to have failed, so [redirects.js on stable-website](https://github.com/hashicorp/boundary/blob/stable-website/website/redirects.js) doesn't reflect the [cleared redirect.js on main](https://github.com/hashicorp/boundary/blob/main/website/redirects.js). This seems to be filtering working as expected 👍
- For `cloud.hashicorp.com`, redirects still need to be applied to that domain. This seems to be filtering working as expected 👍
- For `terraform-website`, this is 🚨 a problem 🚨 & I'm going to add a test to ensure that passing `terraform-website` as the product slug gets corrected 👍

BRKalow:
@zchsh I think `cloud.hashicorp.com` might need the same treatment as `terraform-website`. The redirects for that domain are defined here: https://github.com/hashicorp/cloud.hashicorp.com/blob/main/cloud.hashicorp.com.redirects.js the ones being loaded should apply to developer 👍

zchsh:
@BRKalow ah great catch - have added a step to transform the incoming `repoSlug` into a Dev-Dot-friendly `productSlug` before filtering redirects on `productSlug`, and have added tests for `terraform-website` and `cloud.hashicorp.com`.

For HCP, I do notice that the one redirect that would be filtered goes from `hcp/docs/packer/store-image-metadata/template-configuration` to `hcp/docs/packer/store-image-metadata/packer-template-configuration` (note the missing leading slash!). So, I think that's still going to get filtered out, although it's currently not working anyways (https://developer.hashicorp.com/hcp/docs/packer/store-image-metadata/template-configuration, combo with "prune old pages" problem so it still loads).

I'll open a PR on the `cloud.hashicorp.com` repo to fix up that redirect 👍

zchsh:
https://github.com/hashicorp/cloud.hashicorp.com/pull/555 has merged, so mergin' this in 👍

PR #1317, opened 2022-11-01
feat: run `saveAndLoadAnalytics()` after signin

pbortnick:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-pbsave-and-load-consent-signin-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1203274026595792/f) 🎟️

## 🗒️ What

Call `saveAndLoadAnalytics` at the right place in Dev Dots auth flow

## 🤷 Why

Legal confirmed that by signing in, users were also consenting to analytics and tracking.

## 🛠️ How

Use consent manager updates from https://github.com/hashicorp/react-components/pull/793 to run `saveAndLoadAnalytics()`in `useAuthentication` hook when user is authenticated and preferences are not set

## 🧪 Testing

In order to test you might have to be outside of the US or use a VPN. Open the preview link, make sure you are logged out and the consent manager banner is visible at the bottom of the page. If not, delete the `hashi-consent-preferences` cookie. Then log in and the consent manager should no longer be there.


mwickett (commented):
I tested this and it works great! I'll let others approve the implementation here, but functionality wise it looks good to me. 

Not for this PR, but up next, getting a `segment.identify` call in here somewhere will be great so that the `ajs_user_id` is set. This is related to the Marketo work, but I think we need to shift that connection from the client to the "backend" (here I mean data warehouse -> Marketo sync). I'm writing up a message to that effect.

CleanShot 2022-11-03 at 12 03 07@2x

ashleemboyer (commented):
This code looks great! The test failures are peculiar though. 🤔 Have you had a chance to investigate the errors?

package.json
```diff
@@ -164,6 +164,7 @@
 		"gray-matter": "^4.0.3",
 		"http-status-codes": "^2.2.0",
 		"isomorphic-unfetch": "^3.1.0",
+		"jest-raw-loader": "^1.0.1",
```
BRKalow, line 167: @pbortnick can we move this to `devDependencies`? As it's only used during testing, it makes the most sense to shift it there!

pbortnick: good call. Done! https://github.com/hashicorp/dev-portal/pull/1317/commits/f25dd062367b9994649fd50372dc5fd465c019c7

BRKalow (approved):
👏

PR #1322, opened 2022-11-03
Removing DevDotOptIn and references

ashleemboyer:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1202097197789424/1203272049975738/f) 🎟️

## 🗒️ What

- Removes `DevDotOptIn` component and what should be the remaining references to opt in/out code

## 🤷 Why

- We've GA'd so opt in/out functionality is no longer needed. 🥳 

## 🧪 Testing

- I don't think there's much to manually test in this PR since the UI pages that are affected get redirected to developer.hashicorp.com anyways. We should be okay to merge just based off of the PR checks.


zchsh (approved):
🔪 🎉  This looks good to me!

PR #1326, opened 2022-11-04
[hotfix] Calling signInWrapper on RefreshAccessTokenError

ashleemboyer:
## 🗒️ What

- Updates a `useEffect` in `useAuthentication` that was previously calling `signOutWrapper` on `RefreshAccessTokenError`s. It now calls `signInWrapper` instead.

## 🤷 Why

- Because the `error` on the `token` was never being cleared out, this code and the `signOut` event in `[...nextauth].tsx` were infinitely executing, leading to thousands of messages being logged under the `"failed to refresh token set"` error.
- The `signOutWrapper` invokes next-auth's `signOut` under the hood, where the default functionality is to redirect users to the home page of the app after signing out. This is extremely disruptive for authenticated users with several tabs open to read later when the `"failed to refresh token set"` error occurs. It causes them to lose track of the pages they were on and unnecessary frustration with the app.
- The choice to use `signInWrapper` came from taking another look at next-auth's [Refresh Token Rotation](https://next-auth.js.org/tutorials/refresh-token-rotation#client-side) tutorial, where they invoke `signIn` when this error occurs.

## 🧪 Testing

Since the refresh token cycle is 1 hour and not configurable, and we haven't yet confirmed how to reproduce the errors locally (this is the next thing I'm working on after this PR), I simulated the the `"failed to refresh token set"` error locally and did a screen recording showing via Slack that these changes work as expected.

So there are two options currently if we want to be really sure these changes work as expected:

- [ ] Check the branch out locally and do what I did in the video
- [ ] Hop on a call and I'll demo via screen share

## 💭 Anything else?

More to come today:

- Hopefully a fix for the frequent the `"failed to refresh token set"` errors
- Increased / refined logging monitor to include the `"failed to refresh token set"` error


BRKalow (approved):
Makes sense to me!

thiskevinwang (approved):

PR #1329, opened 2022-11-04
Prefixing RefreshAccessTokenError message with "[next-auth]"

ashleemboyer:
## 🗒️ What

- Prefixes our custom `RefreshAccessTokenError` message with `"[next-auth]"`

## 🤷 Why

- So an existing log monitor picks these errors up
- To group this error with others in hopes that it makes debugging the actual error easier

## 🧪 Testing

Nothing to test ✅

### DIFF
--- a/src/pages/api/auth/[...nextauth].ts
+++ b/src/pages/api/auth/[...nextauth].ts
@@ -91,7 +91,12 @@ export default NextAuth({
 					token.refresh_token = refresh_token
 				} catch (err) {
 					console.error('failed to refresh token set', err)
-					return { ...token, error: AuthErrors.RefreshAccessTokenError }
+					return {
+						...token,
+						// NOTE: temporary prefix of [next-auth] for existing log monitor
+						// TODO: abstract logging so everything is grouped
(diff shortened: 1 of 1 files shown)


mhonnatti (approved):

src/pages/api/auth/[...nextauth].ts
```diff
@@ -91,7 +91,12 @@ export default NextAuth({
 					token.refresh_token = refresh_token
 				} catch (err) {
 					console.error('failed to refresh token set', err)
-					return { ...token, error: AuthErrors.RefreshAccessTokenError }
+					return {
+						...token,
+						// NOTE: temporary prefix of [next-auth] for existing log monitor
+						// TODO: abstract logging so everything is grouped
+						error: `[next-auth] ${AuthErrors.RefreshAccessTokenError}`,
```
Comment (thiskevinwang) [inline src/pages/api/auth/[...nextauth].ts:98]: We should update this check as well:
https://github.com/hashicorp/dev-portal/blob/13562d4a7220b94c01a3cfdf58767519f8e9ea8d/src/hooks/use-authentication/index.ts#L46-L47

PR #1330, opened 2022-11-04
Prefixing actual RefreshAccessTokenError message for string equality checks

ashleemboyer:
## 🗒️ What

Moves the `"[next-auth]"` prefix on `RefreshAccessTokenError` to the definition of the error. The string equality check in `useAuthentication` was broken in #1329.


thiskevinwang (approved):
Noice!

ashleemboyer:
Great catch on this @thiskevinwang!! 👏🏻

PR #1338, opened 2022-11-08
Add auth error dialog

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambadd-auth-error-dialog-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1203346078532406/f) 🎟️

## 🗒️ What

- Adds new `AuthenticationErrorDialog` component
- Renders `AuthenticationErrorDialog` in `BaseNewLayout`
- Adds `error` and `hasError` properties to `UseAuthenticationResult`
- Removes from `useAuthentication` the effect that auto signed out users on `AuthErrors.RefreshAccessTokenError`
- Updates `DialogProps["onDismiss"]` to reference Reach's `DialogProps["onDismiss"]`
- Adds debug log to `NextAuth.events.signOut` in `src/pages/api/auth/[...nextauth].ts`

## 🤷 Why

The purpose of this dialog is to give a notice to authenticated users that their authentication state has changed. There are multiple analytics events added to the dialog so we can gather insights into things like how often the modal is seen and how users interact with it.

## 📸 Design Screenshots

[Figma frame](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=15915%3A212986&viewport=979%2C936%2C0.93&t=xkKbTMtzDfIyqmd0-11)

image

## 🧪 Testing

- [ ] Check out this branch locally
- [ ] Start the local server
- [ ] In a browser, open the app in 10 tabs
- [ ] In an editor, open the `src/pages/api/auth/[...nextauth].ts` file
- [ ] In `callbacks.jwt`, change the `if (isAccessTokenExpired) {` line to `if (true) {`
- [ ] Go back to your browser window
- [ ] Switch between a few tabs
- [ ] The dialog should render
- [ ] Click the "Sign in" button
- [ ] You should be taken to the `auth.hcp.dev` sign in page
- [ ] Complete the sign in flow
- [ ] Switch between a few tabs to render the dialog again
- [ ] Click the close button
- [ ] You should be signed out of the app
- [ ] Check your terminal where the server is running
- [ ] There should be one occurrence of the `"Inside of NextAuth.events.signOut"` log (rather than one for each tab)
- [ ] Sign in to the app again
- [ ] Switch between a few tabs to render the dialog one more time
- [ ] Click on the overlay behind the dialog
- [ ] You should be signed out of the app
- [ ] Check your terminal where the server is running
- [ ] There should be one occurrence of the `"Inside of NextAuth.events.signOut"` log (rather than one for each tab)

### DIFF
--- a/src/components/authentication-error-dialog/index.tsx
+++ b/src/components/authentication-error-dialog/index.tsx
@@ -1,0 +1,97 @@
+import { useCallback, useEffect, useRef } from 'react'
+import { IconSignIn16 } from '@hashicorp/flight-icons/svg-react/sign-in-16'
+import { IconX16 } from '@hashicorp/flight-icons/svg-react/x-16'
+import { AuthErrors } from 'types/auth'
+import { safeAnalyticsTrack } from 'lib/analytics'
+import useAuthentication from 'hooks/use-authentication'
+import Button from 'components/button'
+import Dialog from 'components/dialog'
+import Text from 'components/text'
+import { guessMethodOfDismissal } from './helpers'
+import s from './authentication-error-dialog.module.css'
+
+/**
+ * A dialog that renders for `AuthErrors.RefreshAccessTokenError` errors.
+ * Provides an option for users to sign in again. When dismissed, automatically
+ * signs the user out.
+ */
+const AuthenticationErrorDialog = () => {
+	const closeButtonRef = useRef<HTMLButtonElement>()
+	const { error, signIn, signOut } = useAuthentication()
+	const isRefreshAccessTokenError = error === AuthErrors.RefreshAccessTokenError
+	const label = 'Are you still there?'
+
+	useEffect(() => {
+		if (isRefreshAccessTokenError) {
+			safeAnalyticsTrack('AuthenticationErrorDialog rendered', {
+				error,
+			})
+		}
+	}, [error, isRefreshAccessTokenError])
+
+	/**
+	 * Set up callback for when the dialog is dismissed. Tracks an analytics event
+	 * and signs the user out.
+	 *
+	 * @NOTE The dismissEvent variable can be multiple types of events, but
+	 * Reach's DialogProps only specifies MouseEvent. Probably better to replace
+	 * Reach Dialog with our custom build when the HDS Modal component is stable,
+	 * rather than mess with the TypeScript.
+	 */
+	const handleDismiss = useCallback(
+		// TODO: this can be a click or keydown event
+		(dismissEvent: $TSFixMe) => {
+			const methodOfDismissal = guessMethodOfDismissal({
+				dismissEvent,
+				closeButtonRef,
+			})
+			safeAnalyticsTrack('AuthenticationErrorDialog dismissed', {
+				methodOfDismissal,
+			})
+			signOut()
+		},
+		[signOut]
+	)
+
+	/**
+	 * Set up the callback for the "Sign in" button's click. Tracks an analytics
+	 * event and starts the sign in UI flow.
+	 */
+	const handleButtonClick = useCallback(() => {
+		safeAnalyticsTrack('Click', {
+			category: 'AuthenticationErrorDialog sign in',
+		})
+		signIn()
+	}, [signIn])
+
+	return (
+		<Dialog
+			isOpen={isRefreshAccessTokenError}
+			label={label}
+			onDismiss={handleDismiss}
+			contentClassName={s.content}
+		>
+			<Text asElement="p" size={300} weight="semibold">
+				{label}
+			</Text>
+			<button
+				className={s.closeButton}
+				onClick={handleDismiss}
+				ref={closeButtonRef}
+			>
+				<IconX16 />
+			</button>
+			<Text asElement="p" className={s.body}>
+				You have been signed out due to inactivity.
+			</Text>
+			<Button
+				icon={<IconSignIn16 />}
+				iconPosition="leading"
+				text="Sign in"
+				onClick={handleButtonClick}
+			/>
+		</Dialog>
+	)
+}
+
+export default AuthenticationErrorDialog
--- a/src/components/authentication-error-dialog/index.tsx
+++ b/src/components/authentication-error-dialog/index.tsx
@@ -21,13 +21,16 @@ const AuthenticationErrorDialog = () => {
 	const isRefreshAccessTokenError = error === AuthErrors.RefreshAccessTokenError
 	const label = 'Are you still there?'
 
-	useEffect(() => {
-		if (isRefreshAccessTokenError) {
-			safeAnalyticsTrack('AuthenticationErrorDialog rendered', {
-				error,
-			})
-		}
-	}, [error, isRefreshAccessTokenError])
+	/**
+	 * @TODO uncomment when the events have been architected
+	 */
+	// useEffect(() => {
+	// 	if (isRefreshAccessTokenError) {
+	// 		safeAnalyticsTrack('AuthenticationErrorDialog rendered', {
+	// 			error,
+	// 		})
+	// 	}
+	// }, [error, isRefreshAccessTokenError])
 
 	/**
 	 * Set up callback for when the dialog is dismissed. Tracks an analytics event
@@ -41,13 +44,16 @@ const AuthenticationErrorDialog = () => {
 	const handleDismiss = useCallback(
 		// TODO: this can be a click or keydown event
 		(dismissEvent: $TSFixMe) => {
-			const methodOfDismissal = guessMethodOfDismissal({
-				dismissEvent,
-				closeButtonRef,
-			})
-			safeAnalyticsTrack('AuthenticationErrorDialog dismissed', {
-				methodOfDismissal,
-			})
+			/**
+			 * @TODO uncomment when the events have been architected
+			 */
+			// const methodOfDismissal = guessMethodOfDismissal({
+			// 	dismissEvent,
+			// 	closeButtonRef,
+			// })
+			// safeAnalyticsTrack('AuthenticationErrorDialog dismissed', {
+			// 	methodOfDismissal,
+			// })
 			signOut()
 		},
 		[signOut]
@@ -58,9 +64,12 @@ const AuthenticationErrorDialog = () => {
 	 * event and starts the sign in UI flow.
 	 */
 	const handleButtonClick = useCallback(() => {
-		safeAnalyticsTrack('Click', {
-			category: 'AuthenticationErrorDialog sign in',
-		})
+		/**
+		 * @TODO uncomment when the events have been architected
+		 */
+		// safeAnalyticsTrack('Click', {
+		// 	category: 'AuthenticationErrorDialog sign in',
+		// })
 		signIn()
 	}, [signIn])
 
--- a/src/hooks/use-authentication/index.ts
+++ b/src/hooks/use-authentication/index.ts
@@ -3,12 +3,7 @@ import { useRouter } from 'next/router'
 import { useSession } from 'next-auth/react'
 import { saveAndLoadAnalytics } from '@hashicorp/react-consent-manager'
 import { preferencesSavedAndLoaded } from '@hashicorp/react-consent-manager/util/cookies'
-import {
-	AuthErrors,
-	SessionData,
-	UserData,
-	ValidAuthProviderId,
-} from 'types/auth'
+import { SessionData, UserData, ValidAuthProviderId } from 'types/auth'
 import { UseAuthenticationOptions, UseAuthenticationResult } from './types'
 import { makeSignIn, makeSignOut, signUp } from './helpers'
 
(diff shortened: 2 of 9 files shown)


kellyschummer (approved):
👍 working as expected- Thank you @ashleemboyer !

src/hooks/use-authentication/index.ts
```diff
@@ -23,6 +30,9 @@ export const DEFAULT_PROVIDER_ID = ValidAuthProviderId.CloudIdp
 const useAuthentication = (
 	options: UseAuthenticationOptions = {}
 ): UseAuthenticationResult => {
+	const { flags } = useFlags()
```
BRKalow, line 33: [question] is there a reason we're opting to use HappyKit here vs. the flags from config? Thinking specifically because you mentioned in the video that this should never be on for production environments, in which case a flag from config would be preferable as it allows the build to remove any code behind a conditional.

src/contexts/debugging-state/index.tsx
```diff
@@ -0,0 +1,28 @@
+import { createContext, useContext, useState } from 'react'
+import { DebuggingStateProviderProps } from './types'
+
+const DebuggingStateContext = createContext<$TSFixMe>(undefined)
```
BRKalow, line 4: Do we want to type this? Or just leave it as `any`?

src/components/user-dropdown-disclosure/components/debugging-items.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * A special portion of the `UserDropdownDisclosure` that is intended to only
+ * render if the `showDebugUserMenuItems` feature flag is enabled. The purpose
+ * is to add custom items that can be used to trigger the app to change state.
+ *
+ * For example, an authentication error is tedious and challenging
+ * to organically reproduce. With this component, we can render an item that
+ * emulates an error so we can test how the app handles the error.
+ */
```
BRKalow, line 20: future 💭  these dev-only commands feel like something that would be well suited to integrating into the command dialog.

src/components/authentication-error-dialog/index.tsx
```diff
@@ -0,0 +1,30 @@
+import useAuthentication from 'hooks/use-authentication'
+import Button from 'components/button'
+import Dialog from 'components/dialog'
+import Text from 'components/text'
+import s from './authentication-error-dialog.module.css'
+import { IconSignIn16 } from '@hashicorp/flight-icons/svg-react/sign-in-16'
+
+const AuthenticationErrorDialog = () => {
```
BRKalow, line 8: [question/suggestion] It might be a good idea to add a segment event here so we can get an idea of how often the modal is triggering. Ideally I think the modal is a last resort that we want to see fire rarely, if ever.

I think @kendallstrautman mentioned this in one of our discussion threads as well!

src/hooks/use-authentication/index.ts
```diff
@@ -45,29 +55,23 @@ const useAuthentication = (
...
-
 	// Deriving booleans about auth state
 	const isLoading = status === 'loading'
 	const isAuthenticated = status === 'authenticated'
 	const showAuthenticatedUI = isAuthenticated
 	const showUnauthenticatedUI = !isLoading && !isAuthenticated
 	const preferencesLoaded = preferencesSavedAndLoaded()
 
+	/**
+	 * If the `showDebugUserMenuItems` feature flag is enabled, then we `hasError`
```
BRKalow, line 66:
```suggestion
	 * If the `showDebugUserMenuItems` feature flag is enabled, then `hasError`
```

BRKalow (commented):
The debugging utility is neat, it's nice to have ways to easily trigger hard to recreate bugs.

I'm a little nervous that we're introducing a non-dismissible dialog without clarity on how often it's going to be shown to users. Adding a tracking event and keeping an eye on that after the rollout will help us get a better grasp of the impact though.

src/contexts/debugging-state/index.tsx
ashleemboyer: Oh I forgot about this! I think we want to do `Record<string, unknown>`?

src/components/user-dropdown-disclosure/components/debugging-items.tsx
ashleemboyer: Ooooh I like that idea too! Reminds me of being able to emulate `prefers-reduced-motion` in Chrome dev tools with a command palette.

src/components/authentication-error-dialog/index.tsx
ashleemboyer: She totally did and I forgot about that, apologies! It's been a while since I've done that. I think I need to use `safeAnalyticsTrack` which is in `src/lib/analytics.ts`? 🤔

src/hooks/use-authentication/index.ts
ashleemboyer: Yep! I wanted the flag to be controllable during runtime rather than build time. I set up the production version of the flag to always be false whether it's turned on or off, but I think we can also put in some extra code that checks the current environment to make this more stable in the code.

image

src/components/authentication-error-dialog/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+				{label}
+			</Text>
+			<Text asElement="p" className={s.body}>
+				Your session has expired due to inactivity. Please sign in to continue.
+			</Text>
+			<Button
+				icon={<IconSignIn16 />}
+				iconPosition="leading"
+				text="Sign in"
+				onClick={() => signIn()}
```
kendallstrautman, line 24: 💭 I think we should provide an option if a user doesn't want to sign back in. Majority of our site doesn't require authentication to interact with it. Could we provide an 'x' to dismiss (and sign out the user, or just leave it) or should we have an explicit button, stating an option for not signing in?

src/hooks/use-authentication/index.ts
```diff
@@ -45,29 +55,23 @@ const useAuthentication = (
...
 
+	/**
+	 * If the `showDebugUserMenuItems` feature flag is enabled, then `hasError`
+	 * is determined by whether or not the app's debugging state has an error in
+	 * it. Otherwise, we check whether for an error in the session `data` object.
+	 */
+	const enableDebugging = !!flags?.showDebugUserMenuItems
+	const hasError = enableDebugging
+		? debuggingState && debuggingState[DEBUGGING_ERROR_PROPERTY_NAME]
+		: data && data.error
```
kendallstrautman, line 73: I think we should specifically scope this modal behavior to `AuthErrors.RefreshAccessTokenError` so we don't potentially swallow other auth errors.

src/components/authentication-error-dialog/index.tsx
ashleemboyer: Having two separate buttons was in the original proposal, but design decided to go with one to keep things simplified. I don't think we'd want the `X` to do anything automatically because the user won't know that's what happened. @kellyschummer do you have any additional context on not having a Sign out button present?

src/components/authentication-error-dialog/index.tsx
kendallstrautman: That sounds right! Pretty sure that will send to segment.

kendallstrautman: It would be nice to get more info on the error passed to segment as well. Instead of just 'hasError', perhaps we can store some information about the error to help with debugging if we find this popping up a lot.

src/hooks/use-authentication/index.ts
ashleemboyer: Is your concern about us not being alerted about them? If so, I think we should refine our monitoring and still show this modal for all auth errors since we don't have a good grasp of all the errors that have happened so far.

kendallstrautman: Or if the error is of the token refresh type, we can render the modal with the 'your session has expired' copy...whereas if we got a different error and we wanted to render the modal, we could use your messaging that is just the 'encountered an auth error' message. 

I have a little bit of hesitation telling the user this is a routine session expiry when there could be something more / a bigger issue going on with the way we handle token revalidation 😬 I didn't feel that way until I saw Mike's error reproduction that seemed very different from our previous hypothesis of this issue being only related to multiple tabs going stale / revalidating.

src/hooks/use-authentication/index.ts
kendallstrautman: Are we planning on extracting out the flag code after we feel this is stable?

src/hooks/use-authentication/index.ts
ashleemboyer: The error Mike was having falls in the same bucket as these. It looks like his `refresh_token` became out of date, which can happen multiple ways so it's hard to tell what exactly caused his error. I do feel confident that his error would have gone away if he had continued the sign in flow, but nothing told him to do that (which is where this modal comes in).

I think we should wait until we have more information before we change the messaging of the modal. The error that triggers all other errors has been the refresh token one, and we don't know the other ones yet. This modal will prevent the subsequent errors from happening at all, so our logs will be easier to follow and make an informed decision from.

src/hooks/use-authentication/index.ts
ashleemboyer: Yep!

src/contexts/debugging-state/index.tsx
ashleemboyer: Updated in https://github.com/hashicorp/dev-portal/pull/1338/commits/5c137c2782eafd3865c27689a26faddeccadfec7

src/components/authentication-error-dialog/index.tsx
ashleemboyer: Added an `error` prop to the `useAuthentication` result in https://github.com/hashicorp/dev-portal/pull/1338/commits/7a632cfeb83c768151f2280a35b7d1269d0160de; added `TODO`s for the analytics tracking in https://github.com/hashicorp/dev-portal/pull/1338/commits/fdcc27aa002f51496121ff78d7a0c0be32c6e674 (waiting to hear back from product if the values are acceptable)

src/components/authentication-error-dialog/index.tsx
ashleemboyer: 🗒️ PR has been updated and this convo is no longer relevant, so marking resolved.

src/components/user-dropdown-disclosure/components/debugging-items.tsx
ashleemboyer: 🗒️ PR has been updated and this convo is no longer relevant, so marking resolved.

src/hooks/use-authentication/index.ts
ashleemboyer: 🗒️ PR has been updated and this convo is no longer relevant, so marking resolved.

ashleemboyer:
@kellyschummer @BRKalow @kendallstrautman I've updated this PR with the latest plan we came up with as a group in today's Zoom call! It's ready for re-review. @kellyschummer I'll DM you in Slack to figure out a way to get this VQA'd. 😃

src/hooks/use-authentication/index.ts
```diff
@@ -45,28 +40,14 @@ const useAuthentication = (
...
-	}, [data?.error, signOut])
-
 	// Deriving booleans about auth state
 	const isLoading = status === 'loading'
 	const isAuthenticated = status === 'authenticated'
 	const showAuthenticatedUI = isAuthenticated
 	const showUnauthenticatedUI = !isLoading && !isAuthenticated
 	const preferencesLoaded = preferencesSavedAndLoaded()
+	const error = data?.error
+	const hasError = !!error
```
kendallstrautman, line 50: ⛏️ I don't believe this is used. 🔪 ?

src/components/authentication-error-dialog/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * A dialog that renders for `AuthErrors.RefreshAccessTokenError` errors.
+ * Provides an option for users to sign in again. When dismissed, automatically
+ * signs the user out.
+ */
+const AuthenticationErrorDialog = () => {
+	const closeButtonRef = useRef<HTMLButtonElement>()
+	const { error, signIn, signOut } = useAuthentication()
+	const isRefreshAccessTokenError = error === AuthErrors.RefreshAccessTokenError
```
kendallstrautman, line 21: 🤔 should we store this in state to make sure the analytics tracking isn't fired off if this is redefined on every re-render?

That reminds me we could check the segment debugger to see what is firing.

kendallstrautman (approved):
Followed the testing steps and everything works as described! I do think we should follow up to verify that the segment calls are tracking as expected / intended. If you haven't used the segment debugger yet, we can pair on it. 

But no blocking feedback here, just a few suggestions. Thanks for driving the charge on this 👍🏻

src/components/authentication-error-dialog/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+		},
+		[signOut]
+	)
+
+	/**
+	 * Set up the callback for the "Sign in" button's click. Tracks an analytics
+	 * event and starts the sign in UI flow.
+	 */
+	const handleButtonClick = useCallback(() => {
+		safeAnalyticsTrack('Click', {
```
alexcarpenter, line 61: 💭 this will create a new table in DWH with `Click` as the table name. Should we be more descriptive of what we are tracking here?

kellyschummer (approved):
🙌 💯 🥳

PR #1342, opened 2022-11-09
chore: typing for strip-undefined-props

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Modifies `stripUndefinedProperties` so that it better retains type information.

## 🤷 Why

To ensure we don't lose valuable type information when using `stripUndefinedProperties`.

## 🛠️ How

Takes the approach mentioned in [the task][task]:

```ts
export function stripUndefinedProperties<T = any>(obj: T): T {
  return JSON.parse(JSON.stringify(obj))
}
```

## 💭 Anything else?

- Adding a bunch of `$TSFixMe` was necessary to prevent type errors. Intent is to handle these, and add more accurate types, in future "someday" work.
    - Maybe a co-ordinated effort to root out `$TSFixMe` in the repo could be good, and could tie in with [`no-implicit-any` work](https://app.asana.com/0/1100423001970639/1202025665826439/f) or something

[task]: https://app.asana.com/0/1202097197789424/1202406907340320/f
[preview]: https://dev-portal-git-zssupport-type-for-strip-undefined-hashicorp.vercel.app/

### DIFF
--- a/src/lib/strip-undefined-props.ts
+++ b/src/lib/strip-undefined-props.ts
@@ -1,6 +1,11 @@
-// Hack needed to avoid JSON-Serialization validation error from Next.js https://github.com/zeit/next.js/discussions/11209
-// >>> Reason: `undefined` cannot be serialized as JSON. Please use `null` or omit this value all together.
+/**
+ * Hack needed to avoid JSON-Serialization validation error from Next.js.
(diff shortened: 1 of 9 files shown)


kendallstrautman (approved):
👍🏻

PR #1345, opened 2022-11-09
feat: adds tutorials redirects page support

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksadds-tutorials-redirects-support-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1199634971449915/1202982040929997) 🎟️

## 🗒️ What

This PR adds support for serving tutorial redirects from the Learn API. See related PRs for context on the whole workflow. 

## Related PRs
- [content sync](https://github.com/hashicorp/learn-api-content-sync/pull/65)
- [tutorials repo](https://github.com/hashicorp/tutorials/pull/950)


build-libs/tutorial-redirects.js
```diff
@@ -0,0 +1,24 @@
+async function getTutorialRedirects() {
+	const url = new URL(
+		'/pages/redirects',
+		'http://127.0.0.1:5000'
```
kendallstrautman, line 4: TODO - remove this and uncomment the base url env. Was using this for local testing with the whole setup.

build-libs/tutorial-redirects.js
```diff
@@ -0,0 +1,24 @@
...
+		//	process.env.NEXT_PUBLIC_LEARN_API_BASE_URL
```
kendallstrautman, line 5:
```suggestion
		process.env.NEXT_PUBLIC_LEARN_API_BASE_URL
```

next.config.js
```diff
@@ -90,12 +91,14 @@ module.exports = withSwingset({
...
 			const { simpleRedirects, globRedirects } = await redirectsConfig()
+			const tutorialRedirects = await getTutorialRedirects()
+
 			await fs.promises.writeFile(
 				path.join('src', 'data', '_redirects.generated.json'),
 				JSON.stringify(simpleRedirects, null, 2),
 				'utf-8'
 			)
-			return globRedirects
+			return [...globRedirects, ...tutorialRedirects]
```
BRKalow, line 101: can we shift the tutorial redirects fetching into `redirectsConfig()` in `build-libs/redirects.js`? The idea being we keep the redirect generation logic consolidated and consistent.

BRKalow: I'd like this file to stay unchanged. The redirects would get split yea, but it should all be handled correctly by the existing setup!

zchsh: ah oops, I missed that we have `splitRedirectsByType` to do the work for us 👍 👍 👍 I think I'm tracking now, like we'd add `const tutorialRedirects = await getTutorialRedirects()` in right after [proxiedSiteRedirects](https://github.com/hashicorp/dev-portal/blob/a15f17e311880b0362ccaee170bea5c7bd79b399/build-libs/redirects.js#L442), right?

BRKalow: Yeah, exactly! 😊

kendallstrautman: ah yes, I can do that!

BrandonRomano (approved):

next.config.js
kendallstrautman: @BRKalow since the `simple redirects` are written to the generated file to be served from middleware and have docs-specific logic handled there, I'm thinking of creating a `tutorialRedirects` return value similar to what @zchsh suggested yesterday. So I can just spread here alongside the glob.

Or is there something I'm missing about wanting to serve the simple redirects in middleware?

BRKalow: It's mostly just about being consistent with how we handle redirects. If there's no functional reason to have special handling for the tutorials redirects, it seems reasonable to handle them like all other redirects. The current logic should handle the tutorial redirects without issue. We initially did the simple/glob split to try and limit the number of redirects defined in our config.

kendallstrautman: The issue is that the current setup is very 'product' driven and learn urls don't necessarily abide by that so some refactoring would need to be done I think to make it work. I could look into it.

kendallstrautman: Unless maybe the learn redirects just go under '*', so maybe im mistaken!

BRKalow: @kendallstrautman I think it will work as-is. 💭  The product is determined by a host/cookie condition, but[ otherwise they get grouped under `*`](https://github.com/hashicorp/dev-portal/blob/main/build-libs/redirects.js#L420-L432), which is the [fallback value](https://github.com/hashicorp/dev-portal/blob/main/src/middleware.ts#L24) when running the middleware.

kendallstrautman: Updated it!

PR #1347, opened 2022-11-10
Update Packer API docs target file

sylviamoss:
## 🗒️ What

In this PR, I'm updating the API docs target file to pull from the desired API version. 

## 🤷 Why

The Packer team has beta features in the API preview version `2022-04-11` which we wish to list in our API docs.


BRKalow (approved):
thank you!

PR #1350, opened 2022-11-10
Add new Link component (wrapper for next/link)

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambambadd-link-component-hashicorp.vercel.app/) 🔎 
- [Asana task](https://app.asana.com/0/1202097197789424/1203346623842011/f) 🎟️

## 🗒️ What

- Adds new `Link` component
- Updates references to `next/link` component outside of:
  - `src/components/_proxied-dot-io`
  - `src/pages/_proxied-dot-io`
  - `src/views/_proxied-dot-io`

## 🤷 Why

The "why" of these changes are noted in block comments or can be read about in [DEVDOT-010: Custom Link component](https://docs.google.com/document/d/14embplHRyd2zO1A9a7Zws75xZeNZ6yG6WaXr57DnSuQ/edit).

## 🧪 Testing

Components to test:

- `HitWrapper`
- `BreadcrumbBar`
- `CardLink`
- `CommandBarLinkListItem`
- `DropdownDisclosureLinkItem`
- `InlineLink`
- `MobileUserDisclosure`
- `NavigationDisclosureLink`
- `NavigationHeader`'s:
  - `HomePageHeaderContent`
  - `PrimaryNavLink`
  - `ProductPageHeaderContent`
- `SidebarNavSubmenuItem`
- `StandaloneLink`
- `NotFound`
- `HomePageView`'s:
  - `HeroWithActions`
  - `ProductNav`
- `SidecarMarketingCard`
- `ProductTutorialsSitemap`
- `SignUpView`

## 💭 Anything else?

- A quick-follow to this PR for aligned prop naming: [Rename various "openInNewTab" props to "opensInNewTab"](https://app.asana.com/0/1202097197789424/1203388835781795/f)


ashleemboyer:
Oops! Meant to close other PRs

src/components/link/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+		ariaDescribedBy += opensInNewTabLabelId
+	}
+	if (restProps['aria-describedby']?.length > 0) {
+		ariaDescribedBy += ` ${restProps['aria-describedby']}`
+	}
+
+	return (
+		<>
+			<NextLink href={href} {...nextLinkProps}>
+				<a {...restProps} aria-describedby={ariaDescribedBy} target={target}>
```
BRKalow, line 52: [suggestion] I don't think the `nextLinkProps` separation is necessary here. For a few reasons:

1. It's not currently in use anywhere
2. With next 13, [`NextLink` renders an anchor tag automatically](https://nextjs.org/docs/api-reference/next/link), so the spread would then look like `<NextLink {...restProps} {...nextLinkProps}`
3. Based on the types, we're "extending" the native anchor element and allowing some props to be passed through to `NextLink`. This sort of makes sense when the two were separate. However, with the new behavior I think it will feel more intuitive to use if we treat is as just an extension of `NextLink`.

Extending `NextLink` should make for cleaner types as well, instead of mixing the native anchor props and the `NextLink` props. What do you think?

src/components/link/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+const ARIA_DESCRIBED_BY_PREFIX = 'opens-in-new-tab-label'
+
+/**
+ * A wrapper for the next/link component that also handles unique functionality
+ * with the `opensInNewTab` prop.
+ *
+ * When `opensInNewTab` is passed, a `<span>` with the `g-screen-reader-only`
+ * class is rendered with "(opens in new tab)" as its inner text. This approach
+ * is inspired by WCAG 2.1 Ttechnique G201: Giving users advanced warning when
```
BRKalow, line 14:
```suggestion
 * is inspired by WCAG 2.1 Technique G201: Giving users advanced warning when
```

BRKalow (changes requested):
Nice, this feels like a good abstraction and will definitely make the diff for the Next 13 upgrade much smaller!

I left some feedback on the `nextLinkProps` separation. I think it will make for a more intuitive interface if we do away with it, especially so with the upcoming change in 13.

src/components/link/index.tsx
ashleemboyer: Sweet! I didn't know we would be upgrading this soon, but went ahead and removed `nextLinkProps` for now in https://github.com/hashicorp/dev-portal/pull/1350/commits/454db9ed484dddf51159009e5f2cc26b30c4c8e5. 😊

src/components/link/index.tsx
ashleemboyer: Fixed in https://github.com/hashicorp/dev-portal/pull/1350/commits/c7816ce34214bd25d5c99e63d5227d78722e4042 ✅

ashleemboyer:
@BRKalow ready for re-review! 🥳

BRKalow (approved):
nice work! 🏆

ashleemboyer:
Seems like GitHub may be having some issues in the E2E tests, so going ahead and merging as mentioned in Slack 🚀

PR #1362, opened 2022-11-15
Ks.debugging auth

kendallstrautman:
## 🔗 Relevant links

- [Asana task](url) 🎟️

### DIFF
--- a/src/hooks/use-authentication/helpers/make-sign-in.ts
+++ b/src/hooks/use-authentication/helpers/make-sign-in.ts
@@ -1,11 +1,6 @@
-import { NextRouter } from 'next/router'
 import { signIn, SignInOptions as _SignInOptions } from 'next-auth/react'
 import { ValidAuthProviderId } from 'types/auth'
 import { DEFAULT_PROVIDER_ID } from '..'
-
-interface MakeSignInOptions {
-	routerPath: NextRouter['asPath']
-}
 
 type SignInOptions = Pick<_SignInOptions, 'callbackUrl' | 'redirect'>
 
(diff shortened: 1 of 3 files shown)


src/hooks/use-authentication/helpers/make-sign-in.ts
```diff
@@ -15,14 +10,14 @@ type SignInOptions = Pick<_SignInOptions, 'callbackUrl' | 'redirect'>
...
+const makeSignIn = () => {
 	return (
 		provider: ValidAuthProviderId = DEFAULT_PROVIDER_ID,
 		options: SignInOptions = {}
 	) => {
-		const { callbackUrl = routerPath, redirect = true } = options
+		const { redirect = true } = options
 
-		return signIn(provider, { callbackUrl, redirect })
+		return signIn(provider, { redirect })
```
kendallstrautman, line 20: The default behavior is to redirect to the page that initiated sign in, I don't think this will fix the issue but a small thing to adjust to simplify

src/hooks/use-authentication/helpers/make-sign-out.ts
```diff
@@ -19,6 +19,7 @@ const makeSignOut = ({ routerPath }: MakeSignOutOptions) => {
 			callbackUrl = routerPath === '/profile/bookmarks' ? '/' : routerPath,
 			redirect = true,
 		} = options
+		console.log({ callbackUrl }, { routerPath })
```
kendallstrautman, line 22: Will remove these logs, was just trying to repro the error. to no avail.

kendallstrautman:
We need to clean up the types to get the latest version of next auth working, I think. I'm setting this down for now but noting as a breadcrumb.

PR #1367, opened 2022-11-17
feat: implement TryHcpCallout

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Implements `TryHcpCallout` components.

## 🤷 Why

To build components that will be places as "Try HCP" CTAs on specific areas of Dev Dot.

## 🛠️ How

- Built out small pieces that make up the "variants"
- Composed small pieces into the two "variants" (main, and compact)
- Built wrapper component for authorable contexts, that takes in `productSlug` and handles the rest via baked-in content

## 📸 Design Screenshots

🎨 [Figma design spec][figma]

> **Note**: See [Swingset][swingset] or [Storybook][storybook] to explore all variants and responsive behaviour.

## Design

design-spec

## Implementation

CleanShot 2022-11-17 at 16 26 44@2x

CleanShot 2022-11-17 at 16 26 29@2x

## 🧪 Testing

- [ ] Visit the [swingset page for TryHcpCallout][swingset]
    - The component should look and function as expected, across all viewport sizes
    - Viewport testing may be easier on [Storybook][storybook]

## 💭 Anything else?

The SVG assets here presented some unique challenges. Feels worth sharing some takeaways.

<details>

<summary>Takeaways from SVG processing work in this PR</summary>

### SVG only supports `Radial` and `Linear` gradients

The only gradient types that seem supported in `SVG` are `Radial` and `Linear`. 

Figma provides many more options though. If source artwork in Figma contains other gradient types, such as `Diamond`, these will _not_ be exported correctly. The differences can be subtle and hard to catch - Figma doesn't warn, and kind of fudges the result, approximating all gradients to `Radial` or `Linear` in a very rough way.

This is relatively easy to detect if you know what to do - export the asset, then grab the exported `.svg`, pull it back into Figma, and do a side-by-side comparison. This way it's much easier to see the differences.

To remedy this issue, there are broadly two options:
- Export the asset in `.png` instead. The file will be a lot larger, but will have a better guarantee of rendering as expected. This is a tradeoff though, especially for gradient backgrounds, as a relatively small SVG can balloon into a giant, still-kinda-pixellated PNG.
- Rebuild the artwork with `Radial` and `Linear` gradients only, removing and replacing all other gradient types. You'll likely need to work with a designer to do this. But the payoff may be worth it. In a recent example, a gradient background as a PNG would have been ~ 200 kB, and the save background as an SVG was only 2 kB.

CleanShot 2022-11-17 at 17 06 35@2x

### Figma does not flatten shapes correctly during export

SVGs do not support all the same types of transforms as Figma. Figma seems to "bake" transforms into exports in a subtly different way than it does when flattening shapes directly in Figma.

This comes up a lot in art that uses our 3D isometric grid style. Typically there might be a rectangle that's rotated, and then flattened by a certain amount. If 

To remedy this issue, flatten all shapes before export. Un-flattened shapes can be hard to detect - but one tell is they have a skewed bounding box. Select any un-flattened shapes, and using the contextual menu, select `Flatten`. You can also select text and then use the shortcut `⌘ + E`.

CleanShot 2022-11-17 at 16 58 21@2x

### Text must be outlined before export

Remember to outline text in assets before export - Figma does not automatically do this.

If text is not flattened, there is zero guarantee of correct rendering. Typically this will look like subtle shifts in positioning on your local machine. But can be a completely different typeface when viewed in browser, or on random web visitors' computers.

To remedy this, outline text in Figma before export. This requires edit access - which can be approximated by duplicating an asset you need to export to personal drafts. Then, select the text elements that need to be outlines, and from the right-click contextual menu choose `Outline Text`. You can also select text and then use the shortcut `Shift + ⌘ + O`.

</summary>

[task]: https://app.asana.com/0/1202114367927919/1203186425392266/f
[figma]: https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=15357%3A196108&t=TbSHtGHt6JSdOxr1-4
[preview]: https://dev-portal-git-zstry-hcp-components-hashicorp.vercel.app/
[swingset]: https://dev-portal-git-zstry-hcp-components-hashicorp.vercel.app/swingset
[storybook]: https://dev-portal-with-storybook.vercel.app/?path=/story/components-tryhcpcallout--default


package.json
```diff
@@ -104,6 +104,7 @@
 		"@hashicorp/platform-markdown-utils": "^0.2.2",
 		"@hashicorp/platform-packer-plugins": "^0.3.0",
 		"@hashicorp/platform-product-meta": "^0.1.1",
+		"@hashicorp/platform-util": "^0.2.0",
```
zchsh, line 107: I'm using `mitigateWidows` to try to get the `description` text wrapping nicely, so relying directly on `@hashicorp/platform-util`.

src/components/try-hcp-callout/components/description/index.tsx
```diff
@@ -0,0 +1,14 @@
+import mitigateWidows from '@hashicorp/platform-util/text/mitigate-widows'
+import s from './description.module.css'
+
+/**
+ * Description component, for use in HCP callout cards.
+ */
+export function Description({ description }: { description: string }) {
```
zchsh, line 7: There are a couple of "variants" of this component already. With this in mind, I've broken things down into pretty small pieces, so that the small pieces can be composed to create "variants".

src/components/try-hcp-callout/components/standalone-link-contents/standalone-link-contents.module.css
```diff
@@ -0,0 +1,31 @@
...
+	gap: 6px;
+	line-height: var(--line-height);
+	margin-bottom: 0;
+	margin-top: 24px;
+
+	& svg {
+		flex-shrink: 0;
+		transition: transform 0.2s;
+
+		@nest :global(a):hover & {
```
zchsh, line 15: This is a bit of an interesting one - found this was a neat and probably better alternative to having this component receive an `isHovered` prop or something. Open to other ideas here, of course 👍

src/components/try-hcp-callout/components/try-hcp-callout-prebuilt/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { tryHcpCalloutContent } from './content'
+
+/**
+ * A wrapper around TryHcpCallout.
+ *
+ * Handles content for a set of supported `productSlug` value, which allows
+ * this component to provide a single-prop `productSlug` props API.
+ * Intended for use in authorable contexts.
+ */
+export function TryHcpCalloutPrebuilt({
```
zchsh, line 12: This is component that'll be dropped into authorable contexts - it accepts a `productSlug` and that's it.

src/components/try-hcp-callout/components/standalone-link-contents/standalone-link-contents.module.css
alexcarpenter: Interesting idea, we ran into a similar challenge with our [ExpandableArrow](https://react-components.vercel.app/components/expandablearrow) component that needs to be triggered by a parent component. Couldn't guarantee it'd always be within an `a` tag though. For example our link in the card does not wrap the card content similar to how your CardLink works. Ended up adding a prop to trigger the state change.

alexcarpenter: Now that I mentioned it, this might not be working correctly, since the link tag is not wrapping the content?

zchsh: Ah yeah great catch 💖  this needs to be `[class*="card-link"]`, I had that before but then thought I could clean it up 😅 Updating now! 

Great point re: guarantee for the hover behaviour. This is definitely a pattern we should probably explore further. In the context of this PR, and since this is a lil piece scoped for use in this component only, personally I think the relative brittleness might be worth the trade-off.

alexcarpenter: Agreed!

zchsh: Also happy to switch to the `isHovered` hook approach - I'm totes happy either way, very open if anyone has strong or even lukewarm or weak preference there 🙇

src/components/try-hcp-callout/docs.mdx
```diff
@@ -0,0 +1,41 @@
...
+	ctaText="Try Cloud for free"
+	ctaUrl="https://app.terraform.io/public/signup/account"
+/>
+`}</LiveComponent>
+
+### `<TryHcpCalloutPrebuilt />`
+
+This component wraps `<TryHcpCallout />`. It handles content for a set of supported `productSlug` value, which allows this component to provide a single-prop `productSlug` props API. Intended for use in authorable contexts.
+
+<LiveComponent>{`<TryHcpCalloutPrebuilt productSlug="terraform"/>`}</LiveComponent>
```
ashleemboyer, line 41: Can this be updated to go ahead and show all of the products?

```suggestion
<LiveComponent>{`<>
  <TryHcpCalloutPrebuilt productSlug="boundary"/><br/>
  <TryHcpCalloutPrebuilt productSlug="consul"/><br/>
  <TryHcpCalloutPrebuilt productSlug="hcp"/><br/>
  <TryHcpCalloutPrebuilt productSlug="packer"/><br/>
  <TryHcpCalloutPrebuilt productSlug="terraform"/><br/>
  <TryHcpCalloutPrebuilt productSlug="vault"/><br/>
  <TryHcpCalloutPrebuilt productSlug="waypoint"/>
</>`}</LiveComponent>
```

src/components/try-hcp-callout/components/description/index.tsx
kendallstrautman: Could these building blocks be eventually shared between all callouts? Seeing that we have a few similar callout variants already. That may already be what you are describing. Just future thinking 💭

kendallstrautman: Here are the callouts I see in the repo currently: 
- interactive lab
- callout card
- mdx hcp callout 
- branded callout (on the tutorials product pages)

Might be cool to create a shared abstraction if the designs are super similar (or in come cases the same minus the background assets). Not necessary for this PR though.

src/components/try-hcp-callout/components/try-hcp-callout-prebuilt/content.ts
```diff
@@ -0,0 +1,55 @@
+import { TryHcpCalloutProps } from 'components/try-hcp-callout/types'
+import { ProductSlugWithContent } from '../../types'
+
+type HcpCalloutContent = Pick<
+	TryHcpCalloutProps,
+	'heading' | 'description' | 'ctaText' | 'ctaUrl'
+>
+
+export const tryHcpCalloutContent: Record<
```
kendallstrautman, line 9: Nice, it's nice to see all the data here!

src/components/try-hcp-callout/components/try-hcp-callout-prebuilt/index.tsx
kendallstrautman: Love it. We should make sure to add this to the authoring documentation for available components. We're adding this to product landing pages and in MDX content?

src/components/try-hcp-callout/docs.mdx
```diff
@@ -0,0 +1,41 @@
...
+	heading="Terraform Cloud"
+	description="Automate infrastructure provisioning at any scale for free."
+	ctaText="Try Cloud for free"
+	ctaUrl="https://app.terraform.io/public/signup/account"
+/>
+`}</LiveComponent>
+
+### `<TryHcpCalloutCompact />`
+
+This composition includes logo and text elements only. Intended for use in width-constrained areas, such as the sidecar area of our sidebar-sidecar layout.
```
kendallstrautman, line 25: Do we have guidance from design yet on which pages / places we are adding this? Is that planned for a follow up PR?

kendallstrautman:
For some reason, I don't see the component page in swingset 🤷🏻‍♀️  I do see[ 'hcpcallout' in there](https://dev-portal-git-zstry-hcp-components-hashicorp.vercel.app/swingset/components/hcpcallout) 😆 maybe we should adjust the naming on that one to be more specific since how we have two hcp callouts. But that might be annoying to rollout given that's available in mdx content (not widely used yet though). 

Anyways, storybook though 💯

kendallstrautman:
It looks like there should be a bit more height to accommodate the asset. See the bottom line that is super close to the edge of the container. 

Storybook
Screen Shot 2022-11-17 at 4 47 51 PM

Figma
Screen Shot 2022-11-17 at 4 47 23 PM

src/components/try-hcp-callout/components/product-icon-heading/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+export function ProductIconHeading({
+	productSlug,
+	headingText,
+}: ProductIconHeadingProps) {
+	if (productSlug === 'hcp') {
+		return (
+			<div className={s.rootHcpLogo}>
+				<InlineSvg src={hcpLogo} />
+				<span className="g-screen-reader-only">{headingText}</span>
+			</div>
```
ashleemboyer, line 27: 🚧 The screen reader experience for this variant is very close! It just needs one small adjustment.

Both the `InlineSvg`'s underlying `div` and the visually hidden `span` can be navigated to with the arrow keys with VoiceOver. This means something that is visually presented as one element is non-visually presented as two elements, and that's a mismatched experience. Here's a demonstration (no audio, need to download CleanShot again for better recordings!!):

https://user-images.githubusercontent.com/43934258/202585526-7687c444-efae-48fb-b2fe-85d14164a4e3.mov

To fix this mismatched experience and always present this as one element, I believe we need the code to look like this:

```suggestion
			<>
				<span id="auto-generated-id" className="g-screen-reader-only">
					{headingText}
				</span>
				<div aria-labelledby="auto-generated-id" role="presentation">
					<InlineSvg aria-hidden className={s.rootHcpLogo} src={hcpLogo} />
				</div>
			</>
```

List of the changes:

- Replaces wrapper `div` with a fragment (`InlineSvg` renders a `<div>`)
- Moves `s.rootHcpLogo` class to `InlineSvg` (gets applied to it's rendered outer `div`)
- Moves visually hidden span to be first (not sure if this is needed, but it feels right to place it before the element that references it)
- Adds an `id` to the visually hidden span (We should auto generate this with `useId` from `@react-aria/utils` because multiple HCP variants of this component can `technically` be rendered on the same page)
- Sets `aria-hidden` on the `InlineSvg`, and wrap it with a `div` with `aria-labelledby` that references the visually hidden span and `role="presentation"`. Explanations:
  - When `aria-labelledby` is set on `InlineSvg`, it's still presented as two elements because the rendered `<div>` is the labelled element, and it has an `svg` with a `title` inside of it
  - Without `role="presentation"`, it's still presented as two elements and the first one is presented as "HashiCorp Cloud Platform, empty group".

Here is a demonstration of how these changes sound with VoiceOver in Chrome:

https://user-images.githubusercontent.com/43934258/202598199-f9d10fc6-fb7d-4e80-a796-758226089166.mov

I realize this a long comment for a small code change... 😆 So I'm happy to sync on this via Slack / quick Zoom if that'd be helpful! 😊

ashleemboyer (changes requested):
I'm still reviewing, but wanted to comment on this part before finishing in case you get to it before I can finish my review! 🙏🏻 💟

src/components/try-hcp-callout/components/try-hcp-callout-prebuilt/content.ts
```diff
@@ -0,0 +1,55 @@
...
+		description: 'Simplify your application deployments across platforms',
+		ctaText: 'Try HCP Waypoint for free',
+		ctaUrl: 'https://portal.cloud.hashicorp.com/sign-up',
+	},
+	hcp: {
+		heading: 'HashiCorp Cloud Platform',
+		description: 'The fastest way to get up and running with HashiCorp tools',
+		ctaText: 'Try cloud for free',
+		ctaUrl: 'https://portal.cloud.hashicorp.com/sign-up',
+	},
```
ashleemboyer, line 54: Can we double-check with design and make sure that these slightly different `ctaText`s are really what we want? It's interesting to me that some include "for free" and others don't, and some some say "Try HCP <Product Name>" but some others don't (aside from HCP and Terraform's -- those differing makes sense to me!) The compact version of some of these don't always match their non-compact version either, so just want to double check! ❤️

src/components/try-hcp-callout/components/description/index.tsx
zchsh: I dig it - added a task for the future: 🎟 [Assess abstracting atomic "callout" components](https://app.asana.com/0/1202097197789424/1203396252179033/f)

src/components/try-hcp-callout/components/try-hcp-callout-prebuilt/index.tsx
zchsh: ☝️ I think we won't be adding this to MDX, at least not yet! AFAIK the placements are only:
- On the `/<product>` landing pages (authored in `.json`)
- In the sidecar of install views (authored in JSX)
I can definitely add some docs for the `.json` authoring context, great call 💯 I'll do that in the next PR where I implement the authoring, if that sounds good!

src/components/try-hcp-callout/components/try-hcp-callout-compact/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+}: TryHcpCalloutProps) {
+	return (
+		<CardLink className={s.root} ariaLabel={ctaText} href={ctaUrl}>
+			<div className={s.background} />
+			<ProductIconHeading productSlug={productSlug} headingText={heading} />
+			<Description description={description} />
+			<StandaloneLinkContents text={ctaText} />
+		</CardLink>
+	)
+}
```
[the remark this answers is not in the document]
zchsh: ☝️ I'm on the fence!

IMO adding the `normal | compact` prop has the benefit of drying up the code a bit... But I think it also has the disadvantage of setting up a path of conditional rendering (eg for the "image" part). I figured the duplication of things like `.backgroundBase` is worth the tradeoff - especially since `.background` classes, while similar, use different `background-image` assets and `background-position`. And I feel like `CardLink` does a great job of handling the semantic concerns around `ariaLabel`, and the smaller "atomic" elements like `ProductIconHeading` also do a bit of that handling as well. 

So personally I lean slightly towards keeping the components separate - especially since we have another "variant" in use on the home page (not rolled in this PR, but maybe something we'd want to approach later as Kendall mentioned).

Admittedly it does feel a lil awkward in some ways (eg the `StandaloneLinkContents` interface & `:hover` work in the thread above)... hopefully that'll improve a bit as the approach matures (eg with subsequent work in 🎟 [Assess abstracting atomic "callout" components](https://app.asana.com/0/1202097197789424/1203396252179033/f), see also [ExpandableArrow](https://react-components.vercel.app/components/expandablearrow) as Alex linked).

All that being said, definitely open to refactoring here if there are strong opinions on taking the `normal|compact` & conditional rendering approach 🙇

src/components/try-hcp-callout/components/standalone-link-contents/standalone-link-contents.module.css
```diff
@@ -0,0 +1,31 @@
...
+	margin-bottom: 0;
+	margin-top: 24px;
+
+	& svg {
+		flex-shrink: 0;
+		transition: transform 0.2s;
+
+		@nest :global([class*='card-link']):hover & {
+			transform: translate(2px);
+		}
```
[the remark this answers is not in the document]
zchsh: Sweet, I'll definitely wrap in the `@media` query!

Re: passing in a class string via props, like "standaloneLink", that seems like a neat approach - but how do we get the provided `className` string into CSS? 🤔 As well, I think we'd still need to the `:global` selector? Admittedly I haven't tried very hard, but it seems harder to do this handoff than it might be for things like CSS custom properties.

The `[class*="card-link"]` selector is definitely a bit brittle - but I figure since `StandaloneLinkContents` isn't abstracted yet, and is just for use in these specific components, that brittleness might be worth the tradeoff and fine for this initial implementation. That being said, open to diving deeper if it feels like the right time.

For near-future, what do you think about [the approach Alex mentioned](https://github.com/hashicorp/dev-portal/pull/1367#discussion_r1025773087), via JS & via a prop (eg maybe `isHovered`)? Intent would be to maybe look at that as part of 🎟 [Assess abstracting atomic "callout" components](https://app.asana.com/0/1202097197789424/1203396252179033/f).

src/components/try-hcp-callout/components/try-hcp-callout/try-hcp-callout.module.css
```diff
@@ -0,0 +1,98 @@
+.root {
+	--padding: 24px;
+
+	background: none;
+	display: flex;
+	isolation: isolate;
+	overflow: hidden;
```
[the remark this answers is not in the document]
zchsh: Ah great catch - I'll take another look at this, thanks for diving in here with the refinements 🙇 💖

src/components/try-hcp-callout/docs.mdx
zchsh: Yep, planning a follow-up PR! As far as I know, only placements are on the `/<product>` landing in main content area ("main" version of the callout), & `/<product>/downloads` views in the sidecar ("compact" version of the callout).

src/components/try-hcp-callout/components/product-icon-heading/index.tsx
zchsh: 💖 Thanks for diving in here, I'll work through this list of changes! 💯

src/components/try-hcp-callout/components/try-hcp-callout/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+ */
+export function TryHcpCallout({
+	productSlug,
+	heading,
+	description,
+	ctaText,
+	ctaUrl,
+}: TryHcpCalloutProps) {
+	return (
+		<CardLink className={s.root} ariaLabel={ctaText} href={ctaUrl}>
```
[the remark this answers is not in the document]
zchsh: TBH I wasn't sure exactly what to set it to - but having some very long text (like combining some text representation of the heading, description, and ctaText) didn't seem right, it felt like it would be repetitive when moving through the various elements (the `<a />` tag itself, and then the other text elements).

Also just peeked at [the existing HCP callout](https://github.com/hashicorp/dev-portal/blob/zs.try-hcp-components/src/views/homepage/components/merchandising-slots/slots/logo-cta-slot/index.tsx#L19) and other `CardLink` use cases to confirm that take.

I should definitely tested with VoiceOver though, then the choice would have been a bit more informed 😅

src/components/try-hcp-callout/components/try-hcp-callout-compact/try-hcp-callout-compact.module.css
```diff
@@ -0,0 +1,27 @@
+.root {
+	--padding: 21px;
+
+	background: none;
+	display: flex;
+	flex-direction: column;
+	isolation: isolate;
+	overflow: hidden;
```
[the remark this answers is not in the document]
zchsh: I suspect so! I'll take a look, and see if I can apply a similar fix to the one you came up with 👍

src/components/try-hcp-callout/components/try-hcp-callout-prebuilt/content.ts
zchsh: Yep, I can double-check! I am going off what's in Figma, which is _in theory_ an approved source of truth, but 🤷 . I'll run the other updates in this PR, make sure Swingset is rendering the latest, & post in `#proj-dev-portal` 👍

zchsh:
@kendallstrautman re: [asset height](https://github.com/hashicorp/dev-portal/pull/1367#issuecomment-1319406741), great catch 🦅 👁️ 💯 

For the UI mock positioning, I'd messed up `--image-top-offset`. For the background, I'd had the `size` as `cover` to try to better account for responsive behaviour... but yeah didn't match up right 🤔 I've tweaked it a bit to try to better match design:

CleanShot 2022-11-18 at 11 05 05@2x

src/components/try-hcp-callout/components/product-icon-heading/index.tsx
zchsh: ✅ okey dokes, this should be fixed up, I hope! Thanks for the direct suggestion here - I applied that and then brought in `useId` 👍 I think that covers the list of changes!

```ts
const uniqueId = useId()

if (productSlug === 'hcp') {
	return (
		<>
			<span id={uniqueId} className="g-screen-reader-only">
				{headingText}
			</span>
			<div aria-labelledby={uniqueId} role="presentation">
				<InlineSvg aria-hidden className={s.rootHcpLogo} src={hcpLogo} />
			</div>
		</>
	)
}
```

src/components/try-hcp-callout/types.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Only specific product slugs, being those with HCP offerings,
+ * can be used with this component.
+ */
+export type ProductSlugWithContent = Exclude<
+	ProductSlug,
+	'nomad' | 'sentinel' | 'vagrant'
+>
+
+export interface TryHcpCalloutProps {
+	productSlug: Exclude<ProductSlug, 'nomad' | 'sentinel' | 'vagrant'>
```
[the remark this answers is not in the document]
zchsh: ✅ applied!

src/components/try-hcp-callout/components/standalone-link-contents/standalone-link-contents.module.css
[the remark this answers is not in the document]
zchsh: ✅ updated with the `@media` query!

Let me know what you think about deferring the `[class*="card-link"]` refactor for now 🙇

src/components/try-hcp-callout/components/standalone-link-contents/standalone-link-contents.module.css
```diff
@@ -0,0 +1,31 @@
...
+	composes: hds-typography-body-100 from global;
+
+	--line-height: 1rem;
+}
+
+.medium {
+	composes: hds-typography-body-200 from global;
+
+	--line-height: 1rem;
+}
```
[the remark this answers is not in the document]
zchsh: In any case - ✅ updated! Thanks for the suggestion 💖

src/components/try-hcp-callout/components/try-hcp-callout/try-hcp-callout.module.css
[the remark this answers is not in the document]
zchsh: ✅ should be fixed now!
- Added `--border-radius` on the `.root` of both elements, set it too for clarity. Then can share with contained elements:
- Adds matching border-radius to `backgroundBase`, and to `imageContainer` for the default composition as well 👍

src/components/try-hcp-callout/components/try-hcp-callout-compact/try-hcp-callout-compact.module.css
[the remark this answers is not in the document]
zchsh: ✅ sharing `border-radius` and moving the `overflow-hidden` works here too, I think!

src/components/try-hcp-callout/components/standalone-link-contents/standalone-link-contents.module.css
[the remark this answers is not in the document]
ashleemboyer: Ah! I didn't even mention where to put this code. 🤦🏻

```css
/* Only enable animation if query is supported and value is no-preference */
  /* `.standaloneLink` would be the class passed to `StandaloneLinkContents` */
  &:hover .standaloneLink svg {
    transform: translate(2px);
  }
}
```

This ☝🏻 would live in `src/components/try-hcp-callout/components/try-hcp-callout/try-hcp-callout.module.css`, and be defined in the `.root` class at the top. Then in the `TryHcpCallout` component, the `s.standaloneLink` would be passed as `className` to the `StandaloneLinkContents` component it renders. No global selector or brittle class selector needed! 😊

Here's a diff of what the changes would be for just `TryHcpCallout`. The compact variant would need the same changes as `TryHcpCallout` is getting. This is working locally for me:

https://github.com/hashicorp/dev-portal/compare/zs.try-hcp-components...amb-zs.try-hcp-components

---

> For near-future, what do you think about https://github.com/hashicorp/dev-portal/pull/1367#discussion_r1025773087, via JS & via a prop (eg maybe isHovered)? Intent would be to maybe look at that as part of 🎟 [Assess abstracting atomic "callout" components](https://app.asana.com/0/1202097197789424/1203396252179033/f).

I'm not sure we should try to change styles via JS except as a last resort, especially since this can be handled with CSS. It keeps the CSS and JS concerns separate to their main uses: CSS for style & presentation and JS for adding functionality.

src/components/try-hcp-callout/components/try-hcp-callout/index.tsx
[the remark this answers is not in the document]
ashleemboyer: You're totally all good here!! I think this is the way to go. Concatenating different pieces of text would probably be too verbose.

And in terms of comparing the visual experience and the voice control experience -- my _guess_ is that a voice control user might try to say the text of the CTA to click the card. So setting `ariaLabel` to that exact text prevents us from breaking the guessed voice control experience! (Although we definitely data to confirm the guess)

image

Nicely done! 🙌🏻

src/components/try-hcp-callout/components/try-hcp-callout-prebuilt/index.tsx
kendallstrautman: Oh okay cool! That makes it easier then if its not in MDX. Doing in the next PR sounds great 😄

src/components/try-hcp-callout/components/try-hcp-callout-compact/index.tsx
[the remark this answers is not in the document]
kendallstrautman: I lean towards having a small bit of duplication as opposed to conditional rendering in this case as well. I like the composed variants and makes the code more bite-sized / easier to iterate on IMO.

src/components/try-hcp-callout/components/standalone-link-contents/standalone-link-contents.module.css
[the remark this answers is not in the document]
ashleemboyer: Yes with how the code is currently written, the `translate(2px)` would have to be defined in 2 places:

- `TryHcpCallout`'s `.root` class
- `TryHcpCalloutCompact` `.root` class

As we've chatted about in another thread, these components already have a lot of duplicate code. If defining the CSS twice is now a concern, then we should make `TryHcpCallout` define the main CSS classes, update `TryHcpCalloutCompact` to import that style file and it's own, and apply classes from both files to its markup.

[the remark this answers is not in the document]
zchsh: I think the question of when & how much we want to duplicate code, vs abstract code here, is a great one 💯

I think there are benefits and drawbacks whichever way we approach this. Personally I'd love to defer further iteration here to a subsequent PR, especially since we have an eye on re-using components like `StandaloneLinkContent` in other components.

Would deferring further iteration here be acceptable?

kendallstrautman (approved):
LGTM! Very clean implementation 💯 📯 🤙🏻

src/components/try-hcp-callout/components/try-hcp-callout-compact/index.tsx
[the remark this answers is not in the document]
ashleemboyer: Oh, I don't think we'd need to use conditional rendering! I see the prop being used to apply certain classes, and then the CSS would handle the style differences between the two types.

On the note of semantics and markup order what I meant by divergence is that because these are two separate components, they can easily diverge in what they pass to `CardLink`'s `ariaLabel`, the order in which they render the `children` passed to `CardLink`, and the ability for their CSS properties to change in the nearly identical classes. If these are things we want to remain consistent, then this feels like too much flexibility and too much to manually maintain.

src/components/try-hcp-callout/components/try-hcp-callout/try-hcp-callout.module.css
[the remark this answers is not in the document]
ashleemboyer: Sweeeeet just tested the focus styles and they're working as expected again! 👏🏻

src/components/try-hcp-callout/components/try-hcp-callout-compact/try-hcp-callout-compact.module.css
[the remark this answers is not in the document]
ashleemboyer: As confirmed in another comment, it's working! 🤩

src/components/try-hcp-callout/components/standalone-link-contents/standalone-link-contents.module.css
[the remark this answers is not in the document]
ashleemboyer: Sure 👍🏻

src/components/try-hcp-callout/components/try-hcp-callout-compact/index.tsx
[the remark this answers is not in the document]
zchsh: @ashleemboyer I suppose that's true, we could hide the `image` part in CSS rather than conditionally rendering it. To me it feels pretty similar.

The risk of unintentional divergence feels relatively low in my opinion - I think our `CardLink` component does a great job of handling the order for the link elements.

Rather than combining `TryHcpCallout` and `TryHcpCalloutCompact` here, maybe there's a case building for like `CardLinkWithBackground` that has `.backgroundBase` stuff built in? (Since that's something we use in a few other places, like the callout slot cards on the home page). Maybe that would help reduce some of the duplicative code here, while keeping the composition-y approach - feels like getting backgrounds to work as expected in `CardLink` will definitely come up again 👍

I'd be very on board to explore abstracting out the `CardLink` background work as part of 🎟 [Assess abstracting atomic "callout" components](https://app.asana.com/0/1202097197789424/1203396252179033/f). I'll add a lil breadcrumb note to that task - let me know what y'all think! 🙇

src/components/try-hcp-callout/components/try-hcp-callout-prebuilt/content.ts
ashleemboyer: Fine with deferring this to a subsequent PR since it's not a prod change, just wanted to make sure to note the inconsistencies I saw! 😊

And AFAIK we're not adding these to the content of any pages, but rather just making it available as an MDX component & marketing content block to enable authors to place it where they want to when they're ready.

src/components/try-hcp-callout/components/standalone-link-contents/standalone-link-contents.module.css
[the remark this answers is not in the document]
zchsh: Awesome! I added a note to 🎟 [Assess abstracting atomic "callout" components](https://app.asana.com/0/1202097197789424/1203396252179033/f). For this thread specifically, made a block for `StandaloneLinkContent`. For the other duplicated code here, which I think is mostly related to background styling, made a block for `CardLinkWithBackground` 👍

src/components/try-hcp-callout/components/product-icon-heading/index.tsx
ashleemboyer: Looks great!! Skip the first few seconds of this recording (I _really_ need to download Cleanshot again 🤣) and that's the output I got testing the preview URL 👍🏻 

https://user-images.githubusercontent.com/43934258/202780960-598e1ca0-01a8-46ac-ad1a-fcf1238467cf.mov

src/components/try-hcp-callout/components/try-hcp-callout-prebuilt/content.ts
zchsh: ☝️ Great Q re: MDX - I'm double-checking with Kelly on [this task](). In our handoff sync it sounded like the inverse: 
- We're adding the "default size" CTA to product landing pages (via `.json` authoring, but we're adding them in in specific spots and want to keep them there, so not exactly fully authorable)
- We're adding the "compact" CTA to sidecar on downloads views (JSX)
- We're _not_, at least from the latest info I have, intending to make `TryHcpCallout` available in MDX (?)

src/components/try-hcp-callout/components/try-hcp-callout-compact/index.tsx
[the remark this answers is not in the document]
ashleemboyer: > The risk of unintentional divergence for ariaLabel and critical element ordering feels relatively low in my opinion - I think our CardLink component does a great job of handling the order for the link elements.

To clarify -- I'm not referring to what `CardLink` does under the hood. I'm referring to:

- The actual values of `ariaLabel` passed to `CardLink` by the `TryHcpCallout` and `TryHcpCalloutCompact`. They both currently pass `ctaText` as `ariaLabel`. We don't want that value to diverge between the two components because it would cause them to have different screen reader announcements & voice control commands, and it would likely break the voice control experience.
- The order in which `TryHcpCallout` and `TryHcpCalloutCompact` arrange the elements they pass to `CardLink`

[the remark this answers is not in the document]
zchsh: I think we're referring to the same!

Since both components pass the `ctaText` prop directly, have a shared type, and aren't doing additional duplicate logic to generate the `ariaLabel` (say, like we might do in `TutorialsCard`), I personally think the risk of unintentional divergence is low. Similarly, personally I think the order of elements seems unlikely to be changed unintentionally.

It may be that we introduce new forms of the Try HCP callout that maybe have different compositions of the same elements, or it may be that we change one of the existing compositions (say, we decide to remove the `description` from the "compact" composition). In those cases we can intentionally have the different compositions diverge, and I don't think we'd introduce any new issues 🤔

src/components/try-hcp-callout/components/try-hcp-callout-prebuilt/content.ts
ashleemboyer: > ☝️ Great Q re: MDX - I'm double-checking with Kelly on [this task](https://app.asana.com/0/0/1203365375583141/f). In our handoff sync it sounded like kind of the inverse:
> 
> * We're adding the "default size" CTA to product landing pages (via `.json` authoring, but we're adding them in in specific spots and want to keep them there, so not exactly fully authorable)
> * We're adding the "compact" CTA to sidecar on downloads views (JSX)
> * We're _not_, at least from the latest info I have, intending to make `TryHcpCallout` available in MDX (?)

To clarify on what I meant by "to enable authors to place it where they want to when they're ready" -- I was referring to what landing pages (not the location in each template) and us just adding the slot to the templates, not updating the JSON for authors.

It'd be good to get clarification on whether or not we need an MDX component, because [there's a comment from Kyle in the feedback doc](https://docs.google.com/document/d/1hAD6CV0VKxmJZLBdq1FpS_V1wCWXHCS2Eo5tkzzJ4cs/edit?disco=AAAAiVsmLfQ) that says we do want one (which is why I decided to create [a separate ticket for it](https://app.asana.com/0/0/1203365375583141/f)):

> i think we need two "authoring points" 1 is the primitive (like HCPCallout where we just pass 1 product slug prop). I think additionally is that we need to have support for it within our landing page .json as well. i connected with Kelly about this specifically and she DOES want to have the ability for authors to use them in both places

src/components/try-hcp-callout/components/try-hcp-callout-compact/index.tsx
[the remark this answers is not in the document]
ashleemboyer: To me it feels like an unnecessary risk to already vulnerable users and unnecessary maintenance for us when we have the ability to prevent both and doing so would be low effort and a one-time, initial sort of effort too. 🤷🏻 But we can disagree about it and keep moving forward. Primarily wanted to make sure we were referring to the same thing.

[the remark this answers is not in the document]
zchsh: ☝️ Great point! Would definitely be on board with pursuing any and all efforts to reduce the risk & reduce maintenance - if you want to describe the fix you're suggesting on 🎟 [Assess abstracting atomic "callout" components](https://app.asana.com/0/1202097197789424/1203396252179033/f) (would that be the `normal | compact` prop?), would definitely be open to taking another look at it as part of follow-up work ☺️ I feel like cause of the non-trivial, but not overly complex scope here, it might be a good kind of case study for the tradeoffs of the various approaches we can come up with 💯

ashleemboyer (approved):

PR #1373, opened 2022-11-18
feat: add TryHcpCallout to install views

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Adds a "Try HCP" callout, specifically `TryHcpCalloutCompactPrebuilt`, to product install views that have an HCP offering.

## 🛠️ How

- Adds a `Prebuilt` version of `TryHcpCalloutCompact`, that automatically uses the same content as set up for `TryHcpCallout`.
- Adds a helper function `hasHcpCalloutContent`, as a typeguard and as a way to determine whether to render the callout on a given product's install view
- Adds the prebuilt component to the install view

## 📸 Design Screenshots

> **Note**: Vagrant and Nomad install views do not have any changes, as they don't have an HCP offering.

waypoint
boundary
vault
consul
packer
terraform

## 🧪 Testing

- [ ] Visit an install view that has an HCP-ish offering, such as [/terraform/downloads][/terraform/downloads] or [/vault/downloads][/vault/downloads]
    - An HCP callout should appear in the sidecar, on desktop viewports
    - On smaller viewports, the sidecar and the HCP callout within it should not be visible
- [ ] Visit an install view that does _not_ have an HCP-ish offering, such as [/vagrant/downloads][/vagrant/downloads]
    - The sidecar should appear the same as upstream, without an HCP callout

### On design tweaks

This PR tweaks the "compact" version of the callout to have a slightly smaller heading element, to avoid text wrap. Specifically:
- Uses `Display 200` (rather than `Display 300`)
- Uses `20px` icon size (rather than `24px`)

Have been DM'ing in Slack, these changes are not final! But felt worth implementing for preliminary review.

### On content

I've tweaked content for consistency relative to design mocks, which also helps avoid some text-wrap issues in this limited space scenario. Hopefully this looks alright! Intent in the implementation is to have the pre-built content consistent between the compact and default sizes of the callout.

### On sidecar scrolling

It'd be nice if the sidecar scrolled - on some laptop viewports, the sidecar content on the install view means the HCP callout is getting pretty close to "overflowing" and needing the whole main area to be scrolled through to become visible. However, this concern felt slightly outside the scope of this PR, since it's not a critical issue IMO. But definitely happy to explore it as a follow-up! We have a task here: 🎟 [[bug] Very tall sidecar needs overflow handling](https://app.asana.com/0/1202097197789424/1203060158979316/f)

[task]: https://app.asana.com/0/0/1203365375583147/f
[preview]: https://dev-portal-git-zsadd-try-hcp-to-downloads-hashicorp.vercel.app/
[/terraform/downloads]: https://dev-portal-git-zsadd-try-hcp-to-downloads-hashicorp.vercel.app/terraform/downloads
[/vault/downloads]: https://dev-portal-git-zsadd-try-hcp-to-downloads-hashicorp.vercel.app/vault/downloads
[/vagrant/downloads]: https://dev-portal-git-zsadd-try-hcp-to-downloads-hashicorp.vercel.app/vagrant/downloads


kendallstrautman (approved):
🥳

kendallstrautman:
💭 I wonder if that +9kb add on all the pages is related to the icon component, which could be getting quite large at this point.

kellyschummer (approved):
🙌 💯 - Thank you @zchsh!

anthonyvetter:
Sorry if this was already handled in the Design Review, but we have been getting frequent feedback about the use of the word "free". Try HCP Vault Free, for example. I know we are using that term in the wider in-line banners. And I assume we just didn't have the room in these smaller spaces, especially when it came to Terraform? Is that the case?

emilypersson1 (approved):
beautiful, thank you!

zchsh:
@anthonyvetter Great question on the inclusion of `"free"` - the limited space here is definitely a factor. We can definitely revisit content here, I think that'd be something to co-ordinate at the product level ☺️ 

@kendallstrautman I spent a little time trying to track down the source of the bundle size increase, but other than the component itself I didn't find much 🤔 I think the `ProductIcon` would already be included in the bundle due to its use in `IconTileLogo` in the heading area of the page. I did make a couple changes since the last review, figured maybe worth re-reviewing so tagged you just in case:
- Moved content to `src/content/<product>/install-landing.json` files. This way rather than each install page including content for all possible products, it includes only the content that'll actually be displayed. And, felt a bit better to have the content authorable in `src/content`.
- Split out the HCP logo from `ProductIconHeading` - the compact version of the component used here wouldn't work with the HCP logo anyways, and [the HCP logo alone is ~ 8 kB](https://github.com/hashicorp/mktg-logos/blob/main/product/hcp/primary/black.svg). (However, this change doesn't seem to have changed the bundle size much - I think it's still loaded in elsewhere in the bundle).
- Now ensuring the SVG background gets loaded into CSS as a URL, rather than a data string. I think this is what's trimmed down from 9 kB to the current ~ 7.5 kB bundle size change.

src/components/try-hcp-callout/components/product-icon-heading/index.tsx
```diff
@@ -1,44 +1,29 @@
...
- * the full HCP logo, and we render the provided headingText as visually hidden.
  */
 export function ProductIconHeading({
 	productSlug,
 	headingText,
+	size = 'medium',
 }: ProductIconHeadingProps) {
-	const uniqueId = useId()
-
-	if (productSlug === 'hcp') {
```
zchsh, line 25: Have split out the `"hcp"` case as it's only used in `TryHcpCallout`. This prevents `TryHcpCalloutCompact` from loading in the full HCP logo, when it won't be used.

src/components/try-hcp-callout/components/try-hcp-callout-compact/try-hcp-callout-compact.module.css
```diff
@@ -24,7 +24,7 @@
 
 .background {
 	composes: backgroundBase;
-	background-image: url('./img/try-hcp-callout-compact-background.svg');
+	background-image: url('./img/try-hcp-callout-compact-background.svg?url');
```
zchsh, line 27: `?url` suffix here ensures the SVG background doesn't get loaded in as a data string, which should help bundle size as well.

src/components/try-hcp-callout/components/product-icon-heading/index.tsx
kendallstrautman: nice, I like it!

src/views/product-downloads-view/server.ts
```diff
@@ -106,6 +108,23 @@ const generateGetStaticProps = (
...
+		const isHcpProduct = hasHcpCalloutContent(product.slug)
+		const hasContent = typeof rawSidecarHcpCallout !== 'undefined'
+		if (hasContent && isHcpProduct) {
+			sidecarHcpCallout = {
+				...rawSidecarHcpCallout,
+				productSlug: product.slug,
+			}
+		} else if (hasContent) {
+			console.warn(
+				`In "product-downloads-view", for product slug "${product.slug}", "sidecarHcpCallout" content was provided, but the Try HCP callout component cannot be rendered as "${product.slug}" is not recognized as a product with an HCP offering. Please ensure "sidecarHcpCallout" content only exists for products with HCP offerings, or report this as a bug.`
```
kendallstrautman, line 124: Good call adding this feedback

kendallstrautman (approved):
Looks good! ty for the adjustments 😄

PR #1375, opened 2022-11-22
[wip] refactoring auth state

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksrefactoring-auth-state-hashicorp.vercel.app/) 🔎
- [Asana task](url) 🎟️

### DIFF
--- a/src/hooks/bookmarks/use-all-bookmarks.ts
+++ b/src/hooks/bookmarks/use-all-bookmarks.ts
@@ -3,7 +3,7 @@ import {
 	getAllBookmarks,
 	GetAllBookmarksResult,
 } from 'lib/learn-client/api/bookmark'
-import { useAuthenticationToken } from 'hooks/use-authentication'
+import useAuthentication from 'hooks/use-authentication'
 
 type QueryDataType = GetAllBookmarksResult
 
@@ -26,7 +26,7 @@ const useAllBookmarks = ({
 	const queryClient = useQueryClient()
 
 	// Get the current user's access token
-	const accessToken = useAuthenticationToken()
+	const { isAuthenticated, token: accessToken } = useAuthentication()
 
 	// Set up the `onSuccess` callback
 	const onSuccess = (data: QueryDataType) => {
@@ -41,7 +41,7 @@ const useAllBookmarks = ({
 		['bookmarks'],
 		() => getAllBookmarks({ accessToken }),
 		{
-			enabled: enabled && !!accessToken,
+			enabled: enabled && isAuthenticated && !!accessToken,
 			onSuccess,
 		}
 	)
--- a/src/hooks/bookmarks/use-all-bookmarks.ts
+++ b/src/hooks/bookmarks/use-all-bookmarks.ts
@@ -26,7 +26,7 @@ const useAllBookmarks = ({
 	const queryClient = useQueryClient()
 
 	// Get the current user's access token
-	const { isAuthenticated, token: accessToken } = useAuthentication()
+	const { isAuthenticated, accessToken } = useAuthentication()
 
 	// Set up the `onSuccess` callback
 	const onSuccess = (data: QueryDataType) => {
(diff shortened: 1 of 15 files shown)


src/hooks/bookmarks/use-all-bookmarks.ts
```diff
@@ -42,7 +41,7 @@ const useAllBookmarks = ({
 		['bookmarks'],
 		() => getAllBookmarks({ accessToken }),
 		{
-			enabled: enabled && !!accessToken,
+			enabled: enabled && isAuthenticated && !!accessToken,
```
kendallstrautman, line 44: Instead of relying on the presence of the accessToken to check whether we are in a good state to make an API call, we should rely on something like `isAuthenticated`, which could be adjusted if we are in an error state where an accessToken may exist but it may not be valid.

src/hooks/use-authentication/index.ts
```diff
@@ -58,28 +85,16 @@ const useAuthentication = (
...
 	// Return everything packaged up in an object
 	return {
-		error,
-		hasError,
+		error: data?.error,
 		isAuthenticated,
 		isLoading,
-		session,
-		showAuthenticatedUI,
-		showUnauthenticatedUI,
```
kendallstrautman, line 77: I want to simplify what is returned from this hook, some of these properties weren't used or can be inferred in other ways. 

The `showAuthenticatedUI` properties were only used in one spot, so to simplify things I decided to remove them from the hook for now, since they can be inferred from other values already returned from the hook.

src/pages/api/auth/[...nextauth].ts
```diff
@@ -94,13 +94,16 @@ export default NextAuth({
...
-						`${AuthErrors.RefreshAccessTokenError}: failed to refresh token set`,
-						err
-					)
+					let errorType = AuthErrors.RefreshAccessTokenError
+
+					if (err.error === 'token_inactive') {
+						errorType = AuthErrors.RefreshAccessTokenExpiredError
+					}
+
+					console.error(`${errorType}: failed to refresh token set`, err)
```
Comment (kendallstrautman) [inline src/pages/api/auth/[...nextauth].ts:103]: This was implemented on another PR, may be removed if that other one is merged.

src/hooks/use-authentication/index.ts
```diff
@@ -1,53 +1,80 @@
...
+	// retain authenticated state based on status
+	useEffect(() => {
+		if (!isAuthenticated && status === 'authenticated') {
+			setIsAuthenticated(true)
+		} else if (isAuthenticated && status === 'unauthenticated') {
+			setIsAuthenticated(false)
+		} else if (status !== 'authenticated') {
+			setIsAuthenticated(false)
+		}
+	}, [status, isAuthenticated])
```
kendallstrautman, line 56: Change here is to store `isAuthenticated` in state so that we can adjust this based not only on the session status but if there is an error, to prevent API calls for authenticated data if we are in an errored state.

kendallstrautman: Open to another way, but wanted to see how this behaved.

src/hooks/bookmarks/use-all-bookmarks.ts
kendallstrautman: I suppose I could also use error or status here instead of storing something in state 🤔

src/hooks/use-authentication/index.ts
```diff
@@ -1,53 +1,80 @@
...
+const useAuthentication = (
+	options: UseAuthenticationOptions = {}
+): UseAuthenticationResult => {
+	const { signIn, signOut } = useAuthMethods()
 	// Get option properties from `options` parameter
+	// `onUnauthenticated` function only calls when `isRequired` is true
 	const { isRequired = false, onUnauthenticated = () => signIn() } = options
 
 	// Pull data and status from next-auth's hook, and pass options
 	const { data, status } = useSession({
```
kendallstrautman, line 41: another idea, we could use react query to store the session to cache it properly. https://github.com/nextauthjs/react-query/blob/main/index.js

PR #1382, opened 2022-11-23
chore: upgrade next-auth

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksupgrade-next-auth-hashicorp.vercel.app/) 🔎
- [Asana task](url) 🎟️

## 🗒️ What
Upgrading to the latest version of next auth. Looks like the biggest changes related to types. We should do smoketests to verify that auth is still working as expected.

### DIFF
--- a/src/components/mobile-menu-container/index.tsx
+++ b/src/components/mobile-menu-container/index.tsx
@@ -40,10 +40,11 @@ const MOBILE_MENU_MOTION = {
  */
 const MobileAuthenticationControls = () => {
 	const { asPath } = useRouter()
-	const { showAuthenticatedUI, showUnauthenticatedUI, signIn, signOut, user } =
+	const { isAuthenticated, isLoading, signIn, signOut, user } =
 		useAuthentication()
+	const showUnauthenticatedUI = !isLoading && !isAuthenticated
 
-	if (!showAuthenticatedUI && !showUnauthenticatedUI) {
+	if (!isAuthenticated && !showUnauthenticatedUI) {
 		return null
 	}
 
@@ -68,7 +69,7 @@ const MobileAuthenticationControls = () => {
 				/>
 			</>
 		)
-	} else if (showAuthenticatedUI) {
+	} else if (isAuthenticated) {
 		content = (
 			<MobileUserDisclosure
 				items={getUserMenuItems({ signOut })}
--- a/src/hooks/use-authentication/index.ts
+++ b/src/hooks/use-authentication/index.ts
@@ -56,8 +56,6 @@ const useAuthentication = (
 	// Deriving booleans about auth state
 	const isLoading = status === 'loading'
 	const isAuthenticated = status === 'authenticated'
-	const showAuthenticatedUI = isAuthenticated
-	const showUnauthenticatedUI = !isLoading && !isAuthenticated
 	const preferencesLoaded = preferencesSavedAndLoaded()
 	const error = data?.error
 
@@ -84,8 +82,6 @@ const useAuthentication = (
 		isAuthenticated,
 		isLoading,
 		session,
-		showAuthenticatedUI,
-		showUnauthenticatedUI,
 		signIn,
 		signOut,
 		signUp,
--- a/src/hooks/use-authentication/index.ts
+++ b/src/hooks/use-authentication/index.ts
@@ -6,6 +6,7 @@ import { preferencesSavedAndLoaded } from '@hashicorp/react-consent-manager/util
 import {
 	AuthErrors,
 	SessionData,
+	SessionStatus,
 	UserData,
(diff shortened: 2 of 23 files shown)


src/hooks/use-authentication/index.ts
```diff
@@ -43,7 +44,7 @@ const useAuthentication = (
 	const { data, status } = useSession({
 		required: isRequired,
 		onUnauthenticated,
-	})
+	}) as { data: SessionData; status: SessionStatus }
```
kendallstrautman, line 47: Is there a better way to do this? The `useSession` type is kinda nuts in the next auth code.

src/hooks/use-authentication/types.ts
```diff
@@ -19,7 +24,7 @@ interface UseAuthenticationOptions {
 
 interface UseAuthenticationResult {
 	// https://github.com/nextauthjs/next-auth/blob/next-auth@v4.10.3/packages/next-auth/src/core/types.ts#L422-L440
-	error: undefined | Session['error']
+	error: SessionData['error']
```
kendallstrautman, line 27: Apparently 'error' isn't on the next auth session type anymore, same with `status`

src/pages/api/auth/[...nextauth].ts
```diff
@@ -122,13 +131,19 @@ export default NextAuth({
...
+			token: JWT
+			session: NextAuthSession
+		}): Promise<SessionData> {
 			return {
 				...session,
-				accessToken: token.access_token,
-				id: token.sub,
-				user: { ...session.user, nickname: token.nickname },
-				error: token.error,
+				accessToken: token.access_token as TokenSet['access_token'],
```
Comment (kendallstrautman) [inline src/pages/api/auth/[...nextauth].ts:143]: Again I had a hard time typing this. I can't set the token to `JWT` on the function itself because next auth requires it to be a JWT, so I have to cast as the `TokenSet`. Maybe I can just add a @TSFixMe to address later.

src/types/auth.ts
```diff
@@ -1,32 +1,42 @@
...
+export type SessionData = Omit<NextAuthSession, 'user'> & {
 	accessToken?: string
 	id?: string
 	error?: AuthErrors
+	user: NextAuthSession['user'] & {
+		nickname: string
+	}
+}
+
+export type SessionStatus = 'authenticated' | 'unauthenticated' | 'loading'
```
kendallstrautman, line 19: Defining this type in our app because I couldn't find it defined in next auth types.

src/hooks/use-authentication/index.ts
kendallstrautman: Ty for doing this digging and finding those docs. I updated things to use the modules and it seems to work _much_ better.

src/components/mobile-menu-container/index.tsx
```diff
@@ -40,10 +40,11 @@ const MOBILE_MENU_MOTION = {
  */
 const MobileAuthenticationControls = () => {
 	const { asPath } = useRouter()
-	const { showAuthenticatedUI, showUnauthenticatedUI, signIn, signOut, user } =
```
kendallstrautman, line 43: I chose to remove `showAuthenticatedUI` booleans because they were only used in one place and the logic can be inferred already from values that are passed from `useAuthentication` hook.

src/hooks/use-authentication/index.ts
```diff
@@ -84,16 +77,12 @@ const useAuthentication = (
...
-		error,
 		isAuthenticated,
 		isLoading,
 		session,
-		showAuthenticatedUI,
-		showUnauthenticatedUI,
 		signIn,
 		signOut,
 		signUp,
-		status,
```
kendallstrautman, line 96: All the ones removed here were either not used or could be inferred from data already passed in other properties. In an effort to simplify this hook, I decided to remove them!

src/types/auth.ts
```diff
@@ -1,45 +1,15 @@
...
-
 export enum AuthErrors {
 	// NOTE: temporary prefix of [next-auth] for existing log monitor
 	// TODO: abstract logging so everything is grouped
 	RefreshAccessTokenError = '[next-auth] RefreshAccessTokenError',
 	RefreshAccessTokenExpiredError = '[next-auth] RefreshAccessTokenError: token_inactive',
 }
-
-/** The response shape from `POST {IDENTITY_PROVIDER}/oauth2/token` */
-export interface TokenSet extends JWT {
```
kendallstrautman, line 38: This is defined in the `JWT` module extension now in next-auth.d.ts

zchsh (approved):
🙌 Looks great to me! Quickly tested by logging in and out on the preview, and messing with some of my bookmarks and progress. Happy to dive deeper if there are other testing steps I could take 🙇

PR #1386, opened 2022-11-23
refactor: adjust is authenticated

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksadjust-is-authenticated-hashicorp.vercel.app/) 🔎
- [Asana task](url) 🎟️

## 🗒️ What

This adjusts the `enabled` boolean for authenticated data queries to rely on `isAuthenticated` instead of just the presence of the token, which could be invalid. I adjusted the logic for `isAuthenticated` to account for a token refresh error state. 

This should help reduce api calls when we're in that errored state and a stream of 401 errors in the logs when debugging this issue. Additionally, I think react query fetches errors calls 3 additional times by default, so preventing the errored call is what we're trying to do here. 

## 🧪 Testing

Authenticate on the preview and test that things work as expected. 

If wanting to simulate the errored state pull this down locally and go to [...nextauth].ts where the refresh token error is handled. Adjust the booleans so that the app only throws the refresh token error and you should see in the network tab that 401 calls aren't happening to the Learn API.


src/hooks/use-authentication/index.ts
```diff
@@ -41,26 +41,25 @@ const useAuthentication = (
 		onUnauthenticated,
 	}) as { data: Session; status: SessionStatus }
 
+	// Deriving booleans about auth state
+	const isLoading = status === 'loading'
+	const isAuthenticated =
```
kendallstrautman, line 46: I've considered putting this in `state`...but I think this a sufficient change for now given that we want to access session data via react query in subsequent work. 

Trying to weigh the pros and cons.

src/hooks/use-authentication/index.ts
```diff
@@ -41,26 +41,25 @@ const useAuthentication = (
...
+
 	/**
 	 * Force sign out to hopefully resolve the error. The user is signed out
 	 * to prevent unwanted looping of requesting an expired refresh token
 	 *
 	 * https://next-auth.js.org/tutorials/refresh-token-rotation#client-side
 	 */
 	useEffect(() => {
-		if (
-			data?.error === AuthErrors.RefreshAccessTokenExpiredError ||
```
kendallstrautman, line 52: I removed this since we just need to catch the generic error now.

src/pages/api/auth/[...nextauth].ts
```diff
@@ -103,16 +103,13 @@ export default NextAuth({
 					token.access_token = access_token
 					token.refresh_token = refresh_token
 				} catch (err) {
-					let errorType = AuthErrors.RefreshAccessTokenError
-
-					if (err.error === 'token_inactive') {
-						errorType = AuthErrors.RefreshAccessTokenExpiredError
-					}
```
Comment (BRKalow) [inline src/pages/api/auth/[...nextauth].ts:110]: Why don't we care about the error type here anymore? 💭

Comment (kendallstrautman) [reply src/pages/api/auth/[...nextauth].ts:110]: Well we're not doing anything specific with it anymore. Were just signing out on either error type at the moment.

Comment (kendallstrautman) [reply src/pages/api/auth/[...nextauth].ts:110]: ah yes you beat me to it 😄

zchsh (approved):
✅ This seems like a great idea to me! 💯

PR #1387, opened 2022-11-23
chore: removes link warning

kendallstrautman:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

In local dev, many toasts were popping to show this warning. I believe since this component supports both options we shouldn't show this warning toast as it degrades developer experience. Using the target attribute should be an open preference and the cognitive load could go both ways if one isn't familiar with the prop structure of this link component. 

## To Test

Pull things down locally, start the server and you should see no link warnings pop.


zchsh (approved):
✅ I'm on board with this! Had some thoughts earlier in Slack, but realized I just didn't read the source for `components/link` closely enough - can see now that `opensInNewTab` & `target="_blank"` are already essentially aliases 👍 💯 

---

⛏️  Nitpick outside the changed lines, though kind of related to this PR... I think as-is, with this warning now removed, you could do something like the following, passing different values for the two aliases, and not get a warning:

```tsx
<Link {...otherProps} opensInNewTab={false} target="_blank" />
```

At first I thought we needed to retain some separate warning slash add logic to account for that, but now noticing might be a simpler tweak on line 48 worth making (also added a quick description suggestion since I think that was a pasto-typo over from the main file):

```diff
/**
+ * Ensure only one of `opensInNewTab` or `target="_blank"` is passed.
- * Generate the `target` prop.
 *
 * @NOTE The `opensInNewTab` prop takes precedence over the `target` prop.
 * This is because the `opensInNewTab` prop renders a screen-reader-only label
 * when `true`. If both `opensInNewTab` and `target` were passed, and `target`
 * did not have the value of `"_blank"`, then the screen-reader-only label
 * would not be accurate.
 */
+ if (typeof opensInNewTab !== "undefined" && typeof target !== "undefined")
- if (opensInNewTab && !!target) {
	developmentToast({
		color: ToastColor.critical,
		title: 'Error in src/components/Link',
		description:
			'Both `opensInNewTab` and `target` were passed. Only pass one or the other.',
	})
}
```

PR #1388, opened 2022-11-25
refactor: CardsGridList auto-wrap in <li />

zchsh:
## 🔗 Relevant links

- [Asana task](url) 🎟️

### DIFF
--- a/src/components/cards-grid-list/docs.mdx
+++ b/src/components/cards-grid-list/docs.mdx
@@ -17,18 +17,10 @@ The layout for this component is driven by minimum width settings on card childr
 
 <LiveComponent>{`
 <CardsGridList isOrdered={false}>
-  <li style={{ border: '1px solid magenta', padding: '16px' }}>
-      I am a fake card.
-  </li>
-  <li style={{ border: '1px solid magenta', padding: '16px' }}>
-      I am a fake card.
-  </li>
-  <li style={{ border: '1px solid magenta', padding: '16px' }}>
-      I am a fake card.
-  </li>
-  <li style={{ border: '1px solid magenta', padding: '16px' }}>
-      I am a fake card.
-  </li>
+	<div style={{ border: '1px solid magenta', padding: '16px' }}>I am a fake card.</div>
+	<div style={{ border: '1px solid magenta', padding: '16px' }}>I am a fake card.</div>
+	<div style={{ border: '1px solid magenta', padding: '16px' }}>I am a fake card.</div>
+	<div style={{ border: '1px solid magenta', padding: '16px' }}>I am a fake card.</div>
 </CardsGridList>
 `}</LiveComponent>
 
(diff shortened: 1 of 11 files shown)


src/views/tutorial-view/components/featured-in-collections/index.tsx
```diff
@@ -25,9 +25,10 @@ export function FeaturedInCollections({
 				<CardsGridList fixedColumns={collections.length == 1 ? 2 : null}>
 					{collections.map((cardPropsWithId: CollectionCardPropsWithId) => {
 						return (
-							<li key={cardPropsWithId.id} className={s.listItem}>
```
zchsh, line 28: Note: the `s.listItem` class removed here didn't actually exist, so wasn't doing anything.

You can see this by inspecting the "Featured in collections" are of a live page such as [/terraform/tutorials/aws-get-started/install-cli](https://developer.hashicorp.com/terraform/tutorials/aws-get-started/install-cli) - note that the `<li />` elements do not have a `className` rendering (because here in the code, `s.listItem` is undefined).

src/components/cards-grid-list/docs.mdx
```diff
@@ -6,29 +6,23 @@ Displays cards in a grid. Intended for use in main content areas.
 
 The layout for this component is driven by minimum width settings on card children, rather than by explicit column counts. This reduces brittleness, helps ensure we never shrink cards below their desired minimum width, and opens the possibility of using this component outside of our main content area, if desired.
 
+Note that `CardsGridList` should be passed raw card elements as `children`, such as `<div />`, `<a />`, `<CardLink />`, and so on. Do _not_ wrap card elements in `<li />`, as the component handles this internally.
```
zchsh, line 9: I've updated the docs to try to make the new behaviour clear 👍

src/components/cards-grid-list/index.tsx
```diff
@@ -64,7 +64,9 @@ function CardsGridList({
 				} as CSSProperties
 			}
 		>
-			{children}
+			{Children.map(children, (child) => {
+				return <li>{child}</li>
```
kendallstrautman, line 68: 💭 Does this need a key?

src/views/tutorial-view/components/featured-in-collections/index.tsx
kendallstrautman: good catch!

kendallstrautman (approved):
Nice improvement!

src/components/cards-grid-list/index.tsx
zchsh: I don't think needs `key` - my understanding is `Children.map` will use the `key` of each `child`. My main point of reference was [this StackOverflow question](https://stackoverflow.com/a/50303840), there's also [an example in the beta React docs](https://beta.reactjs.org/apis/react/Children#transforming-children) that I think kind of confirms this.

PR #1396, opened 2022-11-29
chore: adjust asset

kendallstrautman:
## 🔗 Relevant links

- [Asana task](url) 🎟️


BRKalow (approved):

braydenlove (approved):
I am still confused why it is showing up as 375 x 200px here when the .png is 334 x 178px. It is a little bigger but not by a ton, for now I think we can ship as we have a ticket already written up to look at all of these graphics holistically across the site.

kendallstrautman:
@braydenlove I think its because its scaling to fit the container which should match the height of the blurb to the left.

PR #1402, opened 2022-11-30
fix: code block hydration mismatches

### LINKED ISSUE #1401: Slow performance on Nomad API documentation pages
👋,

Since the move to `developer.hashicorp.com` I've been running into a few performance issues compared to the old Nomad documentation (both time to render a page and ongoing memory usage, with things like the nomad jobs api page taking ~500-800MB of RAM alone).

For a specific example: https://developer.hashicorp.com/nomad/api-docs/jobs takes about 3 minutes to render on my 32 core / 64GB RAM desktop in Firefox 107. It also appears that this needs to recompute on every page load.

I'm not really a competent browser person, but I can upload a Firefox profile if that's helpful 😅

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Pulls in this code block change to address hydration issues: https://github.com/hashicorp/react-components/pull/834

Fixes #1401 

## 🧪 Testing

- [ ] Visit https://dev-portal-git-brkfix-hydration-mismatch-codeblock-hashicorp.vercel.app/nomad/api-docs/jobs
- [ ] The page should load in a reasonable time and there should be no react errors in the console

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,7 +28,7 @@
 				"@hashicorp/react-alert-banner": "^7.0.1",
 				"@hashicorp/react-button": "^6.2.1",
 				"@hashicorp/react-call-to-action": "^5.1.0",
-				"@hashicorp/react-code-block": "^6.2.2-canary-20221130173256",
+				"@hashicorp/react-code-block": "^6.2.2",
 				"@hashicorp/react-command-line-terminal": "^2.0.4",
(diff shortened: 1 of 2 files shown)


zchsh (approved):
✅ Checked out https://github.com/hashicorp/react-components/pull/834, and this fix makes a lot of sense to me! 👍

PR #1406, opened 2022-11-30
refactor: adjust tutorial search types

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksadjust-tutorial-search-types-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1203304523389989/1203467793559452) 🎟️

## 🗒️ What

Tweaks types to better reflect the actual hit results. And updates the library hit results to use the shared type returned from the learn index.


src/components/command-bar/commands/search/components/tutorials-tab-contents/types.ts
```diff
@@ -11,15 +16,25 @@ interface TutorialsTabContentsProps {
...
+		| 'name'
+		| 'slug'
+		| 'description'
+		| 'readTime'
+		| 'edition'
+	> & {
+		hasVideo: boolean
+		isInteractive: boolean
+		products: ProductOption[]
+		level: CollectionLevelOption
```
kendallstrautman, line 36: I hate this but I was having trouble with the `BaseHit` requirements if I tried to make this into an interface that extends these types.

src/components/command-bar/commands/search/components/tutorials-tab-contents/types.ts
```diff
@@ -11,15 +16,25 @@ interface TutorialsTabContentsProps {
...
 type TutorialHitObject = Hit<{
 	description: string
-	name: string
-}> & {
-	defaultContext: {
-		slug: string
+	page_title: string
+	name: string // TODO: remove this once page_title is fully seeded
+	headings: string[]
+}> &
```
kendallstrautman, line 22: One question I have and was curious about the previous implementation. When the types are passed to `Hit` does that reflect the searchable attributes? Or why wasn't `defaultContext` included in that first object?

src/components/command-bar/commands/search/components/tutorials-tab-contents/types.ts
```diff
@@ -11,15 +16,25 @@ interface TutorialsTabContentsProps {
 
 type TutorialHitObject = Hit<{
 	description: string
-	name: string
-}> & {
-	defaultContext: {
-		slug: string
+	page_title: string
```
kendallstrautman, line 19: I think I'll wait to do this one as a follow up

src/components/command-bar/commands/search/components/tutorials-tab-contents/types.ts
```diff
@@ -11,15 +16,25 @@ interface TutorialsTabContentsProps {
...
+	name: string // TODO: remove this once page_title is fully seeded
```
kendallstrautman, line 20:
```suggestion
	name: string 
```

src/components/command-bar/commands/search/components/tutorials-tab-contents/types.ts
zchsh: Great question, I'm not sure tbh 🤔 That they might be searchable attributes seems like a good explanation to me 👍 I would think maybe only the `defaultContext.name` or something would be a searchable attribute, so maybe that's why the whole `defaultContext` obj was left out? (i'm speculating, I'm not very familiar with the tutorials index or the command-bar search code here 😅 )

src/components/command-bar/commands/search/components/tutorials-tab-contents/tutorial-hit.tsx
```diff
@@ -41,7 +41,7 @@ const TutorialHit = ({ hit }: TutorialHitProps) => {
 	}
 
 	const badges = products?.map(
-		(productSlug: ProductSlug) => productSlugsToNames[productSlug]
+		(productSlug: LearnProductSlug) => productSlugsToNames[productSlug]
```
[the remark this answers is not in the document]
zchsh: Hmm yeah - i guess they're similar, but `ProductOption` seems like more "correct" i guess since it's specifically around Learn? Personally also on the fence, and think we'd be fine with either 👍

zchsh (approved):

PR #1407, opened 2022-12-01
fix: ensure product redirects do not get dropped

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

It seems like product-specific redirects are getting dropped when we handle grouping our redirects for use in middleware. I've updated the logic and added a test case to ensure that redirects for the developer domain are retained.

### DIFF
--- a/build-libs/redirects.js
+++ b/build-libs/redirects.js
@@ -304,8 +304,14 @@ function splitRedirectsByType(redirects) {
 			) ||
 			(redirect.has && redirect.has.some((has) => has.type !== 'host'))
 		) {
+			if (redirect.source === '/nomad/tools/autoscaling/concepts/checks') {
+				console.log('glob', redirect)
+			}
 			globRedirects.push(redirect)
 		} else {
+			if (redirect.source === '/nomad/tools/autoscaling/concepts/checks') {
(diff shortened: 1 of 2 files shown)


build-libs/redirects.js
```diff
@@ -304,8 +304,14 @@ function splitRedirectsByType(redirects) {
 			) ||
 			(redirect.has && redirect.has.some((has) => has.type !== 'host'))
 		) {
+			if (redirect.source === '/nomad/tools/autoscaling/concepts/checks') {
+				console.log('glob', redirect)
+			}
```
BRKalow, line 309:
```suggestion
```

build-libs/redirects.js
```diff
@@ -304,8 +304,14 @@ function splitRedirectsByType(redirects) {
...
 			globRedirects.push(redirect)
 		} else {
+			if (redirect.source === '/nomad/tools/autoscaling/concepts/checks') {
+				console.log('simple', redirect)
+			}
```
BRKalow, line 314:
```suggestion
```

zchsh (approved):

PR #1414, opened 2022-12-06
patch: don't use link component

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-kshandle-not-found-app-hashicorp.vercel.app/vault/docs/enterprise/redundancy-zones) 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1203192943606795) 🎟️

## 🗒️ What

This feels like a temporary workaround. Basically we need to find out why the `notFound` return value from `getStaticProps` isn't taking precedent here. The server shows that it knows the `notFound` state but the client throws the error first. 

The error is reproducible locally if you run `npm run serve` and use the product switcher to navigate to a page that doesn't exist. 

Handling in `_app` wouldn't change because the clientside error boundary is what catches and the clientside error doesn't know the 404 status that has been determined in `getStaticProps`.

Next steps as I see it is to dig through next.js issues / discussions to see if anyone else has similar issues with pre-fetching links whose data 404s. 

---
Update: I realized that[ next router _should_ be handling this ](https://app.asana.com/0/0/1203192943606795/1203510379551105/f), but our error boundary is catching it instead. I'm going to research a path forward where we can handle this in our custom error page. 

## 🧪 Testing

- [ Visit this path](https://dev-portal-git-kshandle-not-found-app-hashicorp.vercel.app/vault/docs/enterprise/redundancy-zones)
- Hit v1.9.x in the version switcher dropdown
- The page should navigate to the correct 404 view


kendallstrautman:
@ashleemboyer @zchsh opening this patch up as an interim workaround until we can find a better path with the error boundary. 

In looking into the next/link code, the 404 link _should_ route to the custom _error page but instead a clientside error is happening before that occurs. So we need to perhaps create a custom error boundary component, although I'm not sure how we would catch this precise instance since the clientside error is a reference error, trying to access props that aren't defined. 

I think the longer term idea of not rendering links in the dropdown that we _know_ 404 is the best move. What do y'all think about this as a short term patch?

src/components/dropdown-disclosure/components/list-item/index.tsx
```diff
@@ -122,6 +122,31 @@ const DropdownDisclosureLinkItem = ({
 	)
 }
 
+/*
+ * This component provides a temporary workaround using an anchor tag
+ * instead of 'link' component. There has been an issue observed where
+ * 404 links throw uncaught error pages with the version switcher drpodown
+ */
+
+const DropdownDisclosureAnchorItem = ({
```
kendallstrautman, line 131: Created a new component here so that the user dropdown component will still use `Link` and the anchor tag use is contained to just the version switcher dropdown.

zchsh (approved):
🙌 I dig this as a short-term patch! Feels like a very reasonable trade-off to sacrifice a little bit of performance in accessing past versions in order to resolve the errors 💯 

Agree that avoiding rendering "old version" links that will 404 seems like it'd be a nice long-term solution 👍 I'm picturing like a grayed-out disabled link state so that it's clear the old version exists, but also clear that the current document doesn't exist in said old version. I'm unsure on the feasibility details on the content-api side... but I assume it must be possible! 🤞

PR #1422, opened 2022-12-07
RELENG-298: update command array for vagrant

claire-labry:
## 🔗 Relevant links

We've introduced the vagrant cask to hashicorp's homebrew-tap: https://github.com/hashicorp/homebrew-tap/pull/193

## 🗒️ What / Why

This updates the commands to use when brew installing the vagrant cask -- source is now under hashicorp instead of the official homebrew tap.


BRKalow (approved):

mdeggies (approved):

src/pages/_proxied-dot-io/vagrant/downloads/index.jsx
```diff
@@ -46,12 +46,12 @@ function DownloadsPage({ product, releases, latestVersion }) {
...
 				{
 					label: 'Homebrew',
-					commands: [`brew install vagrant`],
+					commands: [`brew install hashicorp/tap/hashicorp-vagrant`],
 					os: 'darwin',
 				},
 				{
 					label: 'Homebrew',
-					commands: [`brew install vagrant`],
+					commands: [`brew install hashicorp/tap/hashicorp-vagrant`],
```
mdeggies, line 54: oh actually, I think we need to revert this since only darwin AMD is supported

mdeggies: @BRKalow are we ok to remove this whole section? https://github.com/hashicorp/dev-portal/blob/64af4af39b67eb9763cefaafa1486c57d71a5ab2/src/pages/_proxied-dot-io/vagrant/downloads/index.jsx#L52-L56 This will not work for users

claire-labry: I went ahead and removed this 👍🏻

PR #1427, opened 2022-12-08
kw.integrations demo 2 generated client

thiskevinwang:
- add customRequest
- commit generated client
- use new client


BrandonRomano:
We'll add this in later (on our other branch) once we've got this released as a package.

PR #1429, opened 2022-12-08
feat: Product Integrations version selector

thiskevinwang:
## 🔗 Relevant links

- https://dev-portal-lgjps1uqf-hashicorp.vercel.app/waypoint/integrations/docker
- [Asana task](https://app.asana.com/0/1202626741675858/1203394837268540/f) 🎟️

## 🗒️ What

This implements version-selector behavior for Product Integrations

https://user-images.githubusercontent.com/26389321/206563684-a1c53221-20d7-4587-af38-3ab559b99359.mp4


src/views/product-integration/components/header/index.tsx
```diff
@@ -24,12 +27,26 @@ export default function Header({
...
 	description,
+	activeRelease,
 }: HeaderProps) {
+	// /waypoint/integrations/docker[/0.10.2]
+	const { asPath } = useRouter()
+
+	// derive some values from the path
+	const productSlug = asPath.split('/')[1]
+	const integrationSlugAndVersion = asPath.split('/').slice(3).join('/')
+	const [integrationSlug, _] = integrationSlugAndVersion.split('/')
```
BrandonRomano, line 38: I don't think this component should be coupled to the router.

The data that you're pulling out of the URL structure should be able to be passed to this header as props, which will keep it independent of where it is rendered in the app.

src/views/product-integration/components/variable-group-list/index.tsx
```diff
@@ -81,7 +81,7 @@ function unflattenVariables(variables: Array<Variable>): Array<Variable> {
 			const keyDepth = key.split('.').length
 			if (keyDepth > maxDepth) {
 				maxDepth = keyDepth
-				console.log(key)
+				// console.log(key)
```
BrandonRomano, line 84:
```suggestion
```

Let's just remove these while we're at it!

src/views/product-integration/components/variable-group-list/index.tsx
```diff
@@ -72,7 +72,7 @@ function unflattenVariables(variables: Array<Variable>): Array<Variable> {
 	// Pull all of the root nodes out
 	const rootNodes: Array<Variable> = []
 
-	console.log(variables)
+	// console.log(variables)
```
BrandonRomano, line 75:
```suggestion
```

BrandonRomano (changes requested):
Largely good!  I don't think the header has to be coupled to the router; I think we should just try to pass it the props it needs.   Will cause less of a headache of the IA ever changes & will make it more reusable elsewhere if that ever happens.

src/views/product-integration/components/header/index.tsx
thiskevinwang: I'll look at generating the links, server side — that might be the most reliable

thiskevinwang: Done serverside: [b681eed](https://github.com/hashicorp/dev-portal/pull/1429/commits/b681eedc96c70776311d40a3c70748fabcf1e68e)

BrandonRomano: I like your resolution here!

BrandonRomano (approved):

PR #1404, opened 2022-11-30
spike: certifications

zchsh:
🚨 DO NOT MERGE 🚨 

This is currently an implementation spike, just experimenting.

## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Adds a certification landing page at [/certifications][/certifications], as well as individual certification landing pages to match those listed under <https://www.hashicorp.com/certification>.

## 🛠️ How

- Sets up some local content in `.json` files
- Drives `/certifications/<program>` pages off local files available
- Uses `zod` schemas for validation and typing
- TODO: polished UI work

## 📸 Design Screenshots

None yet.

## 🧪 Testing

Visit each of the certification pages and ensure they look and function as expected:

- [/certifications][/certifications]
- [/certifications/infrastructure-automation][/certifications/infrastructure-automation]
- [/certifications/networking-automation][/certifications/networking-automation]
- [/certifications/security-automation][/certifications/security-automation]

## 💭 Anything else?

Not at the moment.

[preview]: https://dev-portal-git-zscertifications-spike-hashicorp.vercel.app
[/certifications]: https://dev-portal-git-zscertifications-spike-hashicorp.vercel.app/certifications
[task]: https://app.asana.com/0/1203301890649977/1203186539243924/f

[/certifications/infrastructure-automation]: https://dev-portal-git-zscertifications-spike-hashicorp.vercel.app/certifications/infrastructure-automation
[/certifications/networking-automation]: https://dev-portal-git-zscertifications-spike-hashicorp.vercel.app/certifications/networking-automation
[/certifications/security-automation]: https://dev-portal-git-zscertifications-spike-hashicorp.vercel.app/certifications/security-automation

PR #1430, opened 2022-12-08
feat: add Accordion component

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Implements an `Accordion` component.

## 🤷 Why

The `Accordion` component is needed for the new `/certifications` page work in #1404 .

## 🛠️ How

- Modifies `AccordionDisclosure` component to accept new `groupData` prop, and implements related styles
- ~Adds `DisclosureHeadingWrapper` component, and uses it in `AccordionDisclosure` via a `activatorHeadingLevel` prop. This allows for setting semantic heading levels in `Accordion` use cases, as recommended in the [Accordion pattern](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/)~
    - ~Note: this required modification of some styles, since element structure changes slightly when following the `Accordion` pattern. We now hide and show content using the `isOpen` value from the disclosure provider, rather than  directly relying on the `aria-expanded` property on the `button` element (which may now be wrapped in an `<h{n} />` element).~
    - EDIT: the `DisclosureHeadingWrapper` has been removed from `Accordion` for now. We're deferring until we can resolve accessibility concerns.
- Sets up an `Accordion` component which accepts an array of `items` and renders a group of `AccordionDisclosure` components

## 📸 Design Screenshots

🎨 [Figma link][figma]

> **Note**: design spec shown is from Certifications work, and displays a specific style of content within `Accordion`. The `Accordion` component as implemented in this PR accepts arbitrary content. Content within the Accordion elements will be specific to certifications.

| Design | Implementation |
| - | - |
| CleanShot 2022-12-08 at 20 57 38@2x | CleanShot 2022-12-08 at 20 56 23@2x |

## 🧪 Testing

- [ ] Visit the new [Accordion swingset page][swingset] and confirm the component looks and behaves as expected in all examples
    - In all examples on the Swingset page, each `AccordionDisclosure` `button` should be wrapped in an appropriate `<h{n} />` heading level
    - When nested, as with the existing `AccordionDisclosure`, the `box-shadow` does not change on hover.

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/1201987349274776/1201779497881404/f
[figma]: https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=16598%3A247078&t=lcQilHOMRgwRjXCL-1
[preview]: https://dev-portal-git-zsadd-accordion-component-hashicorp.vercel.app/
[swingset]: https://dev-portal-git-zsadd-accordion-component-hashicorp.vercel.app/swingset/components/accordion


zchsh:
@kendallstrautman @ashleemboyer This PR implements the `Accordion` component.

I've tried to work off the existing `AccordionDisclosure` component, to ensure the work done for the styling, focus states, and keyboard interactions can be shared, while also trying to match the [Accordion pattern](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/). As a result, there are some changes to `AccordionDisclosure` and the underlying `DisclosureActivator` in particular to meet the `<h{n} />` wrapping of `button` as in the recommended pattern.

Main intent here is to get this implemented for Certifications work. I've rebased #1404 onto this PR, so you can see the `Accordion` in use on [the Certifications preview link](https://dev-portal-git-zscertifications-spike-hashicorp.vercel.app/certifications).

Open to any & all feedback 🙇

src/components/disclosure/components/disclosure-activator/components/heading-wrapper/index.tsx
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Given a heading level,
+ * Return a heading component at that level that will wrap children
+ * without adding styles.
+ */
+export function getHeadingWrapper(level: HeadingLevel) {
+	const Heading = level as React.ElementType
+	// eslint-disable-next-line react/display-name
+	return ({ children }: { children: ReactNode }) => (
+		<Heading className={s.headingMarginReset}>{children}</Heading>
```
kendallstrautman, line 14: 💭 [You could use the `Heading` component](https://github.com/hashicorp/dev-portal/blob/main/src/components/heading/types.ts#L19) and just pass a number for the level

src/components/disclosure/components/disclosure-activator/types.ts
```diff
@@ -28,6 +29,15 @@ interface DisclosureActivatorProps {
...
 	'data-heap-track'?: string
+
+	/**
+	 * Optional heading level to contain the `button` element for the
+	 * DisclosureActivator. This is intended for use in establishing
+	 * semantic meaning for the disclosure section, particularly when
+	 * used in an Accordion design pattern. Ref:
+	 * https://www.w3.org/WAI/ARIA/apg/example-index/accordion/accordion.html
+	 */
+	headingLevel?: HeadingLevel
```
kendallstrautman, line 40: nice addition! 👍🏻

src/components/accordion-disclosure/index.tsx
```diff
@@ -11,18 +11,39 @@ import {
...
 const AccordionDisclosure = ({
 	children,
 	description,
 	initialOpen,
 	title,
+	isFirstItem,
+	isLastItem,
+	activatorHeadingLevel,
 }: AccordionDisclosureProps) => {
+	const [hoverRef, isHovered] = useHover<HTMLButtonElement>()
```
kendallstrautman, line 25: Do we have to use the hook because it's hard to capture `:hover` on the right component?

zchsh: Yep, exactly!

When `button` is hovered, I think we want the box-shadow on the enclosing section to change. So `:hover` won't work in CSS since we could only target elements nested within `button`.

For context, the previous implementation had a box-shadow on both `button` and the enclosing section. This kind of worked, but had two issues IMO:
- no hover style when open
- box-shadow from `button` creates an intersecting line with content, which looks a little odd during the animation / hover change as the content opens but the box-shadow on button is still present


https://user-images.githubusercontent.com/4624598/206761864-d0db93b4-0015-48b0-82be-03db61ac0150.mp4

src/components/accordion-disclosure/types.ts
```diff
@@ -28,4 +31,26 @@ export interface AccordionDisclosureProps {
...
+	 */
+	isFirstItem?: boolean
+
+	/**
+	 * Whether this disclosure is the last in a group of similar items.
+	 * If `true`, specific styles for the last item in a group will be applied.
+	 * If `false`, then styles will be applied assuming there is a next item.
+	 * If this prop & `isFirstItem` are `undefined`, no group styles are used.
+	 */
+	isLastItem?: boolean
```
kendallstrautman, line 49: 💭 What if we put these in a `group` prop? 

```
group?: {
  numItems: number
  currentIndex: number
}
```
Or we could pass the `isFirst` `isLast` in the group prop instead of the num / index. Then we are explicitly stating that the item is in a group as a prop instead of inferring whether its in a group from `isFirst` or `isLast`.

src/components/accordion-disclosure/index.tsx
kendallstrautman: okay that makes sense! I like that we have over styles when opened now too. Maybe we could leave a comment here for why we're doing hover like this as a breadcrumb.

kendallstrautman (approved):
Overall looks good to me! Just a few minor comments / suggestions 🚀

src/components/disclosure/components/disclosure-activator/components/heading-wrapper/index.tsx
zchsh: I think that'd add classes & styles that we might not want though 🤔 Could pass a `className` to reset them, similar to how `margin` is being reset here, but figured might be easier to not have to worry about the CSS differences.

I'll add a comment here with the rationale! Definitely happy to adjust if we feel strongly about using the existing `Heading` component for this use case, it certainly seems possible 👍

src/components/accordion-disclosure/index.tsx
zchsh: great call on the breadcrumb comment 💯  ✅ added that in!

src/components/accordion-disclosure/types.ts
zchsh: I dig this! 👍 will update...

zchsh: ✅ updated! I went with `groupData` but otherwise kept things exactly as suggested. Definitely feels cleaner, and dig having the `isFirstItem` & `isLastItem` logic within the component 👍

src/components/disclosure/components/disclosure-activator/components/heading-wrapper/index.tsx
kendallstrautman: Oh okay if the styles don't match up then disregard! 👍🏻

src/components/disclosure/disclosure.module.css
```diff
@@ -19,17 +19,18 @@
...
+*     container is the sibling element immediately after the button. It could
+*     also allow us to use the aria-expanded attribute as a CSS selector.
+*     However, when wrapping the button element in a heading element to meet
+*     the Accordion pattern, the sibling selector no longer works.
 *
 * ref: https://adrianroselli.com/2020/05/disclosure-widgets.html#CSS
 ***
 */
-.activator[aria-expanded='false'] + .contentContainer {
+.closedContentContainer {
```
ashleemboyer, line 34: 🚧 We should retain applying styles using selectors here. Their purpose is to make sure styles are only applied when the programmatic state & DOM order are correct AKA when accessibility is coded correctly.

src/components/accordion/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+							numItems: itemsCount,
+							currentIndex: idx,
+						}}
+						activatorHeadingLevel={activatorHeadingLevel}
+					>
+						{content}
+					</AccordionDisclosure>
+				)
+			})}
+		</div>
```
ashleemboyer, line 32: I think we need some tweaks to the semantics here so that an `Accordion` is conveyed as a group. I think there's also a lot of value in supporting `aria-label`/`aria-labelledby`, but that's a bit trickier when it comes to exposing this as an authoring component, so I think we can iterate on that part later.

An easy-win for achieving the group semantics, while also leaving room for `aria-label`/`aria-labelledby` later, feels like using `<ul>` and `<li>` elements. What do you think?

src/components/accordion-disclosure/index.tsx
```diff
@@ -11,18 +12,48 @@ import {
...
+	 */
+	const [hoverRef, isHovered] = useHover<HTMLButtonElement>()
 	const isNested = useAccordionDisclosureContext()
+
+	/**
+	 * Set up variables related to group styling.
+	 */
+	const isGroupedItem = typeof groupData !== 'undefined'
+	const isFirstItem = groupData?.currentIndex === 0
+	const isLastItem = groupData?.currentIndex === groupData?.numItems - 1
```
ashleemboyer, line 42: I think we should use CSS selectors for the first/last item styles, since this purely has to do with styles instead of functionality. The JS approach also requires that the correct data is passed in for each of these components, whereas the CSS approach would automatically handle styles based on placement and be less work on the DX side.

src/components/accordion-disclosure/index.tsx
```diff
@@ -11,18 +12,48 @@ import {
...
+	/**
+	 * Note: useHover is used here to allow us to change the styling
+	 * of the outer `root` container component when the `<button />`
+	 * within, which is the functional interactive element, is hovered.
+	 *
+	 * Since we're changing styles on an ancestor node of the `<button />`
+	 * whose `:hover` state we want to track, this is not as feasible
+	 * with pure CSS.
+	 */
+	const [hoverRef, isHovered] = useHover<HTMLButtonElement>()
```
ashleemboyer, line 34: I think we should go for a CSS approach here before adding a `useHover` hook that's purely for changing styles and not adding component functionality. What we want to do should be possible with some rearrangement of the HTML elements. Rather than applying some styles to the parent of `<button>`, we could add a sibling element next to it that we apply styles to when `<button>` is hovered. It would look like a container, but really be an absolutely positioned element that's behind `<button>`.

src/components/accordion-disclosure/index.tsx
```diff
@@ -32,22 +63,25 @@ const AccordionDisclosure = ({
...
-					data-heap-track="accordion-disclosure-activator"
-				>
-					<span className={s.labelContainer}>
-						<Text asElement="span" className={s.title} weight="semibold">
-							{title}
-						</Text>
-						{description && (
-							<Text asElement="span" className={s.description} size={200}>
-								{description}
+				<DisclosureHeadingWrapper level={activatorHeadingLevel}>
```
ashleemboyer, line 66: 🚧 Just reiterating a comment I made in Slack earlier -- we should wait to add heading elements to this HTML until we can 100% ensure that the headings rendered to not break the heading outline of a page.

ashleemboyer (dismissed):

src/components/accordion-disclosure/index.tsx
zchsh: For context, I think wrapping `button` in a heading element is necessary to match [the recommended Accordion pattern](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/).

I think it might also be worth clarifying that with `DisclosureHeadingWrapper` as in this PR, `button` will be wrapped in a heading element _only if_ `activatorHeadingLevel` is provided. So for example on the [Boundary API docs pages](https://dev-portal-git-zsadd-accordion-component-hashicorp.vercel.app/boundary/api-docs/account-service), and anywhere else `AccordionDisclosure` might be used without an `activatorHeadingLevel` prop explicitly provided, `button` will _not_ be wrapped in a heading element - we'll retain the same outline as we have upstream on `main`.

I think we can 100% ensure that headings will not break the outline _unless_ we explicitly provide an `activatorHeadingLevel` that is incorrect. This might not be a proper 100% guarantee, but it's as close as I can imagine use getting in a pragmatic way.

Does that feel like an acceptable solution for now? Or would there be a change here that we feel like would decrease the chance of breaking heading outlines on a page?

src/components/accordion/index.tsx
zchsh: I don't have a strong opinion on using `<ul>` & `<li>`! My understanding was that the use of headings would create the correct semantics to match [the Accordion pattern](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/).

For adding `<ul>` & `<li />`, would the implication here be to make `s.root` here a `<ul>` and then wrap each `AccordionDisclosure` in an `<li />`?

src/components/accordion-disclosure/index.tsx
ashleemboyer: A heading element is not required to wrap the `AccordionDisclosure` `<button>`. I understand that the heading only renders with a certain prop, but I think adding this functionality before we can ensure that it doesn't break heading order in a page is taking on unnecessary tech debt that will impact us negatively later. We don't have anything automated in place to prevent broken heading orders, and the risk for breaking heading orders will increase when this component is eventually exposed as an authoring component. We should defer adding the ability to render a heading element until we can ensure accessibility won't be broken.

src/components/disclosure/disclosure.module.css
zchsh: It makes sense that we'd ideally want to do this!

When wrapping in a `<heading>` element, however, I think it's not feasible for us to use the `[aria-expanded='false']` sibling selector, as the `button` should be the only element within the `heading`. We could potentially set `aria-expanded ='false'` on the heading so that the style still works, but it seems like that might really be what we're looking for 🤔 

For clarity, the use of this selector is an extra guarantee, rather than a key part of the functionality, right? In that the DOM order can be correct without the use of this selector? It seems like we're already validating that the DOM order is correct using `validate-disclosure-children`, so given the difficulty in retaining the sibling selector when wrapping `button` in `<h{n} />` to match [the Accordion pattern](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/), my thought was this might be an acceptable compromise in order to get the functionality we need.

src/components/accordion-disclosure/index.tsx
zchsh: Ah, alright! I was going off of this section of the [Accordion pattern](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/):

> Each accordion header button is wrapped in an element with role [heading](https://w3c.github.io/aria/#heading) that has a value set for [aria-level](https://w3c.github.io/aria/#aria-level) that is appropriate for the information architecture of the page.

It looks like other parts of the `WAI-ARIA Roles, States, and Properties` are explicitly marked as optional, so my reading was that the heading wrapping around `button` was a required part of the accordion pattern. If our stance is that that's not the case, I can remove the heading-related code from this PR.

That being said, I don't really see how we can get a better guarantee on the semantics here - with this PR implementing the component only, and not using it anywhere, we're not exposing author control at all to the heading levels in use by the component. In Certifications work, I've been careful to make the heading level part of the code, that would have be changed in a PR in order to cause breakage.

Personally I think there's relatively low risk here to causing accessibility breakage; and implementing the heading logic seems like it might be a recommended part of the accordion pattern.

src/components/accordion-disclosure/index.tsx
zchsh: I like this suggestion! I tried working through it, but hit a few dead ends.

It doesn't seem feasible to me to write styles _within_ `AccordionDisclosure` for first & last styles. For styling the first element in a group, for example, as far as I know we can only target `:first-child` or `:first-of-type`, and the root element here is type `div`. I think we'd have many scenarios where, if we write styles using these selectors, we'd get false positives and see group-like styles in contexts where `AccordionDisclosure` is next to another `div`, for example. (We might be able to use sibling selectors for the middle elements reliably... but the first and last elements seem like a challenge!).

Another option might be reaching into the component and styling it from the outside. We could potentially add styles to the `Accordion` component that target elements within `AccordionDisclosure`, and in that case `:first-child`, `:nth-child`, `:last-child` etc seem like they would work. But, we'd either have to use relatively brittle global-ish selectors (eg `.accordionRoot > div:first-child` or whatnot), or replace the `groupData` prop with a `styleClassName` prop or something that we'd tack onto the element rendering the box shadow.

Personally I think the `groupData` prop approach is slightly preferable to styling from the outside:
- We have a clear and limited set of styles for `AccordionDisclosure`, and the `groupData` prop provides clear documentation as to why these styles exist. Styles that affect the component elements are all contained within the component itself, rather than spread in multiple places in the app.
- When iterating on the component, we don't have to audit and account for all use cases that styling from the outside might be used for. For example, when passing a `styleClassName` prop, as we iterate on the `AccordionDisclosure` styles we might inadvertently change styles in a way that breaks or negatively affects the `styleClassName` styles being passed in by a consumer.

Let me know if the `groupData` prop here seems like an acceptable solution! If we feel like it's blocking, and we want to explore other approaches, I'd be very open to suggestions 🙇

src/components/accordion-disclosure/index.tsx
zchsh: I think this probably relates to the "heading" wrapper and logic. My understanding so far has been that we'll want to add the "heading" wrapper to match [the Accordion pattern](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/), which makes the sibling element selector approach less feasible. It would also seem to break the pattern, which mentions `button` should be the only child of the heading element:

> The button element is the only element inside the heading element. That is, if there are other visually persistent elements, they are not included inside the heading element. 

That being said, if we do feel we need to drop the `heading` wrapper, then I can see how using sibling selectors and an absolutely positioned element could become more feasible 👍

zchsh: Taking another look at this after removing the `heading` stuff. However, I'm still seeing some issues - namely that `validate-disclosure-children` expects exactly `2` children, so adding a absolutely positioned sibling element to `button` would require refactoring the validation function. As well, it would break the `.activator[aria-expanded='false'] + .contentContainer ` in `disclosure.module.css`.

I think the `useHover` hook here is a reliable way to achieve the same thing, while leaving our accessibility checks intact. Does `useHover` seem like an acceptable solution given those constraints?

src/components/accordion-disclosure/index.tsx
zchsh: ✅ have updated to remove all `heading` related logic, per discussion thread in Slack.

src/components/disclosure/disclosure.module.css
zchsh: ✅ update - since dropping the `heading` functionality, have retained this selector pattenr.

src/components/accordion-disclosure/index.tsx
ashleemboyer: I think there are some things we can try, and I'm seeing less value now in `validateDisclosureChildren`, but we can defer.

zchsh: Sounds good - will defer for now, very open to trying stuff later 💯

zchsh:
Based on discussions in Slack, my understanding is there are no blockers to merge here, though we do want to explore follow-up work on improving this component further 🤝 Merging this in, and looking forward to iterating 🙇

PR #1432, opened 2022-12-09
Fix DocsView integration of remark plugins

ashleemboyer:
## 🗒️ What

- Adds `contentType` option to `rewriteTutorialLinksPlugin`, which defaults to `'tutorials'`
- Adds `tutorialMap` option to `rewriteTutorialLinksPlugin` for testing purposes (to avoid executing `getTutorialMap` when it's not needed for a test)
- Adds test case that was failing before the addition of `contentType`
- Updates `DocsView` use of `rewriteTutorialLinksPlugin` to set `contentType` to `"docs"`

## 🤷 Why

- While testing out the docs content link rewrite script, I found that our `DocsView` doesn't correctly handle processing links like `[Waypoint](/waypoint)` in content. Without additional context, the `rewriteTutorialLinksPlugin` recognizes this as a link to legacy Learn's `/waypoint` page, so the link gets rewritten to `/waypoint/tutorials`. The `contentType` option gives the `rewriteTutorialLinksPlugin` the context it needs to know whether or not links like this should be updated in this way.

## 🧪 Testing

- I don't think there is anything to manually test here. I searched Sourcegraph for any of these links in docs content, and we don't have them yet. They will definitely appear after the docs content link rewrite migration though, when links like `https://waypointproject.io` are rewritten to `/waypoint`.

## 💭 Anything else?

- A longer term fix might be to merge the two plugins together, but that will require more time to refactor and test, so I opted not to do it here.


BRKalow (approved):
👏

src/views/docs-view/__tests__/multiple-remark-plugins.test.ts
```diff
@@ -0,0 +1,30 @@
...
+	let waypointProduct
+	let productUrlAdjuster
+
+	beforeAll(() => {
+		waypointProduct = cachedGetProductData('waypoint')
+		productUrlAdjuster = getProductUrlAdjuster(waypointProduct)
+	})
+
+	test('rewriteTutorialLinksPlugin, then remarkPluginAdjustLinkUrls', async () => {
+		const linkToProcess = '[Waypoint](/waypoint)'
```
kendallstrautman, line 17: ❓ are we still running this plugin tutorials content? I can't remember.

src/lib/remark-plugins/rewrite-tutorial-links/index.ts
```diff
@@ -30,29 +30,60 @@ import {
...
 ): string {
 	let newUrl
 
 	try {
 		const urlObject = new URL(url, 'https://learn.hashicorp.com')
 
-		const isExternalLearnLink = getIsExternalLearnLink(url)
+		let isExternalLearnLink
+		if (contentType === 'docs') {
+			isExternalLearnLink = !isRelativeUrl(url)
```
kendallstrautman, line 83: We should check if it _is_ an external learn link right? and not only that its not a relative docs link? for example, if theres a learn.hashicorp.com link in docs content, we still want to rewrite that. 

Also, if "https://kubernetes.io/docs/home/" were passed here, wouldn't `isExternalLearnLink` be true? Or are we only matching on 'ok' domains elsewhere?

src/views/docs-view/__tests__/multiple-remark-plugins.test.ts
ashleemboyer: `DocsView` still uses `rewriteTutorialLinksPlugin`, is that what you're asking?

src/lib/remark-plugins/rewrite-tutorial-links/index.ts
ashleemboyer: I think this'll need a slight adjustment, brb! Will add some more test cases to double-check

ashleemboyer: @kendallstrautman updated!

src/views/docs-view/__tests__/multiple-remark-plugins.test.ts
kendallstrautman: Oh no I was asking if we still run this plugin on tutorials content, but I remember [now that it was removed!](https://github.com/hashicorp/dev-portal/pull/1319)

kendallstrautman (approved):
Awesome, thank you!

src/lib/remark-plugins/rewrite-tutorial-links/index.ts
kendallstrautman: Looks good!

braydenlove:
No comment other than this is really cool!

src/lib/remark-plugins/rewrite-tutorial-links/index.ts
```diff
@@ -30,29 +30,60 @@ import {
...
-function handleRewriteTutorialsLink(node: Link | Definition) {
-	node.url = rewriteTutorialsLink(node.url, TUTORIAL_MAP)
+function handleRewriteTutorialsLink(
+	node: Link | Definition,
+	contentType: RewriteTutorialLinksPluginOptions['contentType'] = DEFAULT_CONTENT_TYPE
+) {
+	node.url = rewriteTutorialsLink(node.url, TUTORIAL_MAP, contentType)
+}
+
+const getIsRelativeUrl = (url) => {
```
zchsh, line 62: ooh i dig this `getIs<BooleanThing>` naming pattern! 🙌

PR #1434, opened 2022-12-09
DocsView fix: passing currentPath option to remarkPluginAdjustLinkUrls

ashleemboyer:
## 🗒️ What

- Adds `currentPath` option to `remarkPluginAdjustLinkUrls`
- Updates `remarkPluginAdjustLinkUrls` to adjust internal paths that do not start with `.` or `/`

## 🤷 Why

- `DocsView` server has not been correctly handling relative links like "variables/input" (which can be seen in the wild here: https://github.com/hashicorp/waypoint/blob/74830fc9bbdca0ee4091f9b97d8f0b04b66af5dc/website/content/docs/waypoint-hcl/variables/index.mdx?plain=1#L35-L37)

## 🧪 Testing

- [ ] Go to `/waypoint/docs/waypoint-hcl/variables` on the preview
- [ ] Scroll down to the Custom Variables section
- [ ] The `custom input variables` link should point to `/waypoint/docs/waypoint-hcl/variables/input` (compared to `https://www.waypointproject.io/variables/input` in prod)
- [ ] The `artifact` link should point to `/waypoint/docs/waypoint-hcl/variables/artifact` (compared to `https://www.waypointproject.io/variables/artifact` in prod)


src/lib/remark-plugin-adjust-link-urls.ts
```diff
@@ -1,15 +1,44 @@
...
+			 * the same thing as "./input" in the context of it's current page.
+			 */
+			if (
+				isInternalPath(node.url) &&
+				!node.url.startsWith('.') &&
+				!node.url.startsWith('/')
+			) {
+				urlToAdjust = path.join(
+					currentPath,
+					node.url.split('/').slice(1).join('/')
```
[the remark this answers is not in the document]
ashleemboyer: @BRKalow Hmmm true. What about something like this instead?

```typescript
// Pretend these are function inputs
// Also pretend we're only doing this for the URLs that don't start with `.`, `/`, `#`, `?`
const currentPath = '/docs/pathA/pathB/pathC'
const urlToChange = 'pathA/pathD'

// Derived from function inputs
const currentPathParts = currentPath.split('/')
const urlToChangeParts = urlToChange.split('/')

const newPathPrefixParts = []
currentPathParts.find(currentPathPart => {
  if (!currentPathPart) {
    return false
  }

  const matchesUrlToChangeFirstPart = currentPathPart === urlToChangeParts[0]
  if (!matchesUrlToChangeFirstPart) {
    newPathPrefixParts.push(currentPathPart)
  }
  
  return matchesUrlToChangeFirstPart
})

// Based on inputs, this should be `/docs/pathA/pathD`
// Also works for currentPath = 'docs/pathA/pathB/pathC' (no / at beginning)
const newPath = `/${newPathPrefixParts.join('/')}/${urlToChange}`
```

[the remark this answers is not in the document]
BRKalow: If we're assuming that the link is intended to be relative to the current dir, can we get by prepending `./`?

`variables/input` -> `./variables/input`

src/lib/remark-plugin-adjust-link-urls.ts
```diff
@@ -1,15 +1,43 @@
...
 		visit(tree, ['link', 'definition'], (node: Link) => {
-			node.url = urlAdjustFn(node.url)
+			let urlToAdjust = node.url
+
+			/**
+			 * This is to handle links like "variables/input", which are effectively
+			 * the same thing as "./input" in the context of its current page.
+			 */
+			const isInternalPath = getIsInternalPath(node.url)
+			const needsPreAdjustment = isInternalPath && /^[./?#]/.test(node.url)
```
kendallstrautman, line 31: Maybe I'm reading this wrong but this would only match `./` links right? so the 'variables/inputs' example wouldn't hit? Also is it trying to match anchor links?

ashleemboyer: Ahh I forgot the `!` 😂 so that it does _not_ match links that start with `.` (includes `./` and `../`), `/`, `?`, or `#`

src/lib/remark-plugin-adjust-link-urls.ts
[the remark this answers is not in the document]
ashleemboyer: @BRKalow so when I "pre-adjust" the URL to just prefix it with `./`, then the output URL is `/waypoint/docs/variables/input` (it should be `/waypoint/docs/waypoint-hcl/variables`)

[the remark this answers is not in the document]
BRKalow: @ashleemboyer if I manually change the link in the browser to `./variables/input`, it works how I would expect (I get routed to https://developer.hashicorp.com/waypoint/docs/waypoint-hcl/variables/input)

Screen Shot 2022-12-09 at 1 28 34 PM

src/lib/remark-plugin-adjust-link-urls.ts
ashleemboyer: This is fixed now!

src/lib/remark-plugin-adjust-link-urls.ts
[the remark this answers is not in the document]
ashleemboyer: Just updated the logic here: https://github.com/hashicorp/dev-portal/pull/1434/commits/1df4d1c5081512c935951818665c3b9c92b2fa27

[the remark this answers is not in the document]
BRKalow: Ooh interesting, yeah it looks like next makes it relative to to the page component 😞

src/lib/remark-plugin-adjust-link-urls.ts
kendallstrautman: oh awesome! 👍🏻 😆

src/lib/remark-plugin-adjust-link-urls.ts
```diff
@@ -1,15 +1,57 @@
...
 		visit(tree, ['link', 'definition'], (node: Link) => {
-			node.url = urlAdjustFn(node.url)
+			let urlToAdjust = node.url
+
+			/**
+			 * This is to handle links like "variables/input", which are effectively
+			 * the same thing as "./input" in the context of its current page.
+			 */
+			const isInternalPath = getIsInternalPath(node.url)
+			const needsPreAdjustment = isInternalPath && !/^[./?#]/.test(node.url)
```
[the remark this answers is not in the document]
kendallstrautman: After re-reading your PR description, it seems like it was intended to capture all those characters! so my b if linking to unhelpful regex playgrounds 😆 . But either way, documenting the intent of what its supposed to capture would help!

src/lib/remark-plugin-adjust-link-urls.ts
```diff
@@ -1,15 +1,57 @@
...
+					const matchesUrlToChangeFirstPart =
+						currentPathPart === urlToChangeParts[0]
+					if (!matchesUrlToChangeFirstPart) {
+						newPathPrefixParts.push(currentPathPart)
+					}
+
+					return matchesUrlToChangeFirstPart
+				})
+
+				urlToAdjust = `/${newPathPrefixParts.join('/')}/${node.url}`
```
kendallstrautman, line 51: I'm having trouble seeing all the cases we are trying to cover here. If its not to much to ask I think this would be a great candidate for some tests?

src/lib/remark-plugin-adjust-link-urls.ts
```diff
@@ -1,15 +1,57 @@
...
+			 * the same thing as "./input" in the context of its current page.
+			 */
+			const isInternalPath = getIsInternalPath(node.url)
+			const needsPreAdjustment = isInternalPath && !/^[./?#]/.test(node.url)
+			if (needsPreAdjustment) {
+				const currentPathParts = currentPath.split('/')
+				const urlToChangeParts = node.url.split('/')
+
+				const newPathPrefixParts = []
+				currentPathParts.find((currentPathPart) => {
```
zchsh, line 37: 💭  I'm finding the use of `find` here as a way to construct `newPathPrefixParts` until we hit an overlapping path segment threw me off a bit. I wonder if a `for` loop through `currentPathParts` with a `break` once `matchesUrlToChangeFirstPart` is found might be more clear?

src/lib/remark-plugin-adjust-link-urls.ts
```diff
@@ -1,15 +1,57 @@
...
 	urlAdjustFn: (url: string) => string
 }): Transformer => {
 	return function transformer(tree) {
 		visit(tree, ['link', 'definition'], (node: Link) => {
-			node.url = urlAdjustFn(node.url)
+			let urlToAdjust = node.url
+
+			/**
+			 * This is to handle links like "variables/input", which are effectively
+			 * the same thing as "./input" in the context of its current page.
```
zchsh, line 28: I'm not actually sure this is correct, but taking at shot at expanding the description here since it wasn't immediately clear to me what the `!/^[./?#]/.test(node.url)` regex was intended for either, or why `currentPath` was needed.

```suggestion
			 * the same thing as "./input" in the context of its current page.
			 *
			 * For context, links that do _not_ start with `.`, `/`, `?`, or `#`,
			 * such as "variables/input", are intended to be relative links.
			 *
			 * These links need to be prefixed with the `currentPath`, 
			 * as `next/link` will otherwise treat relative links as relative to the
			 * current page file, rather than relative to the current page URL.
			 * But, we're also not sure if the path parts in the relative link might
			 * overlap with path parts in the currentPath, so we iterate over
			 * `currentPathParts` until we find overlap, and use the non-overlapping
			 * parts to prefix the relative link.
```

src/lib/remark-plugin-adjust-link-urls.ts
zchsh: 💭 Had a related thought here - it seems like it's possible there might be weird edge cases where `currentPath` contains repeated parts - for example, https://developer.hashicorp.com/consul/docs/api-gateway/usage/usage. Maybe this relates to the broader idea that links like `variables/input` might be worth trying to filter out with content conformance, rather than correct at the this stage.

zchsh (approved):
If I'm following here, this does seem to fix links like `variables/input` as expected 👍 

🎨 +1 to Kendall's suggestion to add some comments to the `needsPreAdjustment`-related code in `./remark-plugin-adjust-links-urls`.

💭 All this has me thinking - is `variables/input` even a valid URL from an authoring standpoint? I don't quite understand how that would have functioned on the previous site 🤔 I wonder if it might be valuable to change course here in a way and use the `/^[./?#]/.test(node.url)` regex as part of content conformance, to enforce that links should start with `./`, `?`, `#`, or `/` only. This encroaches a bit on authoring practices (if `variables/input` is indeed valid, then we're asking folks to limit the types of valid links they write), but I feel like it might be a better balance overall of author effort vs link correction effort.

Either way, followed the testing steps and links seem fixed compared to prod, so would be comfortable shipping this 💯 Would love any other thoughts on whether we want this type of fix to be temporary, and to later try to shift this type of correction to the authoring level via content conformance.

ashleemboyer:
@kendallstrautman @zchsh I've abstracted the pre-adjustment code into a helper, added test cases for the helper, and added many comments on the helper as well. Was able to refactor the logic in a way that didn't need the regex so that it was a little more readable.

For extra context on many things mentioned:

- The regex was testing whether or not the url started with `.` (`./`, `../`, `../../`, etc.), `/` (top-level url), `?` (query string), and `#` (anchor link).
- The `variables/input` url was working on waypointproject.io, see this preview url for a [pre-premanent-redirect deployment](https://vercel.com/hashicorp/waypoint/7t2qvRZWwnEhVwTZLF33uHbP5Va3): https://waypoint-bgeyo4qm2-hashicorp.vercel.app/docs/waypoint-hcl/variables
  - Since this url worked before, I think we should retain it working on developer
- At least in the medium-term, as part of the docs content link rewrite project, all content repos are going to have a PR workflow that will check that link formats conform to [the new format](https://docs.google.com/document/d/1WaSyvoVPS-YCNiSPX0ynGpvc1gySEcpbrRoQVhimQCA/edit)
  - So links like `variables/input` will not be supported in content going forward, but we will still make it work for past versions

PR #1438, opened 2022-12-09
Integrations Pagination

BrandonRomano:
[🔍 Preview](https://dev-portal-git-brintegrations-pagination-2-hashicorp.vercel.app/vault/integrations)

### DIFF
--- a/src/views/product-integrations-landing/components/paginated-integrations-list/index.tsx
+++ b/src/views/product-integrations-landing/components/paginated-integrations-list/index.tsx
@@ -1,0 +1,119 @@
+import { IconArrowLeft16 } from '@hashicorp/flight-icons/svg-react/arrow-left-16'
+import { IconArrowRight16 } from '@hashicorp/flight-icons/svg-react/arrow-right-16'
+import classNames from 'classnames'
+import { Integration } from 'lib/integrations-api-client/integration'
+import { useState, useEffect } from 'react'
+import IntegrationsList from '../integrations-list'
+import s from './style.module.css'
+
+interface PaginatedIntegrationsListProps {
+	integrations: Array<Integration>
+	itemsPerPage?: number
+}
+
+export default function PaginatedIntegrationsList({
+	integrations,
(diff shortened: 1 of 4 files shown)


src/views/product-integrations-landing/components/paginated-integrations-list/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+		(a: Integration, b: Integration): number => {
+			if (a.name < b.name) {
+				return -1
+			} else if (a.name > b.name) {
+				return 1
+			} else {
+				return 0
+			}
+		}
+	)
```
thiskevinwang, line 29: Does the hardcoded alphabetical sorting prevent us from doing other sorting methods? — i.e. by "created date", by "most favorited", etc.?

Either way, no action needed, just asking for clarity, and to maybe leave a comment for the intent/expected limitations here.

src/views/product-integrations-landing/components/paginated-integrations-list/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+			)}
+		</div>
+	)
+}
+
+// =======================
+
+const COLLAPSED = '...'
+
+function paginatedArray(
```
thiskevinwang, line 67: Like `unflattenVariables`, i probably will gloss over this and trust that it works :D 

A quick top-level comment for expected output could be useful here 🙏

src/views/product-integrations-landing/components/paginated-integrations-list/index.tsx
BrandonRomano: It does, but this is the desired sort for now.

In the event in the future we want to add other sorting methods, we would change this here at this level.

src/views/product-integrations-landing/components/paginated-integrations-list/index.tsx
BrandonRomano: Good idea; I'll add a comment!

PR #1439, opened 2022-12-10
chore: adds 404 default to _error page

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksadds-404-default-error-hashicorp.vercel.app) 🔎
- [Asana task](https://app.asana.com/0/1199634971449915/1203192943606795) 🎟️

## 🗒️ What

Expanding off initial exploration https://github.com/hashicorp/dev-portal/pull/1414

This removes the global error boundary, so that we can use the `_error` page to render a view with the appropriate nav and also catch the client-side link issue. It looks like the `getServerSideProps` function wasn't running at all for the error version switcher component for the particular version switcher error. So even though the status code was set as 404 in the server props, this wasn't getting to the error switcher component. This code tests out setting the 404 default status in a few places to see if that fixes the issue. 

Also it sets `prefetch=false` to the link component on the version switcher so that we don't clog the network with known 404s when it attempts to prefetch the page data for all versions. I'm open to removing this change. 

## 🧪 Testing

- [ Visit this path](https://dev-portal-git-ksadds-404-default-error-hashicorp.vercel.app/vault/docs/enterprise/redundancy-zones)
- Hit v1.9.x in the version switcher dropdown
- The page should navigate to the correct 404 view


kendallstrautman:
🤔 hmm I wonder why all the pages increased. something to do with the error boundary removal?

src/pages/_app.tsx
```diff
@@ -105,32 +103,30 @@ export default function App({
 		<QueryClientProvider client={queryClient}>
 			<SSRProvider>
 				<CurrentContentTypeProvider currentContentType={currentContentType}>
-					<ErrorBoundary FallbackComponent={DevDotClient}>
```
kendallstrautman, line 108: The clientside 404 link error was always being caught by this. But the `DevDotClient` doesn't render a layout at all, so the deadend is not preferred UX. 

I'd love feedback here on the reason why we need this in addition to the custom _error page. It seems like next will automatically router everything to the custom error route in the cases we would need for this error boundary. But would love to learn more there. 

The `error-view-switcher` component has the logic to know how to handle the error view properly, so really we want to route folks there in this case.

src/views/error-view-switcher/index.tsx
```diff
@@ -19,7 +19,7 @@ const VERSION_PATTERN = /\/(?<version>v\d+[.]\d+[.](\d+|x))/
  * by matching VERSION_PATTERN in the page URL.
  */
 function ErrorView({
-	statusCode,
+	statusCode = 404,
```
kendallstrautman, line 22: This was the line that seemed to 'fix' everything when building locally.

src/components/dropdown-disclosure/components/list-item/index.tsx
```diff
@@ -112,7 +112,13 @@ const DropdownDisclosureLinkItem = ({
 }: DropdownDisclosureLinkItemProps) => {
 	return (
 		<DropdownDisclosureListItem>
-			<Link className={s.link} href={href} rel={rel} target={target}>
+			<Link
+				className={s.link}
+				href={href}
+				rel={rel}
+				target={target}
+				prefetch={false}
```
kendallstrautman, line 120: Adding prefetch as `false` so that we don't prefetch a bunch of data that doesn't exist for previous versions.

src/pages/_app.tsx
BRKalow: > ErrorBoundary is ideal for handling client-side errors. Without ErrorBoundary, it's hard (impossible?) to catch and log the details of a client-side error.

Yep. The distinction here is errors that are thrown while a component is rendering (as you're seeing in your test PR!).

> pages/404.js & pages/500.js are a way to create custom pages to handle those specific server-side errors
Yep!

> pages/_error.js can handle both client-side and server-side errors... but we won't have the error details we'd otherwise get with ErrorBoundary?

It depends on when the error happens. If we get an error back when fetching page data on navigation from the client, then yeah I think Next would route to our custom error page, but any errors that are thrown during render would not get handled without the error boundary. `_error` is synonymous with the `404.js` / `500.js` error pages, it just allows for SSRing the error page instead of relying on a statically generated one.

kendallstrautman: Interesting, so maybe we could make the error boundary for clientside errors 'smarter' in the same way the _error page is to render the right view.

kendallstrautman: Maybe we could create a unique error boundary for the different views — .io versus dev portal.

PR #1446, opened 2022-12-13
fix: ensure we don't error when setting the happykit cookie

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

When HappyKit returns, it's possible that `cookies` is `null`, in which case we are not handling this and silently erroring when we call `setHappyKitCookie()`. A side-effect seems to be that our experiments are not getting properly served. I've updated the logic to handle this.

We should no longer see treatments incorrectly flashing on our .io sites.

### DIFF
--- a/src/pages/_proxied-dot-io/vault/without-cta-links.tsx
+++ b/src/pages/_proxied-dot-io/vault/without-cta-links.tsx
@@ -50,7 +50,7 @@ export default function Homepage({ data }): React.ReactElement {
 	React.useEffect(() => {
 		abTestTrack({
 			type: 'Served',
-			test_name: 'CRO Vault CTA links 2022-10',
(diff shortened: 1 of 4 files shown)


src/pages/_proxied-dot-io/vault/without-cta-links.tsx
```diff
@@ -50,7 +50,7 @@ export default function Homepage({ data }): React.ReactElement {
 	React.useEffect(() => {
 		abTestTrack({
 			type: 'Served',
-			test_name: 'CRO Vault CTA links 2022-10',
+			test_name: 'CRO home hero CTA links 2022-10',
```
BRKalow, line 53: @alexcarpenter I fixed up the `test_name` value here for Vault to match the others.

alexcarpenter (approved):

PR #1449, opened 2022-12-13
Adds Integration Header

BrandonRomano:


src/views/product-integrations-landing/components/tag-list/index.tsx
```diff
@@ -1,6 +1,22 @@
...
+	Flag,
+	Integration,
+	Tier,
+} from 'lib/integrations-api-client/integration'
 import s from './style.module.css'
 
+export enum Size {
+	SMALL = 'small',
+	MEDIUM = 'medium',
+}
```
thiskevinwang, line 18: Nit: This enum feels like an unnecessary hinderance on the consumer, since they need to import it too if they want to avoid type errors with the `TagListProps.size`.

A more friendly interface, (IMO and especially for components), would be to make this a string union:

```typescript
export type Size = "small" | "medium"
```

I think the compile-time type safety is comparable as well

thiskevinwang (approved):
LGTM! 
Left an optional nit about favoring `union` over `enum` for `tag-list`

src/views/product-integrations-landing/components/tag-list/index.tsx
BrandonRomano: Oh yeah I'll adjust!

PR #1447, opened 2022-12-13
feat: certifications assembly

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Implements new Certifications pages. Specifically:

- [/certifications][/certifications]
- [/certifications/infrastructure-automation][/certifications/infrastructure-automation]
- [/certifications/networking-automation][/certifications/networking-automation]
- [/certifications/security-automation][/certifications/security-automation]

## 🛠️ How

> **Note**: This PR is an assembly branch. Work done in this PR can be examined in more detail in each of the PRs onto this assembly branch.

- https://github.com/hashicorp/dev-portal/pull/1448
- https://github.com/hashicorp/dev-portal/pull/1450
- https://github.com/hashicorp/dev-portal/pull/1452
- https://github.com/hashicorp/dev-portal/pull/1461
- https://github.com/hashicorp/dev-portal/pull/1462
- https://github.com/hashicorp/dev-portal/pull/1463
- https://github.com/hashicorp/dev-portal/pull/1467
- https://github.com/hashicorp/dev-portal/pull/1464
- https://github.com/hashicorp/dev-portal/pull/1469
- https://github.com/hashicorp/dev-portal/pull/1470
- https://github.com/hashicorp/dev-portal/pull/1472
- https://github.com/hashicorp/dev-portal/pull/1473
- https://github.com/hashicorp/dev-portal/pull/1475
- https://github.com/hashicorp/dev-portal/pull/1477
- https://github.com/hashicorp/dev-portal/pull/1478
- https://github.com/hashicorp/dev-portal/pull/1480
- https://github.com/hashicorp/dev-portal/pull/1481
- https://github.com/hashicorp/dev-portal/pull/1482
- https://github.com/hashicorp/dev-portal/pull/1485
- https://github.com/hashicorp/dev-portal/pull/1486
- https://github.com/hashicorp/dev-portal/pull/1487

## 📸 Design Screenshots

The assembly PRs listed above contain detailed design screenshots at each stage of development. The screenshots below show the work currently on this PR.

<details>
<summary>Expand for a table of screenshots</summary>

> **Note**: Pixel width columns refer to viewport widths at which the screenshot was taken.

| Page | 320px | 800px | 1920px |
| - | - | - | - |
| [/certifications][/certifications] | landing-320 | landing-800 | landing-1920 |
[...]

PR #1448, opened 2022-12-13
feat: stub content and getStatic functions

zchsh:
> **Note**: 🏗 This PR targets the Certifications assembly branch. See #1447 for details.

## 🗒️ What

This PR stubs in content `.json` files, `getStaticPaths`, `getStaticProps`, and related utilities for the Certifications pages.

## 🛠️ How

- Adds incomplete placeholder content
- Adds `zod` schemas to validate content
    - Schemas and content will be expanded in subsequent assembly PRs, as sections are added
- Adds content utilities, for reading local file lists for `getStaticPaths`
- Stubs in `getStaticProps` for landing view
- Implements `getStaticPaths` and stubs in `getStaticProps` for landing view
- Renders placeholder content on the landing page and on each program page

## 🔗 Page Links

- [/certifications][/certifications]
- [/certifications/infrastructure-automation][/certifications/infrastructure-automation]
- [/certifications/networking-automation][/certifications/networking-automation]
- [/certifications/security-automation][/certifications/security-automation]

## 🧪 Testing

- [ ] Visit each of the pages listed above.
    - Each page should render without error
    - Each page should display placeholder content from `.json` files

[preview]: https://dev-portal-git-zscertifications-content-base-hashicorp.vercel.app/
[/certifications]: https://dev-portal-git-zscertifications-content-base-hashicorp.vercel.app/certifications
[/certifications/infrastructure-automation]: https://dev-portal-git-zscertifications-content-base-hashicorp.vercel.app/certifications/infrastructure-automation
[/certifications/networking-automation]: https://dev-portal-git-zscertifications-content-base-hashicorp.vercel.app/certifications/networking-automation
[/certifications/security-automation]: https://dev-portal-git-zscertifications-content-base-hashicorp.vercel.app/certifications/security-automation

PR #1450, opened 2022-12-13
feat: certification base layout and hero

zchsh:
> **Note**: 🏗 This PR targets the Certifications assembly branch. See #1447 for details.

## 🗒️ What

This PR implements hero components for the new certifications pages.

## 🛠️ How

- Adds `CertificationsMaxWidth` component, as a re-usable container
- Adds `CertificationsHero` component, for re-use on the landing page and program pages
- Adds `LandingHero` to the landing view, with specific background & image assets
- Adds `ProgramHero` to the program view, with specific background assets per program
     - Note that in Figma, there is ongoing work to refine the hero backgrounds for program pages. Updated assets will be added in subsequent PRs, once they've been finalized by design.
- Adds `BaseNewLayout` to both the landing and program views

## 🔗 Page Links

- [/certifications][/certifications]
- [/certifications/infrastructure-automation][/certifications/infrastructure-automation]
- [/certifications/networking-automation][/certifications/networking-automation]
- [/certifications/security-automation][/certifications/security-automation]

## 🧪 Testing

- [ ] Visit each of the pages listed above.
    - Each page should now render with the base Dev Dot layout
    - Each page should render a hero component, with content from `.json` content files
    - The hero component should look and function as expected across viewports

[preview]: https://dev-portal-git-zscertifications-hero-components-hashicorp.vercel.app/
[/certifications]: https://dev-portal-git-zscertifications-hero-components-hashicorp.vercel.app/certifications
[/certifications/infrastructure-automation]: https://dev-portal-git-zscertifications-hero-components-hashicorp.vercel.app/certifications/infrastructure-automation
[/certifications/networking-automation]: https://dev-portal-git-zscertifications-hero-components-hashicorp.vercel.app/certifications/networking-automation
[/certifications/security-automation]: https://dev-portal-git-zscertifications-hero-components-hashicorp.vercel.app/certifications/security-automation


src/views/certifications/views/[slug]/components/program-hero/program-hero.module.css
```diff
@@ -0,0 +1,26 @@
+.heroBackground {
+	background: var(--background-image);
+	background-position: top center;
+	background-repeat: no-repeat;
+	background-size: cover;
+	bottom: -10%;
+	filter: blur(8px);
```
Comment (zchsh) [inline src/views/certifications/views/[slug]/components/program-hero/program-hero.module.css:7]: We use small dimension images and blur them rather than exporting the blurred gradient images at full size. This lets us greatly reduce the file size of the images. Note: this might change in a future PR, pending ongoing work to refine the hero appearance on the design side.

zchsh:
@kendallstrautman This PR implements the hero components for Certifications pages.

> **Note**: this PR builds on the content stubbed in in #1448. Might be easiest to review #1448 before this PR.

src/views/certifications/components/certifications-max-width/index.tsx
```diff
@@ -0,0 +1,13 @@
...
+import s from './certifications-max-width.module.css'
+
+/**
+ * Re-usable max-width container for Certifications views.
+ *
+ * Note that this corresponds to the Home page layout as well, so this
+ * component could potentially be lifted further and re-used on the homepage.
+ * For now, intent is to keep the scope to Certifications implementation.
+ */
+export function CertificationsMaxWidth({ children }: { children: ReactNode }) {
```
kendallstrautman, line 11: I like this pattern! Good idea to reuse on the homepage.

src/views/certifications/views/[slug]/components/program-hero/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+}: {
+	heading: string
+	description: string
+	slug: ProgramSlug
+}) {
+	return (
+		<CertificationsHero
+			heading={heading}
+			description={description}
+			foreground={slug === 'security-automation' ? 'dark' : 'light'}
```
Comment (kendallstrautman) [inline src/views/certifications/views/[slug]/components/program-hero/index.tsx:21]: 💭 Future thought: I wonder if this setting could be data driven, in case we update the backgrounds in the future. Maybe we could use an enum or const above in the short term?

```
enum ProgramForeground {
	'security-automation' = 'dark',
  //... etc
}
```

src/views/certifications/views/[slug]/components/program-hero/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { CertificationsHero } from 'views/certifications/components'
+import { ProgramSlug } from 'views/certifications/types'
+// Styles
+import s from './program-hero.module.css'
+
+function ProgramHero({
+	heading,
+	description,
+	slug,
+}: {
```
Comment (kendallstrautman) [inline src/views/certifications/views/[slug]/components/program-hero/index.tsx:12]: Can we extract this into an interface?

src/views/certifications/views/[slug]/components/program-hero/index.tsx
Comment (kendallstrautman) [reply src/views/certifications/views/[slug]/components/program-hero/index.tsx:21]: We could also set a default in case a new slug type is introduced?

src/views/certifications/types.ts
```diff
@@ -1,5 +1,14 @@
 import type { RawCertificationProgram } from './content/schemas/certification-program'
 
+/**
+ * Program slugs are mainly used for stylistic tweaks. This type, and related
+ * components,  will need to expanded if additional programs are added.
+ */
+export type ProgramSlug =
```
kendallstrautman, line 7: 💭 follow up thought from my previous PR comment. If we want the slugs to be editable we may want to make this just a `string` to be more flexible. But not sure if we want to open up the slugs for editing.

src/views/certifications/views/[slug]/components/program-hero/program-hero.module.css
Comment (kendallstrautman) [reply src/views/certifications/views/[slug]/components/program-hero/program-hero.module.css:7]: ooooo ✨ v smart 👏🏻

src/views/certifications/types.ts
kendallstrautman: I see now why we want to type this since its used to generate css classes. So no need for action.

src/views/certifications/views/[slug]/server.ts
```diff
@@ -3,13 +3,16 @@ import {
...
+import { ProgramSlug } from 'views/certifications/types'
 import { CertificationProgramViewProps } from './types'
 
 export async function getStaticProps({
 	params: { slug: rawSlug },
 }: GetStaticPropsContext): Promise<{ props: CertificationProgramViewProps }> {
 	// Get the `slug` as a string
-	const slug = Array.isArray(rawSlug) ? rawSlug.join('/') : rawSlug
+	const slug = (
+		Array.isArray(rawSlug) ? rawSlug.join('/') : rawSlug
```
Comment (kendallstrautman) [inline src/views/certifications/views/[slug]/server.ts:14]: Whats the case where this would be an array? I thought only catch-all routes had the potential to be an array?

Comment (kendallstrautman) [reply src/views/certifications/views/[slug]/server.ts:14]: Or is this just a typescript weirdness thing?

src/views/certifications/views/[slug]/components/program-hero/program-hero.module.css
```diff
@@ -0,0 +1,26 @@
...
+	left: -10%;
+	overflow: hidden;
+	position: absolute;
+	right: -10%;
+	top: -10%;
+	z-index: -1;
+
+	&.theme-infrastructure-automation {
+		--background-image: url('./assets/infrastructure-automation.jpg');
```
Comment (kendallstrautman) [inline src/views/certifications/views/[slug]/components/program-hero/program-hero.module.css:16]: Screen Shot 2022-12-13 at 2 37 35 PM

Maybe these are recent changes but it looks like the backgrounds are updated with patterns on top of the gradient?

kendallstrautman (approved):
Overall looking good! I checked the component against the designs. Just a few non-blocking comments / questions 💯

src/views/certifications/views/[slug]/components/program-hero/index.tsx
Comment (zchsh) [reply src/views/certifications/views/[slug]/components/program-hero/index.tsx:21]: I dig this! 💯 Updated with an `enum` as suggested, and added a lil docs comment on how the `foreground` and `theme-<slug>` background classes relate.

We have some defaults within `CertificationsHero` as well - dark text, white background. So in the case a new program page with a new`slug` is added, the hero will default to something like this:

CleanShot 2022-12-14 at 11 35 14@2x

Which obvs means we'll want to remember to add more background images to this component if we add a new page... but the default will at least kind of work & hit the color contrast requirements 👍

Comment (zchsh) [reply src/views/certifications/views/[slug]/components/program-hero/index.tsx:21]: Ooh came up with something slightly different - rather than the enum, I've now got this in place:

```
const programHeroForeground: Record<
	ProgramSlug,
	CertificationsHeroProps['foreground']
> = {
	'security-automation': 'dark',
	'infrastructure-automation': 'light',
	'networking-automation': 'light',
}
```

This should ensure that if a `ProgramSlug` is added, we'll get a TypeScript error unless we've explicitly handled the new `ProgramSlug` within this `ProgramHero` component.

src/views/certifications/views/[slug]/components/program-hero/index.tsx
Comment (zchsh) [reply src/views/certifications/views/[slug]/components/program-hero/index.tsx]: Absolutely! ✅ extracted to an adjacent `types.ts` file 👍

src/views/certifications/views/[slug]/server.ts
Comment (zchsh) [reply src/views/certifications/views/[slug]/server.ts:14]: TypeScript weirdness is a good summary I think 😬 But now I wonder whether there's something I can do with the `GetStaticPropsContext` here to mitigate this, slash signal that we're not in a catch-all route - cause +1 that this shouldn't have the potential to be an array 🤔

Comment (zchsh) [reply src/views/certifications/views/[slug]/server.ts]: Yep yep, I can pass `GetStaticPropsContext<{ slug: ProgramSlug }>` here, I guess to kind of encode the combined `getStaticPaths` and `[slug].tsx` page file setup that NextJS would otherwise be unaware of 👍 This feels way cleaner - thanks for flagging 🙇

src/views/certifications/views/[slug]/components/program-hero/program-hero.module.css
Comment (zchsh) [reply src/views/certifications/views/[slug]/components/program-hero/program-hero.module.css:16]: Great eye 👀 These are recent changes! I'm planning on updating the assets once everything is finalized in the design (I don't think it was yet when I opened this PR, I think it might have just been finalized this morning!)

zchsh:
@kendallstrautman Thanks for the feedback here - updated based on those suggestions! 💯 Going to merge this in, definitely happy to discuss further tweaks here and follow-up in further PRs. (There'll definitely be another PR to update the background assets once I have final exports for those!)

PR #1456, opened 2022-12-14
Product Landing Page navigational links

BrandonRomano:
Appropriately show the Integrations navigation items when enabled.

## Enabled
CleanShot 2022-12-13 at 16 03 20@2x

## Disabled
CleanShot 2022-12-13 at 16 04 04@2x

### DIFF
--- a/src/lib/enabled-integration-products.ts
+++ b/src/lib/enabled-integration-products.ts
@@ -1,0 +1,7 @@
+import { ProductSlug } from 'types/products'
+
+// The Products that have Integrations enabled
+export const ENABLED_INTEGRATION_PRODUCTS: Array<ProductSlug> = [
+	'waypoint',
+	'vault',
+]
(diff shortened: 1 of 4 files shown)


src/lib/enabled-integration-products.ts
```diff
@@ -0,0 +1,7 @@
+import { ProductSlug } from 'types/products'
+
+// The Products that have Integrations enabled
+export const ENABLED_INTEGRATION_PRODUCTS: Array<ProductSlug> = [
```
kendallstrautman, line 4: 💭 We could store this in the[ config data](https://github.com/hashicorp/dev-portal/blob/main/config/base.json) for dev & preview, it's made [accessible via `__config.<property>`](https://github.com/hashicorp/dev-portal/blob/main/next.config.js#L85)

thiskevinwang (approved):

src/lib/enabled-integration-products.ts
BrandonRomano: Oh neat I really like this.

I'll adjust this in the near future to use this strategy; I think I need to think some more about how I'd structure that / what our setups will be so I'll still merge this PR in here, but 100% will switch to this config strategy.

PR #1457, opened 2022-12-14
refactor(integrations): pagination improvements a11y

alexcarpenter:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-acproduct-integrations-paginati-d29047-hashicorp.vercel.app/vault/integrations) 🔎

## 🗒️ What

Refactor pagination to be keyboard accessible and add appropriate aria labels.

- Wrap in `nav` element with aria label "Pagination Navigation"
- Wrap elements in `ul` so VO will read amount of items
- Convert `a` to `button` elements to make them focusable
- Add aria-labels to buttons, example `1 of 10`
- Use disabled attribute vs class
- Use `aria-current="page"` on active element

https://user-images.githubusercontent.com/825855/207608164-bec60201-ee13-42c2-8ed4-81f96235d268.mp4

### DIFF
--- a/src/views/product-integrations-landing/components/paginated-integrations-list/index.tsx
+++ b/src/views/product-integrations-landing/components/paginated-integrations-list/index.tsx
@@ -148,59 +148,68 @@ function Paginator({
 }: PaginatorProps) {
 	paginatedArray(numberOfPages, currentPage)
 	return (
-		<div className={s.paginator}>
-			{/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
-			<a
-				className={classNames({ [s.disabled]: currentPage === 1 })}
-				onClick={(e) => {
-					e.preventDefault()
-					if (currentPage !== 1) {
-						onPageClicked(currentPage - 1)
-					}
-				}}
-			>
-				<IconArrowLeft16 />
-			</a>
+		<nav
+			className={s.paginator}
+			// role="navigation"
+			aria-label="Pagination Navigation"
+		>
+			<ul>
+				<li>
+					<button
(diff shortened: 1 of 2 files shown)


src/views/product-integrations-landing/components/paginated-integrations-list/index.tsx
```diff
@@ -148,59 +148,64 @@ function Paginator({
...
+						)
+					} else {
+						return (
+							<li key={i}>
+								<button
+									onClick={() => {
+										onPageClicked(pageNum as number)
+									}}
+									aria-current={currentPage === pageNum ? 'page' : undefined}
+									aria-label={`Page ${pageNum} of ${numberOfPages}`}
```
ashleemboyer, line 185: Setting the `aria-label` to something other than the visible label of the button will make the button unusable for voice control users. This is because voice commands are based off of the accessible name for an element, which in this case is computed using the `aria-label`. A voice control user would have to say something like "Click page 1 of 10" to activate the button, and they're likely to say "Click 1" based on the visible label. I think it's safe to not set an `aria-label` here since the context of total pages is close by.

ashleemboyer: I'm less sure about the next/previous page buttons. I think we'd need user testing to see what people would most commonly call them, but I think they're fine as is for now.

alexcarpenter: Just seeing these after I closed in favor or the other pagination component already in place, but looks like your comments apply there as well!

PR #1410, opened 2022-12-03
Add docs content link rewrite script

ashleemboyer:
## What

Adds the docs content link rewrite script and it's helper functions. Also adds a reusable workflow that other repos can call in order to check that their MDX content aligns with the new link format.

## Testing

See the workflow runs in Waypoint here:
https://github.com/hashicorp/waypoint/actions/workflows/check-legacy-links-format.yml

These tests are running against a draft PR that has had the rewrite script run on it locally, then pushed up on the branch.

## Anything else

The nav-data checking portion is currently commented out because we need more time in order to update and test our nav-data validation logic that lives in `react-components/docs-sidenav`.

PR #1465, opened 2022-12-15
Add e2e tests for docs content link rewrites

ashleemboyer:
## 🗒️ What

- Adds new spec file that checks all sidebar and main content area links are the same for **all** documentation pages on both a PR's preview URL and a main/prod preview URL.
  - It tests all pages rather than just changed ones because some of the changed MDX files are partials that are imported into other pages. Instead of writing a script that figures out all the pages partials are used on, we can just test all the documentation pages for a product. It takes a while to run, but it's ok since these tests will only run for one PR for each repo.
- Adds constants for sidebar and main element IDs, so those elements can be queried for in a stable way in the e2e tests
- Adds several script helpers for generating data for the new playwright suite, which needs a static file in order to generate all the dynamic `test` cases
- Adds two new npm scripts (2 are needed so the workflow will fail correctly if the test execution script fails):
  - `test:e2e:linkRewrites:loadData` - loads data for the new playwright suite
  - `test:e2e:linkRewrites` - executes the new playwright suite with the `ENABLE_DOCS_CONTENT_LINK_REWRITES_E2E` variable enabled
- Adds reusable workflow (`docs-content-link-rewrites-e2e`) that other repos can call to run the e2e tests
  - This workflow is only supposed to be called in the main migration PR, that's why the required branch name is `docs/amb.migrate-link-formats`

## 🧪 Testing

The main testing we have for this PR is that the workflows are called successfully, and the tests pass in this draft PR for migrating Waypoint's content links:

https://github.com/hashicorp/waypoint/pull/4277


scripts/docs-content-link-rewrites/helpers/normalize-remote-loader-slug.ts
```diff
@@ -0,0 +1,24 @@
...
+		}
+
+		const productData = cachedGetProductData(productSlug)
+		return !!productData.rootDocsPaths.find((rootDocsPath: RootDocsPath) => {
+			return remoteLoaderSlug === rootDocsPath.productSlugForLoader
+		})
+	})
+}
+
+export { normalizeRemoteLoaderSlug }
```
ashleemboyer, line 24: 🗒️ This file and the tests will be coming from #1410

.github/workflows/docs-content-link-rewrites-e2e.yml
```diff
@@ -0,0 +1,71 @@
...
+              commit_sha: process.env.COMMIT_SHA
+            })
+            return data && data[0].head.ref === "docs/amb.migrate-link-formats"
+
+      # @TODO remove `ref` setting before merging
+      - name: Check out dev-portal repo
+        uses: actions/checkout@v3
+        if: ${{ steps.should-run-for-branch.outputs.result == 'true' }}
+        with:
+          ref: amb.link-rewrites-e2e-tests
```
ashleemboyer, line 55: Will apply this suggestion after approval, and before merging

```suggestion
      - name: Check out dev-portal repo
        uses: actions/checkout@v3
        if: ${{ steps.should-run-for-branch.outputs.result == 'true' }}
```

.github/workflows/docs-content-link-rewrites-e2e.yml
```diff
@@ -0,0 +1,77 @@
...
+        uses: actions/checkout@v3
+        if: ${{ steps.should-run-for-branch.outputs.result == 'true' }}
+        with:
+          repository: hashicorp/dev-portal
+          path: dev-portal
+          ref: amb.link-rewrites-e2e-tests
+
+      - name: Change into cloned dev-portal directory
+        if: ${{ steps.should-run-for-branch.outputs.result == 'true' }}
+        run: cd dev-portal
```
ashleemboyer, line 61: Still testing these specific changes in https://github.com/hashicorp/waypoint/pull/4277, but the rest of the PR is ready for review!

BRKalow, line 61: By default, each step uses the default working directory. `cd`ing in a step will not impact subsequent steps AFAIK.

We can either not set a path when checking out on L56 so that the checkout is done in the current dir, or set `working-directory` on each subsequent step ([ref](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun))

thiskevinwang, line 61: I don't think this `cd` persists between steps (you'd have to call it within each step)
It's also favorable to use the [`working-directory`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#defaultsrun:~:text=Using%20the%20working%2Ddirectory%20keyword%2C%20you%20can%20specify%20the%20working%20directory%20of%20where%20to%20run%20the%20command.) field
```suggestion
```

.github/workflows/docs-content-link-rewrites-e2e.yml
```diff
@@ -0,0 +1,77 @@
...
+
+      - name: Install dev-portal deps
+        if: ${{ steps.should-run-for-branch.outputs.result == 'true' }}
```
thiskevinwang, line 64:
```suggestion
        working-directory: dev-portal
        if: ${{ steps.should-run-for-branch.outputs.result == 'true' }}
```

.github/workflows/docs-content-link-rewrites-e2e.yml
```diff
@@ -0,0 +1,77 @@
...
+        run: npm ci
+
+      - name: Install Playwright
+        if: ${{ steps.should-run-for-branch.outputs.result == 'true' }}
```
thiskevinwang, line 68:
```suggestion
        working-directory: dev-portal
        if: ${{ steps.should-run-for-branch.outputs.result == 'true' }}
```

.github/workflows/docs-content-link-rewrites-e2e.yml
```diff
@@ -0,0 +1,77 @@
...
+        run: npx playwright install --with-deps
+
+      - name: Load data for e2e tests
+        if: ${{ steps.should-run-for-branch.outputs.result == 'true' }}
```
thiskevinwang, line 72:
```suggestion
        working-directory: dev-portal
        if: ${{ steps.should-run-for-branch.outputs.result == 'true' }}
```

.github/workflows/docs-content-link-rewrites-e2e.yml
```diff
@@ -0,0 +1,77 @@
...
+        run: npm run test:e2e:linkRewrites:loadData
+
+      - name: Run e2e tests
+        if: ${{ steps.should-run-for-branch.outputs.result == 'true' }}
```
thiskevinwang, line 76:
```suggestion
        working-directory: dev-portal
        if: ${{ steps.should-run-for-branch.outputs.result == 'true' }}
```

BRKalow:
@ashleemboyer it looks like your new test file is getting picked up by this repo's playwright run. We'll most likely need to come up with a way to exclude it. We could use this config option: https://playwright.dev/docs/api/class-testconfig#test-config-test-ignore

ashleemboyer:
> @ashleemboyer it looks like your new test file is getting picked up by this repo's playwright run. We'll most likely need to come up with a way to exclude it. We could use this config option: https://playwright.dev/docs/api/class-testconfig#test-config-test-ignore

Well darn! That's what the env variable was supposed to help with, but it makes sense that I didn't see this issue locally since I have the generated file it's trying to import present. Going to try the ignore!

.github/workflows/docs-content-link-rewrites-e2e.yml
ashleemboyer: This makes sense! I wasn't 100% sure, good to know now. I'm going to try the working-directory suggestion since I'm pushing up changes anyways. :)

ashleemboyer: I think we can not set the path since only one repo is checked out here, I may have gotten overzealous setting the path! Pushed up changes to the Waypoint repo, so we should see if my latest changes here work

ashleemboyer: Forgot to remove the `cd` step until just now! 😅 Just pushed up changes in both repos, so _now_ we should see the tests run soon.

ashleemboyer: https://github.com/hashicorp/waypoint/actions/runs/3716771800/jobs/6303468834

ashleemboyer: Working great! 🤩

ashleemboyer: Ended up removing the `path` setting on the checkout step, and it's working! Going to resolve the other comments now :)

ashleemboyer:
> > @ashleemboyer it looks like your new test file is getting picked up by this repo's playwright run. We'll most likely need to come up with a way to exclude it. We could use this config option: https://playwright.dev/docs/api/class-testconfig#test-config-test-ignore
> 
> Well darn! That's what the env variable was supposed to help with, but it makes sense that I didn't see this issue locally since I have the generated file it's trying to import present. Going to try the ignore!

For posterity sake, we went with configuring separate projects! ✨ 
https://playwright.dev/docs/test-advanced#projects

BRKalow (approved):

PR #1468, opened 2022-12-15
[integrations] Improve markdown rendering

thiskevinwang:
https://dev-portal-git-kw-improve-markdown-rendering-hashicorp.vercel.app

## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🤷 Why

Consistent look/feel of code blocks and markdown in general

## 🛠️ How

This leverages `next-mdx-remote` and our typical suite of remark/rehype plugins and client-side react components

## 📸 Design Screenshots

CleanShot 2022-12-15 at 16 33 54@2x

## 💭 Anything else?

---
  - https://app.asana.com/0/0/1203548351777713

### DIFF
--- a/src/views/product-integration/components/component-tab-content/index.tsx
+++ b/src/views/product-integration/components/component-tab-content/index.tsx
@@ -25,7 +25,9 @@ export default function ComponentTabContent({
 	)
 	return (
 		<div className={s.componentTabContent}>
-			{component.readmeMdxSource ? readmeContent : null}
+			{component.readmeMdxSource ? (
+				<div className={s.mdx_wrapper}>{readmeContent}</div>
+			) : null}
 			{component.variable_groups.length ? (
 				<Tabs allowNestedStyles>
 					{component.variable_groups.map((variableGroup: VariableGroup) => {
--- a/src/views/product-integration/components/component-tab-content/index.tsx
+++ b/src/views/product-integration/components/component-tab-content/index.tsx
@@ -4,21 +4,28 @@ import {
 	Variable as ApiVariable,
 	VariableGroup,
 } from 'lib/integrations-api-client/release'
-import ReactMarkdown from 'react-markdown'
+import { MDXRemote, type MDXRemoteSerializeResult } from 'next-mdx-remote'
+import defaultMdxComponents from 'layouts/sidebar-sidecar/utils/_local_platform-docs-mdx'
 import SearchableVariableGroupList from '../searchable-variable-group-list'
 import { Variable } from '../variable-group-list'
 import s from './style.module.css'
 
 interface ComponentTabContentProps {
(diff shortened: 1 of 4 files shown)


src/views/product-integration/index.tsx
```diff
@@ -11,14 +11,15 @@ import { Integration } from 'lib/integrations-api-client/integration'
 import { Release, ReleaseComponent } from 'lib/integrations-api-client/release'
 import { useRouter } from 'next/router'
 import { useMemo } from 'react'
-import ReactMarkdown from 'react-markdown'
```
BrandonRomano, line 14: @thiskevinwang I think we can remove this from dependencies too.

This should be the only place we were using this & I added `react-markdown` in our integrations PR myself (wasn't already in DevDot).

src/views/product-integration/components/component-tab-content/index.tsx
```diff
@@ -4,21 +4,28 @@ import {
...
+	const readmeContent = (
+		<MDXRemote
+			{...component.readmeMdxSource}
+			components={defaultMdxComponents({})}
+		/>
+	)
 	return (
 		<div className={s.componentTabContent}>
-			{component.readme && <ReactMarkdown>{component.readme}</ReactMarkdown>}
+			{component.readmeMdxSource ? readmeContent : null}
```
BrandonRomano, line 28: There appears to be a margin-top here (I believe coming from the H2).

Is there something we can do about this (e.g. first child no margin top?)

CleanShot 2022-12-19 at 10 12 48@2x

BrandonRomano (changes requested):
- [ ] Can remove margin top on first element
- [ ] Let's remove the react-markdown dependency

src/views/product-integration/index.tsx
thiskevinwang: `react-markdown` is still used by `VariableGroupList` currently. My thought was to leave it for now, at least until Variable Groups' README content are processed by our `next-mdx-remote` pipeline as well.

It seemed like high-effort, low-reward, so I didn't do that part yet last week

BrandonRomano: Oh!  Yes okay -- let's do that my bad!

src/views/product-integration/components/component-tab-content/index.tsx
thiskevinwang: I'll see what I can do!

(note, it's being defined here https://github.com/hashicorp/dev-portal/blob/kw/improve-markdown-rendering/src/components/dev-dot-content/mdx-components/mdx-headings/mdx-headings.module.css#L44-L48, and applied when we opt into using `MdxH1`, `MdxH2`, etc for our markdown heading elements)

BrandonRomano (approved):

PR #1493, opened 2022-12-21
Multi Page Integrations

BrandonRomano:
Breaks up the Integration monolithic page that was previously driven by tabs into multiple pages.

This does not include all visual changes that we plan on making, but this PR handles the lower level adjustments to set this up.

This PR is getting quite big, so I'll polish out other visual adjustments in subsequent pull requests.

[🔍 Preview](https://dev-portal-git-brmulti-page-integrations-hashicorp.vercel.app/waypoint/integrations/docker)

## Overview Page
CleanShot 2022-12-22 at 10 13 05@2x

## Component Page
CleanShot 2022-12-22 at 10 14 13@2x


src/layouts/product-integration-layout/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+	const componentsWithPages: Array<ReleaseComponent> =
+		activeRelease.components.filter((rc: ReleaseComponent) => {
+			return rc.readme || rc.variable_groups.length
+		})
+
+	// Calculate the Sidebar
+	const sidebarNavDataLevels = [
+		generateTopLevelSidebarNavData(currentProduct.name),
+		generateProductLandingSidebarNavData(currentProduct),
+		// TODO: Generate the Integrations Landing Page Level
```
BrandonRomano, line 54: I actually want to do this one before merging this PR in here

src/lib/serialize-integration-markdown.ts
```diff
@@ -0,0 +1,22 @@
...
+	mdxOptions: {
+		remarkPlugins: [paragraphCustomAlerts, typography],
+		rehypePlugins: [
+			[rehypePrism, { ignoreMissing: true }],
+			rehypeSurfaceCodeNewlines,
+		],
+	},
+}
+
+export default async function serializeIntegrationMarkdown(
```
thiskevinwang, line 18:
```suggestion
nice, thank you for lifting this up into its own file :D 
```

BrandonRomano: Hmm I'm not sure about this change 😆

thiskevinwang: > _sir, that's not valid TypeScript_

BrandonRomano: We are now prompt developers, plug that into chatGPT and see what it does

src/pages/[productSlug]/integrations/[integrationSlug]/[integrationVersion]/components/[componentSlug]/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+	if (integrationVersion === integration.versions[0]) {
+		return {
+			redirect: {
+				destination: `/${product.slug}/integrations/${integration.slug}/latest/components/${releaseComponent.component.slug}`,
+				// Not permanent as a new release in the future will turn the
+				// latest release into an older release which should render!
+				permanent: false,
+			},
+		}
+	}
```
Comment (thiskevinwang) [inline src/pages/[productSlug]/integrations/[integrationSlug]/[integrationVersion]/components/[componentSlug]/index.tsx:96]: So we're always redirecting `<latest_version>` to `latest`

I wonder if we actually want this the other way around?

---

Idea comes from
- https://github.com/hashicorp/waypoint/releases/latest
  -  which redirects to https://github.com/hashicorp/waypoint/releases/tag/v0.10.5

Comment (BrandonRomano) [reply src/pages/[productSlug]/integrations/[integrationSlug]/[integrationVersion]/components/[componentSlug]/index.tsx:96]: Oh interesting..

I was thinking we keep `latest` stable so the SEO juices flow into a singular place consistently.

Comment (BrandonRomano) [reply src/pages/[productSlug]/integrations/[integrationSlug]/[integrationVersion]/components/[componentSlug]/index.tsx:96]: I don't know if that is a sound decision though.

Comment (BrandonRomano) [reply src/pages/[productSlug]/integrations/[integrationSlug]/[integrationVersion]/components/[componentSlug]/index.tsx:96]: Thinking about this some more out loud:

Looking at the GitHub context, I think that GitHub would want each version to be identified as **a totally new thing with different content**, as each release will have notes on what was changed from the previous release (and it's not like something that's being slowly incrementally updated.  That being said, I think it makes sense to have *each release* be its own page.

Our use case is slightly different, as our `releases` are more like git commits, slowly incrementing the content over time.  I think our use case here is closer to how our default docs pages latest version get a URL without the version in it (e.g. https://developer.hashicorp.com/packer/docs/templates, vs https://developer.hashicorp.com/packer/docs/v1.7.x/templates).

If we included the version in our docs, I believe that our SEO would be implicated each time we cut a new version.  (Maybe that could be avoided if we used a canonical URL pointing to the new place), but I think it might confuse google into thinking it's a separate document.

Comment (thiskevinwang) [reply src/pages/[productSlug]/integrations/[integrationSlug]/[integrationVersion]/components/[componentSlug]/index.tsx:96]: I also think we should keep stable URLs (ex. `/waypoint/integrations/docker/0.10.3/components/builder`) stable.

Vanity urls (`/waypoint/integrations/docker` or `/waypoint/integrations/docker/latest`) can have the underlying content/data shift.

Comment (thiskevinwang) [reply src/pages/[productSlug]/integrations/[integrationSlug]/[integrationVersion]/components/[componentSlug]/index.tsx:96]: > If we included the version in our docs, I believe that our SEO would be implicated each time we cut a new version. (Maybe that could be avoided if we used a canonical URL pointing to the new place), but I think it might confuse google into thinking it's a separate document.

Hmm, I cannot say I know what Google/Crawlers do in this case `:homer:`, so I defer to your call!

Comment (BrandonRomano) [reply src/pages/[productSlug]/integrations/[integrationSlug]/[integrationVersion]/components/[componentSlug]/index.tsx:96]: I can't say for certain either TBH!  Let's discuss a bit more after you wrap up your review on this!

src/pages/[productSlug]/integrations/[integrationSlug]/[integrationVersion]/components/[componentSlug]/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+				integration,
+				activeRelease,
+				releaseComponent,
+				true
+			),
+		},
+	}
+}
+
+export function integrationComponentBreadcrumbLinks(
```
Comment (thiskevinwang) [inline src/pages/[productSlug]/integrations/[integrationSlug]/[integrationVersion]/components/[componentSlug]/index.tsx:127]: 💭 No need to export this right?

```suggestion
function integrationComponentBreadcrumbLinks(
```

Comment (BrandonRomano) [reply src/pages/[productSlug]/integrations/[integrationSlug]/[integrationVersion]/components/[componentSlug]/index.tsx:127]: Good callout, we should be safe here!

src/pages/[productSlug]/integrations/index.tsx
```diff
@@ -66,6 +73,10 @@ export async function getServerSideProps({ params }) {
 	]
 	return {
 		props: {
+			metadata: {
+				title: `Integrations | ${product.name}`,
+				// description: `TODO`,
```
Comment (thiskevinwang) [inline src/pages/[productSlug]/integrations/index.tsx:78]: 👀

thiskevinwang (approved):
Looks pretty good to me! 

---

Summary for myself:

So `src/layouts/product-integration-layout/index.tsx` effectively does some heavy-lifting here, (lines-of-code-wise)

- `src/views/product-integration/readme-view/index.tsx` uses it
- `src/views/product-integration/components/component-tab-content/index.tsx` uses it

A few new Next page paths were created, and they largely have very similar data-fetching needs

src/pages/[productSlug]/integrations/index.tsx
Comment (BrandonRomano) [reply src/pages/[productSlug]/integrations/index.tsx:78]: Oh!  I'm leaving this here, because I _do_ want a description here, but don't know what it should be yet.  We'll have to sort that out in the new year!

PR #1496, opened 2023-01-04
Adding "api-docs/packer" to hcp's basePaths array

ashleemboyer:
## 🗒️ What

- Adds `api-docs/packer` as to `hcp`'s `basePaths` property

## 🤷 Why

- So that `api-docs/packer` links in HCP docs content are correctly rewritten to `/hcp/api-docs/packer`

## 🧪 Testing

- [ ] Go to [/hcp/docs/packer/manage-image-use/ancestry#ancestry-api](https://dev-portal-git-ambfix-hcp-base-paths-hashicorp.vercel.app/hcp/docs/packer/manage-image-use/ancestry#ancestry-api) on the preview
- [ ] Make sure the `Get Iteration` link goes to `/hcp/api-docs/packer` (not `https://cloud.hashicorp.com/api-docs/packer` [like in staging](https://dev-portal-git-staging-hashicorp.vercel.app/hcp/docs/packer/manage-image-use/ancestry#ancestry-api))
- [ ] Go to [`/hcp/docs/packer/manage-image-use/image-channels#image-channels`](https://dev-portal-git-ambfix-hcp-base-paths-hashicorp.vercel.app/hcp/docs/packer/manage-image-use/image-channels#image-channels) on the preview
- [ ] The `HCP Packer API` link should go to `/hcp/api-docs/packer` (not `https://cloud.hashicorp.com/api-docs/packer` [like in staging](https://dev-portal-git-staging-hashicorp.vercel.app/hcp/docs/packer/manage-image-use/image-channels#image-channels))

PR #1497, opened 2023-01-04
Adjusting link rewrites helper and plugin

ashleemboyer:
## 🗒️ What

- Fixes an accidental `LEARN_DOMAIN` key in the object returned by `getLearnToDevDotUrlMap`
- Writes `mdxUnrewriteableLinks` data to an output file under `src/.generated`
- Comments out nav data JSON file logic and adds a link to the relevant Asana ticket
- Adds a check for `replacementPath` in `getRewrittenNonRelativeLearnUrl`
- Updates the `url.startsWith('/')` case in `handleRelativeUrl` to check if a url starts with a product's basePath, rather than parsing out the first piece of a pathname and seeing if it is a basePath

## 🤷 Why

- The `getRewrittenNonRelativeLearnUrl` change fixes the case where this no replacement path was found and `'undefined'` was getting returned as the rewritten value
- The `handleRelativeUrl` change was necessary for HCP's `api-docs/packer` case. We already know there are some base paths that include a `/`, but this one is the first where there is not a base path that matches the first part.
  - For example, Terraform has several base paths that include a `/`, but it also has a base path for each of those that matches the first part of the complex base paths.

## 🧪 Testing

- [ ] Check out the `cloud.hashicorp.com` repo locally
- [ ] Pull down this branch locally
- [ ] From your copy of this project, run the link rewrite script with (but use the file path for your copy of `cloud.hashicorp.com` in `contentDirectory` and `navDataDirectory`):

```
npx hc-tools ./scripts/docs-content-link-rewrites/rewrite-links.ts -- --repo cloud.hashicorp.com --contentDirectory ../docs-repos/cloud.hashicorp.com/content --navDataDirectory ../docs-repos/cloud.hashicorp.com/data
```

- [ ] `/api-docs/packer` links should be rewritten to `/hcp/api-docs/packer`
- [ ] There should be no learn.hashicorp.com links that are rewritten to `undefined` (they'll stay as-is)


ashleemboyer:
🗒️ This'll be opened for review after #1496 is merged.

scripts/docs-content-link-rewrites/helpers/get-learn-to-dev-dot-url-map.ts
```diff
@@ -116,7 +116,7 @@ const getProductLandingPaths = () => {
 
 const getOtherLandingPaths = () => {
 	const result = {
-		LEARN_DOMAIN: '/',
+		[LEARN_DOMAIN]: '/',
```
BRKalow, line 119: oops 😄 I hate when this happens, good catch!

BRKalow (approved):
LGTM!

PR #1508, opened 2023-01-06
feat: HDS `Pagination`

thiskevinwang:
## 🔗 Relevant links

- [Swingset  Preview][swingset_preview] 🔎
- Asana Task Attached at bottom of description 🎟️

## 🗒️ What

This adds an HDS Pagination component.
Here are some links for reference:
- (Old)
  - ~in-progress ember component: https://github.com/hashicorp/design-system/pull/661~
  - ~preview: https://hds-components-git-hds-943-pagination-component-hashicorp.vercel.app/components/pagination~
- (New)
  - PR: https://github.com/hashicorp/design-system/pull/1020
  - preview: https://hds-components-git-hds-972-pagination-cloud-ui-c491d7-hashicorp.vercel.app/components/pagination
- figma spec: https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=17236%3A219695&t=z5zFye8v2bIQmZZY-0

### TODOs

- [x] mobile styling responsibility can be handled by the consumer. Due to the _composed_ nature of this component, you can wrap `Pagination.Nav|SizeSelector|Info` in `div`s with desired classNames/styling

## 🤷 Why

To achieve design system consistency with HDS.

## 🛠️ How

N/A

## 📸 Design Screenshots

(See [Swingset  Preview][swingset_preview])

## 🧪 Testing

(See [Swingset  Preview][swingset_preview])

Things to look for
- Ensure mouse interactions work
- Ensure keyboard interactions work
- Ensure screenreader interaction works

## 💭 Anything else?

[swingset_preview]: https://dev-portal-git-kw-hds-paginator-hashicorp.vercel.app/swingset/components/pagination

---
  - https://app.asana.com/0/0/1203548351777721
  - https://app.asana.com/0/0/1203653018794425


src/components/pagination/pagination.module.css
```diff
@@ -0,0 +1,177 @@
...
+		margin-left: 12px;
+		padding: 0 24px 0 8px;
+		font-size: var(--token-typography-body-100-font-size);
+		font-family: var(--token-typography-body-100-font-family);
+		line-height: var(--token-typography-body-100-line-height);
+		background-position: center right 5px;
+	}
+}
+
+/* TODO: extract this to global scope */
```
thiskevinwang, line 165: This was copied from the HDS ember implementation

src/components/pagination/docs.mdx
```diff
@@ -0,0 +1,101 @@
...
+	onPageChange={(idx, pagesize) => console.log("Select page: %s; Pagesize: %s", idx, pagesize)}
+>
+	<Pagination.Info showTotalItems={true}/>
+	<Pagination.Nav type="truncated" />
+	<Pagination.SizeSelector sizes={[10, 30, 50]}/>
+</Pagination>
+`}
+</LiveComponent>
+
+## With Elements
```
BrandonRomano, line 64: This demo does not actually render the appropriate items when I switch "Items per page" to anything other than 10.  It still will render the items as if you have 10 selected for me.

src/components/pagination/docs.mdx
```diff
@@ -0,0 +1,101 @@
...
+<LiveComponent>
+	{`<Pagination
+	totalItems={103}
+	pageSize={10}
+	page={3}
+	onPageChange={(idx, pagesize) => console.log("Select page: %s; Pagesize: %s", idx, pagesize)}
+>
+	<Pagination.Info showTotalItems={true}/>
+	<Pagination.Nav type="truncated" />
+	<Pagination.SizeSelector sizes={[10, 30, 50]}/>
```
BrandonRomano, line 59: Curious of the motivation here to use some children items to specify some configuration.  Is this the interface of the other HDS implementation? Also too, curious if there are any further pros/cons to this approach you can share -- I'm not super familiar with approaching managing the config this way (usually I'd just reach for props on the top level component).

BrandonRomano: p.s. I definitely don't object to this setup at all, just genuinely curious!

BrandonRomano (commented):
Looks really good!  I think the example will need to get fixed as it doesn't work when you change to more than 10 per page.

Beyond that though this has my approval.

src/components/pagination/docs.mdx
thiskevinwang: Fixed here: https://github.com/hashicorp/dev-portal/pull/1508/commits/c7a6e0da64687ed18ebaafc1cd33328e3b354faa (no additional testcase needed as it was consumer error)

braydenlove:
lmk when you are ready for VQA

src/components/pagination/docs.mdx
thiskevinwang: Fixed in: https://github.com/hashicorp/dev-portal/pull/1508/commits/5180305452e80c31f12c8f5c1fcf6d4de7c4bcd7

BrandonRomano (approved):

src/components/pagination/docs.mdx
thiskevinwang: I think this — the _composition_ pattern — avoids the case of having a "black-box-ish" top-level `Pagination` component.

Pro: If you go to the Swingset - "With Mobile Styling" example, you'll see that you can very easily conditionally render various composed children components. (read: there's more flexibility/freedom)

Con: You could theoretically and potentially enable passing props to _either_ the top-level or composed-children components, depending on how you write the internals. (which raises the authoring complexity)

I haven't thought too hard or been dogmatic about, _"Oh props can ONLY be passed here or cannot be passed there"_

These my own moderate opinions, loosely held 😂 

Good example in the wild — https://cmdk.paco.me/

thiskevinwang:
@braydenlove This is now ready for VQA 🤝

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,270 @@
...
+		>
+			{icon}
+			{type === 'compact' && (
+				<span className={classNames(s.label)} aria-hidden="true">
+					{label}
+				</span>
+			)}
+		</button>
+	)
+}
```
ashleemboyer, line 209: 🚧 A few things for `ButtonArrow`:

- For the voice control experience to work, the Accessible Name of the rendered `<button>` has to match the visible label of it. In this case, where we're setting the Accessible Name `aria-label`, that means `aria-label` needs to be the same as the visible label (`label`).
- Rather than several maps with `prev/next` properties, we can use just one.
- The `{ next, prev }` map that sets a `className` isn't needed and can be simplified with `s[direction]`.
- `aria-hidden` should be set on the `span`. It's redundant because `aria-label` is being set on the element that contains it.

```tsx
const ButtonArrow = ({ type, direction }: ButtonArrowProps) => {
	const pagination = usePagination()

	const { handleClick, icon, isDisabled, label } = {
		next: {
			handleClick: () => {
				// clamp upper bound
				const nextIdx = Math.min(pagination.page + 1, pagination.totalPages)
				pagination.setPage(nextIdx)
				pagination.onPageChange(nextIdx)
			},
			icon: <IconChevronRight16 />,
			isDisabled: pagination.page === pagination.totalPages,
			label: 'Next',
		},
		prev: {
			handleClick: () => {
				// clamp lower bound (0)
				const prevIdx = Math.max(pagination.page - 1, 0)
				pagination.setPage(prevIdx)
				pagination.onPageChange(prevIdx)
			},
			icon: <IconChevronLeft16 />,
			isDisabled: pagination.page === 1,
			label: 'Previous',
		},
	}[direction]

	return (
		<button
			aria-label={label}
			className={classNames(s.arrow, s.control, s[direction])}
			disabled={isDisabled}
			onClick={handleClick}
		>
			{icon}
			{type === 'compact' && (
				<span className={classNames(s.label)}>{label}</span>
			)}
		</button>
	)
}
```

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,270 @@
...
+		<button
+			className={classNames(s.control)}
+			type="button"
+			onClick={handleClick}
+		>
+			<span className={'g-screen-reader-only'}>page&nbsp;</span>
+			{children}
+		</button>
+	)
+}
```
ashleemboyer, line 233: 🚧 Two things here:

- The `children` and `page` props are the same, so both props aren't needed. I'd vote for keeping just `page` since it's more descriptive in the `handleClick` function.
- The `g-screen-reader-only` `span` is redundant because `button` automatically handles setting its Accessible Name we want using its children.

```tsx
const ButtonNumber = ({ page }: ButtonNumberProps) => {
	const pagination = usePagination()
	const handleClick = () => {
		pagination.setPage(page)
		// don't trigger callback if the current page is already selected
		if (pagination.page !== page) {
			pagination.onPageChange(page)
		}
	}
	return (
		<button
			className={classNames(s.control)}
			type="button"
			onClick={handleClick}
		>
			{page}
		</button>
	)
}
```

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,270 @@
...
+		<div className={s['size-selector']}>
+			<label className={s.label}>
+				Items per page
+				<select className={s.select} onChange={handleChange}>
+					{sizes.map((e) => (
+						<option key={e} value={e} selected={pagination.pageSize === e}>
+							{e}
+						</option>
+					))}
+				</select>
```
ashleemboyer, line 260: 🙋🏻 Can the `e` variable name be updated to `size` for clarity?

```suggestion
				<select className={s.select} onChange={handleChange}>
					{sizes.map((size) => (
						<option
							key={size}
							value={size}
							selected={pagination.pageSize === size}
						>
							{size}
						</option>
					))}
				</select>
```

src/components/pagination/docs.mdx
```diff
@@ -0,0 +1,128 @@
...
+
+## With Mobile Styling
+
+<LiveComponent>
+	{`() => {
+	const {
+		isDesktop,
+		isMobile,
+		isTablet,
+	} = useDeviceSize()
```
ashleemboyer, line 111: I think the function being passed to `LiveComponent` is only being invoked once on initial load, so this demo isn't changing when the viewport changes.

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,270 @@
...
+		}
+	}
+
+	return (
+		<div className={s['size-selector']}>
+			<label className={s.label}>
+				Items per page
+				<select className={s.select} onChange={handleChange}>
+					{sizes.map((e) => (
+						<option key={e} value={e} selected={pagination.pageSize === e}>
```
ashleemboyer, line 256: 🚧 Oh no! I misled you with the `selected` prop. 😅 There's a console message about this happening:

```
Warning: Use the `defaultValue` or `value` props on <select> instead of setting `selected` on <option>
```

src/components/pagination/docs.mdx
ashleemboyer: On initial load, the `Info` component isn't showing the correct numbers. It starts with `21 – 30 of 83` when I'd expect `1 – 30 of 83` since the current page is `1` and the page size is `30`.

image

ashleemboyer (changes requested):

src/components/pagination/docs.mdx
thiskevinwang: Good catch! This was a logic bug — will fix and add a test

thiskevinwang: Fixed in: https://github.com/hashicorp/dev-portal/pull/1508/commits/77fbb35d914fffcdfc756ca5eaf87b21d56cbe94

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,308 @@
...
+		)
+	}
+	if (typeof _pageSize !== 'number') {
+		throw new Error(
+			'Pagination: pageSize is required, but was not specified. Please try adding a value such as `10`.'
+		)
+	}
+
+	const [page, setPage] = useState(() => _page)
+	const [pageSize, setPageSize] = useState(() => _pageSize)
```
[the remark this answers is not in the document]
thiskevinwang: I'll make this change!
(I was on the fence with this idea of wrapping/merging the callbacks & state-setters and you've swayed me towards doing it :D )

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,308 @@
...
+	if (typeof totalItems !== 'number') {
+		throw new Error(
+			'Pagination: totalItems is required, but was not specified. Please try adding a value such as `103`.'
+		)
+	}
+	if (typeof _pageSize !== 'number') {
+		throw new Error(
+			'Pagination: pageSize is required, but was not specified. Please try adding a value such as `10`.'
+		)
+	}
```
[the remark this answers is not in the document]
thiskevinwang: Handled in https://github.com/hashicorp/dev-portal/pull/1508/commits/ba1c100e8c70bdea297b1df036d31f2a0780755a

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,270 @@
...
+		switch (t) {
+			case 'numbered':
+				return rawitems
+			case 'truncated':
+				return generateTruncatedList(rawitems, currentPage)
+			case 'compact':
+			default:
+				return []
+		}
+	})(type)
```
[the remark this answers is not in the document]
thiskevinwang: I'll punt on this change, if that's alright?

src/components/pagination/index.tsx
thiskevinwang: Handled in https://github.com/hashicorp/dev-portal/pull/1508/commits/3af6b68814714e2837b4a3caedfebb88b3ce7f03
- re `s[direction]` — I opted for a `className` field that maps to `s.prev | s.next` . I didn't want to _mask_ a css-module class, behind a dynamic key (`direction`) in TSX, despite them _implicitly_ matching. I opted to be explicit here.

thiskevinwang: Removed redundant `aria-hidden` https://github.com/hashicorp/dev-portal/pull/1508/commits/15630c2c8eff24e4d4267ad7c5ce35d2fef47de1

src/components/pagination/index.tsx
thiskevinwang: Handled in https://github.com/hashicorp/dev-portal/pull/1508/commits/a1da1faafd4ad8fa6570bff9afca0255e9351a12

src/components/pagination/index.tsx
thiskevinwang: Updated!

src/components/pagination/index.tsx
thiskevinwang: This has been fixed! 
- Removed `option.selected` in favor of `select.defaultValue`

src/components/pagination/index.tsx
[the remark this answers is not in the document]
pbortnick: 💡  / ❓ Would it be possible to return null or `<></>` if the `totalItems` is `0` or _maybe_ `undefined || 0` rather than throwing an error? There are situations when using the pagination on a search page where a search can yield 0 results

@thiskevinwang  @ashleemboyer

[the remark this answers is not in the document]
thiskevinwang: @pbortnick Thank you and this is a great callout — I'll go and implement this for `totalItems`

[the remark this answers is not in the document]
pbortnick: FWIW I copy pasted in dotcom in the faceted search project and this was the only issue I encountered 🎉 good stuff!

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,259 @@
...
+	)
+	return (
+		<div className={s.info}>
+			{start} - {end}
+			{showTotalItems ? ` of ${pagination.totalItems}` : null}
+		</div>
+	)
+}
+
+const Nav = ({ type = 'compact' }: NavProps) => {
```
alexcarpenter, line 108: Non-blocking question: Can we expose classNames option for the compound components so that we can use CSS to conditionally hide them vs needing JS media query hook?

thiskevinwang: Yes, 100% — I was on the fence about this, but seems this is the cleaner approach

thiskevinwang: I may actually punt on this for now — the `Nav.type` prop handles some conditional rendering via TS/JS, and the use of classNames doesn't yet handle that case :/

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,259 @@
...
+		<button
+			className={classNames(s.arrow, s.control, className)}
+			aria-label={label}
+			onClick={handleClick}
+			disabled={isDisabled}
+		>
+			{icon}
+			{type === 'compact' && (
+				<span className={classNames(s.label)}>{label}</span>
+			)}
```
ashleemboyer, line 206: 🙋🏻 Just noticing this -- should the `label` be shown if the type is `"compact"`? I'd imagine that the label would _not_ show in this case, to save space.

src/components/pagination/docs.mdx
ashleemboyer: @thiskevinwang did you get a chance to look at this? I don't think we want to include this in the Swingset page since it doesn't behave as expected.

src/components/pagination/index.tsx
thiskevinwang: Based on the [HDS implementation](https://hds-components-git-hds-943-pagination-component-hashicorp.vercel.app/components/pagination#how-to-use), yea the "Previous/Next" label will be shown _only_ on `compact`

ashleemboyer: Ohhh, the `type` passed in to `ButtonArrow` in the pagination type! I thought it was just scoped to the button itself. Thanks! 😊

src/components/pagination/docs.mdx
thiskevinwang: @ashleemboyer https://dev-portal-git-kw-hds-paginator-hashicorp.vercel.app/swingset/components/pagination

I'm seeing that it is working as expected. Can you double check?

https://user-images.githubusercontent.com/26389321/211910376-74d7b4bd-d29d-47c6-b39a-f05ac35c83bd.mp4

src/components/pagination/index.tsx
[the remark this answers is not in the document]
ashleemboyer: The part I feel most strongly about is that there is a function being created and immediately executed, making it difficult to understand what's happening. It doesn't need to be resolved in the way I wrote it in my comment, but I do think we should avoid creating a function like this.

src/components/pagination/docs.mdx
ashleemboyer: Just checked again! It's working as expected. Who knows what happened yesterday? 😆

src/components/pagination/index.tsx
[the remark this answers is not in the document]
thiskevinwang: @pbortnick: implemented here https://github.com/hashicorp/dev-portal/pull/1508/commits/ca227325be5b3a2fa3b1a2e584b57dfc3eedd8f9

Screenshot
CleanShot 2023-01-11 at 15 50 16@2x

src/components/pagination/index.tsx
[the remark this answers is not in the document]
ashleemboyer: Thanks for the links! I know what these are and how they work, but didn't know the official names.

src/components/pagination/index.tsx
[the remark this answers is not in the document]
pbortnick: That works for me. I actually prefer that to rendering nothing. Thank you 😄 !

src/components/pagination/index.tsx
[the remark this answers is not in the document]
thiskevinwang: Applied your suggestion in https://github.com/hashicorp/dev-portal/pull/1508/commits/7ebe76ec85e95aba5ee8dcf50ff0b5787da01751

thiskevinwang:
> Screen Shot 2023-01-11 at 2 14 24 PM
> 
> Curious why the x - x of xx isn't left aligned to the container of elements like the select is on the right.

@braydenlove it's due to stray padding on the demo `ul` element, which won't be a problem in actual usage. Regardless I can update this

CleanShot 2023-01-11 at 16 18 59@2x

braydenlove:
> > Screen Shot 2023-01-11 at 2 14 24 PM
> > Curious why the x - x of xx isn't left aligned to the container of elements like the select is on the right.
> 
> @braydenlove it's due to stray padding on the demo `ul` element, which won't be a problem in actual usage. Regardless I can update this
> 
> CleanShot 2023-01-11 at 16 18 59@2x

Awesome thanks for the info!

braydenlove:
One visual dif I am seeing, curious why the focus styles are different and the total area is different between the HDS stuff and swingset.

Can you also ensure that the numbers and icons are `foreground-primar`y? They appear to be `#000000`. 
image

Outside of that this is ready to ship. Lmk if it is easier to jump on a quick zoom.

src/components/pagination/index.tsx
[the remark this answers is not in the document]
ashleemboyer: Thank you for making the refactor! Greatly appreciated ❤️

ashleemboyer (approved):
✨

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+const ButtonNumber = ({ page }: ButtonNumberProps) => {
+	const pagination = usePagination()
+	const handleClick = () => {
+		pagination.setPage(page)
+	}
+	return (
+		<button
+			className={classNames(s.control)}
+			type="button"
```
alexcarpenter, line 222: 💭 can we add `aria-current="page"` to the active button? This will announce the to user which page is active when focusing the active button.

CleanShot 2023-01-12 at 09 13 51@2x

thiskevinwang: Done: https://github.com/hashicorp/dev-portal/pull/1508/commits/e779e23e279e70368d9278e36ccd5a1de5018dbf

ashleemboyer: According to [the `aria-current` documentation](https://www.w3.org/TR/wai-aria-1.2/#aria-current), I think the `"page"` token is only valid when it's attached to an `<a>` element.

> A page token used to indicate a link within a set of pagination links, where the link is visually styled to represent the currently-displayed page.

In this component, I think we'll need to use [`aria-selected`](https://www.w3.org/TR/wai-aria-1.2/#aria-selected) (value is a boolean) instead. From the `aria-current` docs:

> Authors SHOULD NOT use the [aria-current](https://www.w3.org/TR/wai-aria-1.2/#aria-current) attribute as a substitute for [aria-selected](https://www.w3.org/TR/wai-aria-1.2/#aria-selected) in widgets where [aria-selected](https://www.w3.org/TR/wai-aria-1.2/#aria-selected) has the same meaning. For example, in a [tablist](https://www.w3.org/TR/wai-aria-1.2/#tablist), [aria-selected](https://www.w3.org/TR/wai-aria-1.2/#aria-selected) is used on a [tab](https://www.w3.org/TR/wai-aria-1.2/#tab) to indicate the currently-displayed [tabpanel](https://www.w3.org/TR/wai-aria-1.2/#tabpanel).

src/components/pagination/pagination.module.css
```diff
@@ -0,0 +1,182 @@
...
+		content: '';
+	}
+
+	/* tab focus or click focus */
+	&:focus {
+		color: var(--token-color-foreground-action);
+		&::before {
+			box-shadow: var(--token-focus-ring-action-box-shadow);
+		}
+	}
```
ashleemboyer, line 103: 🙋🏻 Is it possible to use the [`.g-focus-ring-from-box-shadow` helper](https://github.com/hashicorp/dev-portal/blob/main/src/pages/style.css#L176-L185) here to decrease the maintenance effort of the shared style? It avoids using `:focus` since that adds an unwanted ring after clicking, also handles setting `outline: transparent` so that the focus ring works in high contrast modes.

thiskevinwang:
@braydenlove selected/focus styles are updated

src/components/pagination/pagination.module.css
thiskevinwang: @ashleemboyer done in https://github.com/hashicorp/dev-portal/pull/1508/commits/02005095f2c1c6be25531993bae9560ac5c5dcd8

PR #1494, opened 2022-12-23
Variable Group Styling

BrandonRomano:

PR #1512, opened 2023-01-09
Variable Group Styling

BrandonRomano:
Recreation of #1494 to avoid a messy rebase.

---
https://dev-portal-git-brvariable-group-hashicorp.vercel.app/waypoint/integrations/kubernetes/latest/components/platform
CleanShot 2023-01-10 at 12 13 19@2x


src/views/product-integration/component-view/components/variable-group-list/index.tsx
```diff
@@ -35,32 +39,52 @@ export function VariableGroupList({
...
-						<code className={s.type}>{variable.type}</code>
-						{typeof variable.description !== 'undefined' && (
-							<div>
+						{isNested ? (
+							<div className={s.arrowIcon}>
+								<IconCornerDownRight16 />
+							</div>
+						) : (
+							<></>
+						)}
```
thiskevinwang, line 52: Nit: We can use [logical &&](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator) instead of ternary+fragment for conditional rendering

```suggestion
						{isNested && (
							<div className={s.arrowIcon}>
								<IconCornerDownRight16 />
							</div>
						)}
```

src/views/product-integration/component-view/components/variable-group-list/index.tsx
```diff
@@ -35,32 +39,52 @@ export function VariableGroupList({
...
+										>
+											{variable.required ? 'Required' : 'Optional'}
+										</span>
+									)}
+								</span>
+								{variable.type ? (
+									<Badge color="highlight" text={variable.type} />
+								) : (
+									<></>
+								)}
```
thiskevinwang, line 71: Same nit: [logical &&](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator) instead of ternary + fragment

thiskevinwang (changes requested):
Looks pretty good! Calling out some discrepancies based on recent [Loom](https://www.loom.com/share/f8a76bcab23c48e697a785fac4a6059b) and [Figma](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=17236%3A219695&t=RRtJcfbCbdHJlwrp-0)

The <kbd>category</kbd> badges appear to have _inverted_ styling
- CleanShot 2023-01-10 at 12 38 04@2x

---

Side note (out of scope of this PR): I also added an asana task for implementing the **Required only** filtering
https://app.asana.com/0/1202626741675858/1203684611651045/f

- CleanShot 2023-01-10 at 12 38 29@2x

BrandonRomano:
Good call-outs.  I'll adjust today.

src/views/product-integration/component-view/components/variable-group-list/index.tsx
BrandonRomano: Ah!  cc @thiskevinwang looks like this article is suggesting ternary over &&, whereas you suggested logical && over ternary.

I see how you can get the job done with either; curious if there is an explicit preference that we've agreed upon on the team?  @kendallstrautman might be a good thread to open up for discussion in the channel (will leave that to you to do though if you want / not).

thiskevinwang: I'm not super tied to either approach!

I've historically reached for both `x && y` as well as `x ? y : z` ...(:homer:)
- The article sounds like it advocates for ternary for mitigating a small class of bugs.

thiskevinwang: (Feel free to disregard my "nits")

BrandonRomano: I don't have a strong opinion either (but I do have an opinion that *we* should pick what the default favored approach should be)

src/views/product-integration/component-view/components/variable-group-list/index.tsx
BrandonRomano: Going to resolve this as the case is not out on the preference here.  Will adjust once we land on something as a team!

src/views/product-integration/component-view/components/variable-group-list/index.tsx
BrandonRomano: Going to resolve this for _this PR_ -- maybe let's bring this one to the team to see if we can land on a preferred style.

BrandonRomano:
Going to merge this in as all feedback is addressed for now (& I want a mainline preview URL with this work in for an RFC).

PR #1515, opened 2023-01-10
Fix DocsView handling of some links with query params and/or anchors

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-amburl-adjuster-tweaks-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202114367927919/1203641801270770/f) 🎟️

## 🗒️ What

- Replaces some manual test cases with automated ones that handle having a case for every combination of when query params or an anchor are present/omitted in:
  - `rewrite-external-docs-link.test.ts`
  - `multiple-remark-plugins.test.ts`
- Fixes the logic for the new, failing test cases in:
  - `rewrite-external-docs-link.ts`
  - `rewrite-external-learn-link.ts`

## 🧪 Testing

No manual testing to do here.


src/lib/remark-plugins/rewrite-tutorial-links/__tests__/rewrite-external-docs-link.test.ts
```diff
@@ -6,14 +6,23 @@ const getTestURLObject = (url: string) => {
 	return new URL(url, __config.dev_dot.canonical_base_url)
 }
 
+const URL_SUFFIXES = ['', '?param=value', '#heading', '?param=value#heading']
```
kendallstrautman, line 9: A note would be good here as it makes the test a bit harder to see at first glance since none of the assertions talk about the hash / query param inputs / expectations.

kendallstrautman: 💭 Rather than having the `testEachCase` always add all suffixes, it could be more clear to separate this into its own function, so that we can read the intent. So there would be a base `testEachCase` that doesn't append or modify the paths, and a `testEachCaseWithParams` — to show that there's a manipulation happening with those paths.

src/views/docs-view/__tests__/multiple-remark-plugins.test.ts
```diff
@@ -4,42 +4,62 @@ import { rewriteTutorialLinksPlugin } from 'lib/remark-plugins/rewrite-tutorial-
...
+				.process(`[test](${input})`)
+			expect(result.contents).toMatch(`[test](${expected})`)
+		}
+	)
+}
 
-			// The link should be unchanged
-			expect(result.contents).toMatch(expectedValue)
-		})
+describe('multiple remark plugins', () => {
```
kendallstrautman, line 37: Nice, man the multiple remark plugins is nasty to track down this stuff! well done

src/lib/remark-plugins/rewrite-tutorial-links/__tests__/rewrite-external-docs-link.test.ts
kendallstrautman: That said, I think a note would totally suffice as well 😄 its up to you

kendallstrautman (approved):
Overall looks good, just a few non-blocking comments on making it more clear whats happening with the search / hash testing!

PR #1503, opened 2023-01-05
feat: adds enhanced mdx alert

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksadds-enhanced-mdx-alert-hashicorp.vercel.app/swingset/components/mdxinlinealert) 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1201321706676523) 🎟️
- [Figma File](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=16287%3A225825) 🎨 

## 🗒️ What

This PR adds new components 'InlineAlert' & `MdxInlineAlert`, which is used for authors in MDX content. This is an update from the [previous alert components](https://github.com/hashicorp/remark-plugins/tree/main/plugins/paragraph-custom-alerts) which were referenced via [custom syntax](https://github.com/hashicorp/tutorials/blob/31a605985c4fc17811dfee69f6c3b8ff6ab258c6/content/tutorials/cloud/vault-ops.mdx?plain=1#L19 )

The base [`InlineAlert` component ](https://dev-portal-git-ksadds-enhanced-mdx-alert-hashicorp.vercel.app/swingset/components/inlinealert)should correspond to the HDS Alert spec. 

> Note: this does not make the `MdxInlineAlert` component available for authors to use. That will occur in a subsequent PR after feedback is received on the authoring interface.

The mdx alert has 4 variants: 'tip' (default), 'highlight', 'note' and 'warning'. Below is the _mapping_ from the old alerts to these new ones: 
```
info → Tip or Highlight
warning → Note
danger → Warning
```

Screen Shot 2023-01-04 at 5 27 33 PM

## 🧪 Testing

- [Visit the swingset page ](https://dev-portal-git-ksadds-enhanced-mdx-alert-hashicorp.vercel.app/swingset/components/mdxinlinealert)to view all the variants

## 💭 Anything else?

Still need to update docs in the tutorials repo on how to use this component!

PR #1517, opened 2023-01-11
feat: adds inline alert to mdx components

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksadds-mdx-alert-to-markdown-hashicorp.vercel.app/swingset/components/mdxinlinealert) 🔎
- [Asana task](https://app.asana.com/0/1203412256229472/1203676883553328) 🎟️

## 🗒️ What

Follow up from https://github.com/hashicorp/dev-portal/pull/1503

This PR adds the `MdxInlineAlert` component to the authoring options. I seeded my local database to test out with real content and found a few things. 

First: there are a ton of interesting use cases for how these alerts are being used today that I haven't considered. 
- Authors put images in them?
- They include them in top level list structures
- I'm sure there are more we could root out

Luckily since this is a net new component, we won't have any breakage issues, but there could be use cases we didn't consider that may render weird as authors pressure test this. 

**Biggest note**

Since these components need to be able to handle markdown syntax, we have to do some extra type style overrides to keep things cohesive if a paragraph or anchor tag renders.

## Authoring Interface

📹 [see my rambling commentary here ](https://www.loom.com/share/4c623d0df2f349cca008420b41f2cdf6)on whether or not we finalize with this interface, comments welcome 😄 

The authors wanted unique components for each alert type. 

```mdx
<Note>Hi I am a note</Note>

<Tip>

  **If they want to render** md styles or eleements, they need to adhere to the spaced conventions

</Tip>

<Highlight title="My custom title>This is a great highlight</Highlight>

<Warning>Don't do this</Warning>
```

I'm a bit on the fence whether this is an okay choice from a maintenance position, but I definitely see it as a better authoring experience. Below is a picture of some test content and the rendered output. 

Screen Shot 2023-01-10 at 5 32 37 PM
Screen Shot 2023-01-10 at 5 25 30 PM


src/components/dev-dot-content/mdx-components/mdx-alert/docs.mdx
```diff
@@ -4,6 +4,8 @@ componentName: 'MdxInlineAlert'
 
 The `MdxInlineAlert` component is used in long-form MDX content to highlight various message types. Default titles and icons are provided for each 'type', however these can be overridden via the `title` prop.
 
+_Note_ Authors use specific variants of this component in content where the type is set: `Tip`, `Highlight`, `Note`, & `Warning`: see `./variants.tsx`
```
kendallstrautman, line 7: I will write detailed docs in the tutorials repo on how exactly to use this component once we 100% land on the authoring interface.

src/components/dev-dot-content/mdx-components/mdx-alert/mdx-inline-alert.module.css
```diff
@@ -1,3 +1,22 @@
...
+* This components chidlren can render regular markdown
+* that inherits the global long form content type styles
+* however this component needs to maintain specific type
+* styles defined in inline-alert.module.css 
+* 
+* Currently only anchor and paragraph elements are accounted for
+* but depending on authoring behavior, additional adjustments
+* could be made
+**/
+.typographyOverride {
```
kendallstrautman, line 15: Ideas of how to circumvent this welcome 🤷🏻

kendallstrautman: See my comments in description: 
> Since these components need to be able to handle markdown syntax, we have to do some extra type style overrides to keep things cohesive if a paragraph or anchor tag renders.

src/components/dev-dot-content/mdx-components/mdx-alert/docs.mdx
ashleemboyer: 💭 Can we add the detailed docs here and then link to it where you're thinking of in `hashicorp/tutorials`? That'd make the information available to both tutorials and docs authors in a way that they share the source of truth.

src/components/dev-dot-content/mdx-components/mdx-alert/mdx-inline-alert.module.css
ashleemboyer: 🙋🏻 It's not clear to me what cases need this override class in it. Is there an example in the Swingset page where I can manually remove this class name and see broken styles?

src/components/dev-dot-content/mdx-components/mdx-alert/variants.tsx
```diff
@@ -0,0 +1,23 @@
...
+	return <MdxInlineAlert type="highlight" {...props} />
+}
+
+export function MdxNote(props: Omit<MdxInlineAlertProps, 'type'>) {
+	return <MdxInlineAlert type="note" {...props} />
+}
+
+export function MdxWarning(props: Omit<MdxInlineAlertProps, 'type'>) {
+	return <MdxInlineAlert type="warning" {...props} />
+}
```
ashleemboyer, line 23: 💭 I think we should set `type` after spreading `props` for each of these, that way it can't be accidentally overwritten in MDX (where the TS checks can't happen).

```suggestion
export function MdxTip(props: Omit<MdxInlineAlertProps, 'type'>) {
	return <MdxInlineAlert {...props} type="tip" />
}

export function MdxHighlight(props: Omit<MdxInlineAlertProps, 'type'>) {
	return <MdxInlineAlert {...props} type="highlight" />
}

export function MdxNote(props: Omit<MdxInlineAlertProps, 'type'>) {
	return <MdxInlineAlert {...props} type="note" />
}

export function MdxWarning(props: Omit<MdxInlineAlertProps, 'type'>) {
	return <MdxInlineAlert {...props} type="warning" />
}
```

Jasmine-hashi (approved):

src/components/dev-dot-content/mdx-components/mdx-alert/docs.mdx
kendallstrautman: For sure! I was planning on linking the two just haven't written the tutorials docs yet.

src/components/dev-dot-content/mdx-components/mdx-alert/mdx-inline-alert.module.css
kendallstrautman: The swingset page wont parse markdown content that is passed in the same way it does on the long form content pages so its not possible to see this in swingset. The cases where this is needed can be seen in the screenshots in the description. When md syntax is used within the body of the alert, the content will be turned into `p` or `a` tags depending on the markup. This component shouldn't support headings so I've not included those.

kendallstrautman: The easiest way to test this is to seed a local api and database and run the frontend from that local api, its a bit cumbersome. I could take a video if that would be helpful?

zchsh (approved):
✅ Looks good to me from an implementation standpoint!

💭 Had some thoughts on trying to apply `body-200` to _all_ alerts, I think the content will be hard to keep cohesive-looking since spacing for all MDX elements has been designed for `body-300`. Maybe for longer alerts it'd be easier and more cohesive-looking to stick with `body-300`. But at the same time, maybe a non-issue and more of an authoring-level change (perhaps we're trying to move towards short text-only alerts?), so not at all blocking IMO 👍

src/components/dev-dot-content/mdx-components/mdx-alert/mdx-inline-alert.module.css
braydenlove: Interesting point. Where I get stuck is that we should not be adding images + captions to alerts. Ideally inline-alerts should be shorter and be text only. When I look at alerts across Developer they are all using `body-200` except for the [~>](url) alerts of old. Keep me honest though as I could be missing something. Although authors can add `code` elements, `code-blocks`, images and other mdx components into an alert I don't think we should encourage it by making it look intentional if that makes sense. `body-200` in alerts comes straight from HDS and if we are going to divert I want to make sure it is for really good reasons. Open to discuss more though!

braydenlove: I also feel like body-300 in alerts feels too strong, with the decreased content width an alert with a paragraph of plain text takes up the entire screen and blows out the hierarchy imo.
image

src/components/dev-dot-content/mdx-components/mdx-alert/docs.mdx
kendallstrautman: Okay I added some authoring docs with example mdx content. I'll open PRs to add links to this in the tutorials repo. I'll need to find out where docs authors reference the available components to use 🤔

src/components/dev-dot-content/mdx-components/mdx-alert/mdx-inline-alert.module.css
kendallstrautman: I added some examples in the authoring docs of the cases where these styles are needed, mainly when mdx syntax is used instead of plain strings.

src/components/dev-dot-content/mdx-components/mdx-alert/docs.mdx
```diff
@@ -68,33 +68,72 @@ The `MdxInlineAlert` component is used in long-form MDX content to highlight var
 	soufflé dessert chocolate gingerbread pie.
 </MdxInlineAlert>
 
-## Examples
+## Authoring in MDX
```
kendallstrautman, line 71: Link to these docs in tutorials repo here https://github.com/hashicorp/tutorials/pull/1050

src/components/dev-dot-content/mdx-components/mdx-alert/mdx-inline-alert.module.css
kendallstrautman: Applying the body-200 styles to plain text alerts would definitely make things simpler, but then the alert title needs to be adjusted so it doesn't look out of balance. 

It's tricky because the ideal use case is just the short plain text string, but we need this type override anytime md syntax is used. So for example if there is a link in the body, but it is still a short string, a `p` and `a` tag will be generated that the larger styles apply.

kendallstrautman: Also when I dig through the content with the old syntax, there are soooo many examples of varied content being used in these alerts. Again I don't want to encourage the use this way but we need to be realistic for how authors will actually use these.

zchsh: ☝️ I think that's mainly my concern - I'm a little worried it might not be realistic to think we'll be able to fully transition away from images, lists, and other varied content within alerts. If that's the case, then I think that varied content within alerts might be something we want to dive into a little deeper. Less of an immediate blocking concern, and more of a future thought!

💭  Related thought: if we want to avoid varied content in these new alerts, and are strongly aiming for alerts to be text-only (but allowing bold, italics, links, and other "inline" markdown), would it make sense for the new MDX component prevent use of anything but `<p />` children? This could similar to how eg. `CodeBlockConfig` or `ImageConfig` throws an error for unexpected children. (This might not be a great idea if we're aiming to really push for migrating old alert sigils to the new MDX syntax... or if it just feels too restrictive generally... but it might be a possibility if we're aiming to refactor content to meet a new "alerts should have text only" authoring requirement).

kendallstrautman: Good idea @zchsh, something @braydenlove and I have been chatting through. I think I don't know enough about the authors needs to feel good about putting hard restrictions on this component like that. But I guess we should do whatever will give us the least headache later on. If we leave it open for authors and then come back to adjust styles for the varied content situations based on real world use cases, then it could be a better time to make those calls.

If we restrict this new component but we still support the old one (even though we call it deprecated), and if the new one is too restrictive so authors continue to use the old syntax, that would be not good. So imo we should leave this open, but encourage folks to not use images in them. And if we need to adjust styles later on based on how they're actually using it, I think thats ok.

zchsh: 💯 I dig the above logic, +1 on not being too restrictive. Thanks y'all for the extra context 🙇

ashleemboyer: I think the easiest path forward is to start flexible so we're able to support existing content. Then when content-conformance checks are stable, we can add a rule for what kind of content is allowed inside of these alerts. 😊

ashleemboyer (approved):

src/components/dev-dot-content/mdx-components/mdx-alert/docs.mdx
```diff
@@ -68,33 +68,72 @@ The `MdxInlineAlert` component is used in long-form MDX content to highlight var
...
+The only available prop to pass in authorable contexts with these specific variants is `title`. _**Note**_: When using markdown syntax within the body of an alert, there needs to be a space between the opening / closing
+tags and the content — similar to the Tabs or Accordion components.
+
+### Guidelines
+
+- Keep the title short, as this will be the most prominent element when users scan the alert.
+- Do not end the title with a period, regardless if you use a custom headline via title=“my title”
+- Alert messages should be short but clear and descriptive enough to explain what is happening. They should guide users on how to prevent or fix the issue.
+- We recommend keeping messages under 270 characters.
+- Text formatting capabilities via markdown syntax such as inline links, bold, italic, and lists are supported. Please use bold/italics sparingly to ensure consistency.
```
kendallstrautman, line 84: TODO - add note here re: not using images

kendallstrautman, line 84:
```suggestion
- Text formatting capabilities via markdown syntax such as inline links, bold, italic, and lists are supported. Please use bold/italics sparingly to ensure consistency.
- Please don't use images within the alerts, render them below or above the alert instead.
```

kendallstrautman: @braydenlove thoughts on wording?

PR #1524, opened 2023-01-12
[e2e link rewrites] Wrap fetch that 404s in try/catch

ashleemboyer:
## 🗒️ What

Wraps a `fetch` that happens with `npm run test:e2e:linkRewrites:loadData` with `try/catch`.

## 🤷 Why

These can sometimes 404 ([see example](https://github.com/hashicorp/boundary/actions/runs/3898125840/jobs/6656531561#step:6:17)), which is OK for this script. Wrapping in `try/catch` will keep the workflows that call this from unnecessarily failing.


kendallstrautman (approved):
👍🏻

ashleemboyer:
Disabled auto-merge so I can merge #1526 first 😊

PR #1526, opened 2023-01-12
Adding "Redeploy Content Repo Previews" workflow

ashleemboyer:
## 🗒️ What

- Adds a new workflow that triggers redeploys various content repos' `main` branches.

## 🤷 Why

- This will speed up some of the process for the docs content link rewrites project. In order for the E2E tests to be accurate in each main migration PR, the `main` branch PR's URL for a repo needs to be built from the latest of `dev-portal/main`. With bug fixes happening every few days that fix issues identified in failing E2E tests, it takes a lot of time and context switching to make sure the `main` branch preview URLs get updated.
- A side benefit of this is that content repos's `main` branch preview URLs will always have the latest `dev-portal/main` features & fixes without having to update their `main` branches or perform manual deploys in the Vercel UI.

## 🛠️ How

For each content repo:

- Created a [Deploy Hook](https://vercel.com/docs/concepts/git/deploy-hooks) in the Vercel project's Git settings named `"Re-deploy when dev-portal/main updates"`
- Created an Actions Secret in the `dev-portal` repo settings named `"VERCEL_DEPLOY_HOOK_<vercel-project-slug>"`, with the value being the Deploy Hook URL
- Added a step in the new workflow that runs `curl -X POST` on each Deploy Hook URL through the GitHub `secrets` variable

## 🧪 Testing

I tested this workflow by:

- [ ] Adding a temporary label trigger
- [ ] Adding & removing the test label after pushes to trigger the workflow

The main test to see if triggering the Deploy Hook worked was sucessful in:

- [Commit b6387b5e57270a5283430f5b2d563414907a07da](https://github.com/hashicorp/dev-portal/pull/1526/commits/b6387b5e57270a5283430f5b2d563414907a07da)
- [Run 3903606269](https://github.com/hashicorp/dev-portal/actions/runs/3903606269)

## 💭 Anything else?

- We should refine the triggers for this workflow to save on Vercel builds. To keep the docs content link rewrites project moving, it'd be helpful if we could do that in a separate PR. ❤️

### DIFF
--- a/.github/workflows/redeploy-content-repos.yml
+++ b/.github/workflows/redeploy-content-repos.yml
@@ -20,6 +20,8 @@ jobs:
   redeploy-content-repo-previews:
     runs-on: ubuntu-latest
     steps:
+      - name: Trigger "boundary" deploy
+        run: curl -X POST ${{ secrets.VERCEL_DEPLOY_HOOK_BOUNDARY }}
       - name: Trigger "cloud-docs-developer-preview" deploy
         run: curl -X POST ${{ secrets.VERCEL_DEPLOY_HOOK_CLOUD_DOCS_DEVELOPER_PREVIEW }}
       - name: Trigger "vagrant" deploy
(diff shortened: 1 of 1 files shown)


BRKalow (approved):
Good thinking! 🚀

kendallstrautman (approved):
💯 🥳

PR #1529, opened 2023-01-13
feat: plug in integrations landing header

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

This PR adds a branded header card to `/<product>/integrations` landing pages.

## 🛠️ How

- Adds a generic `BrandedCard` component in `components/branded-card`
    - Renders some `children` content in a card with a branded background
    - Derived from the `BrandedCallout` view-specific component from tutorial landing pages
- Adds a `BrandedHeaderCard` components, specific to `views/product-integrations-landing`
- Updates the integrations landing `getStaticProps` to return full `product` data
    - `product.slug` & `product.name` are needed to render the `BrandedHeaderCard`
    - 🎗️  Passing the full product data means **the top nav changes to be product-specific**.
- Adds the `BrandedHeaderCard` to the top of `views/product-integrations-landing`

> **Note**: This PR also refactors the tutorial landing component `BrandedCallout` to use the now-abstracted `BrandedCard` component. This is slightly tangential to this PR, but felt worth doing to avoid duplicate asset loading and to make it easier to maintain branded card backgrounds if / when they're updated.

## 📸 Design Screenshots

🎨 [Figma design spec][figma]

| Page | Before | After |
| - | - | - |
| [/vault/integrations][/vault/integrations] | vault-before | vault-after |
| [/waypoint/integrations][/waypoint/integrations] | waypoint-before | waypoint-after |

## 🧪 Testing

- [ ] Visit integrations landing pages, namely [/vault/integrations][/vault/integrations] & [/waypoint/integrations][/waypoint/integrations]
    - A branded card should appear at the top of the page, with the expected appearance and content
    - The branded card should render an `<h1 />`
    - The old `Integrations` `<h1 />`  should be removed from the page
- [ ] Confirm that the product-specific top-nav renders as expected on `/<product>/integrations` pages
    - See the `Anything else` section below for details
- [ ] Visit a tutorials landing page, such as [/vault/tutorials][/vault/tutorials]
    - The branded card at the top of the page, which has been refactored as part of this PR, should appear identical to the branded card [upstream][main/vault/tutorials]

## 💭 Anything else?

> **Warning**

The changes in this PR also result in the top nav on `/<product>/integrations` becoming _product-specific_. This seems like the desired top nav configuration (detail of screenshots in the `Design Screenshots` section): 

| | |
| - | - |
| Before | vault-nav-before  |
| After | vault-nav-after |

If we want to avoid this change, we could revise the `getStaticProps` for the page to return a renamed `productData` prop or something. It's the `product` naming specifically that triggers [currentProduct context setting in _app.tsx](https://github.com/hashicorp/dev-portal/blob/0fe3f22ed0b473abcdcf212cfe3408ebeff1f4a7/src/pages/_app.tsx#L90).

[figma]: https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=17267%3A218514&t=38PfUjXUEOnUfHw6-4
[task]: https://app.asana.com/0/1202097197789424/1203709015029198/f
[preview]: https://dev-portal-git-zsintegrations-landing-header-hashicorp.vercel.app/
[/vault/integrations]: https://dev-portal-git-zsintegrations-landing-header-hashicorp.vercel.app/vault/integrations
[/waypoint/integrations]: https://dev-portal-git-zsintegrations-landing-header-hashicorp.vercel.app/waypoint/integrations
[/vault/tutorials]: https://dev-portal-git-zsintegrations-landing-header-hashicorp.vercel.app/vault/tutorials
[main/vault/tutorials]: https://developer.hashicorp.com/vault/tutorials


zchsh:
@BrandonRomano This PR implements the branded Landing Header for `/<product>/integrations` pages.

🎗️ Wanted to call out: it also updates the top nav of `/<product>/integrations` pages to be product-specific. Hopefully this is as intended, but figured worth flagging just in case 🙇

src/views/product-integrations-landing/index.tsx
```diff
@@ -26,6 +30,11 @@ export default function ProductIntegrationsLanding({
 				sidecarSlot={<></>}
 			>
 				<div className={s.mainArea}>
+					<BrandedHeaderCard
+						productSlug={product.slug}
+						heading={`${product.name} Integrations`}
+						description={`A curated collection of official, partner, and community ${product.name} Integrations.`}
+					/>
```
zchsh, line 37: Forgot to call this out - we're generating the heading and description based on the product. If we're expecting this to be less strictly templated for future products, we could consider moving to a more directly authorable interface. For now this felt sufficient 👍

BrandonRomano: Ah, I had a comment in the Asana ticket that was probably burred in the long description.

> I think the title of this page will always be ${product.name} Integrations, but I think the description may be something we want to be configurable on a per-product basis.  We probably will want to add a key to the [integrations config](https://github.com/hashicorp/dev-portal/blob/product-integrations/src/data/vault.json#L169-L185) which will allow authors to specify the description they want here.

I already am in talks with the Vault PMs who would like to change the description here to something more bespoke, so maybe we adjust in this PR before merging to be configuration driven.

BrandonRomano (changes requested):
Let's move the description to the config file before merging (see comment)!

src/views/product-integrations-landing/index.tsx
zchsh: ah cool, sorry I missed that 😬 This sounds great to me, I'll adjust now!

BrandonRomano: No worries at all!

zchsh:
@BrandonRomano I've updated so the `description` for the header element can be authored via `integrationsConfig.description` in the `src/data/<product>.json` file. I've also updated to make it optional, by default no description will be shown.

BrandonRomano (approved):
That'll do it!

PR #1530, opened 2023-01-13
Abstracting logic in tutorials & docs link rewrite plugins for reuse

ashleemboyer:
## 🗒️ What/Why

- Moves `remarkPluginAdjustLinkUrls` to `src/lib/remark-plugins/remark-plugin-adjust-link-urls/`, so it is co-located with other remark plugins and so it could have additional files associated with it under a shared folder
- Abstracts the body of `remarkPluginAdjustLinkUrls` to `processDocsLinkNode`, primarily so it can be reused be the link rewrite script
- Moves helpers that lived in the same file as `remarkPluginAdjustLinkUrls` to a `helpers.ts` file, because importing from the plugin file for the link rewrite script has dependency clashes 
- Abstracts the body of `rewriteTutorialLinksPlugin` to `rewriteTutorialsLink`, also so it can be reused by the link rewrite script

## 🧪 Testing

There is nothing to manually test here. We just need the automated tests for these plugins to continue passing.


src/lib/remark-plugins/remark-plugin-adjust-link-urls/helpers.ts
```diff
@@ -12,6 +10,23 @@ const getIsInternalPath = (url: string) => {
...
+	node,
+	currentPath,
+	urlAdjustFn,
+}: {
+	node: Link | Definition
+	currentPath: string
+	urlAdjustFn: (url: string) => string
+}) => {
+	const urlToAdjust = preAdjustUrl({ currentPath, url: node.url })
+	return urlAdjustFn(urlToAdjust)
```
ashleemboyer, line 27: 🗒️ This is the old body of `remarkPluginAdjustLinkUrls`, only adjusted to return the result of `urlAdjustFn` rather than set the `node.url` variable to the result. The plugin handles setting the `node.url` variable to the result of this new function.

kendallstrautman (approved):
Clean refactor! 👏🏻

PR #1531, opened 2023-01-16
feat: add required-only boolean filter

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Implements a "Required only" filter dimension for integration component variables.

## 🤷 Why

To make it easier to see which variables are required, and to generally make it easier for visitors to find the variable information they're looking for. 

## 🛠️ How

- Implements a `requiredOnly` filter in `SearchableVariableGroupList`
- Splits out `applyQueryFilter`, `applyRequiredFilter`, and `includeMatchAncestors` helpers
- Implements styling to position the "Required only" checkbox across viewports

## 📸 Design Screenshots

🎨 [Figma designs][figma]

### Design

designs-large-viewports

### Implementation

https://user-images.githubusercontent.com/4624598/213031942-91dd16ca-6a7b-4614-b476-e2a974da280b.mp4

#### Small viewports

screenshot of required filter on a 320px wide viewport

## 🧪 Testing

- [ ] Visit an integrations component page, such as [/waypoint/integrations/aws-ecs/latest/components/platform][/waypoint/integrations/aws-ecs/latest/components/platform]
    - A new `Required only` filter should be visible
    - The `Required only` filter should function as expected, hiding any optional properties
    - The `Required only` filter should function as expected when combined with search query filtering
    - The `Required only` filter should not be shown where it would not be relevant, such as in `Output` tabs

## 💭 Anything else?

> **Note**: we haven't yet implemented [HDS Toggle](https://design-system-components-hashicorp.vercel.app/components/form/toggle) in Dev Dot, so this PR uses a checkbox input instead, as suggested in [🎟 the Asana task][task]. We've set up a separate task, [🎟 Implement HDS Form::Toggle component](https://app.asana.com/0/1202097197789424/1203757707724870/f), to implement this later.

> **Note**: some mobile styles aren't fully finished. We have a task [🎟 Mobile styling for Integrations Filtering](https://app.asana.com/0/1202626741675858/1203705051952457/f) that I think covers this.

[task]: https://app.asana.com/0/1202626741675858/1203684611651045/f
[figma]: https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=17236%3A219695&t=1K1EbppAzsrhhCeY-0
[preview]: https://dev-portal-git-zsintegrations-required-only-hashicorp.vercel.app
[/waypoint/integrations/aws-ecs/latest/components/platform]: https://dev-portal-git-zsintegrations-required-only-hashicorp.vercel.app/waypoint/integrations/aws-ecs/latest/components/platform

---
  - https://app.asana.com/0/0/1203684611651045


src/views/product-integration/component-view/components/searchable-variable-group-list/style.module.css
```diff
@@ -31,6 +45,6 @@
 	font-size: 12px;
 	text-transform: uppercase;
 	font-weight: 600;
-	margin-top: 8px;
-	margin-bottom: 24px;
+	margin-top: 24px;
+	margin-bottom: 12px;
```
zchsh, line 49: Tweaked these spacing values based on the closest final-looking design artboard I could find:

mobile-spacing-design

zchsh:
@BrandonRomano I just realized I should probably refactor this slightly to split out the "include parent properties of matches" from the query filtering. It seems like an edge case, but if there's an optional object variable that has a nested required variable for some reason, I think it won't display as expected with the PR as-is 🤔 So moving this back to draft, but should be ready for proper review shortly!

zchsh:
@BrandonRomano  Okey dokes, have updated to split out an `includeMatchAncestors` helper, which is now applied after filtering. Realized during the process that I think I was wrong, things would've worked the same, but definitely felt less clear with the way I previously had the "include ancestors" part wrapped up in the query filtering.

(tbh this is probably not super relevant unless you were already reviewing the PR before this change, so feel free to ignore this comment unless the extra context feels helpful 🙇)

src/views/product-integration/component-view/components/searchable-variable-group-list/style.module.css
BrandonRomano: I know @thiskevinwang is working on mobile for this one here over here: https://github.com/hashicorp/dev-portal/pull/1533.

BrandonRomano (commented):
This is looking good; I think two things:

- [ ] Ensure no conflict w/ Kevin's styling PR
- [ ] When selecting "Required only", I think that we should not highlight everything (I think highlighting should be reserved for search match).

src/views/product-integration/component-view/components/searchable-variable-group-list/style.module.css
thiskevinwang: 💭 (no action) Looks like we're not touching the same files, but there are certainly some duplicated styles.

zchsh: Indeed - the styles for `.searchBar` have a `/* TODO */` comment that notes some duplication, for one. I can set up a quick task & work on a follow-up PR to clean those up if that'd be helpful & worth prioritizing? cc @BrandonRomano

BrandonRomano: I think that's a good task to **scope**.  I would say tag it as low though for now, and maybe let's get to that one a little later?  I have a few more things in the queue that I would say are higher priority.

That being said, if you find yourself particularly energized to tackle that sooner rather than later, I never want to get in the way of motivation to make improvements (so you could tackle it soon if you really wanted to).

zchsh:
@BrandonRomano I think this is properly ready for review now:

> - [ ] Ensure no conflict w/ Kevin's styling PR

Looks like we're not touching any of the same files, so I think we're good to go! #1533 has now merged into `product-integrations`, and have merged in changes into this branch 👍

> - [ ]  When selecting "Required only", I think that we should not highlight everything (I think highlighting should be reserved for search match).

Great call, ty 💖 I've updated the codez here to only highlight `searchQuery` matches 👍 Have also updated the screengrab in the PR description as a demo.

BrandonRomano (approved):
Sweet, works as expected.  Code looks good from my end!  :shipit:

PR #1535, opened 2023-01-17
fix: scroll to page top on pagination change

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Updates the `PaginatedIntegrationsList` to scroll to the top of the page when pages are changed.

## 🛠️ How

- Adds a `useEffect` hook that calls `window.scrollTo(0, 0)` when `currentPage` changes after the initial render

## 🧪 Testing

- [ ] Visit a page with `PaginatedIntegrationsList`, such as [/waypoint/integrations][/waypoint/integrations]
    - On initial render, scroll position should not change. For example, if you scroll partway down the page, and then reload the page, the browser behaviour of remembering your scroll position should still work.
    - When interacting with the pagination bar, if the current page is changed, the window should scroll back up to the top

### Note on positioning

This PR implements scrolling to the very top of the page. If we'd prefer to scroll to a different position, such as the top of the paginated results, we could add a `ref` that we tack onto a target element, and call `ref.current.scrollIntoView()` rather than `window.scrollTo(0, 0)`.

### Thought on alternative approach

In implementing and testing this, I found the behaviour after the change to still feel a bit jarring. Another option that might address the same concern as the original issue could be to move the pagination controls to _above_ the results, rather than below.

<details>
<summary>📺 ⬇️ Demo of pagination at bottom, with scroll reset <em>(this PR)</em></summary>

https://user-images.githubusercontent.com/4624598/213020247-88c849b3-ea44-49b6-a5d7-80dd83b3ebdf.mp4

</details>

<details>
<summary>📺 ⬆️ Demo of pagination at top, without scroll reset<em>(potential alternate approach)</em></summary>

https://user-images.githubusercontent.com/4624598/213020239-37465ee6-c184-4d38-87b6-51dfb18d7227.mp4

</details>

As a related thought, which might be helpful with pagination controls above or below the results list, we could reconsider the default number of items per page - or perhaps even make the default number of items per page dynamically adjust based on viewport height, such that all results on a "page" can be viewed within one viewport height. This could help alleviate the relatively cumbersome process of `scroll → change page → scroll → ... etc` when page results are too tall to be viewed in a single viewport height.

[preview]: https://dev-portal-git-zsintegrations-paginator-scroll-reset-hashicorp.vercel.app/
[task]: https://app.asana.com/0/1202626741675858/1203717383445373/f
[/waypoint/integrations]: https://dev-portal-git-zsintegrations-paginator-scroll-reset-hashicorp.vercel.app/waypoint/integrations

---
  - https://app.asana.com/0/0/1203717383445373

### DIFF
--- a/src/views/product-integrations-landing/components/paginated-integrations-list/index.tsx
+++ b/src/views/product-integrations-landing/components/paginated-integrations-list/index.tsx
@@ -58,6 +58,8 @@ export default function PaginatedIntegrationsList({
 		if (isFirstRender.current) {
 			isFirstRender.current = false
 		} else {
+			// Scroll to the top of the page
+			window.scrollTo(0, 0)
 			// Try to find the first result link, and focus it
 			const targetElement = containerRef.current?.querySelector('a')
 			if (targetElement) {
--- a/src/views/product-integrations-landing/components/paginated-integrations-list/index.tsx
+++ b/src/views/product-integrations-landing/components/paginated-integrations-list/index.tsx
@@ -1,4 +1,5 @@
 import { Integration } from 'lib/integrations-api-client/integration'
+import getFullNavHeaderHeight from 'lib/get-full-nav-header-height'
 import { useEffect, useRef, useState } from 'react'
 import IntegrationsList from '../integrations-list'
 import s from './style.module.css'
@@ -58,12 +59,18 @@ export default function PaginatedIntegrationsList({
 		if (isFirstRender.current) {
 			isFirstRender.current = false
 		} else {
-			// Scroll to the top of the page
-			window.scrollTo(0, 0)
 			// Try to find the first result link, and focus it
 			const targetElement = containerRef.current?.querySelector('a')
 			if (targetElement) {
-				targetElement.focus()
+				targetElement.focus({ forceVisible: true })
+				/**
+				 * We need to scroll up a bit, as the focused item may be slightly
+				 * hidden behind the top navigation bar. Note this approach is slightly
+				 * brittle, as --navigationHeader. We also add an extra 64px,
+				 * which makes it more clear we've scrolled to the top of results.
+				 */
+				const navScrollCompensation = getFullNavHeaderHeight() + 64
+				window.scrollBy(0, navScrollCompensation * -1)
 			}
 		}
 	}, [currentPage])
--- a/src/views/product-integrations-landing/components/paginated-integrations-list/index.tsx
+++ b/src/views/product-integrations-landing/components/paginated-integrations-list/index.tsx
@@ -1,5 +1,5 @@
 import { Integration } from 'lib/integrations-api-client/integration'
-import { useEffect, useState } from 'react'
+import { useEffect, useRef, useState } from 'react'
 import IntegrationsList from '../integrations-list'
 import s from './style.module.css'
 import Pagination from 'components/pagination'
@@ -11,6 +11,7 @@ interface PaginatedIntegrationsListProps {
 export default function PaginatedIntegrationsList({
 	integrations,
 }: PaginatedIntegrationsListProps) {
+	const containerRef = useRef(null)
 	const [itemsPerPage, setItemsPerPage] = useState(8)
 	// Sort integrations alphabetically. Right now this is our
 	// preferred way of sorting. In the event we want to add different
@@ -40,13 +41,26 @@ export default function PaginatedIntegrationsList({
 		setCurrentPage(1)
 	}, [sortedIntegrations])
 
+	/**
+	 * If our pagination page changes, scroll up to the top of the wrapper.
+	 *
+	 * We also focus the search input, since otherwise, keyboard users would
+	 * be scrolled to the top of the page (due to scrollTo), and then
+	 * immediately scrolled to the bottom of the page (since )
+	 */
 	function setPageWithScrollReset(page: number) {
-		setCurrentPage(1)
+		setCurrentPage(page)
+		// Scroll to the top of the page
 		window.scrollTo(0, 0)
+		// Try to find the first result link, and focus it
+		const targetElement = containerRef.current?.querySelector('a')
+		if (targetElement) {
+			targetElement.focus({ focusVisible: true })
+		}
 	}
 
 	return (
-		<div className={s.paginatedIntegrationsList}>
+		<div className={s.paginatedIntegrationsList} ref={containerRef}>
 			<IntegrationsList integrations={currentPageIntegrations} />
 			{/* It's possible that all integrations display on a single page
       in that case, just don't even show the paginator. */}
--- a/src/views/product-integrations-landing/components/paginated-integrations-list/index.tsx
+++ b/src/views/product-integrations-landing/components/paginated-integrations-list/index.tsx
@@ -12,6 +12,7 @@ interface PaginatedIntegrationsListProps {
 export default function PaginatedIntegrationsList({
 	integrations,
 }: PaginatedIntegrationsListProps) {
+	const isFirstRender = useRef(true)
 	const containerRef = useRef(null)
 	const [itemsPerPage, setItemsPerPage] = useState(8)
 	// Sort integrations alphabetically. Right now this is our
@@ -49,16 +50,17 @@ export default function PaginatedIntegrationsList({
 	 * be scrolled to the top of the page (due to scrollTo), and then
 	 * immediately scrolled to the bottom of the page (since )
 	 */
-	function setPageWithScrollReset(page: number) {
-		setCurrentPage(page)
-		// Scroll to the top of the page
-		window.scrollTo(0, 0)
-		// Try to find the first result link, and focus it
-		const targetElement = containerRef.current?.querySelector('a')
-		if (targetElement) {
-			targetElement.focus()
+	useEffect(() => {
+		if (isFirstRender.current) {
+			isFirstRender.current = false
+		} else {
+			// Try to find the first result link, and focus it
+			const targetElement = containerRef.current?.querySelector('a')
+			if (targetElement) {
+				targetElement.focus()
+			}
 		}
-	}
+	}, [currentPage])
 
 	const { isDesktop, isMobile, isTablet } = useDeviceSize()
 
@@ -73,7 +75,7 @@ export default function PaginatedIntegrationsList({
 						totalItems={integrations.length}
 						pageSize={itemsPerPage}
 						page={currentPage}
-						onPageChange={setPageWithScrollReset}
+						onPageChange={setCurrentPage}
 						onPageSizeChange={setItemsPerPage}
 					>
 						{(isDesktop || isTablet) && <Pagination.Info />}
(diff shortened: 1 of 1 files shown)


zchsh:
@BrandonRomano This PR implements the "scroll to top on page change" for `PaginatedIntegrationsList`. As mentioned in the task, I've implemented this specifically for `PaginatedIntegrationsList`, _not_ in a generic way for all pagination.

💭 Wanted to call out the potential alternate approach of moving pagination _above_ results. I added some details in the `Anything else` section in the PR description. Happy to modify this PR to take that approach, or add a `Post MVP` task to the board to explore the idea, or just drop it if it feels like a non-issue 👍

BrandonRomano (approved):
Nice and simple, nice work ❇️

ashleemboyer (changes requested):
While `scrollTo` works for sighted mouse users, it doesn't work well for other users, such as keyboard-only users. Here's a screen recording of how I tested:

- `TAB` to the `Pagination` component
- Activate the button for the second page, with `ENTER`/`SPACE`
- Observe the page scroll up
- Hit the `TAB` key again
- Observe being scrolled down to the bottom of the page again

https://user-images.githubusercontent.com/43934258/213021164-336509f8-cedf-4d30-8901-625c1ffdd624.mp4

We should be able to do this with some manual `.focus()` calls on either an existing element we can query for / using a `ref`.
  - I think we might also want to do this logic in the `onPageChange` callback `Pagination` exposes. This would prevent needing to keep track of the first render. What do you think?

src/views/product-integrations-landing/components/paginated-integrations-list/index.tsx
```diff
@@ -40,6 +41,18 @@ export default function PaginatedIntegrationsList({
...
+	 * If our pagination page changes, scroll up to the top of the wrapper
+	 * However, don't scroll to the wrapper if this is the first render.
+	 */
+	useEffect(() => {
+		if (isFirstRender.current) {
+			isFirstRender.current = false
+		} else {
+			window.scrollTo(0, 0)
+		}
+	}, [currentPage])
```
alexcarpenter, line 54: Can we rely on the `onPageChange` callback from pagination vs needing to detect first render?

BrandonRomano:
Great callout / nice catch @ashleemboyer!

zchsh:
@ashleemboyer Thanks for catching that 💖 We don't have much design guidance on the `focus()` behaviour after pagination changes, and Brayden is out sick today 🤔

I've revised this PR to focus the first `<a />` result link element in the results page after the `scrollTo` occurs. I tested that out and it seems to get the scrolling result we want (which didn't seem to fully happen on focus alone), without then jumping back to the pagination controls:

https://user-images.githubusercontent.com/4624598/213024908-989f52bd-9381-4330-9587-f64260d5d59a.mp4

I've also revised to use the `onPageChange` function, rather than a `useEffect` as previously 👍 (thanks @alexcarpenter for catching that too!)

Let me know if that looks alright 🙇

ashleemboyer:
@zchsh taking another look! 😊

ashleemboyer:
> @ashleemboyer Thanks for catching that 💖 We don't have much design guidance on the `focus()` behaviour after pagination changes, and Brayden is out sick today 🤔

Let's make a follow-up ticket to check in with the HDS team / Melanie Sumner on this UX! 👍🏻

> I've revised this PR to focus the first `<a />` result link element in the results page after the `scrollTo` occurs. I tested that out and it seems to get the scrolling result we want (which didn't seem to fully happen on focus alone), without then jumping back to the pagination controls:
> 
> I've also revised to use the `onPageChange` function, rather than a `useEffect` as previously 👍 (thanks @alexcarpenter for catching that too!)
> 
> Let me know if that looks alright 🙇

So as it turns out... we'll actually need to do that `isFirstRender` check you were initially doing in a `useEffect`. At the time when we call `.focus()` on the first `<a>` in the results list, we're calling it on the first `<a>` element listed _before_ the list re-renders after the page change happens. So then the element that was focused is removed, the browser keeps the focus in that area of the page, and then when we hit `TAB` again, the first element of the newly selected page gets focused. We shouldn't need `scrollTo` at all!

ashleemboyer:
> > @ashleemboyer Thanks for catching that 💖 We don't have much design guidance on the `focus()` behaviour after pagination changes, and Brayden is out sick today 🤔
> 
> Let's make a follow-up ticket to check in with the HDS team / Melanie Sumner on this UX! 👍🏻
> 
> > I've revised this PR to focus the first `<a />` result link element in the results page after the `scrollTo` occurs. I tested that out and it seems to get the scrolling result we want (which didn't seem to fully happen on focus alone), without then jumping back to the pagination controls:
> > I've also revised to use the `onPageChange` function, rather than a `useEffect` as previously 👍 (thanks @alexcarpenter for catching that too!)
> > Let me know if that looks alright 🙇
> 
> So as it turns out... we'll actually need to do that `isFirstRender` check you were initially doing in a `useEffect`. At the time when we call `.focus()` on the first `<a>` in the results list, we're calling it on the first `<a>` element listed _before_ the list re-renders after the page change happens. So then the element that was focused is removed, the browser keeps the focus in that area of the page, and then when we hit `TAB` again, the first element of the newly selected page gets focused. We shouldn't need `scrollTo` at all!

@zchsh I think maybe the easiest thing to do here, so that we can leverage the more intuitive `onPageChange`, is call `.focus()` on something that is not going to be removed from the view on page changes. For example, we could focus the search input in `onPageChange`, and that should always work. It may not be the best UX right away, but we can check in with the HDS team this week and follow-up?

zchsh:
@ashleemboyer Sounds good - one challenge with trying to focus the search input is that it is outside of the `PaginatedIntegrationsList` component. So I think we'd need a relatively significant refactor compared to what we have in this PR so far.

Do we feel like it's a requirement to leverage `onPageChange`? Or would it be acceptable to `useEffect` so that the `focus()` on the `<a />` element works as expected?

ashleemboyer:
> @ashleemboyer Sounds good - one challenge with trying to focus the search input is that it is outside of the `PaginatedIntegrationsList` component. So I think we'd need a relatively significant refactor compared to what we have in this PR so far.
> 
> Do we feel like it's a requirement to leverage `onPageChange`? Or would it be acceptable to `useEffect` so that the `focus()` on the `<a />` element works as expected?

@zchsh how about we go with the `useEffect` and then put a `TODO` comment above it so we don't lose track of the idea? Ideally we would leverage `onPageChange` because it makes it a little clearer why/when the focus is changing.

zchsh:
@ashleemboyer Sweet - I've updated the PR with the `isFirstRender` & `useEffect` approach, focusing the first result link when the page changes.

I noticed when I removed `window.scrollTo`, that the scroll position didn't seem to be exactly what we're looking for - sometimes the element is only partially visible:

https://user-images.githubusercontent.com/4624598/213031310-825262b3-f0e6-42f9-a1f2-79ba5c92daf4.mp4

With that in mind, I've kept `scrollTo` in place for now, for this result:

https://user-images.githubusercontent.com/4624598/213031334-a7eeae3d-ad8c-44e9-b075-9e615ea4d3cd.mp4

Let me know if that seems alright. Happy to explore further.

zchsh:
Forgot to mention: I've also set up a task 🎟 [[discuss] Determine desired focus behaviour for HDS pagination](https://app.asana.com/0/1202097197789424/1203752518527704/f) to dig into the desired default behaviour for HDS pagination.

ashleemboyer:
> @ashleemboyer Sweet - I've updated the PR with the `isFirstRender` & `useEffect` approach, focusing the first result link when the page changes.
> 
> I noticed when I removed `window.scrollTo`, that the scroll position didn't seem to be exactly what we're looking for - sometimes the element is only partially visible:
> 
>  2023-01-17-without-scroll-to_opt.mp4 
> With that in mind, I've kept `scrollTo` in place for now, for this result:
> 
>  2023-01-17-with-scroll-to_opt.mp4 
> Let me know if that seems alright. Happy to explore further.

@zchsh I imagine this is happening for the same reason we have the scroll offsets on heading elements. Could you try using `window.scrollBy` to scroll up a little bit after calling `.focus()`, rather than scrolling to the top of the page every time? I think the latter may be more brittle when it comes to supporting different viewport sizes and page zooms.

src/views/product-integrations-landing/components/paginated-integrations-list/index.tsx
```diff
@@ -58,12 +59,18 @@ export default function PaginatedIntegrationsList({
 		if (isFirstRender.current) {
 			isFirstRender.current = false
 		} else {
-			// Scroll to the top of the page
-			window.scrollTo(0, 0)
 			// Try to find the first result link, and focus it
 			const targetElement = containerRef.current?.querySelector('a')
 			if (targetElement) {
-				targetElement.focus()
+				targetElement.focus({ forceVisible: true })
```
alexcarpenter, line 65: Since you have access to the `targetElement` can you use `targetElement.scrollIntoView()`?

zchsh:
@ashleemboyer ah, that makes sense! I've updated to use `scrollBy`, offsetting based on the `--navigation-header-height` variable. Let me know if this looks alright.

https://user-images.githubusercontent.com/4624598/213033602-f71ae2ad-563c-4b02-8686-65ed3e23af7a.mp4

src/views/product-integrations-landing/components/paginated-integrations-list/index.tsx
```diff
@@ -58,12 +59,18 @@ export default function PaginatedIntegrationsList({
...
+				/**
+				 * We need to scroll up a bit, as the focused item may be slightly
+				 * hidden behind the top navigation bar. Note this approach is slightly
+				 * brittle, as --navigationHeader. We also add an extra 64px,
+				 * which makes it more clear we've scrolled to the top of results.
+				 */
+				const navScrollCompensation = getFullNavHeaderHeight() + 64
+				window.scrollBy(0, navScrollCompensation * -1)
```
ashleemboyer, line 73: LGTM! ✨

ashleemboyer (approved):

src/views/product-integrations-landing/components/paginated-integrations-list/index.tsx
zchsh: I think `.focus()` does this automatically (in most browsers, at least). As Ashlee mentioned, I think the issue in the screengrab I shared earlier was that while we're already scrolling the element into view, a good portion of it is hidden behind the stick top nav bar.

I've updated to grab the (brittle, imperfect, but still useful) `--navigation-header-height` variable, and added that to a `scrollBy` offset (as Ashlee suggested! 💯 ). I guess we could also do `targetElement.scrollIntoView()` before that `scrollBy` as a kind of insurance, but I don't think it's necessary.

alexcarpenter: ah interesting, TIL scrollBy

PR #1536, opened 2023-01-17
chore: adds heading anchors to certifications accordions

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksadd-anchors-hashicorp.vercel.app/certifications/infrastructure-automation) 🔎
- [Asana task](https://app.asana.com/0/1202801212949828/1203714554456192) 🎟️

## 🗒️ What

Adds anchor links to the accordion headings so the education team can directly link to a specific program details. 

## 🧪 Testing

- Visit a [certifications program page](https://dev-portal-git-ksadd-anchors-hashicorp.vercel.app/certifications/infrastructure-automation)
- Click on an accordion heading, validate that an anchor link renders in your browser tab
- Visit the [anchor link directly](https://dev-portal-git-ksadd-anchors-hashicorp.vercel.app/certifications/infrastructure-automation#terraform-associate-(003)-details) to make sure it works. 

## 💭 Anything else?

We should decide whether we want to render the link icon, similar to how we do anchor links in long form content.

### DIFF
--- a/src/views/certifications/views/[slug]/components/heading-permalink/heading-permalink.module.css
+++ b/src/views/certifications/views/[slug]/components/heading-permalink/heading-permalink.module.css
@@ -1,3 +1,9 @@
+/** 
+* The permalink icon focus / hover styles should reflect 
+* the same behavior as the mdx-heading-permalink component
+* /src/components/dev-dot-content/mdx-components/mdx-heading-permalink
+**/
+
 .heading {
 	composes: hds-typography-display-400 from global;
 	composes: g-offset-scroll-margin-top from global;
@@ -28,11 +34,6 @@
 		}
 	}
 }
-
-/** 
-* These styles should reflect the same behavior as 
-* the mdx-heading-permalink component
-**/
 
 .icon {
 	width: 24px;
--- a/src/views/certifications/views/[slug]/components/heading-permalink/heading-permalink.module.css
+++ b/src/views/certifications/views/[slug]/components/heading-permalink/heading-permalink.module.css
@@ -1,0 +1,53 @@
+.heading {
+	composes: hds-typography-display-400 from global;
+	composes: g-offset-scroll-margin-top from global;
+	font-weight: var(--token-typography-font-weight-bold);
+	color: var(--token-color-foreground-strong);
+	margin: 40px 0 32px 0;
+
+	&:hover {
+		& .icon {
+			color: var(--token-color-foreground-action-hover);
+			opacity: 1;
+		}
+	}
+}
+
+.anchorLink {
+	composes: g-focus-ring-from-box-shadow from global;
+	border-radius: 6px;
+	color: inherit;
+	display: flex;
+	align-items: center;
+
+	&:focus {
+		& .icon {
+			color: var(--token-color-foreground-action-active);
+			background: var(--token-color-surface-faint);
+			opacity: 1;
+		}
+	}
+}
+
+/** 
+* These styles should reflect the same behavior as 
+* the mdx-heading-permalink component
+**/
+
+.icon {
+	width: 24px;
+	height: 24px;
+	border-radius: 6px;
+	margin-left: 8px;
+	opacity: 0;
+	color: var(--token-color-foreground-action);
+	padding: 4px;
+
+	@media (prefers-reduced-motion: no-preference) {
+		transition: opacity 0.2s ease;
+	}
+
+	&:hover {
+		box-shadow: var(--token-surface-mid-box-shadow);
+	}
+}
--- a/src/views/certifications/views/[slug]/components/heading-permalink/heading-permalink.module.css
+++ b/src/views/certifications/views/[slug]/components/heading-permalink/heading-permalink.module.css
@@ -10,6 +10,7 @@
 	font-weight: var(--token-typography-font-weight-bold);
 	color: var(--token-color-foreground-strong);
 	margin: 40px 0 32px 0;
+	width: fit-content;
 
 	&:hover {
 		& .icon {
--- a/src/views/certifications/views/[slug]/components/heading-permalink/index.tsx
+++ b/src/views/certifications/views/[slug]/components/heading-permalink/index.tsx
@@ -1,8 +1,10 @@
 import slugify from 'slugify'
 import { IconLink16 } from '@hashicorp/flight-icons/svg-react/link-16'
 import s from './heading-permalink.module.css'
+
 // Future thought: refactor this to use a shared `HeadingPermalink`
 // base component that can also be used in the mdx heading component
+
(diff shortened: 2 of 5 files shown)


src/views/certifications/views/[slug]/index.tsx
```diff
@@ -41,8 +47,20 @@ function CertificationProgramView({
 										versionTested={exam.versionTested}
 										slug={slug}
 									/>
-									<h2 className={s.examAccordionHeading}>
-										{`${title}${examCode ? ` ${examCode}` : ''} Details`}
+
+									<h2
+										className={classNames(
+											s.examAccordionHeading,
+											'g-offset-scroll-margin-top'
```
Comment (kendallstrautman) [inline src/views/certifications/views/[slug]/index.tsx:54]: Is this the right way to set the offset? I just copied what was done with the long form content anchor links.

Comment (zchsh) [reply src/views/certifications/views/[slug]/index.tsx:54]: This looks right to me!

⛏️ Nitpick, but personally I dig doing `composes: g-offset-scroll-margin-top from global` in the CSS for consistency & to keep the JSX just a lil cleaner... but very nitpicky, seems functional either way so probably the farthest thing from blocking 👍

zchsh (approved):
🙌 This looks great to me, based on the testing steps!

💭 Great Q re: whether to render the link icon. I don't have all the context on timeline & priority for this change, but IMO that seems like something we could follow up with if we want. So approving this either way ✅

ashleemboyer:
> 🙌 This looks great to me, based on the testing steps!
> 
> 💭 Great Q re: whether to render the link icon. I don't have all the context on timeline & priority for this change, but IMO that seems like something we could follow up with if we want. So approving this either way ✅

+1 to this question! If we need the link icon like we have for other headings, then we probably need to leverage the `Heading` component and overwrite the styles it sets that we don't want to retain. One large benefit of the link icon is that that heading text will be selectable. With the code as-is on this preview, it's not easy to select the heading text. (I haven't found a way yet.)

kendallstrautman:
Okay sweet, pinging @emilypersson1 & @kellyschummer re: whether we want to render the link icon next to the heading.

emilypersson1 (approved):
I think we're okay without the link icon, but I'll wait for Kelly to confirm. Thank you so much!

kellyschummer (commented):
This is wonderful, thank you @kendallstrautman !! Everything is rendering correctly on my side. I'd like to keep the link icon for consistency and visual affordance. 
+1 to @ashleemboyer comment too.

src/views/certifications/views/[slug]/index.tsx
Comment (kendallstrautman) [reply src/views/certifications/views/[slug]/index.tsx:54]: wonderful suggestion, will do that!

src/views/certifications/views/[slug]/components/heading-permalink/heading-permalink.module.css
```diff
@@ -0,0 +1,54 @@
...
+**/
+
+.heading {
+	composes: hds-typography-display-400 from global;
+	composes: g-offset-scroll-margin-top from global;
+	font-weight: var(--token-typography-font-weight-bold);
+	color: var(--token-color-foreground-strong);
+	margin: 40px 0 32px 0;
+
+	&:hover {
```
Comment (kendallstrautman) [inline src/views/certifications/views/[slug]/components/heading-permalink/heading-permalink.module.css:14]: I referenced this behavior from the mdx permalinks component since that was Kelly's preference for the link icon!

src/views/certifications/views/[slug]/components/heading-permalink/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { IconLink16 } from '@hashicorp/flight-icons/svg-react/link-16'
+import s from './heading-permalink.module.css'
+
+export function HeadingPermalink({ heading }: { heading: string }) {
+	const id = `${slugify(heading, {
+		lower: true,
+	})}`
+	return (
+		<h2 className={s.heading} id={id}>
+			<a className={s.anchorLink} href={`#${id}`}>
```
Comment (kendallstrautman) [inline src/views/certifications/views/[slug]/components/heading-permalink/index.tsx:11]: I see the current anchor links plugin as an aria-label on the anchor. If we did that here it would read `${heading} permalink`. Would that be helpful? 💭

src/views/certifications/views/[slug]/program-view.module.css
```diff
@@ -10,13 +10,6 @@
 	margin-top: -72px;
 }
 
-.examAccordionHeading {
```
Comment (kendallstrautman) [inline src/views/certifications/views/[slug]/program-view.module.css:13]: Moved this into `heading-permalink`

kendallstrautman:
@ashleemboyer @kellyschummer @zchsh I added the icon styles.
Screen Shot 2023-01-18 at 4 02 12 PM

src/views/certifications/views/[slug]/components/heading-permalink/index.tsx
Comment (ashleemboyer) [reply src/views/certifications/views/[slug]/components/heading-permalink/index.tsx:11]: I don't think we need it here since the interactivity of this element isn't communicated until someone hovers over or focuses on this link. We'd also not want to include `"permalink"` for the voice control experience, since the Accessible Name of an element (can be set by `aria-label`) needs to match the visible label. The Accessible Name is already the heading text based on how that's calculated by browsers, so I think we're good here!

ashleemboyer (approved):
LGTM! 👍🏻

kellyschummer (approved):
Beautiful! Thank you!!

src/views/certifications/views/[slug]/components/heading-permalink/index.tsx
```diff
@@ -0,0 +1,17 @@
+import slugify from 'slugify'
+import { IconLink16 } from '@hashicorp/flight-icons/svg-react/link-16'
+import s from './heading-permalink.module.css'
+
```
Comment (kendallstrautman) [inline src/views/certifications/views/[slug]/components/heading-permalink/index.tsx:4]:
```suggestion
// Future thought: refactor this to use a shared `HeadingPermalink`
// base component that can also be used in the mdx heading component
```

src/views/certifications/views/[slug]/components/heading-permalink/index.tsx
```diff
@@ -0,0 +1,18 @@
+import slugify from 'slugify'
+import { IconLink16 } from '@hashicorp/flight-icons/svg-react/link-16'
+import s from './heading-permalink.module.css'
+// Future thought: refactor this to use a shared `HeadingPermalink`
+// base component that can also be used in the mdx heading component
+export function HeadingPermalink({ heading }: { heading: string }) {
```
Comment (kendallstrautman) [inline src/views/certifications/views/[slug]/components/heading-permalink/index.tsx:6]:
```suggestion
import s from './heading-permalink.module.css'

// Future thought: refactor this to use a shared `HeadingPermalink`
// base component that can also be used in the mdx heading component

export function HeadingPermalink({ heading }: { heading: string }) {
```

kendallstrautman:
[Follow up ticket](https://app.asana.com/0/1202097197789424/1203754924050712) to refactor this to use a shared component with the mdx-heading-permalink.

PR #1537, opened 2023-01-18
feat: optional sidecarSlot on integrations landing

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Updates the integration landing view to remove blank space where the sidecar would be. This increases the space available for the main content area.

## 🤷 Why

To better match design specs, and avoid an off-centered main content area.

## 🛠️ How

- In `SidebarSidecarLayout`, update the `sidecarSlot` prop to be truly optional (can be omitted from props, rather than passing `null` or `<></>`)
- Use this new capability for the `ProductIntegrationsLanding` view

## 📸 Design Screenshots

🎨 [Figma link][figma]

### Design

CleanShot 2023-01-18 at 10 57 09@2x

#### Before

https://user-images.githubusercontent.com/4624598/213223147-05240e69-e3fa-4207-b742-bacb0f0838ca.mp4

#### After

https://user-images.githubusercontent.com/4624598/213288686-b61a0a56-48f0-44f1-8f44-5fc461f1c914.mp4

## 🧪 Testing

- [ ] Visit an integrations landing page, such as [/vault/integrations][/vault/integrations]
    - There should _not_ be a blank column-ish space where the sidecar usually is (compare to [upstream][upstream/vault/integrations])
    - When toggling between pages with and without a sidecar, such as between [/vault][/vault] and [/vault/integrations][/vault/integrations], the main content area should be right-aligned regardless of whether the sidecar is shown

## 💭 Anything else?

Not at the moment!

<details>
<summary>Previously `max-width` in this PR didn't match spec, but have since updated 👍</summary>

In this PR, we've used the same `main` content area max-width as in all other `SidebarSidecarLayout` instances. This differs from the max-width in Figma designs, which looks to be `1024px` (rather than the current `896px`). If implementing the wider `1024px` max-width is a strict requirement, we could either:

1. Add another option to our `mainWidth` prop (currently either `wide / 896px` or `narrow / 680px`), which would be low-effort but doesn't feel like a great solution long term, or
2. Refactor `SidebarSidecarLayout` for better composability, which would be higher effort (since our existing component has a few things we'd need to detangle)

For now I've gone with the third option of not worrying to hard about the exact pixel value of the `main` area max-width - especially considering we had recent pushes to reduce the main area width for readability, it seems like `1024px` wide cards might be slightly less readable than `896px` wide cards anyways.

</details>

[preview]: https://dev-portal-git-zsrm-sidecar-space-from-landing-hashicorp.vercel.app
[task]: https://app.asana.com/0/1202626741675858/1203746230354963/f
[figma]: https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=17267%3A218514&t=JYmt6p7PplivfYHF-4
[/vault/integrations]:  https://dev-portal-git-zsrm-sidecar-space-from-landing-hashicorp.vercel.app/vault/integrations
[upstream/vault/integrations]:  https://dev-portal-git-product-integrations-hashicorp.vercel.app/vault/integrations

---
  - https://app.asana.com/0/0/1203746230354963

### DIFF
--- a/src/layouts/sidebar-sidecar/index.tsx
+++ b/src/layouts/sidebar-sidecar/index.tsx
@@ -107,9 +107,15 @@ const SidebarSidecarLayoutContent = ({
 	} else {
 		SidecarContent = null
 	}
+	// We'll include the sidecar only if `SidecarContent` is not null
+	const includesSidecar = SidecarContent !== null
 
 	return (
-		<div className={classNames(s.root, s[`mainWidth-${mainWidth}`])}>
+		<div
+			className={classNames(s.root, s[`mainWidth-${mainWidth}`], {
+				[s.includesSidecar]: includesSidecar,
+			})}
+		>
 			<MobileMenuContainer className={s.mobileMenuContainer} ref={sidebarRef}>
 				<div className={s.sidebarContentWrapper}>
 					<MobileAuthenticationControls />
@@ -153,7 +159,7 @@ const SidebarSidecarLayoutContent = ({
 							/>
 						)}
 					</main>
-					{SidecarContent !== null ? (
+					{includesSidecar ? (
 						<div className={s.sidecarWrapper}>
 							<SidecarContent />
 						</div>
--- a/src/layouts/sidebar-sidecar/index.tsx
+++ b/src/layouts/sidebar-sidecar/index.tsx
@@ -92,14 +92,20 @@ const SidebarSidecarLayoutContent = ({
 		sidebarContent = <Sidebar {...sidebarProps} />
 	}
 
-	const SidecarContent = (): ReactElement => {
-		if (typeof sidecarSlot !== 'undefined') {
-			return sidecarSlot
-		}
-
-		return (
+	/**
+	 * Optionally render sidecar content. If no sidecar content is provided,
+	 * the main content will not have sidecar space rendered alongside it.
+	 */
+	let SidecarContent: () => ReactElement | null
+	if (typeof sidecarSlot !== 'undefined') {
+		SidecarContent = () => sidecarSlot
+	} else if (typeof headings !== 'undefined') {
+		// eslint-disable-next-line react/display-name
+		SidecarContent = () => (
 			<TableOfContents headings={filterTableOfContentsHeadings(headings)} />
 		)
+	} else {
+		SidecarContent = null
 	}
 
 	return (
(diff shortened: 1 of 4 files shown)


zchsh:
@BrandonRomano This PR implements the full removal of the sidecar slot from the integrations landing view.

For now, I haven't implemented a new `mainWidth` option on the `main` content area - but can do that if we definitely want a new `1024px` max-width setting for the content on the integrations landing pages. Let me know what you think 🙇

BrandonRomano:
> For now, I haven't implemented a new mainWidth option on the main content area - but can do that if we definitely want a new 1024px max-width setting for the content on the integrations landing pages. Let me know what you think 🙇

I think the new max width on the content container when no sidecar is passed should be this:

```css
max-width: calc(var(--main-element-max-width) + var(--sidecar-width) + 64px); /* 64px to account for padding */
```

This effectively has the main content area take up what the sidecar would have taken.

CleanShot 2023-01-18 at 09 16 14@2x

This will ensure the same alignment of content on the website. Note how the breadcrumb bars are placed in the same place when flipping between a tab with this width configured, and one with a sidecar:

https://user-images.githubusercontent.com/2105067/213250430-ef5e866e-7005-4e1f-96ef-f36923c611d0.mp4

We'll ultimately have to run this by design -- if this feels too wide, we'll likely have to set a wrapper container with this width to ensure alignment, and then have a left-aligned container inside with a separate max-width for the actual content area.

zchsh:
@BrandonRomano Thanks for catching that, with the context on alignment the large `max-width` makes a lot more sense 💖 

I've updated this PR to set the `max-width` as you recommended - have also updated the screengrab in the PR description to demo toggling between `/vault` (has sidecar) and `/vault/integrations` (no sidecar). Also onboard with waiting for final design review before shipping this, if we want 👍

BrandonRomano:
That's a good idea - let's just leave this open and lets have Brayden review here to see if this matches his original intent, once he returns.  I'll tag him for review.

BrandonRomano (approved):
This has my code approval, let's wait for design approval here too.

BrandonRomano:
Okey @zchsh I chatted with Brayden, and we ultimately landed on deciding to not do this and we'll keep the standard sidebar / sidecar layout (without a sidecar).  We'll probably in the future come up with something to put there.

Sorry for the fuss here!

PR #1538, opened 2023-01-18
[wip] patch: fix safari card link

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksfix-safari-card-link-hashicorp.vercel.app/packer) 🔎
- [Asana task](url) 🎟️

### DIFF
--- a/src/components/card-link/index.tsx
+++ b/src/components/card-link/index.tsx
@@ -40,7 +40,9 @@ const CardLink = ({
 				data-heap-track={`card-link ${dataHeapTrack ?? ''}`}
 				href={href}
 				target={target}
-			/>
+			>
+				&nbsp;
+			</Link>
 			{children}
 		</Card>
 	)
--- a/src/components/card-link/card-link.module.css
+++ b/src/components/card-link/card-link.module.css
@@ -33,6 +33,10 @@ style applied to anchor elements.
 		z-index: 0;
 	}
 
+	&:focus {
+		outline: transparent;
+	}
+
 	/* only styling the ::before element on focus */
 	&:focus::before {
 		outline: transparent;


ashleemboyer:
The elements are focusable, but cards aren't rendering the way we want to in either Safari or Chrome:

CleanShot 2023-01-17 at 21 12 28

kendallstrautman:
haha dammit! so weird they didn't look like that for me locally. I'll keep playing with this

src/components/card-link/index.tsx
```diff
@@ -33,14 +33,21 @@ const CardLink = ({
...
 			 */}
 			<Link
 				aria-label={ariaLabel}
 				className={s.anchor}
 				data-heap-track={`card-link ${dataHeapTrack ?? ''}`}
 				href={href}
 				target={target}
-			/>
+			>
+				<span>&nbsp;</span>
```
ashleemboyer, line 49: Should we set `aria-hidden` on this `span` for good measure? 🤔

ashleemboyer (approved):
Tested in Safari, Chrome, and Firefox and works for me! 👍🏻 

- Navigating each element with VO + arrow keys: only the text of the current element is read ("link" is not added to it) ✅ 
- Navigating each focusable element with the `TAB` and `SHIFT+TAB` keys: they're in the correct order (`CardLink` then bookmark button) and the focus rings still show as expected. ✅
- Using a mouse: the `CardLink` goes to the link, the bookmark button is clickable, and the `Badge`s still show a tooltip on hover ✅ 

https://user-images.githubusercontent.com/43934258/213271533-00f15826-e678-45ea-9789-f645d4bbac70.mp4

src/components/card-link/index.tsx
kendallstrautman: oh yes! somehow I removed that in my testing process. Will add back 😄

PR #1544, opened 2023-01-18
[preAdjustUrl] Handling developer.hashicorp.com links and no-dots folder relative links starting with "api"

ashleemboyer:
## 🗒️ What

- Adds test cases and handling for `developer.hashicorp.com` links and no-dots folder relative links starting with "api" (such as `api/some/docs/path`)

### DIFF
--- a/src/lib/remark-plugins/remark-plugin-adjust-link-urls/helpers.ts
+++ b/src/lib/remark-plugins/remark-plugin-adjust-link-urls/helpers.ts
@@ -1,6 +1,15 @@
 import path from 'path'
 import { Definition, Link } from 'mdast'
 
+const getIsInternalPath = (url: string) => {
+	try {
+		new URL(url)
+		return false
+	} catch (e) {
+		return true
+	}
+}
+
 /**
  * Handles processing a Link or Definition node in documentation content. Returns
  * the adjusted node URL.
@@ -16,6 +25,25 @@ const processDocsLinkNode = ({
 }) => {
 	const urlToAdjust = preAdjustUrl({ currentPath, url: node.url })
 	return urlAdjustFn(urlToAdjust)
+}
+
+/**
+ * Handles processing full URLs that may or may not be external to dev dot. If
+ * the URL is under the developer.hashicorp.com hostname, then the URL is
+ * returned as an internal path to dev dot. Otherwise, URLs that are totally
+ * external to dev dot are returned as-is.
+ */
+const handleFullUrl = ({ url }) => {
+	// Construct a URL object
+	const { hostname, pathname = '', search = '', hash = '' } = new URL(url)
+
+	// If it's a developer.hashicorp.com url, return it as an internal path
+	if (hostname === 'developer.hashicorp.com') {
+		return `${pathname}${search}${hash}`
+	}
+
+	// Otherwise, there is nothing to pre-adjust for external URLs
+	return url
 }
 
 /**
@@ -123,16 +151,10 @@ const preAdjustUrl = ({ currentPath, url }): string => {
 		return url
 	}
 
-	// Handle developer.hashicorp.com links if needed
-	try {
-		const { hostname, pathname, search = '', hash = '' } = new URL(url)
-		if (hostname === 'developer.hashicorp.com') {
-			return `${pathname}${search}${hash}`
-		} else {
-			return url
-		}
-	} catch (e) {
-		// is an internal path, so continue to rest of function
+	// Handle full URL that may link externally or internally
+	const isFullUrl = getIsInternalPath(url) === false
+	if (isFullUrl) {
+		return handleFullUrl({ url })
 	}
 
 	// Do nothing if url is a top-level path
--- a/src/lib/remark-plugins/remark-plugin-adjust-link-urls/helpers.ts
+++ b/src/lib/remark-plugins/remark-plugin-adjust-link-urls/helpers.ts
@@ -76,6 +76,44 @@ const handleDotSlashFolderRelativeUrl = ({
 }
 
 /**
+ * Handles folder-relative URLs that do not start with any dots syntax or
+ * punctuation.
+ */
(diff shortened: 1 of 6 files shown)


src/lib/remark-plugins/remark-plugin-adjust-link-urls/helpers.ts
```diff
@@ -84,6 +75,44 @@ const handleDotSlashFolderRelativeUrl = ({
 	return newUrl
 }
 
+/**
+ * Handles folder-relative URLs that do not start with any dots syntax or
+ * punctuation.
```
kendallstrautman, line 80:
```suggestion
 * Handles folder-relative URLs that do not start with any dots syntax or
 * punctuation. For example: api-docs/some-path 
```
An example could be nice!

src/lib/remark-plugins/remark-plugin-adjust-link-urls/helpers.ts
```diff
@@ -94,10 +123,16 @@ const preAdjustUrl = ({ currentPath, url }): string => {
 		return url
 	}
 
-	// Do nothing if url is not an internal path
-	const isInternalPath = getIsInternalPath(url)
```
kendallstrautman, line 98: At this point have we already weeded out external links?

src/lib/remark-plugins/remark-plugin-adjust-link-urls/helpers.ts
```diff
@@ -94,10 +123,16 @@ const preAdjustUrl = ({ currentPath, url }): string => {
...
-	if (!isInternalPath) {
-		return url
+	// Handle developer.hashicorp.com links if needed
+	try {
+		const { hostname, pathname, search = '', hash = '' } = new URL(url)
+		if (hostname === 'developer.hashicorp.com') {
+			return `${pathname}${search}${hash}`
```
kendallstrautman, line 130: Do you want to return the path here? Wouldn't you want to keep rewriting it?

src/lib/remark-plugins/remark-plugin-adjust-link-urls/helpers.ts
ashleemboyer: Updated here: https://github.com/hashicorp/dev-portal/pull/1544/commits/73d5a96615d7192c2e24b7eff5de35f93eda2a06

src/lib/remark-plugins/remark-plugin-adjust-link-urls/helpers.ts
ashleemboyer: The `try/catch` that is replacing this call to `getIsInternalPath` is functionally the same as the one in `getIsInternalPath`. In the `try`, if `new URL(url)` doesn't throw an error, then we know we're looking at an external link. So we check the `hostname` and only pre-adjust the URL if the `hostname` is `developer.hashicorp.com`. Otherwise, we don't do anything in the `catch` and continue on in the function because it's an internal url.

src/lib/remark-plugins/remark-plugin-adjust-link-urls/helpers.ts
ashleemboyer: This function only pre-adjusts URLs before they're actually rewritten. The `remarkPluginAdjustLinkUrls` calls `processDocsLinkNode` (defined at the top of this file), which calls `preAdjustUrl`, then passes the result of `preAdjustUrl` to `urlAdjustFn` (which does the rewriting).

ashleemboyer: No, because there's nothing to adjust for top-level paths. This can be seen in the next `if` block after this `tr/catch`:

```ts
// Do nothing if url is a top-level path
if (url.startsWith('/')) {
  return url
}
```

I'll add some more comments for clarity!

ashleemboyer: Added a `handleFullUrl` helper in https://github.com/hashicorp/dev-portal/pull/1544/commits/b0ba9d6a739e48aed9b7ff17b137c4afda1a616e!

src/lib/remark-plugins/remark-plugin-adjust-link-urls/helpers.ts
ashleemboyer: The `handleFullUrl` added in https://github.com/hashicorp/dev-portal/pull/1544/commits/b0ba9d6a739e48aed9b7ff17b137c4afda1a616e should help bring clarity here as well!

kendallstrautman: Those comments are great! Thank you 😄

src/lib/remark-plugins/remark-plugin-adjust-link-urls/helpers.ts
```diff
@@ -94,10 +151,10 @@ const preAdjustUrl = ({ currentPath, url }): string => {
...
 	}
 
-	// Do nothing if url is not an internal path
-	const isInternalPath = getIsInternalPath(url)
-	if (!isInternalPath) {
-		return url
+	// Handle full URL that may link externally or internally
+	const isFullUrl = getIsInternalPath(url) === false
+	if (isFullUrl) {
+		return handleFullUrl({ url })
```
kendallstrautman, line 157: Not sure if you need to pass it as an object?

kendallstrautman (approved):
Just one final comment. Thanks for bearing with me as I try to understand what's going on 🙇🏻

src/lib/remark-plugins/remark-plugin-adjust-link-urls/helpers.ts
ashleemboyer: Just doing it for consistency with the majority of the other helpers in this file. 😊

I've found that passing a single object param rather than many different types of params has are easier to work with long term. When adding new options, they don't change the API in such a way that other calls to the function have to be refactored.

PR #1549, opened 2023-01-19
[link rewrite plugins] Update to no longer rewrite .io home page links in content

ashleemboyer:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1202114367927919/1203750517664204/f) 🎟️

## 🗒️ What

- Adds test cases for both `rewriteTutorialLinksPlugin`- and `remarkPluginAdjustLinkUrls`- related logic to:
  - not rewrite .io home page links
  - also not rewrite `/` links (which were previously rewritten to .io home page links by `rewriteDocsUrl`)
- Updates the logic to no longer rewrite .io home page links to developer product landing pages

## 🤷 Why

To retain the marketing abilities that come with linking to .io home pages

### DIFF
--- a/src/views/docs-view/__tests__/multiple-remark-plugins.test.ts
+++ b/src/views/docs-view/__tests__/multiple-remark-plugins.test.ts
@@ -41,8 +41,12 @@ describe('multiple remark plugins', () => {
 
 		testEachCase(productUrlAdjuster, [
 			{
+				input: '/',
+				expected: 'https://waypointproject.io/',
+			},
+			{
 				input: 'https://waypointproject.io/',
-				expected: '/waypoint',
+				expected: 'https://waypointproject.io/',
 			},
 			{
 				input: '/waypoint',
--- a/src/views/docs-view/__tests__/multiple-remark-plugins.test.ts
+++ b/src/views/docs-view/__tests__/multiple-remark-plugins.test.ts
@@ -42,7 +42,7 @@ describe('multiple remark plugins', () => {
 		testEachCase(productUrlAdjuster, [
 			{
 				input: '/',
-				expected: 'https://waypointproject.io/',
+				expected: '/',
 			},
 			{
 				input: 'https://waypointproject.io/',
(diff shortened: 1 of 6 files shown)


src/views/docs-view/__tests__/multiple-remark-plugins.test.ts
```diff
@@ -40,9 +40,13 @@ describe('multiple remark plugins', () => {
 		const productUrlAdjuster = getProductUrlAdjuster(waypointProduct)
 
 		testEachCase(productUrlAdjuster, [
+			{
+				input: '/',
```
kendallstrautman, line 44: 🤔 hmmm I don't think we would want the root link to go to the .io page? Let's say a docs author wants to link to the dev dot home after the links are updated with this project, we wouldn't want it routing back to .io homepage right? 

This one is tricky because I see how with the previous docs versions, we would want what you have here. 🧠 What do you think?

ashleemboyer: Chatted over Slack, and we came up with an alternative solution!

- The 2 link rewrite plugins will _not_ rewrite `/` urls
- As part of the docs content link rewrite project, we will update all `/` urls in MDX content to reference the .io home pages as pre-work to migrating the links

Making the updates now 😊

ashleemboyer: @kendallstrautman Updates made in these three commits ✨ 

- https://github.com/hashicorp/dev-portal/pull/1549/commits/ad64660b6a587a3804db6d16c721a9d5db742a86
- https://github.com/hashicorp/dev-portal/pull/1549/commits/a57232c76b2219c4fcb14ceb75d91e06aa6fa01d
- https://github.com/hashicorp/dev-portal/pull/1549/commits/c01df755e72e29e193ee06b8cd84a74f5ebde4ed

kendallstrautman (approved):

PR #1552, opened 2023-01-23
feat: styled overview item in sidebars

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Updates sidebars on multiple pages to add a brand-themed highlighted item.

## 🤷 Why

To improve sidebar hierarchy.

## 🛠️ How

- Adds a new `SidebarNavHighlightItem` component
- Adds the overview item to sidebars as designed
- Splits out `generateCollectionSidebarNavData` helper function, since it's shared between views

## 📸 Design Screenshots

🎨 [Figma link][figma]

### Design Spec

2023-01-24-ia-quick-wins-export

### Implementation

| Page | Example | Implementation |
| - | - | - |
| Product landing | [/nomad][/nomad] | product-landing | 
| Docs landing | [/nomad/docs][/nomad/docs] | docs-landing |
| Docs page | [/nomad/api-docs/json-jobs][/nomad/api-docs/json-jobs] | api-page |
| Tutorials landing | [/nomad/tutorials][/nomad/tutorials] | tutorials-landing |
| Install view | [/nomad/downloads][/nomad/downloads] | downloads-view |
| Integrations view | [/vault/integrations][/vault/integrations] | integrations-after |

## 🧪 Testing

- [ ] Visit each of the pages listed in the above section
    - Each of the above pages should have a newly styled item at the top of the sidebar
    - The mobile menu should also show a newly styled item, in the same location
    - Outside of the changes, the pages should look & function in the same way

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/1203461693655903/1203524851357319/f
[preview]: https://dev-portal-git-zssidebar-hierarchy-hashicorp.vercel.app
[figma]: https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=17788%3A230417&t=tNfAxNFRPCcUV4KH-4
[/nomad]: https://dev-portal-git-zssidebar-hierarchy-hashicorp.vercel.app/nomad
[/nomad/docs]: https://dev-portal-git-zssidebar-hierarchy-hashicorp.vercel.app/nomad/docs
[/nomad/api-docs/json-jobs]: https://dev-portal-git-zssidebar-hierarchy-hashicorp.vercel.app/nomad/api-docs/json-jobs
[/nomad/tutorials]: https://dev-portal-git-zssidebar-hierarchy-hashicorp.vercel.app/nomad/tutorials
[/nomad/downloads]: https://dev-portal-git-zssidebar-hierarchy-hashicorp.vercel.app/nomad/downloads
[/vault/integrations]: https://dev-portal-git-zssidebar-hierarchy-hashicorp.vercel.app/vault/integrations

### DIFF
--- a/src/components/sidebar/components/sidebar-nav-highlight-item/index.tsx
+++ b/src/components/sidebar/components/sidebar-nav-highlight-item/index.tsx
@@ -1,3 +1,6 @@
+import Link from 'components/link'
+import classNames from 'classnames'
+import ProductIcon from 'components/product-icon'
 import { ProductSlug } from 'types/products'
 import s from './sidebar-nav-highlight-item.module.css'
 
@@ -17,12 +20,17 @@ export default function SidebarNavHighlightItem({
 	href: string
 }) {
 	return (
-		<div className={s.root}>
-			{text}
-			<br />
-			{`(fancy highlight item)`}
-			<br />
-			{`(theme: ${theme}, href: ${href})`}
-		</div>
+		<Link
+			// aria-current={ariaCurrent}
+			// aria-label={ariaLabel}
+			className={classNames(s.root, s[`theme-${theme}`])}
+			// data-heap-track="sidebar-nav-link-item"
+			href={href}
+			// opensInNewTab={isExternal}
+			// rel={rel}
+		>
+			<ProductIcon productSlug={theme} />
+			<span className={s.text}>{text}</span>
+		</Link>
 	)
 }
--- a/src/components/sidebar/components/sidebar-nav-highlight-item/index.tsx
+++ b/src/components/sidebar/components/sidebar-nav-highlight-item/index.tsx
@@ -14,14 +14,16 @@ export default function SidebarNavHighlightItem({
 	theme,
 	text,
 	href,
+	isActive,
 }: {
 	theme: ProductSlug
 	text: string
 	href: string
+	isActive?: boolean
 }) {
 	return (
 		<Link
-			// aria-current={ariaCurrent}
+			aria-current={isActive ? 'page' : undefined}
 			// aria-label={ariaLabel}
 			className={classNames(s.root, s[`theme-${theme}`])}
 			// data-heap-track="sidebar-nav-link-item"
--- a/src/components/sidebar/components/sidebar-nav-highlight-item/index.tsx
+++ b/src/components/sidebar/components/sidebar-nav-highlight-item/index.tsx
@@ -1,0 +1,28 @@
+import { ProductSlug } from 'types/products'
+import s from './sidebar-nav-highlight-item.module.css'
+
+/**
+ * Render a fancy-looking, product themed linked sidebar item.
+ *
+ * Intended to be used as a kind of title-ish element, to establish hierarchy
+ * within sidebar contents.
+ */
+export default function SidebarNavHighlightItem({
+	theme,
+	text,
+	href,
+}: {
+	theme: ProductSlug
+	text: string
+	href: string
+}) {
+	return (
+		<div className={s.root}>
+			{text}
+			<br />
+			{`(fancy highlight item)`}
+			<br />
+			{`(theme: ${theme}, href: ${href})`}
+		</div>
+	)
+}
--- a/src/components/sidebar/components/sidebar-nav-highlight-item/index.tsx
+++ b/src/components/sidebar/components/sidebar-nav-highlight-item/index.tsx
@@ -24,12 +24,8 @@ export default function SidebarNavHighlightItem({
 	return (
 		<Link
 			aria-current={isActive ? 'page' : undefined}
-			// aria-label={ariaLabel}
 			className={classNames(s.root, s[`theme-${theme}`])}
-			// data-heap-track="sidebar-nav-link-item"
 			href={href}
-			// opensInNewTab={isExternal}
-			// rel={rel}
 		>
 			<ProductIcon productSlug={theme} />
 			<span className={s.text}>{text}</span>
--- a/src/components/sidebar/components/sidebar-nav-highlight-item/index.tsx
+++ b/src/components/sidebar/components/sidebar-nav-highlight-item/index.tsx
@@ -1,7 +1,8 @@
 import Link from 'components/link'
 import classNames from 'classnames'
+import { isProductSlug } from 'lib/products'
 import ProductIcon from 'components/product-icon'
-import { ProductSlug } from 'types/products'
+import { NavHighlightItem } from 'components/sidebar/types'
 import s from './sidebar-nav-highlight-item.module.css'
 
 /**
@@ -16,7 +17,7 @@ export default function SidebarNavHighlightItem({
 	href,
 	isActive,
 }: {
-	theme: ProductSlug
+	theme: NavHighlightItem['theme']
 	text: string
 	href: string
 	isActive?: boolean
@@ -27,7 +28,7 @@ export default function SidebarNavHighlightItem({
 			className={classNames(s.root, s[`theme-${theme}`])}
 			href={href}
 		>
-			<ProductIcon productSlug={theme} />
+			{isProductSlug(theme) ? <ProductIcon productSlug={theme} /> : null}
 			<span className={s.text}>{text}</span>
 		</Link>
 	)
--- a/src/components/sidebar/components/sidebar-nav-highlight-item/sidebar-nav-highlight-item.module.css
+++ b/src/components/sidebar/components/sidebar-nav-highlight-item/sidebar-nav-highlight-item.module.css
@@ -1,0 +1,3 @@
+.root {
+	border: 1px solid magenta;
+}
--- a/src/components/sidebar/components/sidebar-nav-highlight-item/sidebar-nav-highlight-item.module.css
+++ b/src/components/sidebar/components/sidebar-nav-highlight-item/sidebar-nav-highlight-item.module.css
@@ -1,4 +1,10 @@
 .root {
+	/* Generic default theme, eg. for HCP and Sentinel */
+	--gradient-start: var(--token-color-palette-neutral-100);
+	--gradient-stop: var(--token-color-palette-neutral-50);
+	--inset-border-color: var(--token-color-palette-neutral-200);
+
+	/* Set up themed property helpers */
 	--inset-border-shadow: inset 0 0 0 1px var(--inset-border-color);
 	--gradient-background: linear-gradient(
 		315deg,
--- a/src/components/sidebar/components/sidebar-nav-highlight-item/sidebar-nav-highlight-item.module.css
+++ b/src/components/sidebar/components/sidebar-nav-highlight-item/sidebar-nav-highlight-item.module.css
@@ -1,32 +1,44 @@
 .root {
-	--border-width: 1px;
+	--inset-border-shadow: inset 0 0 0 1px var(--inset-border-color);
+	--gradient-background: linear-gradient(
+		315deg,
+		var(--gradient-start) 0%,
+		var(--gradient-stop) 100%
+	);
 
 	composes: g-focus-ring-from-box-shadow from global;
 	align-items: center;
 	background: var(--token-color-surface-primary);
-	border: var(--border-width) solid transparent;
 	border-radius: 5px;
 	color: var(--token-color-foreground-primary);
 	display: flex;
 	gap: 8px;
-	padding: calc(8px - var(--border-width));
+	margin-bottom: 2px;
+	padding: 8px;
+	position: relative;
 	width: 100%;
 	z-index: 0;
 
 	&:hover {
-		background: linear-gradient(
-			315deg,
-			var(--hover-gradient-start) 0%,
-			var(--hover-gradient-stop) 100%
-		);
-	}
-
-	&:active {
-		border-color: var(--active-border-color);
+		background: var(--gradient-background);
 	}
 
 	&:focus {
 		z-index: 1;
+	}
+
+	&:active {
+		/* Using box-shadow for an inset border. Also plays nice with focus styles. */
+		box-shadow: var(--inset-border-shadow);
+	}
+
+	/* AKA active */
+	&[aria-current='page'] {
+		background: var(--gradient-background);
+
+		&:not(:focus) {
+			box-shadow: var(--inset-border-shadow);
+		}
 	}
 }
 
@@ -40,7 +52,7 @@
  */
 
 .theme-nomad {
-	--hover-gradient-start: var(--token-color-nomad-gradient-faint-stop);
-	--hover-gradient-stop: var(--token-color-nomad-gradient-faint-start);
-	--active-border-color: var(--token-color-nomad-border);
+	--gradient-start: var(--token-color-nomad-gradient-faint-stop);
+	--gradient-stop: var(--token-color-nomad-gradient-faint-start);
+	--inset-border-color: var(--token-color-nomad-border);
 }
--- a/src/components/sidebar/components/sidebar-nav-highlight-item/sidebar-nav-highlight-item.module.css
+++ b/src/components/sidebar/components/sidebar-nav-highlight-item/sidebar-nav-highlight-item.module.css
@@ -1,3 +1,46 @@
 .root {
-	border: 1px solid magenta;
+	--border-width: 1px;
+
+	composes: g-focus-ring-from-box-shadow from global;
+	align-items: center;
+	background: var(--token-color-surface-primary);
+	border: var(--border-width) solid transparent;
+	border-radius: 5px;
+	color: var(--token-color-foreground-primary);
+	display: flex;
+	gap: 8px;
+	padding: calc(8px - var(--border-width));
+	width: 100%;
+	z-index: 0;
+
+	&:hover {
+		background: linear-gradient(
+			315deg,
+			var(--hover-gradient-start) 0%,
+			var(--hover-gradient-stop) 100%
+		);
+	}
+
+	&:active {
+		border-color: var(--active-border-color);
+	}
+
+	&:focus {
+		z-index: 1;
+	}
+}
+
+.text {
+	composes: hds-typography-body-200 from global;
+	font-weight: var(--token-typography-font-weight-medium);
+}
+
+/*
+ * Themeing
+ */
+
+.theme-nomad {
+	--hover-gradient-start: var(--token-color-nomad-gradient-faint-stop);
+	--hover-gradient-stop: var(--token-color-nomad-gradient-faint-start);
+	--active-border-color: var(--token-color-nomad-border);
 }
--- a/src/components/sidebar/components/sidebar-nav-highlight-item/sidebar-nav-highlight-item.module.css
+++ b/src/components/sidebar/components/sidebar-nav-highlight-item/sidebar-nav-highlight-item.module.css
@@ -55,7 +55,7 @@
 }
 
 /*
- * Themeing
+ * Theming
  */
 
 .theme-terraform {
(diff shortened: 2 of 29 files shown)


BRKalow:
@zchsh I expect this might be out of scope, but the Boundary API docs do not have the new overview item: https://dev-portal-git-zssidebar-hierarchy-hashicorp.vercel.app/boundary/api-docs

zchsh:
@BRKalow Ah great catch, that probably is in scope, I'll try to patch that up 👍 Realizing I probably missed Packer Plugins too 😬 🏃

zchsh:
Okey dokes, have updated to fill the Boundary API docs and Packer Plugins gaps:

### Boundary API docs

CleanShot 2023-01-25 at 14 12 45@2x
CleanShot 2023-01-25 at 14 30 15@2x

### Packer Plugins

CleanShot 2023-01-25 at 14 29 36@2x
CleanShot 2023-01-25 at 14 29 53@2x

src/components/sidebar/components/sidebar-nav-highlight-item/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+}) {
+	return (
+		<Link
+			aria-current={isActive ? 'page' : undefined}
+			className={classNames(s.root, s[`theme-${theme}`])}
+			href={href}
+		>
+			<ProductIcon productSlug={theme} />
+			<span className={s.text}>{text}</span>
+		</Link>
```
ashleemboyer, line 32: 💭 Should we render `SidebarNavLinkItem` here so we can more easily maintain the same functionality for things like `aria-current`, `data-heap-track`, leading icon positioning, font size/weight, a bunch of redundant & complex CSS properties?

src/components/sidebar/types.ts
```diff
@@ -102,11 +103,21 @@ interface FilteredLinkNavItem extends EnrichedLinkNavItem {
...
+
 /**
  *
  * For reference: this is also defined in react-components/docs-sidenav:
  * https://github.com/hashicorp/react-components/blob/main/packages/docs-sidenav/types.ts
  *
+ * TODO: it might be worth refactoring this to be a union of multiple types,
+ * one for each distinct type of menu item, rather than a single interface
+ * with optional properties to cover all possible menu item types.
+ *
```
ashleemboyer, line 120:
```suggestion
 * TODO: it might be worth refactoring this to be a union of multiple types,
 * one for each distinct type of menu item, rather than a single interface
 * with optional properties to cover all possible menu item types.
 * ref: https://app.asana.com/0/1202097197789424/1202405210286689/f
```

src/pages/packer/plugins/[...page].tsx
```diff
@@ -104,7 +104,16 @@ export async function getStaticProps({ params, ...ctx }) {
...
-			menuItems: navData,
+			menuItems: [
+				{
+					title: 'Plugins',
+					fullPath: `/${productData.slug}/plugins`,
+					theme: productData.slug,
+				},
+				// Remove the authored title element and "Overview" link, these
+				// would be redundant with the styled overview item that we add.
+				...navData.slice(2),
```
Comment (ashleemboyer) [inline src/pages/packer/plugins/[...page].tsx:115]: 🙋🏻 Will this break when these items are changed at the source?

src/pages/packer/plugins/index.tsx
```diff
@@ -42,8 +42,11 @@ async function getStaticProps(ctx) {
 	 */
 	if ('props' in staticProps) {
 		// Partial nav data is provided from base getStaticProps, in menuItems
-		const partialNavData =
+		const baseMenuItems =
 			staticProps.props.layoutProps.sidebarNavDataLevels[2].menuItems
+		// Remove the authored title element and "Overview" link, these
+		// would be redundant with the styled overview item that we add.
+		const partialNavData = [baseMenuItems[0], ...baseMenuItems.slice(3)]
```
ashleemboyer, line 49: 🙋🏻 Same question here, and also wondering if there is a way we can abstract these two instances of slicing since they're special cases and could easily become out of sync?

src/components/sidebar/components/sidebar-nav-highlight-item/sidebar-nav-highlight-item.module.css
```diff
@@ -0,0 +1,101 @@
...
+		z-index: 1;
+	}
+
+	&:active {
+		/* Using box-shadow for an inset border. Also plays nice with focus styles. */
+		box-shadow: var(--inset-border-shadow);
+	}
+
+	/* AKA active */
+	&[aria-current='page'] {
```
kendallstrautman, line 37: Nice, TIL about `aria-current`.

src/pages/packer/plugins/index.tsx
```diff
@@ -63,7 +66,9 @@ async function getStaticProps(ctx) {
 		})
 
 		// Replace our original navData with our prepared navData
-		staticProps.props.layoutProps.sidebarNavDataLevels[2].menuItems = navData
+		staticProps.props.layoutProps.sidebarNavDataLevels[2].menuItems = [
+			...navData,
+		]
```
ashleemboyer, line 71: 🙋🏻 Did you see some behavior that required updating this to copy `navData`? Just curious!

src/components/sidebar/components/sidebar-nav-highlight-item/sidebar-nav-highlight-item.module.css
```diff
@@ -0,0 +1,101 @@
...
+	}
+}
+
+.text {
+	composes: hds-typography-body-200 from global;
+	font-weight: var(--token-typography-font-weight-medium);
+}
+
+/*
+ * Themeing
```
kendallstrautman, line 52:
```suggestion
 * Theming
```
⛏️

src/components/sidebar/components/sidebar-nav-highlight-item/sidebar-nav-highlight-item.module.css
```diff
@@ -0,0 +1,101 @@
...
+	--inset-border-color: var(--token-color-nomad-border);
+}
+
+.theme-waypoint {
+	--gradient-start: var(--token-color-waypoint-gradient-faint-stop);
+	--gradient-stop: var(--token-color-waypoint-gradient-faint-start);
+	--inset-border-color: var(--token-color-waypoint-border);
+}
+
+.theme-vagrant {
```
kendallstrautman, line 97: ❓ should we have a generic variant for sidebars like ['well-architected-framework'](https://dev-portal-git-zssidebar-hierarchy-hashicorp.vercel.app/well-architected-framework)? Also, maybe this could be follow up, but we should ask design how to handle that overview item. 

Screen Shot 2023-01-25 at 1 41 18 PM

src/components/sidebar/helpers/generate-product-landing-nav-items.ts
```diff
@@ -75,5 +83,6 @@ export const generateProductLandingSidebarNavData = (product: ProductData) => {
 		menuItems,
 		showFilterInput,
 		title,
+		visuallyHideTitle: true,
```
kendallstrautman, line 86: A note here re: that we insert the 'themed' title would be helpful.

src/components/sidebar/types.ts
```diff
@@ -102,11 +103,21 @@ interface FilteredLinkNavItem extends EnrichedLinkNavItem {
...
+	title: string
+	fullPath: string
+}
+
 /**
  *
  * For reference: this is also defined in react-components/docs-sidenav:
  * https://github.com/hashicorp/react-components/blob/main/packages/docs-sidenav/types.ts
  *
+ * TODO: it might be worth refactoring this to be a union of multiple types,
```
kendallstrautman, line 117: Agreed! 👍🏻

src/components/tutorials-sidebar/index.tsx
```diff
@@ -15,6 +16,8 @@ import {
 	TutorialSidebarProps,
 } from './types'
 import s from './tutorials-sidebar.module.css'
+import { useRouter } from 'next/router'
+import { ProductSlug } from 'types/products'
```
kendallstrautman, line 20: ⛏️ can we move these imports up to keep the order of external --> local

src/pages/packer/plugins/index.tsx
kendallstrautman, line 49: 🎏  Feels a little risky since this logic could break if the order of the menu items is edited, right? Not sure how else to go about it though 🤔 Have we considered removing those items in the source content instead? Or at least the 'overview'?

kendallstrautman: Mostly speaking to the `slice` here and in the previous file.

src/views/collection-view/helpers/generate-collection-sidebar-nav-data.tsx
```diff
@@ -0,0 +1,32 @@
...
+		levelButtonProps: {
+			levelUpButtonText: `${product.name} Home`,
+			levelDownButtonText: 'Previous',
+		},
+		backToLinkProps: {
+			text: `${product.name} Home`,
+			href: `/${product.slug}`,
+		},
+		title: 'Tutorials',
+		visuallyHideTitle: true,
```
kendallstrautman, line 24: 💭 It might be a bit confusing tracking down how the themed title is generated. What have been your thoughts on the data driving the themed title item versus automating it at the component level? IOW if we defined the themed title here instead of lower down.

src/views/docs-view/server.ts
```diff
@@ -275,49 +287,40 @@ export function getStaticGenerationFunctions<
...
-			/**
-			 * Exception: If the first navData node is a `heading`,
-			 * we'll avoid adding an overview item even if there's
-			 * no overview item match.
+			 * We always include a brand-themed "highlight" item,
+			 * which serves as a menu item representing the top-level docs category.
+			 *
+			 * If the first item in docsSidebarLevel.menuItems is an "Overview" item,
+			 * then we remove it, as it would be redundant with the "highlight" item
+			 * we're adding.
```
kendallstrautman, line 313: Nice, this is good context.

src/views/docs-view/server.ts
```diff
@@ -275,49 +287,40 @@ export function getStaticGenerationFunctions<
...
 			 */
-			if (!overviewItemMatch && !firstItemIsHeading) {
-				docsSidebarLevel.overviewItemHref = versionPathPart
-					? `/${product.slug}/${basePath}/${versionPathPart}`
-					: `/${product.slug}/${basePath}`
+			if (isOverviewItem(docsSidebarLevel.menuItems[0])) {
```
kendallstrautman, line 315: 💭 Again, here I wonder why we can't edit the source files. I'm sure there's a bigger conversation about that I haven't seen.

kendallstrautman: Probably because coordinating that effort would be difficult? I'd just be worried about the application taking on too many of these types of things, instead of updating source content, which would be cleaner /easier in the long run.

src/views/docs-view/server.ts
```diff
@@ -275,49 +287,40 @@ export function getStaticGenerationFunctions<
...
+				docsSidebarLevel.menuItems.shift()
 			}
+			docsSidebarLevel.menuItems.unshift({
+				title: docsSidebarTitle,
+				fullPath: docsBasePathFullPath,
+				theme: product.slug, // this makes this a "highlighted" themed item
+				// TODO: would be great to fix up Menu Item types at some point.
```
kendallstrautman, line 322: 💡 [Might be related ticket?](https://app.asana.com/0/1202097197789424/1202405210286689) I'm not sure.

src/views/docs-view/server.ts
zchsh: This is a great question. For docs views, I think editing the source files could be an option. For other views we don't necessarily have the same author interface anyways, but I'm assuming this question was mainly focused on docs.

I think what made me lean towards implementing on our side is how restrictive I think we'd be with adding in "highlight" items. I think:
- We'd only ever want authors to include these "highlight" items as the first item in a sidebar
- We always want the "highlight" item's title to be the name of the docs category
- We always want the "highlight" item's URL to be the root of the docs category

If that all seems right, then I think it might be a little odd to expose highlight items as authorable in source - we'd want authors to use them in a very specific way, so they wouldn't exactly be "authorable" in that sense.

All that being said -- for the Packer plugins situation I think you're absolutely right, we should ideally update in source 😅 During QA there are also a few spots in Terraform content where we have similar situations.

💭 Of course, the biggest challenge as you mentioned is co-ordination.... I think I'll work on a clearer workflow for adding a "highlight" item to `menuItems` - it could handle the removal of "Overview" items for now, and ideally handle the Terraform and Packer plugin cases in a more clean and robust way. Once content is updated, then we can remove those extra processing step on our end. Hopefully this should address the Packer plugin case brittleness as well. Will update here if I get anywhere with that!

src/views/collection-view/helpers/generate-collection-sidebar-nav-data.tsx
zchsh: I dig this! I'll give it a shot 👍

src/pages/packer/plugins/index.tsx
zchsh: Ah thanks for catching this 💖 It's some unintentional cruft from getting Packer plugin pages working. Not needed so I'll revert the change here 🔪

src/components/sidebar/helpers/generate-product-landing-nav-items.ts
zchsh: 💖 great call - added! ✍️

src/views/collection-view/helpers/generate-collection-sidebar-nav-data.tsx
zchsh: Ah wait so I misinterpreted this - I'm actually not sure I follow. Currently, for collection sidebar levels (as used on product-tutorials views, collection views, and tutorial views for the mobile sidebar), we render the themed title item through the `CollectionViewSidebarContent` component.

Is there somewhere else we want to move that rendering? I guess we could pass it here as a first item in `children`, rather than within `<CollectionViewSidebarContent />`?

(I hadn't given it too much thought, `CollectionViewSidebarContent` just seemed like the most intuitive place at first!)

src/components/sidebar/components/sidebar-nav-highlight-item/sidebar-nav-highlight-item.module.css
zchsh: I like this idea! I've stubbed something in for design's consideration. Will add screenshots in a collection "feedback and changes so far" comment 👍

src/components/sidebar/components/sidebar-nav-highlight-item/index.tsx
zchsh: I don't have a strong opinion here - they felt like different enough items, with potential for further divergence, that I figured it might be slightly easier to reason about the styling if they were separate rather than intertwined, with the measured downside of some duplicate code (some differences: font weight difference, background gradient and border, brand theming prop and styling, related hover and active states, no possibility of being disabled or external or having a dropdown, no possibility of trailing icon).

That being said, happy to take another look and dive deeper into this if it feels blocking 🙇

src/pages/packer/plugins/[...page].tsx
Comment (zchsh) [reply src/pages/packer/plugins/[...page].tsx]: Great callout, they would 💖 I've updated to use a shared `addBrandedOverviewSidebarItem`, which is significantly more robust in how it checks for and removes potentially redundant items before adding the "branded" overview item:
- If first item is an "Overview" item, removes it
- Else if first item is a heading `MenuItem`, and second item is and "Overview" item, removes them both
- Else leaves items unchanged

src/pages/packer/plugins/index.tsx
zchsh: Great point - took a shot at this with `addBrandedOverviewSidebarItem` 👍

src/components/tutorials-sidebar/index.tsx
zchsh: 👍 I dig it, I lean on VSCode auto-importing and forget how messy it ends up 😅 Re-ordered these! ✅

src/pages/packer/plugins/index.tsx
zchsh: Have added a `addBrandedOverviewSidebarItem` helper to make the logic more robust - will add detail in the main comment thread shortly 👍

zchsh:
@ashleemboyer @kendallstrautman Thanks for the excellent feedback y'all 💖 

Here are the changes I've made based on feedback (in addition to smaller suggestions, thanks for those too!)

### 🛠 Added new shared `addBrandedOverviewSidebarItem` utility

This utility adds in the "branded overview" item, and _also_ removes redundant items in a more reliable way. Specifically, before adding the branded overview item:

- If first item is an "Overview" item, we drop it (this logic was already in place for Docs)
- Else if first item is a heading item, and second item is an "Overview" item, we drop both those (new logic!)
- Else we leave items untouched

This is now used for Packer plugins pages (replacing the brittle `slice` approach), and for all other Docs pages. It's also used for product landing pages, and WAF (maybe, more on that later).

And, this also handles cases like [/terraform/language](https://dev-portal-git-zssidebar-hierarchy-hashicorp.vercel.app/terraform/language), which were previously showing redundant heading + overview items (Mike mentioned this in DMs).

### 🖤 Added "neutral" theming, used on WAF pages

Thanks Kendall for catching this 💯  I've taken a shot at a "neutral" theme, and used it on the WAF pages.

| Before | After |
| - | - |
| before | after |

I've looped in @heitzke to get design help here, to decide if we want to take this past a first cut if it feels close enough, or abandon it otherwise.

src/views/docs-view/server.ts
zchsh: Update: implemented a helper function `addBrandedOverviewSidebarItem` to do this. See [this comment](https://github.com/hashicorp/dev-portal/pull/1552#issuecomment-1404565590) for details.

src/components/sidebar/components/sidebar-nav-highlight-item/index.tsx
ashleemboyer: Nope, not blocking! Just pondering. 😊 And now I'm pondering we should get rid of that disabling... I'm almost certain that was only there for the "Coming soon" section leading up to beta! I tossed a low-priority ticket on the backlog. 😎

[[SidebarNavLinkItem] Remove ability to disable](https://app.asana.com/0/1202097197789424/1203833247481790/f)

src/pages/packer/plugins/[...page].tsx
Comment (ashleemboyer) [reply src/pages/packer/plugins/[...page].tsx]: That sounds sweet! I'll take a look at that 🤩

heitzke (approved):
Changes look good to me, thanks for getting TF and WAF all fixed up too 🔥

ashleemboyer (approved):
LGTM! 👍🏻

src/components/sidebar/components/sidebar-nav-highlight-item/sidebar-nav-highlight-item.module.css
zchsh: Update: have implemented this! ✅

src/views/collection-view/helpers/generate-collection-sidebar-nav-data.tsx
kendallstrautman: Let's roll with what you have!

kendallstrautman (approved):
Nice work! 🚀

PR #1525, opened 2023-01-12
Refactor docs-content-link-rewrites/rewrite-script.ts

ashleemboyer:
## 🗒️ What/Why

- Updates `scripts/docs-content-link-rewrites/rewrite-links-plugin.ts` to leverage logic from our two server-side link rewrite plugins:
  - `remark-plugin-adjust-link-urls`
  - `rewrite-tutorial-links`
- Simplifies how "statistics" are tracked for the script's runs
  - There's now one conditional that checks if a link gets rewritten, and it's added to the appropriate data structure under the `statistics` argument 
- Removes now unneeded helpers:
  - `scripts/docs-content-link-rewrites/helpers/get-dot-io-to-dev-dot-url-map.ts`
  - `scripts/docs-content-link-rewrites/helpers/get-learn-to-dev-dot-url-map.ts`
- Shortens some of the less important `console.log`s so they don't take up as many lines in a terminal
- In `get-rewrite-links-script-arguments.ts`:
  - Removes unused `addPrefixToFilePaths`
  - Replaces the `contentDirectory` and `navDataDirectory` arguments with a `localCopyLocation` argument for easier CLI use
- Removes commented-out portions of the script that were intended for handling nav data (but leaves the `TODO` in case we come back to this part via this script)

## 🧪 Testing

We've been testing this script on several draft migration PRs:

- [vagrant#13048](https://github.com/hashicorp/vagrant/pull/13048)
- [boundary#2777](https://github.com/hashicorp/boundary/pull/2777)
- [packer#12192](https://github.com/hashicorp/packer/pull/12192)
- [nomad#15779](https://github.com/hashicorp/nomad/pull/15779)
- [consul#15976](https://github.com/hashicorp/consul/pull/15976)
- [vault#18696](https://github.com/hashicorp/vault/pull/18696)

PR #1555, opened 2023-01-23
[link format checker] Adding check for VERCEL_URL

ashleemboyer:
## 🗒️ What

- Adds `"Get target_url hostname"` step to pull `hostname` from `github.event.deployment_status.target_url`
- Updates `.github/workflows/docs-content-check-legacy-links-format.yml` workflow to set `VERCEL_URL` inline when it runs `npm run rewrite-docs-content-links`

## 🤷 Why

Since refactoring the link rewrite script in https://github.com/hashicorp/dev-portal/pull/1525, the script calls `getTutorialMap` which relies on a VERCEL_URL, or defaults the baseUrl for the api endpoint to `http://localhost:3000`. The latest runs for this workflow aren't setting VERCEL_URL, so the default is being used. Since no local server is running, an error happens when trying to fetch the tutorial map.

## 🧪 Testing

- [ ] Go to the test run of this workflow in HCP: https://github.com/hashicorp/cloud.hashicorp.com/actions/runs/3990754564/jobs/6844770507
- [ ] The run should no longer error with `ECONNREFUSED`
  - It's OK if the workflow fails because old link formats were detected

## 💭 Anything else?

Once this PR is merged, I will update the workflows that call this one to:

- Be triggered with `on: [deployment_status]`
- Only run the job `if: github.event.deployment_status.state == 'success'`
- Have the latest commit hash in the `uses` line

### DIFF
--- a/.github/workflows/docs-content-check-legacy-links-format.yml
+++ b/.github/workflows/docs-content-check-legacy-links-format.yml
@@ -13,9 +13,6 @@ on:
       commit-sha:
         required: true
         type: string
-      vercel-url:
-        required: true
-        type: string
       mdx-directory:
         required: true
         type: string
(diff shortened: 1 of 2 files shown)


BRKalow (approved):
Makes sense!

BRKalow:
@ashleemboyer just had a thought, could we use the deployed api endpoint on `developer.hashicorp.com`?

ashleemboyer:
> @ashleemboyer just had a thought, could we use the deployed api endpoint on `developer.hashicorp.com`?

@BRKalow I'm not sure if that'll always work, let's chat in Slack so others' can chime in too!

PR #118, opened 2022-02-18
feat: continue Packer bring-up

dstaley:
## "Ready for Review" checklist

- [x] The Vercel preview link has been added to this PR's description
- [ ] The Asana task has been added to this PR's description
- [ ] This PR's link has been added to the "📐 Pull Request" field on the Asana task
- [ ] This Vercel preview link has been added to the "🔍 Preview" field on the Asana task
- [ ] The description template has been filled in below

---

## Relevant links

- [Preview link](https://dev-portal-git-dsadd-packer-hashicorp.vercel.app/) 🔎
- [Asana task](url) 🎟️

## What

This PR continues the bring-up of Packer's site within dev-portal, including the enablement of external plugin docs.

## Testing

- [ ] Navigate to the preview: https://dev-portal-git-dsadd-packer-hashicorp.vercel.app/
- [ ] Select Packer from the product switcher in the bottom left
- [ ] Validate that the site matches https://www.packer.io/
- [ ] Validate that the docs are rendering as expected by clicking "Docs" in the menu
- [ ] Validate that plugin docs are rendering as expected by clicking "Plugins" in the menu, and clicking on one of the plugins listed in the sidebar

PR #1558, opened 2023-01-23
[docs-content-check-legacy-links-format] Skip steps if no PR number found; set VERCEL_URL to prod

ashleemboyer:
## 🗒️ What/Why

- Replaces the `get-target-url-hostname` step with a hardcoded value for `VERCEL_URL` in the `run-rewrite-docs-content-links-script` step
- Changes the return value of `get-pr-number` when no PR is found to `false`
- Updates all subsequent steps to `get-pr-number` to only run if the return value was not false, so the workflow doesn't fail if a commit has been pushed up but does not have an associated PR yet

## 🧪 Testing

Testing with [hashicorp/cloud.hashicorp.com#593](https://github.com/hashicorp/cloud.hashicorp.com/pull/593):

- [ ] See that the steps are skipped in this workflow run because no PR number was found: [cloud.hashicorp.com run #118](https://github.com/hashicorp/cloud.hashicorp.com/actions/runs/3991278846/jobs/6845904473)
- [ ] See that the `"Run rewrite-docs-content-links script"` step is successful and passes the right value for VERCEL_URL in this run: [cloud.hashicorp.com run #119](https://github.com/hashicorp/cloud.hashicorp.com/actions/runs/3991293595/jobs/6845939853#step:7:1)

### DIFF
--- a/.github/workflows/docs-content-check-legacy-links-format.yml
+++ b/.github/workflows/docs-content-check-legacy-links-format.yml
@@ -38,9 +38,15 @@ jobs:
               repo: process.env.REPO,
               commit_sha: process.env.COMMIT_SHA
             })
-            return data[0].number
+            const prNumber = data && data[0] && data[0].number
+            if (prNumber) {
+              return prNumber
+            } else {
+              return false
+            }
 
       - name: Get list of PR files
+        if: ${{ steps.get-pr-number.outputs.result != 'false' }}
         id: get-lists-of-pr-files
         uses: actions/github-script@v6
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/docs-content-check-legacy-links-format.yml
```diff
@@ -28,17 +28,6 @@ jobs:
...
-          DEPLOYMENT_STATUS_TARGET_URL: ${{ github.event.deployment_status.target_url }}
-        with:
-          script: |
-            const { DEPLOYMENT_STATUS_TARGET_URL } = process.env
-            const { hostname } = new URL(DEPLOYMENT_STATUS_TARGET_URL)
-            return hostname
-
       - name: Get PR number
         id: get-pr-number
         uses: actions/github-script@v6
```
kendallstrautman, line 33: Just a future thought, we could set a job output and use[ `if: success() ....` ](https://docs.github.com/en/actions/creating-actions/setting-exit-codes-for-actions)as a way to approach conditionals as well. [Example in tutorials.](https://github.com/hashicorp/tutorials/blob/main/.github/workflows/check-uuid.yml#L36)

kendallstrautman (approved):
Makes sense to me! Perhaps a quick comment next to that vercel url setting would help down the line so we remember why that is needed.

PR #1560, opened 2023-01-24
Adding LandingPageBlocks component, abstracting components from /${productSlug}/docs views

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambadd-lpb-component-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1203820004767055/f) 🎟️ 

## 🗒️ What

- Abstracts the marketing content block components used in `src/views/product-root-docs-path-landing/components/marketing-content/index.tsx` into a new, reusable `LandingPageBlocks` module/component that lives in `src/components`:
  - `AutosizedHeadingBlock`
  - `CalloutCardBlock`
  - `CardGridBlock`
  - `IconCardGridBlock`
  - `ParagraphBlock`
  - `SectionHeadingBlock`

## 🤷 Why

- This is the first step in consolidating the authoring interfaces (JSON files in `src/content`) & components used on our three main landing pages for consistency and lower engineering maintenance

## 🧪 Testing

The following pages should be unchanged compared to prod:

| Page slug | Preview | Prod |
| - | - | - |
| `/hcp/docs` | [preview url](https://dev-portal-git-ambadd-lpb-component-hashicorp.vercel.app/hcp/docs) | [prod url](https://developer.hashicorp.com/hcp/docs) |
| `/terraform/docs` | [preview url](https://dev-portal-git-ambadd-lpb-component-hashicorp.vercel.app/terraform/docs) | [prod url](https://developer.hashicorp.com/terraform/docs) |
| `/packer/docs` | [preview url](https://dev-portal-git-ambadd-lpb-component-hashicorp.vercel.app/packer/docs) | [prod url](https://developer.hashicorp.com/packer/docs) |
| `/consul/docs` | [preview url](https://dev-portal-git-ambadd-lpb-component-hashicorp.vercel.app/consul/docs) | [prod url](https://developer.hashicorp.com/consul/docs) |
| `/vault/docs` | [preview url](https://dev-portal-git-ambadd-lpb-component-hashicorp.vercel.app/vault/docs) | [prod url](https://developer.hashicorp.com/vault/docs) |
| `/boundary/docs` | [preview url](https://dev-portal-git-ambadd-lpb-component-hashicorp.vercel.app/boundary/docs) | [prod url](https://developer.hashicorp.com/boundary/docs) |
| `/nomad/docs` | [preview url](https://dev-portal-git-ambadd-lpb-component-hashicorp.vercel.app/nomad/docs) | [prod url](https://developer.hashicorp.com/nomad/docs) |
| `/waypoint/docs` | [preview url](https://dev-portal-git-ambadd-lpb-component-hashicorp.vercel.app/waypoint/docs) | [prod url](https://developer.hashicorp.com/waypoint/docs) |
| `/vagrant/docs` | [preview url](https://dev-portal-git-ambadd-lpb-component-hashicorp.vercel.app/vagrant/docs) | [prod url](https://developer.hashicorp.com/vagrant/docs) |

## 💭 Anything else?

- Next up: abstracting components from the blocks from `/${productSlug}` landing pages and/or reusing some components abstracted in this PR
- In a few PRs: documentation will be added for `LandingPageBlocks`. Waiting to fully document until its subcomponents list is more complete and their names are stable.

---
  - https://app.asana.com/0/0/1203820004767057

### DIFF
--- a/src/components/landing-page-blocks/index.tsx
+++ b/src/components/landing-page-blocks/index.tsx
@@ -1,0 +1,7 @@
+import { ParagraphBlock } from './components'
+
+const LandingPageBlocks = () => null
+
+export default Object.assign(LandingPageBlocks, {
+	ParagraphBlock,
+})
--- a/src/components/landing-page-blocks/index.tsx
+++ b/src/components/landing-page-blocks/index.tsx
@@ -1,5 +1,6 @@
 import {
 	AutosizedHeadingBlock,
+	CardGridBlock,
 	IconCardGridBlock,
 	ParagraphBlock,
 	SectionHeadingBlock,
@@ -9,6 +10,7 @@ const LandingPageBlocks = () => null
 
 export default Object.assign(LandingPageBlocks, {
 	AutosizedHeadingBlock,
+	CardGridBlock,
 	IconCardGridBlock,
 	ParagraphBlock,
 	SectionHeadingBlock,
--- a/src/components/landing-page-blocks/index.tsx
+++ b/src/components/landing-page-blocks/index.tsx
@@ -1,5 +1,6 @@
 import {
 	AutosizedHeadingBlock,
+	CalloutCardBlock,
 	CardGridBlock,
 	IconCardGridBlock,
 	ParagraphBlock,
@@ -10,6 +11,7 @@ const LandingPageBlocks = () => null
 
 export default Object.assign(LandingPageBlocks, {
 	AutosizedHeadingBlock,
+	CalloutCardBlock,
 	CardGridBlock,
 	IconCardGridBlock,
 	ParagraphBlock,
--- a/src/components/landing-page-blocks/index.tsx
+++ b/src/components/landing-page-blocks/index.tsx
@@ -1,5 +1,6 @@
 import {
 	AutosizedHeadingBlock,
+	IconCardGridBlock,
 	ParagraphBlock,
 	SectionHeadingBlock,
 } from './components'
@@ -8,6 +9,7 @@ const LandingPageBlocks = () => null
 
 export default Object.assign(LandingPageBlocks, {
 	AutosizedHeadingBlock,
+	IconCardGridBlock,
 	ParagraphBlock,
 	SectionHeadingBlock,
 })
--- a/src/components/landing-page-blocks/index.tsx
+++ b/src/components/landing-page-blocks/index.tsx
@@ -1,8 +1,13 @@
-import { AutosizedHeadingBlock, ParagraphBlock } from './components'
+import {
+	AutosizedHeadingBlock,
+	ParagraphBlock,
+	SectionHeadingBlock,
+} from './components'
 
 const LandingPageBlocks = () => null
 
 export default Object.assign(LandingPageBlocks, {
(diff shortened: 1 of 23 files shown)


src/components/landing-page-blocks/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+	CalloutCardBlock,
+	CardGridBlock,
+	IconCardGridBlock,
+	ParagraphBlock,
+	SectionHeadingBlock,
+} from './components'
+
+const LandingPageBlocks = () => null
+
+export default Object.assign(LandingPageBlocks, {
```
kendallstrautman, line 12: Could you share your thoughts on storing all of these in an object instead of importing the components directly? Also have you considered using dynamic import? And do you know how the code splitting will work here, if I only need the `ParagraphBlock`, will I be importing and loading all the others when I use this `LandingPageBlocks` object?

ashleemboyer: @kendallstrautman When @thiskevinwang was working on the new `Pagination` component, I noticed he took this approach. Maybe he can give insight into any other reasons, but what I liked is that it made it possible to render the subcomponents of the main one all in the same Swingset page, which is something we presently can't do without putting them inside of a `LiveComponent`. Having to include `LiveComponent` for every single subcomponent adds a lot of unnecessary clutter to the page if you're just trying to see all the components visually.

I also like that in `ProductRootDocsPathLandingMarketingContent` (where these subcomponents are abstracted from) we can clearly see where the subcomponents "live" or "belong" and we can also see that they're closely related to each other. And then their relationship implies that they're intended to be used together.

As for the code splitting and dynamic imports, I'm not 100% sure yet how the bundle will be impacted in subsequent tasks that will consolidate components used in other landing pages. It's something I plan to keep my eye on as I work through those tasks. As for this PR the bundle size isn't impacted so I don't see this unknown as a blocker right now. In the case that one or a couple of the subcomponents needs used, it should be possible to directly import them from their folders!

kendallstrautman: Okay cool, this is good context to know! It's definitely an interesting pattern. My guess is that we aren't seeing much bundle activity because most of these components were already being imported and used on each product view. But it would be good to dig in more with the bundle analyzer to see how this behaves before we adopt this pattern more broadly. 

Based on the fact that things did increase slightly, makes me wonder if in fact all components are always loaded when the parent component is used. I don't think this is a blocker, but if you want to try and get clarity on whats going on there as you work through the other tasks, that would help inform whether we adopt this pattern more broadly!

kendallstrautman (approved):

src/components/landing-page-blocks/index.tsx
ashleemboyer: @kendallstrautman we're having the same bundle convo on two separate PRs so I just want to double-check this! I'm not seeing that the bundle has changed in this PR based on the most recent from our bundle analysis action: https://github.com/hashicorp/dev-portal/pull/1560#issuecomment-1401327732.

PR #1567, opened 2023-01-25
Adds an Aria label to the integrations card

BrandonRomano:
[🎟️ ARIA Label for Integration Card item](https://app.asana.com/0/1203792610600085/1203824021594914)
[🔍 Preview](https://dev-portal-git-braria-label-hashicorp.vercel.app/vault/integrations)

### DIFF
--- a/src/views/product-integrations-landing/components/integrations-list/index.tsx
+++ b/src/views/product-integrations-landing/components/integrations-list/index.tsx
@@ -43,7 +43,7 @@ function IntegrationCard({ integration }: IntegrationCardProps) {
 
 	return (
 		<CardLink
-			ariaLabel={`${integration.name} by ${integration.organization.slug}, ${integration.description}`}
+			ariaLabel={integration.name}
 			className={s.integrationCard}
 			href={url}
 			opensInNewTab={isExternalLink}
--- a/src/views/product-integrations-landing/components/integrations-list/index.tsx
+++ b/src/views/product-integrations-landing/components/integrations-list/index.tsx
@@ -43,7 +43,7 @@ function IntegrationCard({ integration }: IntegrationCardProps) {
 
 	return (
 		<CardLink
-			ariaLabel="TODO"
+			ariaLabel={`${integration.name} by ${integration.organization.slug}, ${integration.description}`}
 			className={s.integrationCard}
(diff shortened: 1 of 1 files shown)


ashleemboyer (changes requested):
This is super close to what we need, just needs a couple of small tweaks. 😊 

- Concatenating the name, slug, and description for each `CardLink` is more verbose than needed. VoiceOver users are able to access all the text within these cards using their arrow keys, so the information presented ends up being redundant.
- For voice control users, they are required to say the [Accessible Name](https://www.w3.org/TR/accname-1.1/#dfn-accessible-name) of an interactive element in order to activate it. Using a live example on the preview URL, that means they would have to say something like "Click AWS Auth Method by hashicorp, Provides an automated mechanism to retrieve a Vault token for IAM principals and AWS EC2 instances." in order to activate each `CardLink`. It's a lot to say and it's also not predictable based on the visual design.

With the visual design as-is, we should be able to set `ariaLabel` as just the integration's name since it's the first element and the most prominent font-weight.

- If we want more information to be presented by a screen reader when these `CardLink`s come into focus, I'd suggest setting the `aria-describedby` attribute and passing in `id`s from the `organization` and `description` elements.

---

Here's a screen recording of the VoiceOver testing I did:

https://user-images.githubusercontent.com/43934258/214664833-2d4ce22b-d4cd-45a6-9904-ca28100486b8.mp4

BrandonRomano:
That video is super helpful!  I'll adjust to make it just be the title; after listening to the recording I definitely agree what we have is _too verbose_!

src/views/product-integrations-landing/components/integrations-list/index.tsx
```diff
@@ -43,7 +43,7 @@ function IntegrationCard({ integration }: IntegrationCardProps) {
 
 	return (
 		<CardLink
-			ariaLabel="TODO"
+			ariaLabel={`${integration.name} by ${integration.organization.slug}, ${integration.description}`}
```
BrandonRomano, line 46:
```suggestion
			ariaLabel={integration.name}
```

ashleemboyer (approved):
Looks great! ✨ 🥳

PR #832, opened 2022-08-16
feat: tutorial library page

BRKalow:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-brkfeat-tutorial-library-hashicorp.vercel.app/tutorials/library) 🔎
- [Asana task](https://app.asana.com/0/1202520168081556/1202841351070141/f) 🎟️
- [Figma](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=12619%3A155727)

## 🗒️ What

Introduces the tutorial library. Uses Algolia's `react-instantsearch-dom` library to get search data.

---

TODO:
- [x] pagination
- [x] refined filter styles
- [x] no results state
- [x] accessibility adjustments to better support the filtering behavior
- [x] mobile filtering

## 🧪 Testing

- [ ] Visit https://dev-portal-git-brkfeat-tutorial-library-hashicorp.vercel.app/tutorials/library
- [ ] Apply filters, click around
- [ ] Validate the URL is updated
- [ ] Validate the results are filtered accordingly

PR #1570, opened 2023-01-26
feat: refine search bar

zchsh:
- feat: use shared FilterInput component
- feat: implement hover style
- feat: border color for non-empty values
- feat: update integrations to use search icon

## 🔗 Relevant links

- [Asana task](url) 🎟️

### DIFF
--- a/src/components/filter-input/filter-input.module.css
+++ b/src/components/filter-input/filter-input.module.css
@@ -48,9 +48,10 @@
 	composes: g-focus-ring-from-box-shadow from global;
 	appearance: none;
 	background-color: transparent;
-	border: none;
+	border: var(--token-form-control-border-width) solid
+		var(--token-form-control-base-border-color-default);
 	border-radius: 5px;
-	box-shadow: var(--token-surface-inset-box-shadow);
+	box-shadow: var(--token-elevation-inset-box-shadow);
 	color: var(--token-color-palette-neutral-700);
 	font-size: 16px;
 	line-height: 18px;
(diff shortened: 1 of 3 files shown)


zchsh:
@BrandonRomano @braydenlove This PR refines the input styling for searchable integrations list (as used on [/vault/integrations](https://dev-portal-git-zsrefine-integrations-search-bar-hashicorp.vercel.app/vault/integrations), for example).

Two notes:
1. ⚠️ `font-size` for the input is `14px` in designs, but I've left it as `16px` as implemented in #832 . I think the reason we've used `16px` previously is related to browser zoom issues, which come up in some browsers when input font-sizes are less than `14px`. We can definitely update this PR to do `14px`, but wanted to double-check first here given the related issues, and err on the side of trusting previous work and approvals 🙇
2. 💭  Have implemented so that the Tutorials Library inputs (and other inputs) retain the "filter" icon, while the integrations input uses a custom "search" icon. Consistency between the two could simplify things a little, but definitely a smaller issue - we can ship either way, just wanted to check this is the desired icon setup 👍

BrandonRomano (approved):
Looks great!

RE: the Icon inconsistency, I agree and will create a low priority backlog ticket.

braydenlove (approved):
Awesome work @zchsh, super thorough. I agree with all the decisions you made here and triple checked the component specs. Let's follow up on the icon stuff as @BrandonRomano suggested and move forward. Many thanks 🙏🏼

BrandonRomano:
FYI @braydenlove it's filed here: 🎟️ [Tutorial Library search icon differs from Integrations Library](https://app.asana.com/0/1202097197789424/1203838760523793)

PR #1578, opened 2023-01-27
[rewrite-links script] Preventing rewrites of files in root of content directory

ashleemboyer:
## 🗒️ What

- Prevents files that live in the root of content directories from being rewritten

## 🤷 Why

- These files are used for top-level pages in a content repo's website

### DIFF
--- a/scripts/docs-content-link-rewrites/helpers/get-mdx-links-to-rewrite.ts
+++ b/scripts/docs-content-link-rewrites/helpers/get-mdx-links-to-rewrite.ts
@@ -25,6 +25,19 @@ const getMdxLinksToRewrite = async ({
 
 		const filePathWithoutPrefix = filePath.replace(filePathPrefix, '')
 		const currentPath = filePathWithoutPrefix.replace(/(index)?\.mdx$/, '')
+
+		/**
+		 * Don't rewrite links in files that are at the root of the content
+		 * directory. These files are used for content rendered at the top-level of
+		 * various sites.
+		 *
+		 * Example: https://github.com/hashicorp/cloud.hashicorp.com/blob/f7e1db477f8c2b842c1e8a18047a54297e9b295d/content/roles-responsibilities.mdx?plain=1#L60
+		 */
+		const isAtContentRoot = !currentPath.slice(1).includes('/')
+		if (isAtContentRoot) {
+			return
+		}
+
 		const fileContent = fs.readFileSync(filePath, 'utf-8')
 		const data = {
 			linksToRewrite: {},
--- a/scripts/docs-content-link-rewrites/helpers/get-mdx-links-to-rewrite.ts
+++ b/scripts/docs-content-link-rewrites/helpers/get-mdx-links-to-rewrite.ts
@@ -31,7 +31,8 @@ const getMdxLinksToRewrite = async ({
 		 * directory. These files are used for content rendered at the top-level of
 		 * various sites.
 		 *
-		 * Example: https://github.com/hashicorp/cloud.hashicorp.com/blob/f7e1db477f8c2b842c1e8a18047a54297e9b295d/content/roles-responsibilities.mdx?plain=1#L60
+		 * Example of a file that should not have its links rewritten:
+		 * https://github.com/hashicorp/cloud.hashicorp.com/blob/f7e1db477f8c2b842c1e8a18047a54297e9b295d/content/roles-responsibilities.mdx
 		 */
 		const isAtContentRoot = !currentPath.slice(1).includes('/')
 		if (isAtContentRoot) {


scripts/docs-content-link-rewrites/helpers/get-mdx-links-to-rewrite.ts
```diff
@@ -25,6 +25,19 @@ const getMdxLinksToRewrite = async ({
...
 		const currentPath = filePathWithoutPrefix.replace(/(index)?\.mdx$/, '')
+
+		/**
+		 * Don't rewrite links in files that are at the root of the content
+		 * directory. These files are used for content rendered at the top-level of
+		 * various sites.
+		 *
+		 * Example: https://github.com/hashicorp/cloud.hashicorp.com/blob/f7e1db477f8c2b842c1e8a18047a54297e9b295d/content/roles-responsibilities.mdx?plain=1#L60
+		 */
+		const isAtContentRoot = !currentPath.slice(1).includes('/')
```
kendallstrautman, line 36: 🤔 Nice catch. Just want to make sure with all product repos we assume that unless its within a subdirectory, we don't want to rewrite it? 

Just to make sure, we aren't using this script anymore with other repos now that they are 'migrated', this will only apply moving forward for terraform??

kendallstrautman: For a breadcrumb, could you add a note here on an example of a path you would expect to catch here? Or add a test case? Are the content root paths like 'content.mdx'?

ashleemboyer: Yep & yep! And the way this will affect other repos in the sense that they use the link format checker workflow, which runs this script.

ashleemboyer: > For a breadcrumb, could you add a note here on an example of a path you would expect to catch here? Or add a test case? Are the content root paths like 'content.mdx'?

@kendallstrautman there's an example link in block comment. 😊

kendallstrautman: @ashleemboyer that link is super helpful but I also think a note with an example of exactly the path we are trying to hit would be good. I find that helpful especially when using things like `slice` -- like what path exactly would we expect to catch here? Which a test case would also be good to communicate. 

And gotcha, the legacy link checker workflow runs this, I follow now.

ashleemboyer: @kendallstrautman this link is to a file that should be excluded. I'll update the `"Example"` text to be more clear!

scripts/docs-content-link-rewrites/helpers/get-mdx-links-to-rewrite.ts
```diff
@@ -25,6 +25,19 @@ const getMdxLinksToRewrite = async ({
...
 		const filePathWithoutPrefix = filePath.replace(filePathPrefix, '')
 		const currentPath = filePathWithoutPrefix.replace(/(index)?\.mdx$/, '')
+
+		/**
+		 * Don't rewrite links in files that are at the root of the content
+		 * directory. These files are used for content rendered at the top-level of
+		 * various sites.
+		 *
+		 * Example: https://github.com/hashicorp/cloud.hashicorp.com/blob/f7e1db477f8c2b842c1e8a18047a54297e9b295d/content/roles-responsibilities.mdx?plain=1#L60
+		 */
```
ashleemboyer, line 35:
```suggestion
		/**
		 * Don't rewrite links in files that are at the root of the content
		 * directory. These files are used for content rendered at the top-level of
		 * various sites.
		 *
		 * Example of a file that should not have its links rewritten:
		 * https://github.com/hashicorp/cloud.hashicorp.com/blob/f7e1db477f8c2b842c1e8a18047a54297e9b295d/content/roles-responsibilities.mdx
		 */
```

scripts/docs-content-link-rewrites/helpers/get-mdx-links-to-rewrite.ts
kendallstrautman: I found a case in packer too! on the /community-tools page.

kendallstrautman: @ashleemboyer I get that 😆 I was more looking to document the logic here like: 'remove the leading '/' and if there are no other '/' we can assume this is a single file at the root of the dir, such as /roles-responsibilities.mdx'.

kendallstrautman (approved):
Left a suggestion on adding notes on the logic, which I think should be added. Approving to unblock.

PR #1579, opened 2023-01-27
[rewrite-links.ts] Updating output for clearer feedback

ashleemboyer:
## 🗒️ What

- Formats the output of `scripts/docs-content-link-rewrites/rewrite-links.ts` so it's clearer what needs to be fixed, where, and how

## 🤷 Why

- Just the file names is helpful for _us_ locally when we have all the context, but folks seeing this output for the first time need more information than that


BrandonRomano (approved):
Good improvement!

PR #1580, opened 2023-01-27
feat: add versioned alert, adjust versioned routes

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksadd-versioned-alert-hashicorp.vercel.app/waypoint/integrations/BrandonRomano/aws-ecr/v0.10.4) 🔎
- [Asana task](https://app.asana.com/0/1199634971449915/1203713943353901) 🎟️

## 🗒️ What

This PR adds support to detect versioned paths in the sidebar sidecar layout, so the versioned alert banner renders. 

Additionally, the paths to access versioned routes have been updated to align with the rest of developer. These paths are access with the prefix 'v'. A new function was introduced `getTargetVersion` to handle the mapping between what the API expects and the version slug from the route. 

## 🧪 Testing

- [Visit this path](https://dev-portal-git-ksadd-versioned-alert-hashicorp.vercel.app/waypoint/integrations/BrandonRomano/aws-ecr/v0.10.4), validate the banner renders as expected.
- Try to hit other versions
- All versioned behavior and UI should be the same, besides the new alert.
- [Visit another versioned path](https://dev-portal-git-ksadd-versioned-alert-hashicorp.vercel.app/waypoint/docs/v0.8.x) to ensure it also works the same.

### DIFF
--- a/src/constants/version-path.ts
+++ b/src/constants/version-path.ts
@@ -3,9 +3,3 @@ export const VERSION_IN_PATH_REGEX = /v\d+\.\d+\.(\d+|\w+)/i
 
 /** This REGEX is used to parse a Terraform Enterprise version from a URL */
 export const TFE_VERSION_IN_PATH_REGEXP = /v[0-9]{6}-\d+/i
-
-/**
- * TODO - validate the expected versioned integrations paths for other products
- * Left the 'v' as optional here in case there is variation between the products
- */
-export const VERSION_IN_INTEGRATIONS_PATH_REGEX = /v?\d+\.\d+\.(\d+|\w+)/i
--- a/src/constants/version-path.ts
+++ b/src/constants/version-path.ts
@@ -3,3 +3,9 @@ export const VERSION_IN_PATH_REGEX = /v\d+\.\d+\.(\d+|\w+)/i
 
 /** This REGEX is used to parse a Terraform Enterprise version from a URL */
 export const TFE_VERSION_IN_PATH_REGEXP = /v[0-9]{6}-\d+/i
+
+/**
+ * TODO - validate the expected versioned integrations paths for other products
+ * Left the 'v' as optional here in case there is variation between the products
+ */
+export const VERSION_IN_INTEGRATIONS_PATH_REGEX = /v?\d+\.\d+\.(\d+|\w+)/i
--- a/src/lib/get-version-from-path.ts
+++ b/src/lib/get-version-from-path.ts
@@ -1,6 +1,7 @@
 import {
 	VERSION_IN_PATH_REGEX,
 	TFE_VERSION_IN_PATH_REGEXP,
+	VERSION_IN_INTEGRATIONS_PATH_REGEX,
 } from 'constants/version-path'
 
 /**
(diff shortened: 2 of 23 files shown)


src/constants/version-path.ts
```diff
@@ -3,3 +3,9 @@ export const VERSION_IN_PATH_REGEX = /v\d+\.\d+\.(\d+|\w+)/i
 
 /** This REGEX is used to parse a Terraform Enterprise version from a URL */
 export const TFE_VERSION_IN_PATH_REGEXP = /v[0-9]{6}-\d+/i
+
+/**
+ * TODO - validate the expected versioned integrations paths for other products
+ * Left the 'v' as optional here in case there is variation between the products
+ */
+export const VERSION_IN_INTEGRATIONS_PATH_REGEX = /v?\d+\.\d+\.(\d+|\w+)/i
```
kendallstrautman, line 11: @BrandonRomano, all of the other docs versioned paths have `v` in the path itself. Is this something we want to align for consistency across the platform with integrations paths? 

If not, we'll have to treat these versioned paths as a one-off, which is okay but I want to make sure we made an intentional choice to remove it.

BrandonRomano: Ah, we should probably include the v in the path, then!

src/lib/get-version-from-path.ts
```diff
@@ -20,9 +20,10 @@ export function getVersionFromPath(path: string): string | undefined {
 		if (i === 2 && TFE_VERSION_IN_PATH_REGEXP.test(el)) {
 			return true
 		}
-		if ((i === 2 || i === 3) && VERSION_IN_PATH_REGEX.test(el)) {
+		if ((i === 2 || i === 3 || i === 4) && VERSION_IN_PATH_REGEX.test(el)) {
```
kendallstrautman, line 23: Integration versioned paths are one step deeper than the rest of the docs.

src/lib/get-version-from-path.ts
```diff
@@ -20,9 +20,10 @@ export function getVersionFromPath(path: string): string | undefined {
...
 			return true
 		}
+
```
kendallstrautman, line 26:
```suggestion
```

kendallstrautman:
Interestingly, the old paths (without the 'v' prefix) still work, since they can fetch to the API 🤔 I sort of think this is _okay_ but we could specifically 404 on those if we want?

src/lib/get-version-from-path.ts
BrandonRomano: There's a comment above we should update as well as this code here:
> 	// version is only expected to be at index 2, or 3 in the case of TF-Plugins

BrandonRomano (approved):
This looks good!  I caught a comment that is stale that we probably should update on one of the changes that you made to the thing that parses the version out of the URL.

BrandonRomano:
> Interestingly, the old paths (without the 'v' prefix) still work, since they can fetch to the API 🤔 I sort of think this is okay but we could specifically 404 on those if we want?

Hmm; I feel like we would want to 404, or we would want to do a permanent redirect to the location with the `v`.  I think I lean more towards a permanent redirect towards the page with the `v` in it.

kendallstrautman:
@BrandonRomano, yeah I agree. I'll update this to not accept that path and 404 if just the param (without 'v') is used. I don't think we need a redirect at this point since its not shipped yet.

src/views/product-integration/component-view/server.ts
```diff
@@ -135,13 +136,22 @@ async function getStaticProps({
...
 		}
 	}
+
+	const [targetVersion] = getTargetVersion({
+		versionSlug: integrationVersion,
+		latestVersion: integration.versions[0],
+	})
+
+	// if the version slug is not prefix with 'v', return 404
+	if (targetVersion === null) {
```
kendallstrautman, line 146: @BrandonRomano Adjusted this so we 404 on improperly formatted versioned slugs.

BrandonRomano (approved):

PR #1585, opened 2023-01-27
add stable redirect for Nomad envoy troubleshooting

gulducat:
## 🔗 Relevant links

Nomad PR: hashicorp/nomad#15908

## 🗒️ What

In that ^ PR, I'm adding a link to an error message and would like to use this short and stable redirect link.


lgfa29 (approved):
Maybe the @hashicorp/dev-portal can clarify this, but I think we can still add redirects to https://github.com/hashicorp/nomad/blob/main/website/redirects.js? 

I wonder if these `/s` redirects should be kept there so we can find them more easily while searching the codebase.

kendallstrautman:
@lgfa29 Do you want this redirect to root from developer? The redirects in this context are from the .io domain. Since all the docs paths are redirected from that domain, we should only add new redirects from the developer root. Unless this is an old unsupported .io redirected path? 

If you want to add a redirect for developer, you add it in the redirects.js file in the nomad repo (I saw you added some on your PR).

lgfa29:
Hum...not sure if I follow so let me explain our use case for these.

We have some URLs that are hardcoded in our codebase to point users to more expansive documentation. For example here: https://github.com/hashicorp/nomad/blob/v1.4.3/nomad/plan_apply.go#L551

The tricky part is that the documentation we point to may change, like if we write better docs somewhere else, so these `/s` redirects allow us to change the content users land on without having to worry about modifying code.

So what we need is that these redirects work, because that's what we're using in our code:
 - https://www.nomadproject.io/s/port-plan-failure -> https://developer.hashicorp.com/nomad/docs/operations/monitoring-nomad#progress
 - https://www.nomadproject.io/s/envoy-bootstrap-error -> https://developer.hashicorp.com/nomad/docs/integrations/consul-connect#troubleshooting 

What I mentioned was that having the redirects declared in our repo would be helpful since they would be closer to the code that references them.

So are there any concerns in always adding these `/s` redirects to https://github.com/hashicorp/nomad/blob/main/website/redirects.js instead of here?

BRKalow (approved):

PR #1589, opened 2023-01-30
feat: remark process variable descriptions

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Updates variable descriptions to render as rich markdown content processed on the server with `next-mdx-remote`. This replaces the previous client-side processing approach with `react-markdown`.

## 🛠️ How

- Adds a helper to get processed markdown for each variable
- Update `getStaticProps` to process variable descriptions into MDX using the above helper, and pass a map of processed content to the view
- Updates `VariableGroupList` to use `DevDotContent` to render content the processed variable description map
- Adds some style tweaks to account for smaller `body-200` base font size (for detail see `Anything else`)
    - This includes updating `MdxInlineCode` to support `code-100` size styling, via a new optional `size` prop
- Uninstalls `react-markdown`, no longer used
- Tangent fix: Updates "readme" and "component" content rendering to use `DevDotContent` for consistency

## 📸 Design Screenshots

| View | Before | After |
| - | - | - |
| Variable description | variables-before | variables-after |
| Readme | readme-before | readme-after |
| Component description | component-before | component-after |

> **Note**: the "readme" and "component description" fixes are more minor, and somewhat tangential, but felt close enough to the scope of this PR to be worth including.

## 🧪 Testing

> **Warning**: I couldn't easily find variable descriptions that included markdown. So as a temporary measure for testing, this PR prefixes each variable description with some markdown before it is processed, as a way to make it easier to confirm that markdown processing will work when markdown is included in variable descriptions.

- [ ] Visit a component page with many variables with descriptions, such as [/waypoint/integrations/BrandonRomano/kubernetes/latest/components/platform][/waypoint/integrations/BrandonRomano/kubernetes/latest/components/platform]
    - Variables with markdown content should render the markdown as properly styled elements (via `MDXRemote`).
    - Variables that have a `null` description, such as `pod.security_context.run_as_group`, should not display any description element.
    - Aside: the top-level description, above all the variable groups, should now be rendering with our shared long-form content components, similar to Docs or Tutorials content. For example, code blocks should have a dark background and a `Copy` button, rather than being a simple unstyled `<pre />` element.
- [ ] Visit a readme page, such as [/waypoint/integrations/BrandonRomano/kubernetes][/waypoint/integrations/BrandonRomano/kubernetes]
    - The "readme" content should now be rendering with our shared long-form content components, similar to Docs or Tutorials content.

### Note on styling of processed markdown

Setting up `remark` & `next-mdx-remote` processing alone gives us processed markup, but in its simplest form does not give us any of the styling we apply to other "processed markdown" contexts, such as Docs and Learn content.

As well, for variable `description` display, we're using a **smaller base font size** than in typical long-form content (the smaller `body-200` rather than then standard `body-300`). So, MDX components will need to be adjusted slightly so that their font-sizes match. For now, I've made tweaks to the following components to ensure they work a base `body-200` font-size:
- `<strong />` and `<em />`, for bold text and italic text
- `<a />`, for inline links
- `<code />`, for inline code

📌 In the future, we likely want to go back to design and ask if `body-200` content with rich markdown content needs will be a consistent theme. We've seen this already in the MDX Alert component work (within alerts, the font-size is bumped down to `body-200`). So, if we have a strong need for this variation, this may be a use case worth thinking about and building a spec for in more detail. I've made a task for this at 🎟 [[discuss] smaller "body-200" mode for DevDotContent](https://app.asana.com/0/1202097197789424/1203869000261274/f).

> **Note**: The "readme" and "component" descriptions suffered from a similar issue, where MDX content was being rendered, but without our styled components. I've updated those areas to use the recently consolidated `DevDotContent` component (props to Ashlee for that consolidation work!).

### Note on heading and anchor-link processing

💭 On "readme" views, we are rendering MDX content, similar to Docs and Learn. However, we're not running all the same `remark` plugins, so permalinks are not generated for the headings in "readme" content. We could potentially generate permalinks, and even include a sidecar with a table of contents, if we want. Maybe this is something we want to add in a subsequent PR?

[task]: https://app.asana.com/0/1203792701577283/1203697277089836/f
[preview]: https://dev-portal-git-zsensure-variable-group-md-processing-hashicorp.vercel.app/
[/waypoint/integrations/BrandonRomano/kubernetes/latest/components/platform]: https://dev-portal-git-zsensure-variable-group-md-processing-hashicorp.vercel.app/waypoint/integrations/BrandonRomano/kubernetes/latest/components/platform
[/waypoint/integrations/BrandonRomano/kubernetes]: https://dev-portal-git-zsensure-variable-group-md-processing-hashicorp.vercel.app/waypoint/integrations/BrandonRomano/kubernetes


zchsh:
@BrandonRomano This PR implements server-side markdown processing for `variable.description` values via `next-mdx-remote`.

As a tangential fix, I've also updated the "readme" and "component" MDX content to use the shared `DevDotContent` component, which ties things together so we get all the markdown element styles we need (similar to Docs or Tutorials content).

I have some other notes in the `Anything else` section, mainly related to some future thoughts around more first-class support for "smaller" rich content (as we're doing here for `variable.description`), as well as for heading permalinks and table-of-contents for the "readme" view (felt out of scope of this PR, but worth mentioning as it's very `remark`-related).

Let me know what you think 🙇

package.json
```diff
@@ -195,7 +195,6 @@
 		"react-hot-toast": "^2.2.0",
 		"react-instantsearch-hooks-web": "^6.33.0",
 		"react-intersection-observer": "^8.33.1",
-		"react-markdown": "^8.0.3",
```
BrandonRomano, line 198: :chef-kiss:

src/views/product-integration/component-view/components/variable-group-list/index.tsx
```diff
@@ -1,12 +1,24 @@
...
+const smallDescriptionMdxComponents = {
+	inlineCode: (props) => <MdxInlineCode {...props} size="100" />,
+	p: (props) => <MdxP {...props} size="200" />,
+}
+
 export interface Variable {
 	key: string
 	type: string
+	descriptionMdx?: MDXRemoteSerializeResult
 	description?: string
```
BrandonRomano, line 22: Should this be removed now?

src/views/product-integration/component-view/helpers/get-processed-variables-markdown.ts
```diff
@@ -0,0 +1,39 @@
...
+	const processedVariablesMarkdown: ProcessedVariablesMarkdown = {}
+	const { variable_groups } = releaseComponent
+	for (const variableGroup of variable_groups) {
+		for (const variable of variableGroup.variables) {
+			const { key, description, variable_group_id } = variable
+			const uniqueKey = `${variable_group_id}.${key}`
+			processedVariablesMarkdown[uniqueKey] = {}
+			if (description !== null) {
+				// TODO: remove dev_markdown test values here
+				const dev_markdown = `**Test bold**, _emphasis_, also [a link](https://dev.hashicorp.com/not-visited-yet), and \`some inline code\`, heck maybe [\`linked code\`](https://dev.hashicorp.com/yoink)`
```
BrandonRomano, line 30: I would also like to see what a code block looks like 🤔 

```sh
echo "something" # like this
echo "wow"
```

BrandonRomano (approved):
This looks good to me!  Two very small comments, but this is looking very good.

src/views/product-integration/component-view/components/variable-group-list/index.tsx
zchsh: Ah good callout, I needed to add comments here to clarify similar to the one in `component-view/index.tsx` - `descriptionMdx` is used to render the description visibly, while the plain `description` string is used for searching and filtering purposes 👍

src/views/product-integration/component-view/helpers/get-processed-variables-markdown.ts
zchsh: Great question - I temporarily added it in to the dev value, here's a preview:

2023-02-02-code-block-demo

Caveat: with the reduced "base" font size here compared to other long-form content (eg Docs and Learn), elements _other_ than the inline ones we've tweaked styles for in this PR won't match the reduced font-size used for rendering `description` content. We've flagged this previously with design during "alert" work (alerts also use `body-200` rather than the larger `body-300`) but might be worth flagging again if it seems like a priority to address this use case 🎏 

If we want to more fully account for "block" MDX components with this smaller `body-200` base size, I think we'd need a good chunk design involvement to figure out spacing and necessary tweaks to components such as code-block (unless we're happy with the differing relative font sizes & other metrics!). I've got a task set up with details: 🎟 [[discuss] smaller "body-200" mode for DevDotContent](https://app.asana.com/0/1202097197789424/1203869000261274/f)

zchsh:
Update: I've rebased this onto `main`, and removed the dev content test - will merged once I double-check the deploy preview to make sure it's all good ☺️

PR #1590, opened 2023-01-30
[wip] chore: adds AB test for hcp callout

kendallstrautman:
## 🔗 Relevant links

- [Asana task](url) 🎟️

### DIFF
--- a/src/components/try-hcp-callout/components/a-b-test.tsx
+++ b/src/components/try-hcp-callout/components/a-b-test.tsx
@@ -1,0 +1,15 @@
+import { useFlags } from 'flags/client'
+
+const ctaSuffix = 'for free'
+
+export function useABTestCta(ctaText: string) {
+	// use the flag here
+	const { flags } = useFlags()
+	let final = ctaText
+
+	if (flags?.tryCloudFreeCta && !ctaText.endsWith(ctaSuffix)) {
+		final = `${ctaText} ${ctaSuffix}`
+	}
+
+	return final
+}
--- a/src/components/try-hcp-callout/components/try-hcp-callout-compact/index.tsx
+++ b/src/components/try-hcp-callout/components/try-hcp-callout-compact/index.tsx
@@ -5,6 +5,7 @@ import {
 	StandaloneLinkContents,
 } from 'components/try-hcp-callout/components'
 import { TryHcpCalloutCompactProps } from 'components/try-hcp-callout/types'
+import { useABTestCta } from '../a-b-test'
 import s from './try-hcp-callout-compact.module.css'
 
 /**
@@ -27,7 +28,7 @@ export function TryHcpCalloutCompact({
 				size="small"
 			/>
 			<Description description={description} />
-			<StandaloneLinkContents text={ctaText} size="small" />
+			<StandaloneLinkContents text={useABTestCta(ctaText)} size="small" />
 		</CardLink>
 	)
 }
--- a/src/components/try-hcp-callout/components/try-hcp-callout-prebuilt/index.tsx
+++ b/src/components/try-hcp-callout/components/try-hcp-callout-prebuilt/index.tsx
@@ -1,6 +1,7 @@
 import { TryHcpCallout } from '..'
 import { ProductSlugWithContent } from '../../types'
 import { tryHcpCalloutContent } from '../../content'
+import { useABTestCta } from '../a-b-test'
 
 /**
  * A wrapper around TryHcpCallout.
@@ -18,7 +19,7 @@ export function TryHcpCalloutPrebuilt({
 		tryHcpCalloutContent[productSlug]
 	return (
 		<TryHcpCallout
-			ctaText={ctaText}
+			ctaText={useABTestCta(ctaText)}
 			ctaUrl={ctaUrl}
 			description={description}
 			heading={heading}
--- a/src/components/navigation-header/components/product-page-content/index.tsx
+++ b/src/components/navigation-header/components/product-page-content/index.tsx
@@ -14,11 +14,22 @@ import {
 import { ProductIconTextLink } from './components'
 import { getNavItems, getProductsDropdownItems, NavItem } from './utils'
 import s from './product-page-content.module.css'
+import { useABTestCta } from 'components/try-hcp-callout/components/a-b-test'
 
 const ProductPageHeaderContent = () => {
 	const currentProduct = useCurrentProduct()
 	const allProductsItems = getProductsDropdownItems()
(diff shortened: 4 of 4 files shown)


src/components/try-hcp-callout/components/a-b-test.tsx
```diff
@@ -0,0 +1,15 @@
+import { useFlags } from 'flags/client'
+
+const ctaSuffix = 'for free'
+
+export function useABTestCta(ctaText: string) {
+	// use the flag here
```
kendallstrautman, line 6:
```suggestion
```

src/components/try-hcp-callout/components/a-b-test.tsx
```diff
@@ -0,0 +1,27 @@
...
+	if (isInUS() && flag && !ctaText.endsWith(ctaSuffix)) {
+		final = `${ctaText} ${ctaSuffix}`
+	}
+
+	useEffect(() => {
+		abTestTrack({
+			type: 'Served',
+			test_name: 'HCP CTA Trial 2023-02',
+			variant: flag.toString(),
+		})
```
kendallstrautman, line 23: Need to validate if this is the correct way to handle a GM campaign. cc @alexcarpenter

kendallstrautman: 🤔 Thinking through whether I should add a plain 'click' track event for the CTA link, or if I use this function again with 'Result'. Open to feedback!

src/components/try-hcp-callout/components/a-b-test.tsx
```diff
@@ -0,0 +1,27 @@
...
+	const { flags } = useFlags()
+	const flag = flags?.tryCloudFreeCta || false
+
+	let final = ctaText
+
+	if (isInUS() && flag && !ctaText.endsWith(ctaSuffix)) {
+		final = `${ctaText} ${ctaSuffix}`
+	}
+
+	useEffect(() => {
```
kendallstrautman, line 18: TODO send this after `settled`

src/components/try-hcp-callout/components/a-b-test.tsx
```diff
@@ -0,0 +1,27 @@
+import { useEffect } from 'react'
+import { isInUS } from '@hashicorp/platform-util/geo'
+import { useFlags } from 'flags/client'
+import { abTestTrack } from 'lib/ab-test-track'
+
+const ctaSuffix = 'for free'
+
+export function useABTestCta(ctaText: string) {
+	const { flags } = useFlags()
```
kendallstrautman, line 9:
```suggestion
	const { flags, settled } = useFlags()
```

src/components/try-hcp-callout/components/a-b-test.tsx
kendallstrautman: I think I'll wait to hear back from the GM folks to see if they need more events and not add the 'result' yet. Because i'm sure we already have events with product intent firing to the cloud platform.

kendallstrautman:
💭 I should probably update that ab test hook filename to be more clear and maybe put it in `lib`

emilypersson1 (approved):
@kendallstrautman thank you! looks good on my end, and thanks for calling out the jump in the text. I think it's totally fine

zchsh (approved):
✅ ran through the testing steps! 

... and can confirm it is US-only  😅  took me a minute or two to be like "there must be some reason I'm not seeing the suffix" and then realized.... oh yeah I'm not in the US 🤦  used a VPN to confirm it does indeed work for US folks 👍

src/components/try-hcp-callout/components/try-hcp-callout-compact/index.tsx
```diff
@@ -18,6 +19,7 @@ export function TryHcpCalloutCompact({
 	ctaText,
 	ctaUrl,
 }: TryHcpCalloutCompactProps) {
+	const trialCtaText = useABTestCta(ctaText)
```
BRKalow, line 22: Maybe we can leave a breadcrumb here to remove this when the test is finished:
```suggestion
	// TODO: remove this when the HCP CTA Trial 2023-02 test is finished
	const trialCtaText = useABTestCta(ctaText)
```

src/components/try-hcp-callout/components/try-hcp-callout-prebuilt/index.tsx
```diff
@@ -16,9 +17,10 @@ export function TryHcpCalloutPrebuilt({
 }) {
 	const { ctaText, ctaUrl, description, heading } =
 		tryHcpCalloutContent[productSlug]
+	const trialCtaText = useABTestCta(ctaText)
```
BRKalow, line 20:
```suggestion
		// TODO: remove this when the HCP CTA Trial 2023-02 test is finished
	const trialCtaText = useABTestCta(ctaText)
```

src/components/try-hcp-callout/components/a-b-test.tsx
```diff
@@ -0,0 +1,29 @@
+import { useEffect } from 'react'
+import { isInUS } from '@hashicorp/platform-util/geo'
+import { useFlags } from 'flags/client'
+import { abTestTrack } from 'lib/ab-test-track'
+
+const ctaSuffix = 'for free'
+
+export function useABTestCta(ctaText: string): string {
+	const { flags, settled } = useFlags()
```
BRKalow, line 9: I think this hook might be intended to be only called once per page. Maybe not an issue now, but if we add more tests we'll need to lift this up to _app or a page component and pass the flag data down.

BRKalow (approved):
Implementation LGTM, left a few notes

src/components/try-hcp-callout/components/a-b-test.tsx
kendallstrautman: 🤔 good point, i forgot about that! I'll check to see if other flags are in use now and reconsider if so.

src/components/try-hcp-callout/components/try-hcp-callout-compact/index.tsx
kendallstrautman: 🧠 will do!

src/components/try-hcp-callout/components/a-b-test.tsx
kendallstrautman: I made a PR to[ remove the `enableAuth` check here,](https://github.com/hashicorp/dev-portal/pull/1607) which was the only other call for `useFlags` that I saw. Also I added a comment with your suggestion as a future note. Since this is the only test running atm, I'm okay with not lifting it up.

PR #1599, opened 2023-01-31
fix(middleware): handle simple redirects with hash segment

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Fixes an edge case in our middleware redirect handling where we were not properly assigning the hash segment of a URL. This caused the hash to be url encoded as part of the path, breaking the intended destination.

## 🧪 Testing

- Visit the preview, select boundary in the product switcher
- visit: `/help/admin-ui/targets/worker-filters`
- Validate the redirect worked and the URL contains a non-encoded `#`

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -68,7 +68,14 @@ export async function middleware(req: NextRequest, ev: NextFetchEvent) {
 		// Next.js doesn't support redirecting to a pathname, so we clone the
 		// request URL to adjust the pathname in an absolute URL
 		const url = req.nextUrl.clone()
-		url.pathname = destination
+
+		// Simple redirects _can_ contain a hash, which we need to detect and handle
+		const [pathname, hash] = destination.split('#')
+		url.pathname = pathname
+		if (hash) {
+			url.hash = hash
+		}
+
 		console.timeEnd(label)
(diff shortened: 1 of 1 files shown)


src/middleware.ts
```diff
@@ -68,7 +68,14 @@ export async function middleware(req: NextRequest, ev: NextFetchEvent) {
...
 		// request URL to adjust the pathname in an absolute URL
 		const url = req.nextUrl.clone()
-		url.pathname = destination
+
+		// Simple redirects _can_ contain a hash, which we need to detect and handle
+		const [pathname, hash] = destination.split('#')
+		url.pathname = pathname
+		if (hash) {
+			url.hash = hash
+		}
```
BRKalow, line 77: This was an uncaught edge case with how we handle our "simple redirects" that are delegated to middleware. Redirects with query params are not handle here, so it seems reasonable to only worry about the hash.

ashleemboyer:
I followed the test steps, and am still seeing the old .io view of docs. Is that expected because I'm viewing the page via the product switcher?

image

BRKalow:
@ashleemboyer yeah that would be expected in the preview environment. 👍

ashleemboyer (approved):

PR #1601, opened 2023-02-01
[WIP] spike: dark mode token map

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksdark-mode-spike-hashicorp.vercel.app/) 🔎
- [Asana task](url) 🎟️

## 🗒️ What

Testing some new tokens from HDS.


src/components/_proxied-dot-io/boundary/footer/style.css
```diff
@@ -10,7 +10,7 @@
 	}
 
 	& a {
-		color: black;
+		color: var(--token-color-palette-neutral-700);
```
kendallstrautman, line 13: We don't have a clear value for #000 in the tokens, so I chose the 'darkest' neutral color for now.

package.json
```diff
@@ -184,6 +184,7 @@
 		"next-mdx-remote": "^3.0.8",
 		"next-query-params": "^4.1.0",
 		"next-sitemap": "^3.1.21",
+		"next-themes": "^0.2.1",
```
kendallstrautman, line 187: Just refactored this to test out next-themes 🖤

src/components/theme-switch/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+	useEffect(() => {
+		setMounted(true)
+	}, [])
+
+	if (!mounted) {
+		return null
+	}
+
+	return (
+		<select value={theme} onChange={(e) => setTheme(e.target.value)}>
```
kendallstrautman, line 18: [Copied this from their docs](https://github.com/pacocoursey/next-themes#avoid-hydration-mismatch), the system default and having that as an item in the dropdown select is slick.

PR #1600, opened 2023-02-01
feat: redirect waypoint plugins to integrations

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Adds redirects from `/waypoint/plugins` to `/waypoint/integrations`.

🚧 **TODO**: Determine whether we also want to remove `/waypoint/plugins` content, and links to Waypoint plugins from within Dev Dot (such as in the top nav). If so, should probably follow this PR up pretty quickly with that work. We'd remove `src/pages/waypoint/plugins`, and the top nav link, and there are probably some other stray links we'd want to update. As well, we should ensure the Waypoint team knows that content in `website/content/plugins` will no longer be rendered to Dev Dot.

## 🛠️ How

Adds a redirect from `/waypoint/plugins/:slug` to `/waypoint/integrations/BrandonRomano/:slug`, as well as a simple redirect from`/waypoint/plugins` to `/waypoint/integrations`.

 🚧 **TODO**: likely want to update `BrandonRomano` to be `HashiCorp` in URLs. Or `hashicorp`? Not sure on casing. This change probably makes sense to do _after_ moving integrations over so the API reflects this change - which I think would be captured by 🎟 [Migrate `BrandonRomano/waypoint` fork to `HashiCorp/waypoint` branch](https://app.asana.com/0/1203792701577283/1203868826798341).

## 🧪 Testing

Visit all of the following `/waypoint/plugins` links, and confirm they redirect as expected, to pages that exist:

- [/waypoint/plugins][/waypoint/plugins]
- [/waypoint/plugins/aws-ec2][/waypoint/plugins/aws-ec2]
- [/waypoint/plugins/aws-ecs][/waypoint/plugins/aws-ecs]
- [/waypoint/plugins/aws-lambda][/waypoint/plugins/aws-lambda]
- [/waypoint/plugins/aws-ssm][/waypoint/plugins/aws-ssm]
- [/waypoint/plugins/azure-container-instance][/waypoint/plugins/azure-container-instance]
- [/waypoint/plugins/consul][/waypoint/plugins/consul]
- [/waypoint/plugins/docker][/waypoint/plugins/docker]
- [/waypoint/plugins/exec][/waypoint/plugins/exec]
- [/waypoint/plugins/google-cloud-run][/waypoint/plugins/google-cloud-run]
- [/waypoint/plugins/helm][/waypoint/plugins/helm]
[...]

PR #1602, opened 2023-02-01
feat: rewrite waypoint plugins → integrations

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

This PR adds rewriting logic for URLs redirected in #1600. The intent here is to avoid rendering links on our site that we know will redirect.

> **Note**: ideally we'd update these links in content, and this is something we'll want to do regardless. However, we won't catch past versions, and we'll also have to co-ordinate those rewrites in content with the launch of Waypoint integrations, so this approach of automatically fixing the links at `getStaticProps` time felt like it would be a helpful solution for initial launch.

## 🛠️ How

🚧 TODO

- feat: waypoint plugins → integrations for docs
- chore: move rewrite fn to lib/content-adjustments
- feat: waypoint plugins → integrations in tutorials

## 🧪 Testing

🚧 TODO

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/1203792701577283/1203812579407375/f
[preview]: https://developer.hashicorp.com


src/lib/content-adjustments/rewrite-waypoint-plugins-to-integrations.ts
```diff
@@ -0,0 +1,49 @@
...
+		// Handle the root path
+		return '/waypoint/integrations'
+	} else if (inputUrl.startsWith('/waypoint/plugins')) {
+		// Handle all plugins paths
+		const pluginSlug = inputUrl.replace('/waypoint/plugins/', '')
+		// Note: under `/plugins` we only ever had official `hashicorp` plugins
+		// TODO: confirm whether the above is true, I'm actually not sure
+		//       (if this isn't true, could maybe do an old plugin name -> org map?)
+		// TODO: update from `BrandonRomano` to `hashicorp` once API content updates
+		const pluginOrg = 'BrandonRomano'
```
BrandonRomano, line 43: I've included this in https://app.asana.com/0/1203792701577283/1203879536690930

BrandonRomano (approved):
This feels like it gets the job done for now! 

I spun up a few tasks to eventually unwind from this, but this feels like a decent temporary solution for now.

https://app.asana.com/0/1203792616809419/1203879536690933

PR #1603, opened 2023-02-01
Tweak rewrite-links script for Terraform content

ashleemboyer:
## 🗒️ What

- Updates the rewrite links script `getMdxLinksToRewrite` helper to check if there is a `source` property in the frontmatter of an MDX file. If the value for that property is different than the repo having its links rewritten, then the file will be ignored.

## 🤷 Why

- The `ptfe-releases` repo is one that copies some MDX files directly from another repo. When the copy happens, the files are completely overwritten, so changes need to be made at the source.

## 🧪 Testing

- [ ] Make sure you have a local clone of the `ptfe-releases` repo
- [ ] With `dev-portal/main` checked out locally, run the link rewrite script on your `ptfe-releases` copy (`localCopyLocation` is the _folder_ where `ptfe-releases` is located):
  ```
   VERCEL_URL="developer.hashicorp.com" npx hc-tools ./scripts/docs-content-link-rewrites/rewrite-links.ts -- --localCopyLocation <FOLDER_WITH_REPO_CLONE> --repo ptfe-releases
  ```
- [ ] Note how many files `ptfe-releases` were changed by the script
- [ ] Stash the changes
- [ ] Check out `dev-portal/amb.ignore-some-rewrites`
- [ ] Run the script again, just like above
- [ ] There should be significantly less files changed
- [ ] None of the files changed should have a `source` referencing a different repo


scripts/docs-content-link-rewrites/helpers/get-mdx-links-to-rewrite.ts
```diff
@@ -40,6 +43,22 @@ const getMdxLinksToRewrite = async ({
...
+		 *
+		 * Example file in the `ptfe-releases` repository:
+		 * https://github.com/hashicorp/ptfe-releases/blob/ba4dc2af20cc9d0f46247c919be8b2ed90d1e6f8/website/docs/enterprise/registry/index.mdx?plain=1#L6
+		 */
+		const { data: frontmatter } = matter(fileContent)
+		const isSourcedFromDifferentRepo =
+			typeof frontmatter?.source === 'string' && frontmatter.source !== repo
+		if (isSourcedFromDifferentRepo) {
+			continue
+		}
```
BRKalow, line 60: @thiskevinwang Is the existence of the `source` frontmatter property enough to know we should ignore the document, or do we need to do a check like this? ☝️

thiskevinwang: Is the goal/intent to ignore a class of files managed via automation?

If so it _feels like_ simply checking for existence of `source` should suffice here.

thiskevinwang (approved):

PR #1604, opened 2023-02-01
refactor: use config flag

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksuse-config-flag-hashicorp.vercel.app/vault/integrations) 🔎
- [Asana task](https://app.asana.com/0/1199634971449915/1203869414967131) 🎟️

## 🗒️ What

Refactor to use our config flags to enable product integrations. This way we can toggle which integrations are enabled per environment, and it aligns with the way we handle global config in the app, as opposed to in the data files. 

## 🧪 Testing

- Validate that both the vault and waypoint integrations paths are enabled [on this preview](https://dev-portal-git-ksuse-config-flag-hashicorp.vercel.app/). 
- Pull this branch down locally and remove `waypoint` from the `config/development.json` file, start the dev server and you should see that waypoint integrations paths or UI are no longer rendered. This is to validate that when waypoint isn't in that array (as on the production config file), it won't get built. 
- Test search, validate that the tabs show up as expected, no changes

## 💭 Anything else?

We can close this after this is merged https://github.com/hashicorp/dev-portal/pull/1606

PR #1605, opened 2023-02-01
Add Integrations tab to Global Search

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambgs-add-integrations-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1203847455428145/f) 🎟️

## 🗒️ What/Why

- Adds `product_slugs_with_integrations` config array, so there is a client-side way to know which products have integrations enabled on an environment level
  - Enables global search for "vault" in all environments
  - Enables global search for "waypoint" in non-prod environments
- Adds a temporary `integrationsHits` prop to `CustomHitsContainer`, which is used to conditionally render `IntegrationHit` components
- Updates `SearchCommandBarDialogBodyContent` to render `Tab` components based on a map of `SearchableContentType` to objects with props for each `Tab`.
  - This new approach makes it possible to keep `activeTabIndex` and the order of the `Tab`s in automatically in sync. 
- Abstracts a new `TabContentsCta` subcomponent, for consistent use in the tutorials and integrations tabs
- Adds support for `trailingIcon` to `CommandBarListItemContent`, for an external link icon on some `IntegrationHit`s.
- Adds `useIntegrationsByProductSlugs` hook to leverage using React Query to fetch integrations a minimal number of times for global search's use
- Sets `contentType = 'integrations'` on a few views:
  - `ProductIntegrationComponentView`
  - `ProductIntegrationReadmeView`
  - `ProductIntegrationsLanding`
- Abstracts `getFilteredIntegrations` for use in global search in addition to `SearchableIntegrationsList`

## 🎨 Designs

https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=18759%3A236272&t=Nj3ZnYyA5UvX1V0T-11

### When not within a product's context

- [ ] Go to [the home page](https://dev-portal-git-ambgs-add-integrations-hashicorp.vercel.app/)
- [ ] Open command bar
- [ ] Under "Suggested Pages", you should **not** see any links having to do with integrations
- [ ] Search for something that will get results in all three tabs (like "engine")
- [ ] The Documentation & Tutorials tabs should still work as before
- [ ] The Integrations tab should show results as designed
- [ ] Scroll to the bottom of the Integrations tab
- [ ] There should **not** be a CTA pointing to an integrations search page

### When within a product's context and `contentType` is not `"integrations"`

- [ ] Go to [/vault](https://dev-portal-git-ambgs-add-integrations-hashicorp.vercel.app/vault)
- [ ] Open command bar
- [ ] Under "Suggested Pages", you should see a new link after "Vault Tutorials" for the integrations page
  - [ ] The link should look and behave as designed
- [ ] Search for something that will get results in all three tabs (link "engine")
- [ ] The Documentation & Tutorials tabs should still work as before
- [ ] The Integrations tab should show results as designed
- [ ] Scroll to the bottom of the Integrations tab
- [ ] There should be a CTA pointing to /vault/integrations that says "See all Vault integrations"

### When within a product's context and `contentType` is `"integrations"`

- [ ] Go to [/vault/integrations](https://dev-portal-git-ambgs-add-integrations-hashicorp.vercel.app/vault/integrations)
- [ ] Open command bar
- [ ] Search for something that will get results in all three tabs (link "engine")
- [ ] The Integrations tab should be the one that is currently selected

## 💭 Anything else?

- Since the integrations pages aren't based off of the new `product_slugs_with_integrations` config array, those pages are going to 404. But since #1604 will be refactoring that logic to use this new variable, I decided to go ahead and enable `"waypoint"` in non-prod environments.


config/base.json
```diff
@@ -34,7 +34,8 @@
 			"applicationId": "40a2019f-88ad-4076-ab52-ffd074064eed",
 			"clientToken": "pubd5cf774aa1cbbd001accd50cb463925b",
 			"service": "non-prod.developer.hashicorp.com"
-		}
+		},
+		"product_slugs_with_integrations": ["vault", "waypoint"]
```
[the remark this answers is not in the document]
ashleemboyer: The main difference is that `flags` has been used for just booleans, and this object is used for other data types.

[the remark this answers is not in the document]
BrandonRomano: Cool cool;  I like your placement then of this object here better.  I do like the language usage of `enabled_*` used in the key on the other PR though!

I will let you & @kendallstrautman decide though what the best key is to manage this, just wanted to share my opinions!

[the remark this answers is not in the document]
ashleemboyer: If it's alright, it'd be convenient if the rename was handled in the next PR since this isn't my primary focus this week 🙏🏻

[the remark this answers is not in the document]
BrandonRomano: Yeah that's totally fine!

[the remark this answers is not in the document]
kendallstrautman: I'm cool with it not being in `flags`, that makes sense. But I do think having this under `dev_dot` is a better place, since integrations don't apply to .io config. But agreed we can adjust that in the refactor PR!

src/components/command-bar/commands/search/helpers/generate-suggested-pages.tsx
```diff
@@ -63,6 +64,16 @@ const generateBasicSuggestedPages = (productSlug: ProductSlug) => {
...
+	const hasIntegrationsEnabled =
+		__config.dev_dot.product_slugs_with_integrations.includes(productSlug)
+	if (hasIntegrationsEnabled) {
+		pages.push({
+			icon: <IconPipeline16 />,
+			text: `${productSlugsToNames[productSlug]} Integrations`,
+			url: `/${productSlug}/integrations`,
+		})
+	}
+
```
kendallstrautman, line 76: Ooo I need to make a TODO for myself to refactor this to use the [helpers in this file](https://github.com/hashicorp/dev-portal/pull/1604/files#r1093562661)

kendallstrautman: Sorry, just refactor the `hasIntegrationsEnabled` check, not the whole block :smile

src/hooks/integrations/use-integrations-by-product-slugs.ts
```diff
@@ -0,0 +1,40 @@
...
+interface UseIntegrationsByProductSlugsResult
+	extends Omit<UseQueryResult<QueryDataType>, 'data'> {
+	integrations: UseQueryResult<QueryDataType>['data']
+}
+
+/**
+ * Handles fetching and storing integrations with React Query for the given
+ * `productSlugs`.
+ */
+const useIntegrationsByProductSlugs = ({
```
kendallstrautman, line 24: Nice 💯

src/views/product-integrations-landing/components/searchable-integrations-list/helpers/get-filtered-integrations.ts
```diff
@@ -0,0 +1,23 @@
...
+}) => {
+	return integrations.filter((integration: Integration) => {
+		return (
+			integration.name.toLowerCase().includes(filterQuery.toLowerCase()) ||
+			integration.description
+				.toLowerCase()
+				.includes(filterQuery.toLowerCase()) ||
+			integration.organization.slug
+				.toLowerCase()
+				.includes(filterQuery.toLowerCase())
```
kendallstrautman, line 18: 💭 ⛏️  I see this was copied over from what already was there in the file below. But just thinking outloud, there could be opportunity to DRY this up a bit to make it easier to read. Maybe we could start by extracting the lowercased filter query into a constant. Or making a function that runs the `includes` logic. Understandable if you just want to refactor into the function for now.

src/components/command-bar/commands/search/components/dialog-body/index.tsx
```diff
@@ -45,28 +64,77 @@ const SearchCommandBarDialogBodyContent = ({
...
+			<div className={s.tabsWrapper}>
+				<Tabs
+					showAnchorLine={false}
+					initialActiveIndex={activeTabIndex}
+					variant="compact"
+				>
+					{searchableContentTypes.map((contentType: SearchableContentType) => {
+						if (
+							contentType === 'integrations' &&
+							!SHOULD_RENDER_INTEGRATIONS_TAB
```
kendallstrautman, line 119: I think there's an issue here, where we need to filter out the integrations tab when the `currentProductTag` isn't one of the valid slugs. See the photo where consul results show the integrations tab, it would make more sense to me for this tab to only render when a product has integrations available. 

If we want to adjust, maybe we could handle the product filtering at the `tabsBySearchableContentType` object generation level and not as a part of the render function 💭 

Screen Shot 2023-02-01 at 3 10 56 PM

kendallstrautman: I guess I can see the argument for having it for consistency's sake though, so if this was already discussed / decided I get it!

kendallstrautman:
❓ Does it feel weird to anyone that we're linking search results to external sources? To me, search results should link to content within the site. I guess vault is the outlier here since waypoint has landing pages, but it's just a bit funky UX, IMO.

src/components/command-bar/commands/search/components/dialog-body/index.tsx
zchsh: 💭 Very related - when you tab over to that Integrations tab under Consul, for example, there's a `See all Consul integrations` link which when followed results in an error (and at best i think would 404).

So +1 to maybe hiding the tab, or if the tab is to be shown consistently, maybe show "coming soon" for products within integrations activated or something? (Or at least just not rendering the `See all` link).

src/components/command-bar/commands/search/components/integrations-tab-contents/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+	}, [currentInputValue, integrations])
+
+	return (
+		<>
+			<CustomHitsContainer
+				integrationsHits={filteredIntegrations}
+				type="integrations"
+				noResultsSlot={<NoResultsMessage />}
+			/>
+			{currentProductTag ? (
```
zchsh, line 35: 🚧 Potentially blocking, I'm not sure: I think we should avoid rendering this link for products that don't have integrations enabled. For example, for Consul the link ends up leading to a broken page:

https://user-images.githubusercontent.com/4624598/216212653-47c907b5-38d0-4d0d-b4ac-856cc063639a.mp4

src/components/command-bar/commands/search/components/tab-contents-cta/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+interface TabContentsCtaProps {
+	href: string
+	icon: JSX.IntrinsicElements['svg']
+	text: string
+}
+
+const TabContentsCta = ({ href, icon, text }: TabContentsCtaProps) => {
+	return (
+		<Card className={s.cta} elevation="base">
```
zchsh, line 15: 💭 This is maybe more a design slash UX question, but have we considered a `CardLink` here, for consistency in the focus styling and pointer target area of the link?

Because this card is styled in a way where it takes up the same kind of space as a result, personally I find it a little unexpected that the results are these nice large target areas, but then this CTA is only clickable on the `StandaloneLink` part:

https://user-images.githubusercontent.com/4624598/216212841-82c3fcdc-1ffa-4a00-8cb3-8ca61bc49c68.mp4

zchsh (commented):
👍 I ran through the testing steps, everything looks good to me on that front! 🙌 

🚧 I found one potentially blocking thing, which is the `See more` links in the Integrations tab for products that don't have integrations enabled. Maybe this relates to the note on in-flux config in the PR description, but I think we should avoid rendering `See more` links unless integrations are enabled, as otherwise the link will 404 at best (and seems to be error'ing as-is, though I'm not sure exactly why!)

💭 Had a somewhat related thought on `See more` link, just on making the activation area more consistent with search hit items, since they're beside each other and IMO _feel_ like they should have similar activation areas (might be off here though, very optional to address, and probably more a design UX question).

💭 +1 on the amount of "external" stuff for Vault feeling a little odd sometimes. Personally it doesn't seem like a deal-breaker... but I am kind of assuming it's a relatively temporary state. If that's not the case, then I think we should dig deeper into more clearly signaling the difference between external links (which are generally GitHub I guess? so not exactly as polished) vs internal links.

The external link icon does a lil of this, but to me it still seems like "external" and internal hits will lead to the same kind of content... maybe we could have styles for "external" items that make them look like smaller or reduced or something, or maybe even render them with the GitHub icon, so it's clear that engaging with the content will be more involved 🤔

config/base.json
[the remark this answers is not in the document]
ashleemboyer: It's already under `dev_dot`! 😊

src/views/product-integrations-landing/components/searchable-integrations-list/helpers/get-filtered-integrations.ts
ashleemboyer: Since this abstraction affects prod and there's already a lot of changes in this PR, I think it'd be good to refactor this in a separate PR where the scope can be focused and tested more easily!

src/components/command-bar/commands/search/components/dialog-body/index.tsx
ashleemboyer: This was just oversight! We can easily hide the tab if there are no integrations for the `currentProductTag`, and come back to the UX part later. 😊

src/components/command-bar/commands/search/components/tab-contents-cta/index.tsx
ashleemboyer: Can you create a ticket for this to be discussed? This is functionality that existed before this PR, so I think it's out of scope to change for this one.

src/components/command-bar/commands/search/components/integrations-tab-contents/index.tsx
ashleemboyer: I think we should just not render the Integrations tab in these cases! It'll always say "No results match your search", so it feels misleading to make the user do extra work for nothing. I think we can revisit in a later PR if we find alignment on the UX.

braydenlove: I agree, let's not show an integrations tab for products that don't have integrations.

src/components/command-bar/commands/search/components/tab-contents-cta/index.tsx
braydenlove: Yeah I agree with you @zchsh and think we could improve this. Let's just not improve it now so @ashleemboyer can keep rolling. Thanks for the feedback.

braydenlove:
Curious if you all think we should show an integrations tab at all when no product is selected? Reasoning: We can find integration links via the documentation tab but aren't showing anything in the integrations tab when searching by integrations (keyword). This might be a funky edge-case as we only have 2 products in this PR, happy to go either direction just feels a bit odd.

image

ashleemboyer:
@kendallstrautman @zchsh @braydenlove I've just pushed up commits that change when the Integrations tab shows. It now shows all the time except when there is a product tag for a product that does not have integrations. 😊

ashleemboyer:
> Curious if you all think we should show an integrations tab at all when no product is selected? Reasoning: We can find integration links via the documentation tab but aren't showing anything in the integrations tab when searching by integrations (keyword). This might be a funky edge-case as we only have 2 products in this PR, happy to go either direction just feels a bit odd.

@braydenlove I think we should show it when no product is selected. The "integrations" keyword might show results in the Integrations tab later after getting an Algolia index hooked up, but for now the search logic strictly searches against the `name`, `description`, and `organization` of each integration. So if that word isn't in one of those properties, we won't see results for it under the tab.

src/components/command-bar/commands/search/components/tab-contents-cta/index.tsx
zchsh: Sweet, sounds good! Here's a ticket in case it's helpful: 🎟 [[discuss] integrations search results "Show more" link activation target area](https://app.asana.com/0/1202097197789424/1203885784918902/f)

zchsh (approved):
✅ 🙌 Ran through the testing steps again, looks good to me!

braydenlove (approved):
@ashleemboyer thanks for the additional context/feedback. @zchsh thanks for writing up that follow up ticket. I think we should refine the empty state to be more consistent as well as look at the external link complexity. For now though this looks great and I approve, appreciate the hard work!

config/base.json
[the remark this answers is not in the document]
kendallstrautman: Oh I must have misread it! Sorry 😅 all good then

kendallstrautman (approved):
💯 👍🏻 LGTM! Thanks for making those updates. I still think the external linking is odd UX, but not blocking!

BrandonRomano:
Merging in some interface changes to the API, I want to test this out again before merging.

BrandonRomano (approved):

PR #1614, opened 2023-02-03
Remove plugin navigation elements

BrandonRomano:
---
  - https://app.asana.com/0/0/1203879536690957


src/data/waypoint.json
```diff
@@ -59,7 +59,7 @@
 			"type": "inbound"
 		}
 	],
-	"basePaths": ["commands", "docs", "plugins", "tutorials", "downloads"],
+	"basePaths": ["commands", "docs", "integrations", "tutorials", "downloads"],
```
zchsh, line 62: I think `basePaths` is [being deprecated](https://github.com/hashicorp/dev-portal/blob/main/src/data/README.md#devdot-properties), and is also used for old link rewrites, so I think we might want to maintain `"plugins"` in this case 🤔

BrandonRomano: Ah!  Okay, yep I'll return this back to normal.

FWIW, I didn't observe any changes when I switched this so we're definitely safe to go back.

BrandonRomano, line 62:
```suggestion
	"basePaths": ["commands", "docs", "plugins", "tutorials", "downloads"],
```

zchsh:
I think we might need to remove the Waypoint `src/pages/waypoint/plugins` page file to avoid the build error here 🤔 I think that page file is attempting to render, but can't find the matching `rootDocsPath` entry.

src/data/waypoint.json
zchsh: yeah tbh i'm not sure around impact! I think it `basePaths` would mostly apply to automatically rewriting any links to `waypointproject.io/plugins` within content... but we've since rewritten those types of links at source thanks to Ashlee & Kendall's efforts... and with #1600 & #1602 we rewrite plugins links in the newer format so this might be a non-issue either 🤷

BrandonRomano:
@zchsh If I remove plugins page does that mean I should remove it as a `basePaths` entry as well?

zchsh:
@BrandonRomano I'd say leave it in `basePaths` to be safe, I think? I think we might need it for docs links rewrites (in `getIsRewriteableDocsLink`). Although it looks like we do run some playwright tests based on it as well 🤔 Those might fail if we keep it in place. I'm honestly not 100% given the latter! Maybe we'd skip over `waypoint/plugins` in `docs-content-link-rewrites`?

BrandonRomano:
I'm OK not 100% figuring that out today; I've remove the pages which should fix the build & I should be able to get a preview up to test out the base purpose of this PR, at least.

zchsh (approved):
✅ Looks good, pending successful build 👍

BrandonRomano:
Oh whomp I'm going to undo that test change; committed that on the wrong branch 😢

PR #1615, opened 2023-02-06
[preAdjustUrl] Handle with currentPath has a trailing slash

ashleemboyer:
## 🗒️ What

- Updates `preAdjustUrl` to ignore a trailing slash on `currentPath` if it's present
- Adds test cases for the new logic

## 🤷 Why

- When `currentPath` has a trailing slash, a duplicate path part is placed in the final URL.
  - For example, when `currentPath` is `'/plugin/sdkv2/logging/'` and `url` is `'./logging/http-transport'`, the result was `'/plugin/sdkv2/logging/logging/http-transport'`.
- This bug was found while debugging this exact link in the link rewrites for [`hashicorp/terraform-plugin-sdk`](https://github.com/hashicorp/terraform-plugin-sdk/pull/1137/files#r1097404820)


kendallstrautman (approved):
👍🏻

PR #1623, opened 2023-02-07
fix: normalize product slug for tutorial path 

kendallstrautman:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Fixes an issue where 'cloud' product slugs were being used in the tutorial page generation. Since we aren't statically generating tutorial paths with this function right now, we didn't see the issue.


src/views/tutorial-view/server.ts
```diff
@@ -180,6 +180,9 @@ export async function getTutorialPagePaths(): Promise<TutorialPagePaths[]> {
 		 * https://app.asana.com/0/1201903760348480/1201932088801131/f
 		 */
 		const isCloud = productSlugFromCollection == 'cloud'
+		const normalizedProductSlug = normalizeSlugForDevDot(
```
kendallstrautman, line 183: [Function reference ](https://github.com/hashicorp/dev-portal/blob/1e2fc9273d116ec5eb9ecae9b495486c788d784e/src/lib/tutorials/normalize-product-like-slug.ts#L18)

BRKalow (approved):
lgtm

PR #1625, opened 2023-02-07
chore: support node 18, set v18 in .nvmrc

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Updates the required node version to `>=16 <=18`, which is the [current active LTS](https://github.com/nodejs/release#release-schedule). Also updates our `.nvmrc` file to specify v18.

Our [`Managing Node Versions` team doc](https://www.notion.so/hashicorp/Managing-Node-versions-30d9ac89d35043f8b5c70bbc7aef1c44) has been updated as well.

## 💭 Anything else?

---
  - https://app.asana.com/0/0/1203912955325817


src/lib/fetch-file-string.js
```diff
@@ -1,16 +1,6 @@
...
-
-async function bufferFromStream(stream) {
-	return new Promise((resolve, reject) => {
-		const _buf = []
-		stream.on('data', (chunk) => _buf.push(chunk))
-		stream.on('end', () => resolve(Buffer.concat(_buf)))
-		stream.on('error', (err) => reject(err))
-	})
+	return res.text()
 }
```
BRKalow, line 4: I was getting an error here: `stream.on is not a function`. Shifted to use the `.text()` method on the fetch response instead, as I believe it accomplishes the same result.

package.json
```diff
@@ -182,7 +182,7 @@
 		"mdast-util-to-string": "^2.0.0",
 		"moize": "^6.1.0",
 		"next": "^13.0.5",
-		"next-auth": "^4.17.0",
+		"next-auth": "^4.19.2",
```
BRKalow, line 185: This fixed an issue with a transitive dependency that didn't allow node 18 in its engines.

kendallstrautman (approved):

kendallstrautman:
TY 🙌🏻

PR #1632, opened 2023-02-08
[🚧 WIP] Add loading state to ProductIntegrationsLanding

ashleemboyer:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1203792616809419/1203868487235738/f) 🎟️
- [Preview URL](https://dev-portal-git-ambintegrations-loading-state-hashicorp.vercel.app/vault/integrations?tiers=official%2Cpartner%2Ccommunity&components=auth-method%2Cdatabase) 🔎

### DIFF
--- a/src/views/product-integrations-landing/index.tsx
+++ b/src/views/product-integrations-landing/index.tsx
@@ -3,16 +3,17 @@
  * SPDX-License-Identifier: MPL-2.0
  */
 
-import s from './style.module.css'
-import SearchableIntegrationsList from './components/searchable-integrations-list'
-import { IntegrationsSearchProvider } from './contexts/integrations-search-context'
-import { type Integration } from 'lib/integrations-api-client/integration'
-
-import SidebarSidecarLayout from 'layouts/sidebar-sidecar'
-import { type SidebarProps } from 'components/sidebar'
-import { type BreadcrumbLink } from 'components/breadcrumb-bar'
+import { useEffect, useState } from 'react'
 import { ProductData } from 'types/products'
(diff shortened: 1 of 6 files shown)


ashleemboyer:
Going to see what it might look like to use loading skeletons, so here's a screen recording of the current state of this work.

https://user-images.githubusercontent.com/43934258/218858647-8c5c2193-6b0a-4f21-ba50-448e36f8c8a8.mov

src/views/product-integrations-landing/index.tsx
```diff
@@ -27,6 +28,18 @@ export default function ProductIntegrationsLanding({
...
+	/**
+	 * Show a loading spinner until 1500ms after the `useEffect` runs to prevent a
+	 * hydration error.
+	 */
+	const [isLoadingClient, setIsLoadingClient] = useState(true)
+	useEffect(() => {
+		setTimeout(() => {
+			// TODO uncommonent before merging
+			// setIsLoadingClient(false)
+		}, 1500)
```
kendallstrautman, line 40: Is there anyway we can link this to the actual api call to know when we have the data to render? I assume the dynamic call is what's causing the flash?

ashleemboyer: @kendallstrautman the API call is happening server-side and the data is being passed into the view component. The flash and hydration mismatch errors are happening when there are query parameters in the page URL. The query parameters affect what HTML is rendered on the page, and since those aren't accessible until the client starts loading, there's a brief flash from the default/unfiltered view of this page to the filtered view.

PR #1635, opened 2023-02-11
chore: remove old docsCta from learn landing schema

zchsh:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-zswarn-tutorial-docs-cta-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1202225913259475/f) 🎟️

## 🗒️ What

Removes the `docsCta` part of the Joi schema for tutorials landing pages.

## 🤷 Why

The `docsCta` content property was formerly supported, but no longer does anything. So we want to ensure authors will be warned if they use it.

## 💭 Anything else?

Note that at present, none of our product page content uses the `docsCta` property. So we'll only expect to see warnings if authors add back the `docsCta` property for some reason. With this in mind, this PR is really just cleaning up some stale lines of schema code we don't need anymore.


ashleemboyer (approved):

kendallstrautman (approved):

src/lib/learn-client/schemas.ts
```diff
@@ -120,15 +120,5 @@ export const ProductPageSchema = Joi.object({
 			)
 			.required(),
 		showProductSitemap: Joi.boolean(),
-		docsCta: Joi.object({
```
kendallstrautman, line 123: 🤔 We should check whether this is still accepted in the learn api content sync repo. There could be an issue if an author _tries_ to use this, the sync succeeds and then we get an error in building the app.

kendallstrautman: Perhaps its just an additional cleanup task for removing it in the content sync 🧼

zchsh: Ah great call! 🙌  I've opened https://github.com/hashicorp/learn-api-content-sync/pull/68 to clean this up in the `learn-api-content-sync` repo.

PR #1655, opened 2023-02-15
Experiment preselected tabs

im2nguyen:
## 🔗 Relevant links

- [Asana task](url) 🎟️


ashleemboyer:
@im2nguyen did you mean to set the target branch to `amb.spike-preselected-tabs` instead of `main`?

im2nguyen:
This addresses some of the issues on the spike doc -- if you use `router.replace` (or even `router.push`) it will re-render the page. `shallow` would will re-render it, it just won't refetch the data

Using the native Javascript windows.history addresses this. The main issue right now is that tabs with huge variances in length will cause the page to jump. even though this is not ideal behavior, this is the existing behavior ([example](https://developer.hashicorp.com/vault/tutorials/secrets-management/pki-engine#step-3-create-a-role)). preselected tabs does NOT introduce this behavior

as a result, i think we should consider shipping this as a "phase 1" approach

ashleemboyer:
Just converting to draft as an extra protection to the "do not merge" label, so this isn't accidentally merged while we're discussing things! 💖

PR #1662, opened 2023-02-17
feat: add OutlineNav component

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Implements `OutlineNav`, working off the proposed approach in [Digital RFC: Sidecar Outline Architecture Proposal][digital-rfc].

> **Note**: this PR focuses solely on building the `OutlineNav`component. It does not try to handle the concern of modifying `SidebarSidecarLayout` or of using the new `OutlineNav` component. For that work, see #1661.

## 🛠️ How

- Builds a new `OutlineNav` component in `components/outline-nav`
- Adds a [Swingset page][swingset] for `OutlineNav`

## 📸 Design Screenshots

🎨 [Link to Figma][figma]

### Design

2023-02-21-outline-nav-design

### Implementation ([Swingset][swingset])

2023-02-21-outlinenav-swingset

## 🧪 Testing

- [ ] Visit the [OutlineNav swingset page][swingset]
   - The `OutlineNav` component should render a `nav` element with correct semantics, similar to our existing `TableOfContents` component
   - The `OutlineNav` should match design specs, including hover, active, and focus states. Final design is in the [`sidecar-v3` artboard][figma].
   - The interface of `OutlineNav` should be purely presentation, without built-in opinions on other functionality (as discussed in [Digital RFC: Sidecar Outline Architecture Proposal][digital-rfc]). To recreate the opinionated approach on highlighting active sections and handling the `headings` data we need to swap in the component, additional `OutlineNavWithActive` and `OutlineNavFromHeadings` components have been constructed.

> **Note**: for further testing, see #1661, which builds off this PR and swaps in `OutlineNav` to all existing uses of `TableOfContents`. Note that PR also includes modifications to `SidebarSidecarLayout` necessary to make the swap, which were significant enough that they seemed to be worth splitting out into a separate PR.

[figma]: https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=18661%3A236003&t=80Ghe75yweoqNSUA-1
[task]: https://app.asana.com/0/1203810296555970/1203964362474949/f
[digital-rfc]: https://docs.google.com/document/d/1JF4T-oZiZTzv2qt4WitlaWgDxwQGNsKyqLZSUT46ios/edit
[preview]: https://dev-portal-git-zsadd-outline-nav-hashicorp.vercel.app
[swingset]: https://dev-portal-git-zsadd-outline-nav-hashicorp.vercel.app/swingset/components/outlinenav


src/components/outline-nav/components/outline-link/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Renders an individual outline nav link.
+ */
+function OutlineLink({ title, url, isActive }: OutlineLinkProps) {
+	return (
+		<a
+			className={classNames(s.root, { [s.isActive]: isActive })}
+			href={url}
+			/* TODO: confirm specifics for data-heap-track here, if we want it */
```
zchsh, line 13: @BrandonRomano Do you know if we need any specific `data-heap-track` attributes for `OutlineNav` links? Here I've kept something analogous to `table-of-contents`. Let me know what you think, wanted to confirm this as I wasn't 100% sure 🙇

src/components/outline-nav/components/outline-link/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+import s from './outline-link.module.css'
+
+/**
+ * Renders an individual outline nav link.
+ */
+function OutlineLink({ title, url, isActive }: OutlineLinkProps) {
+	return (
+		<a
+			className={classNames(s.root, { [s.isActive]: isActive })}
+			href={url}
```
zchsh, line 12: 💭 Thought: we could update the props interface here to accept `slug` rather than `url`, and then do `href={`#${slug}`}. This would add some opinion to the component (ie it'd no longer be able to render arbitrary URLs), but could simplify other parts of the code here (eg `get-item-slugs` would be much simpler, would work more smoothly with `isActiveSection`).

I'm a little on the fence about which direction to go here, I initially leaned towards baking in as little opinion as possible, so that's what I've implemented so far. Open to thoughts 🙇 cc @BrandonRomano

src/components/outline-nav/components/outline-link/index.tsx
BrandonRomano: This is a good call-out, but I am not sure of the answer.

I would bubble this question up to the team to see if anyone has any idea.

src/components/outline-nav/components/outline-link/index.tsx
BrandonRomano: It's taking me a while to type this out because I myself am conflicted.

On one hand, I think that the lack of opinion is the right move here.  Right now we have plans to use the OutlineNav _only_ in the sidecar, but I could imagine at some point in the future this component could be leveraged and may have a different set of business rules. We're really only going to interface with the OutlineNav from a few higher level implementations, so the interface doesn't need to really be optimized (and we can provide the _slug_ as the interface at one of those higher order implementations and do the mapping).

On the other hand, I think that it's probably unlikely that we ever use this OutlineNav anywhere else than where we plan, so I could see the argument for the other way and implementing the business logic of in-page-links only here.

I think in the spirit of this being a view-model & agnostic to business logic, we probably go the way you have implemented.

zchsh: Sweet - yeah I was kind of thinking the same thing, maybe the issue here is more that the business logic and higher level components could use a little refinement. I'll leave as-is 👍

src/components/outline-nav/components/outline-link/index.tsx
kendallstrautman: cc @anthonyvetter ☝🏻

src/components/outline-nav/components/outline-link/index.tsx
kendallstrautman: Agreed, I'd lean towards keeping this component unaware of the rendering context and leaving it as `url`.

src/components/outline-nav/components/outline-link/index.tsx
kendallstrautman: What was the attribute value for table of contents? We may want to keep it consistent across all sidecars. Perhaps we could pass in the attribute as a prop 🤔

src/components/outline-nav/components/outline-nav-from-headings/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { useMemo } from 'react'
+
+/**
+ * Given a flat array of headings,
+ * Return an array of outline items, for use with OutlineNav components.
+ */
+function buildOutlineItems(headings: OutlineHeading[]): OutlineLinkItem[] {
+	return headings.map((heading: OutlineHeading) => {
+		/* TODO: maybe consider resolving the backticks issue closer to server? */
+		const titleWithoutBackticks = heading.title.replace(/`/g, '')
```
kendallstrautman, line 12: 👍🏻  Nodding to this TODO comment. Ideally this could happen in the heading generation function? Perhaps this could be a note to pick up once you actually integrate it into the sidebar in the next PR.

src/components/outline-nav/components/outline-nav-from-headings/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+		return {
+			title: titleWithoutBackticks,
+			url: `#${heading.slug}`,
```
kendallstrautman, line 15: Perhaps we could do all this outline formatting from headings in the server 💭 Then this component could truly be presentational.

src/components/outline-nav/components/outline-nav-with-active/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+	items,
+}: {
+	items: Omit<OutlineLinkItem, 'isActive'>[]
+}) {
+	/**
+	 * We useDeviceSize to avoid running active heading calculations
+	 * if we're on a viewport where we won't render the nav anyways.
+	 *
+	 * TODO: isDesktop isn't fully accurate, we may want to adjust this
+	 * based on the specific breakpoint at which we show the sidecar.
```
kendallstrautman, line 25: 💡 I remember some logic to work around this with the mobile menu that could be worth referencing.

src/components/outline-nav/components/outline-nav-with-active/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+	 *
+	 * Thought: could we monitor the visibility of `OutlineNav` instead,
+	 * ie the root element we're rendering in this component?
+	 * If it's visible in CSS, then we should enable useActiveSection.
+	 * If it's not visible, then we should NOT enable useActiveSection.
+	 * Maybe a `useVisibility` hook or something would be useful here.
+	 * Maybe this could reduce reliance on specific breakpoint metrics.
```
kendallstrautman, line 32: This could be cool if we want to keep this component aware of the layout level scaling. Personally I think it would be ideal for the layout to decide whether this is `enabled` based on whatever logic applies within that context and pass that down? Then the outline isn't concerned with layout level breakpoints. Just an idea, and maybe doesn't make as much sense in practice when we actually use this within the layout slot. 

 Also, if this is to prevent extra calculations for the active state, depending on how we _hide_ the sidecar at smaller breakpoints, would that (assuming its `display:none`) take care of this issue?

src/components/outline-nav/components/outline-link-with-nesting/index.tsx
```diff
@@ -0,0 +1,25 @@
+import { OutlineListItems, OutlineLink } from '..'
+import type { OutlineLinkWithNestingProps } from '../../types'
+import s from './outline-link-with-nesting.module.css'
+
+/**
+ * Render an outline nav link,
+ * with additional outline links indented in a nested section.
+ */
+function OutlineLinkWithNesting({
```
kendallstrautman, line 9: ❓  Have we considered limiting the levels of nesting allowed?

src/components/outline-nav/index.tsx
```diff
@@ -0,0 +1,26 @@
+import { OutlineListItems } from './components'
+import type { OutlineNavProps } from './types'
+import s from './outline-nav.module.css'
+import { useId } from '@react-aria/utils'
```
kendallstrautman, line 4: TIL about this hook. 👏🏻 Also could we move this import to the top ⛏️

src/components/outline-nav/props.js
```diff
@@ -0,0 +1,33 @@
+module.exports = {
```
kendallstrautman, line 1: Oh man swingset 😭😎 Can we add a breadcrumb on the type file to keep these two in sync? 🍞

kendallstrautman (approved):
Overall looks great! Just have a few non-blocking comments / questions but they can be addressed in follow-up work as well 🚀

src/components/outline-nav/components/outline-link/index.tsx
zchsh: In `TableOfContents`, [it was `data-heap-track={"toc-list-item-index-" + index}`](https://github.com/hashicorp/dev-portal/blob/dd70254ff2ff1cb97080e6b0b78794f641db1b92/src/components/table-of-contents/index.tsx#L89). I wasn't sure about the `index` part there, and it'll perhaps be a bit different now with nesting. I'm not familiar with how these analytics are used, so would be curious as well to get input.

Re: passing in as a prop, I feel like a consistent prop value might be useful in some way (eg tracking interactions with this component at a global level might give some insight?), and feels like generally the prop value + URL will let us dive into more narrow contexts if we need to. Again though would defer to analytics folks - if there are specific contexts where we want to render different analytics values, then maybe we take a more props-based approach 👍

src/components/outline-nav/components/outline-link-with-nesting/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+	url,
+	isActive,
+	items,
+}: OutlineLinkWithNestingProps) {
+	return (
+		<>
+			<OutlineLink title={title} url={url} isActive={isActive} />
+			<div className={s.nested}>
+				<OutlineListItems items={items} />
+			</div>
```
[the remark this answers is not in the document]
ashleemboyer: I think we may also want to add an `id` to the `OutlineLink` rendered here, and programmatically associate it with the `<ol>` element using the `aria-labelledby` property. I'm less sure about this part. I think it'd be good to ask about in the #talk-a11y-guild channel!

src/components/outline-nav/components/outline-link/index.tsx
ashleemboyer: 💭 I'd also suggest keep this component as-is!

src/components/outline-nav/components/outline-list-items/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+ * Each item can be either a plain link item, or
+ * a link item with nested links.
+ *
+ * All items are wrapped in an <li /> element.
+ */
+function OutlineListItems({ items }: { items: OutlineLinkItem[] }) {
+	return (
+		<>
+			{items.map((item: OutlineLinkItem) => {
+				const uniqueKey = `${item.title}${item.url}`
```
ashleemboyer, line 16: 🙋🏻 Do we need both `title` and `url` in the key or is `url` enough? There shouldn't be any heading elements on a page with duplicate `id`s. If we do have cases of this, it'd be good to have an Asana ticket for fixing those, and a comment here as a breadcrumb!

src/components/outline-nav/props.js
ashleemboyer: I think I'd rather see us link to the `types.ts` file from the Swingset page rather than have multiple things to maintain with the component.

src/components/outline-nav/components/outline-link/index.tsx
zchsh: @kendallstrautman Thanks for the suggestion re: attribute-as-prop, took a second for me to catch up but it seems to work way better 💯 I've updated as follows:
- The `OutlineLinkItem` interface now accepts an optional `dataHeapTrack` string attribute, which is passed to `data-heap-track`. (Note this also applies to `OutlineLinkWithNesting`). `OutlineNav` has no opinion on what `dataHeapTrack` should be.
- In `OutlineNavFromHeadings`, which is what we'll be swapping for existing `TableOfContents` use cases, we tack on `dataHeapTrack` properties when we map the flat array `headings` to outline items. This achieves parity with the previous `data-heap-track` attributes in `TableOfContents.

**Note**: this PR does _not_ yet address what `data-heap-track` attributes will be used for nested cases (it is impossible to create nesting with `OutlineNavFromHeadings`). Heap tracking on nested items will be handled in a future PR where we add nesting (#1667).

src/components/outline-nav/components/outline-nav-from-headings/index.tsx
zchsh: Great call - I think this'll be a relatively high-effort refactor, since it'll involve moving the abstraction from client-side to server-side, which will involve changing a few more places. I've added a follow-up task 🎟 [Clean up headings-related OutlineNav uses, to move formatting closer to server](https://app.asana.com/0/0/1204024966433303/f) to do this at some point (hopefully soon!)

src/components/outline-nav/components/outline-nav-from-headings/index.tsx
zchsh: 👍 Added a follow-up task 🎟 [Clean up headings-related OutlineNav uses, to move formatting closer to server](https://app.asana.com/0/0/1204024966433303/f)

src/components/outline-nav/utils/highlight-active-items.ts
```diff
@@ -0,0 +1,25 @@
+import { OutlineLinkItem } from '../types'
+
+/**
+ * Given an array of outline items, each of which has a `url` property,
+ * and a string representing #hash portion of the current URL,
+ *
+ * Return the array of items with `isActive` property set to `true` on any
+ * items that have a matching URL.
+ */
+function highlightActiveItems(
```
ashleemboyer, line 10: 🙋🏻 Having trouble picturing it at the moment, when are there multiple active items?

src/components/outline-nav/components/outline-link-with-nesting/index.tsx
zchsh: In a way, yes! We considered writing code-level checks, such as throwing an error where nesting is too deep. However, this felt a little strange to do at this point, since the only way nesting could become too deep is if we go in and change code to generate deeply nested items and then pass them to `OutlineNav`.

Personally my hot take is: code-level protections seem most useful when dealing with authored content, where the limitations might not be clear, and won't have to go through code review to be bypassed. If we expose `OutlineNav` in a way where authors can somehow generate the levels of nesting, then perhaps code-level protection would be useful... however even then, I think we'd probably lean more in the "content conformance" direction, at the "extract" part of ETL rather than during rendering 🤔

ashleemboyer (changes requested):
Oops, I meant to mark request changes instead of submitting as a comment! ❤️

ashleemboyer:
> Oops, I meant to mark request changes instead of submitting as a comment! ❤️

Oh also! Forgot to mention a visual change I'd like to see on the `<a>` elements: `width: fit-content`. Otherwise, they're very wide and the hover/focus interactions are mismatched from what's expected.

image

image

src/components/outline-nav/props.js
zchsh: ah yep... added a comment! and linked to 🎟 [Support extracting props from TS](https://app.asana.com/0/1199715139452823/1199702038395513/f).

src/components/outline-nav/utils/highlight-active-items.ts
zchsh: We don't expect there to be multiple active items. I suppose if we wanted to we could add some logic here to try to exit early once we've found the first active item, but personally I felt this might not be worth the tradeoff in additional logic & complexity.

src/components/outline-nav/utils/get-item-slugs.ts
```diff
@@ -0,0 +1,21 @@
...
+		const slug = urlObject.hash.replace('#', '') // get hash without leading #
+		if (slug !== '') {
+			acc.push(slug)
+		}
+		if ('items' in item) {
+			acc = acc.concat(getItemSlugs(item.items))
+		}
+		return acc
+	}, [])
+}
```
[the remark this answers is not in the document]
zchsh: We are recursing here already - but agree, `reduce` can be a bit hard to read sometimes. I can switch this to a `for` loop, and split out the `#hash` extraction part for clarity.

The reason we don't reference raw data here (slash why the `#hash` logic is necessary) is that we want `OutlineNavWithActive` to accept items with `{ title, url }` rather than `{ title, slug }`, as mentioned in the other thread on `slug` vs `url` in `OutlineLink`. We also have a follow-up task 🎟 [Clean up headings-related OutlineNav uses, to move formatting closer to server](https://app.asana.com/0/0/1204024966433303/f) to move more of the logic closer to the server, however it'll be higher effort so I'm thinking of it as a future optimization. All new uses of `OutlineNav` (eg in Integrations) will ideally generate outline items directly (kind of using `getStaticProps` as the business logic layer, rather than a mix of `getStaticProps` and logic within components).

src/components/outline-nav/index.tsx
zchsh: ✅ moved to the top!

src/components/outline-nav/props.js
zchsh: ah oops Ashlee i didn't see your comment - we could link to `types.ts`, but then we wouldn't get the props table on the Swingset page. Happy to do either, if we want to avoid using `props.js` and the props table in Swingset, i can go that route 👍

src/components/outline-nav/components/outline-link-with-nesting/index.tsx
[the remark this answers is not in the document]
zchsh: Ah, thanks for catching this! Updated to use an `<ol />` in place of the `<div />`.

I'll ask about the `aria-labelledby` property for nested lists. Is sorting that out something that feels blocking for this PR, or something we could address later?

src/components/outline-nav/components/outline-link/index.tsx
```diff
@@ -0,0 +1,21 @@
+import classNames from 'classnames'
+import type { OutlineLinkProps } from '../../types'
+import s from './outline-link.module.css'
+
+/**
+ * Renders an individual outline nav link.
+ */
+function OutlineLink({ title, url, isActive }: OutlineLinkProps) {
+	return (
+		<a
```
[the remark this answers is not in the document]
zchsh: ✅ great call - updated!

src/components/outline-nav/components/outline-list-items/index.tsx
zchsh: Ah true, `url` should be enough! I'll update accordingly 👍

src/components/outline-nav/utils/get-item-slugs.ts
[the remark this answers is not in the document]
zchsh: ✅ updated to use a `for` loop, and split out `getHash(url: string)` 👍 Let me know if that works!

zchsh:
@ashleemboyer Thanks for the feedback! 💖 I've done my best to address everything - let me know if things look alright now 🙇

src/components/outline-nav/components/outline-nav-with-active/index.tsx
zchsh: Yeah good call - I've added a follow-up task 🎟 [Rethink SidebarSidecarLayout responsive implementation to avoid useActiveSection calculations, at layout level](https://app.asana.com/0/0/1204024966433307/f) to address this. It seems like more of an optimization, so thinking I'll dive into it only if I have time, but open to other approaches 🙇

Re: hiding with `display: none`, I think the hard part is we can't call hooks conditionally. So even if the CSS says hide the sidecar, React will still be rendering it (with that CSS) and still make the `useActiveSection` calls. So +1 to having the layout decide when `useActiveSection` should be running or not 👍

src/components/outline-nav/components/outline-link-with-nesting/index.tsx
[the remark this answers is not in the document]
zchsh: Update: Melanie responded in Slack, I've got with `<ol aria-label={title} />` which looks like it yields the same accessible name for the list in browser inspect. Let me know if that work! ☺️

ashleemboyer:
> @ashleemboyer Thanks for the feedback! 💖 I've done my best to address everything - let me know if things look alright now 🙇

@zchsh taking another look now! 😊

src/components/outline-nav/props.js
ashleemboyer: I don't feel strongly that we need to have the props table on the Swingset page. I don't think we've done it much yet?

src/components/outline-nav/components/outline-link-with-nesting/index.tsx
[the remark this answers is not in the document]
ashleemboyer: Both of these things look good! 👏🏻

src/components/outline-nav/components/outline-link/index.tsx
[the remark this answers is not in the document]
ashleemboyer: Looks good! 👍🏻

src/components/outline-nav/components/outline-list-items/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+						</li>
+					)
+				} else {
+					return (
+						<li key={item.url}>
+							<OutlineLink {...item} />
+						</li>
+					)
+				}
+			})}
```
ashleemboyer, line 29: 🎨 Totally optional! Might be a place to shorten the code a little if you want.

```tsx
			{items.map((item: OutlineLinkItem) => {
				return (
					<li key={item.url}>
						{'items' in item ? (
							<OutlineLinkWithNesting {...item} />
						) : (
							<OutlineLink {...item} />
						)}
					</li>
				)
			})}
```

ashleemboyer (approved):

src/components/outline-nav/components/outline-list-items/index.tsx
zchsh: Oooo yeah great call! Added this in, ty for the suggestion 💖

PR #1667, opened 2023-02-21
feat: integrations nested outline for variable groups

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

This PR implements nested outline links for variable groups on Integrations component pages. Common variable groups include Parameters and Outputs.

## 🤷 Why

To match designs, providing a more detailed outline of component pages.

## 🛠️ How

- Updates business logic to generate nested `outlineNavItems` rather than flat `variableGroupHeadings`
- Swaps out `OutlineNavFromHeadings`, swaps in `OutlineNavWithActive`

## 📸 Design Screenshots

🎨 [Figma file][figma]

### Design

CleanShot 2023-02-23 at 16 27 13@2x

### Implementation

CleanShot 2023-02-23 at 16 28 28@2x

## 🧪 Testing

- [ ] Visit a component page, such as [/waypoint/integrations/hashicorp/docker/latest/components/builder][/waypoint/integrations/hashicorp/docker/latest/components/builder]
    - The sidecar should show an `OutlineNav` listing all parameters and outputs
    - Jump links to `Parameters`, `Outputs`, and all nested items should function as expected

## 💭 Anything else?

> **Note**: on integration component pages, we are not yet extracting headings and adding them to the `OutlineNav`. Due to sanitization requirements, this may be a slightly more involved effort, and regardless felt reasonable to treat as a separate task from `Parameters` and `Outputs` nesting. See 🎟 [[discuss] sanitize-friendly heading anchor links for user content](https://app.asana.com/0/1203792616809419/1203896228323108/f) for details.

[task]: https://app.asana.com/0/1203810296555970/1203970622632444/f
[figma]: https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=19816%3A232842&t=Rb5wf1DnjDv9sTsf-4
[preview]: https://dev-portal-git-zsint-component-nested-outline-hashicorp.vercel.app/
[/waypoint/integrations/hashicorp/docker/latest/components/builder]: https://dev-portal-git-zsint-component-nested-outline-hashicorp.vercel.app/waypoint/integrations/hashicorp/docker/latest/components/builder

### DIFF
--- a/src/components/outline-nav/utils/highlight-active-items.ts
+++ b/src/components/outline-nav/utils/highlight-active-items.ts
@@ -12,11 +12,11 @@ function highlightActiveItems(
 	currentUrl: string
 ): OutlineLinkItem[] {
 	return items.map((item) => {
+		const isActive = item.url === currentUrl
 		if ('items' in item) {
 			const activeItems = highlightActiveItems(item.items, currentUrl)
-			return { ...item, items: activeItems }
+			return { ...item, items: activeItems, isActive }
 		} else {
-			const isActive = item.url === currentUrl
 			return { ...item, isActive }
 		}
 	})
--- a/src/views/product-integration/component-view/index.tsx
+++ b/src/views/product-integration/component-view/index.tsx
@@ -53,10 +53,15 @@ export default function ProductIntegrationComponentView({
 		(variableGroup: VariableGroup) => {
 			const groupName = variableGroup.variable_group_config.name
 			const groupSlug = getVariableGroupSlug(groupName)
-			const nestedItems = variableGroup.variables.map((v) => {
-				const variableSlug = getVariableSlug(groupName, v.key)
-				return { title: v.key, url: `#${variableSlug}` }
-			})
+			const nestedItems = variableGroup.variables
+				.filter((v) => {
+					const isTopLevelVariable = v.key.indexOf('.') === -1
+					return isTopLevelVariable
+				})
+				.map((v) => {
+					const variableSlug = getVariableSlug(groupName, v.key)
+					return { title: v.key, url: `#${variableSlug}` }
+				})
 			return { title: groupName, url: `#${groupSlug}`, items: nestedItems }
 		}
 	)
--- a/src/views/product-integration/component-view/index.tsx
+++ b/src/views/product-integration/component-view/index.tsx
@@ -21,7 +21,7 @@ import { MDXRemoteSerializeResult } from 'next-mdx-remote'
 import { ProductData } from 'types/products'
 import SearchableVariableGroupList from './components/searchable-variable-group-list'
 import { Variable } from './components/variable-group-list'
-import { getVariableGroupSlug } from './helpers'
+import { getVariableGroupSlug, getVariableSlug } from './helpers'
 import type { ProcessedVariablesMarkdown } from './helpers/get-processed-variables-markdown'
 import s from './style.module.css'
 import VersionAlertBanner from 'components/version-alert-banner'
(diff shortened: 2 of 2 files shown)


zchsh:
@hashicorp/dev-portal @kellyschummer This PR implements a nested `OutlineNav` on Integrations component pages, to add detailed `#jump-links` for `Parameters` and `Outputs`. Let me know what you think 🙇

kendallstrautman (approved):
👏🏻 💯 The preview looks great

braydenlove:
@zchsh This is looking really good, we are running into the problem where we can't highlight the correct item due to the height of the item as well as the proximity to the bottom of the page. Do you have any ideas on how we could make this a bit better? I know it is difficult but I want to pick your brain! 
image

zchsh:
@braydenlove Great question re: ensuring the last table-of-contents item gets highlighted. I think the TLDR is, if we have a taller footer similar to hashicorp.com, that would solve the issue in a clean and simple way.

There's a little more detail in 🎟 [[table of contents] Address non-highlighting of last items](https://app.asana.com/0/1202097197789424/1202975829609765/f). It's _technically_, _maybe_ possible to come up with a way to try to get items highlighted even if they're near the bottom of the page... however it greatly complicates things, and the logic of where to jump to on the page would have to change, which means we'd no longer be able to use native `#jump-link` functionality. So the taller footer could really help here. Here's a quick demo of how a taller footer solves the issues:

https://user-images.githubusercontent.com/4624598/221681959-e24cc912-0d2a-4563-a0bc-ab22e01759f6.mp4

braydenlove:
@zchsh taller footer totally makes sense, it could also be something like a taller _next steps_ section that shows up on each page (where feedback is living on the tutorials). The issue is that we currently don't have a taller footer and am not sure when we will have one! It might be me but it feels pretty clunky in this specific context, I wonder if we remove the active state on the integrations pages until we get a taller footer? I am not a huge fan of being inconsistent but am a bit worried that these specific pages feel a bit borked compared to a normal docs or tutorial page.

zchsh:
@braydenlove I definitely agree it's not ideal, and does feel a bit clunky... it feels like a tough compromise to me, I can see how there's kind of inconsistency any way we slice it. Very much agreed that a taller footer isn't in scope here at all. To try to lay out some options that would be in scope:

1. Do nothing
    - Upside: Simple, consistent across pages
    - Downside: items near the bottom of the page won't be highlighted on all pages
2. Remove the active section highlighting
    - Upside: integration component views might not seem as broken
    - Downside: inconsistency, integration component view has no active highlighting but all other pages do
3. Add whitespace at the bottom of integration component views
    - Upside: active section highlighting works much better
    - Downside: whitespace at bottom of the page looks a little odd

I think we have a clear idea of what option `1` looks like (it's how this PR previews currently are), and `2` is pretty easy to imagine as well. Here's a demo of option `3`:

https://user-images.githubusercontent.com/4624598/221689823-d6fb7550-6367-47af-97b0-209efc29ab36.mp4

braydenlove:
@zchsh thanks for laying all of this out. I feel like we want to go with `1` if we are trying to be pragmatic. I feel like we will hear from real users if they feel like it is broken and if so can use that energy to push for a more holistic fix (taller footer could be one of these efforts). Thoughts? cc @heitzke @kellyschummer @BrandonRomano

braydenlove (approved):
Will monitor user feedback to see if others notice the active state friction, will elevate tasks that could help alleviate this as needed.

zchsh:
@braydenlove - wanted to note that Brandon caught an interesting question here that I wanted to confirm before merging. Is the design intent here to include all variables and include their nested properties on object variables (results in a longer nav, some repetition of top-level object names), or only top-level variables _without_ nested properties (results in a shorter name, less repetition). Here's a preview of each:

| Variables & their properties | Variables only |
| - | - |
| vars-with-properties | vars-only |

Note that both variables and their properties have permalink icons, so regardless of what the outline shows, visitors will still be able to share permalinks not just to specific variables but also to specific variable properties nested at any depth.

braydenlove:
@zchsh @BrandonRomano Variables only please!

PR #1674, opened 2023-02-22
Add sitemap xml page

nayres:
## 🔗 Relevant links
- [Sitemaps Improvements RFC](https://docs.google.com/document/d/1d2XFyZWiqRfW1zWI4cKKdkgMWBbmDKV-fsAtONUds7k)
- [Asana task](https://app.asana.com/0/1203590180322427/1204015770310372) 🎟️

## 🗒️ What
This is a phase I solution for generating a more complete sitemap. In the near future, when there is more clarity on [this issue](https://github.com/iamvishnusankar/next-sitemap/issues/534), we will use next-sitemap to dynamically generate this file.

Additionally, I configured next-sitemap to ignore the eight product paths that are included in the phase I sitemaps implementation, and linked the main sitemap.xml to each relevant sitemap in the project.

## 🤷 Why

This is phase I of a larger effort to create more complete sitemaps!

## 🛠️ How

The sitemap.xml.tsx file will live in the pages directory in the DevPortal repository (dev-portal/src/pages/sitemap.xml.tsx), and will be treated as a NextJS page. The all-docs-paths endpoint will be called, and the sitemap XML will be dynamically mapped from the API response at request time using NextJS’s getServerSideProps. The sitemap.xml file will then be accessible at developer.hashicorp.com/sitemap.xml and reachable by web crawlers.


public/sitemap-0.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
BRKalow, line 1: @nayres Should this file be checked in, or part of `.gitignore`?

next-sitemap.config.js
```diff
@@ -4,9 +4,30 @@
...
+		'/profile*',
+		'sitemap.xml',
+		'/boundary*',
+		'/consul*',
+		'/nomad*',
+		'/packer*',
+		'/terraform*',
+		'/vagrant*',
+		'/waypoint*',
+		'/vault*',
```
thiskevinwang, line 30: by excluding these, are we now missing out on sitemap entries that we were previously generating? Ex. anything "tutorials" related:

CleanShot 2023-02-22 at 17 44 18@2x

(@ https://developer.hashicorp.com/sitemap-0.xml)

---

It looks like the new sitemap only has HCP tutorials, which seems like a undesired regression here:
- https://dev-portal-git-na-create-docs-content-sitemap-hashicorp.vercel.app/sitemap-0.xml

thiskevinwang: In the theme of incremental adoption, maybe these `exclude` values can be narrowed down a bit, like:

```diff
- /waypoint*
+ /waypoint/docs*
+ /waypoint/commands*
+ // etc... I.e. only exclude "docs"
```

This way we're only excluding docs, since they're now coming from your new API endpoint, but we're still using dev-portal's build process for tutorials, integrations, etc.

nayres: added

src/pages/sitemap-docs.xml.tsx
```diff
@@ -0,0 +1,31 @@
+import { GetServerSidePropsContext } from 'next'
+
+export default function SiteMap() {
+	return null
+}
+
+const siteUrl = process.env.SITE_URL || 'https://developer.hashicorp.com'
```
BRKalow, line 7: We can use our config value, defined [here](https://github.com/hashicorp/dev-portal/blob/main/config/base.json#L22)

```suggestion
const siteUrl = __config.dev_dot.canonical_base_url
```

src/pages/sitemap-docs.xml.tsx
```diff
@@ -0,0 +1,31 @@
...
+        ${result
+					.map(
+						(sitePath: { path: string; created_at: string }) =>
+							`<url><loc>${siteUrl}/${sitePath.path}</loc><lastmod>${sitePath.created_at}</lastmod><changefreq>daily</changefreq><priority>1.0</priority></url>
+              `
+					)
+					.join('')}
+    </urlset>
+    `
+	res.setHeader('Content-Type', 'text/xml')
```
BRKalow, line 25: It seems like we should add a cache header here 🤔 ([docs](https://vercel.com/docs/concepts/edge-network/caching#serverless-functions)).

I think ideally we'd have a fairly long TTL, like an hour?

src/pages/sitemap-docs.xml.tsx
```diff
@@ -0,0 +1,31 @@
+import { GetServerSidePropsContext } from 'next'
+
+export default function SiteMap() {
+	return null
+}
+
+const siteUrl = process.env.SITE_URL || 'https://developer.hashicorp.com'
+
+export async function getServerSideProps({ res }: GetServerSidePropsContext) {
```
BRKalow, line 9: @nayres Are you thinking that the ideal state will be that `next-sitemap` handles the generation at build time? I saw your note about the issue you ran into, but just confirming that your intention is that this is a "first step."

FWIW, I'm sort of undecided on if this runtime solution, or a build-time solution, is better 🤔

public/robots.txt
```diff
@@ -0,0 +1,11 @@
...
+User-agent: *
+Allow: /
+
+# Host
+Host: https://developer.hashicorp.com
+
+# Sitemaps
+Sitemap: https://developer.hashicorp.com/sitemap.xml
+Sitemap: https://developer.hashicorp.com/hcp-sitemap.xml
+Sitemap: https://developer.hashicorp.com/sitemap-docs.xml
```
BRKalow, line 11: This is generated by `next-sitemap` too, right? Should we `.gitignore` this file?

src/pages/sitemap-docs.xml.tsx
```diff
@@ -0,0 +1,31 @@
...
+    <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+        ${result
+					.map(
+						(sitePath: { path: string; created_at: string }) =>
+							`<url><loc>${siteUrl}/${sitePath.path}</loc><lastmod>${sitePath.created_at}</lastmod><changefreq>daily</changefreq><priority>1.0</priority></url>
+              `
+					)
+					.join('')}
+    </urlset>
+    `
```
BRKalow, line 24: Did you see [this method](https://github.com/iamvishnusankar/next-sitemap#server-side-sitemap-getserversidesitemap) exported from `next-sitemap`? It seems like it could be helpful here.

next-sitemap.config.js
```diff
@@ -4,9 +4,37 @@
...
+		'/packer/docs*',
+		'/packer/commands*',
+		'/terraform/docs*',
+		'/terraform/commands*',
+		'/vagrant/docs*',
+		'/vagrant/commands*',
+		'/waypoint/docs*',
+		'/waypoint/commands*',
+		'/vault/docs*',
+		'/vault/commands*',
```
BRKalow, line 38: These all have different base paths, you can see the data files in `src/data/{product}` for a list of applicable base docs paths (minus `downloads`)

BRKalow (commented):
@nayres How much time did we spend investigating the `export-maker` issue you encountered? It seems like it might be worth timeboxing an attempt to resolve the issue with `next-sitemaps` when using the `additionalPaths` option. If we got that working it would simplify the implementation quite a bit.

src/pages/sitemap-docs.xml.tsx
nayres: Added thanks!

src/pages/sitemap-docs.xml.tsx
nayres: Removed file in favor of next-sitemap, but thanks for the comment!

public/robots.txt
nayres: Ignored all the files generated by sitemap stuff 😄

src/pages/sitemap-docs.xml.tsx
BRKalow: @nayres I think we still want to set a cache header here, it doesn't look like `next-sitemap` sets one: https://github.com/iamvishnusankar/next-sitemap/blob/master/packages/next-sitemap/src/ssr/response.ts

next-sitemap.config.js
nayres: I should have known this. I went through and added the correct paths. Thanks for catching this.

src/pages/sitemap-docs.xml.tsx
nayres: Figured it out!

src/data/vault.json
```diff
@@ -145,14 +145,7 @@
 			"os": "linux"
 		}
 	],
-	"basePaths": [
-		"docs",
-		"api-docs",
-		"intro",
-		"tutorials",
-		"downloads",
-		"downloads"
```
BRKalow, line 154: good catch 😄

next-sitemap.config.js
```diff
@@ -4,9 +4,47 @@
...
-	// ...other options
+	siteUrl,
+	generateRobotsTxt: true,
+	robotsTxtOptions: {
+		additionalSitemaps: [
+			`${siteUrl}/hcp-sitemap.xml`,
+			`${siteUrl}/server-sitemap.xml`,
+		],
+	},
+	exclude: [
```
BRKalow, line 18: Maybe we can add a comment here explaining why we have the extensive exclude list 💭

src/pages/server-sitemap.xml/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+) => {
+	const response = await fetch(
+		`https://content.hashicorp.com/api/all-docs-paths`
+	)
+
+	const { result } = await response.json()
+	const fields = result.map((page: { path: string; created_at: string }) => ({
+		loc: `${__config.dev_dot.canonical_base_url}/${page.path}`,
+		lastmod: page.created_at,
+		priority: 0.7,
```
BRKalow, line 15: Is there a specific reason we chose `0.7` for the value here, or is it a default?

nayres: 0.7 is default for next-sitemap, but since we are compiling an additional sitemap manually, I needed to add it in.

next-sitemap.config.js
nayres: Great idea, will add in

PR #1677, opened 2023-02-23
feat: adds ISR to tutorial paths

kendallstrautman:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What
[
View diff without whitespace](https://github.com/hashicorp/dev-portal/pull/1677/files?diff=split&w=1)

This adds ISR to all tutorial related routes, product tutorials landing, collection and tutorial pages.

### DIFF
--- a/src/pages/[productSlug]/tutorials/[...tutorialSlug]/index.tsx
+++ b/src/pages/[productSlug]/tutorials/[...tutorialSlug]/index.tsx
@@ -83,7 +83,7 @@ async function _getStaticProps({
 	const { productSlug, tutorialSlug } = params
 
 	const productData = cachedGetProductData(productSlug) as LearnProductData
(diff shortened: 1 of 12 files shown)


src/pages/[productSlug]/tutorials/[...tutorialSlug]/index.tsx
```diff
@@ -88,7 +88,7 @@ async function _getStaticProps({
 	if (!props) {
 		return { notFound: true }
 	}
-	return props
+	return { props, revalidate: __config.dev_dot.revalidate }
```
Comment (kendallstrautman) [inline src/pages/[productSlug]/tutorials/[...tutorialSlug]/index.tsx:91]: Just an opinion but I do prefer the meaningful `getStaticProps` logic on the page file itself. In other places we return revalidate with the same function that we create the props on. Personally, I find it helpful to go directly to the page file to check out paths / props config like this are being set.

kendallstrautman:
jk

PR #1678, opened 2023-02-24
[ProductIntegrationsLanding] Adding loading skeleton

ashleemboyer:
## 🗒️ What

- Adds a LoadingSkeleton component scoped to `ProductIntegrationsLanding`.

## 🤷 Why

- This both prevents a page flash and hydration mismatch error when a user lands on this page with query string parameters in the URL.

## 🛠️ How

- The CSS for the placeholder is mostly borrowed from our [`@hashicorp/react-placeholder`](https://react-components.vercel.app/components/placeholder) component.

### Wide viewports

- [ ] Make sure your viewport is `729px` or wider
- [ ] Go to [/waypoint/integrations](https://dev-portal-git-ambintegrations-loading-skeleton-hashicorp.vercel.app/waypoint/integrations) (without query parameters)
- [ ] The loading skeleton should match the following shape:
  `<TODO INSERT SCREENSHOT>`
- [ ] Go to [/waypoint/integrations?tiers=official&components=builder%2Cconfig-sourcer%2Cplatform](https://dev-portal-git-ambintegrations-loading-skeleton-hashicorp.vercel.app/waypoint/integrations?tiers=official&components=builder%2Cconfig-sourcer%2Cplatform) (with query parameters)
- [ ] The loading skeleton should match the following shape:
  `<TODO INSERT SCREENSHOT>`

### Narrow viewports

- [ ] Change your viewport to `728px` or narrower
- [ ] Go to [/waypoint/integrations](https://dev-portal-git-ambintegrations-loading-skeleton-hashicorp.vercel.app/waypoint/integrations) (without query parameters)
- [ ] The loading skeleton should match the following shape:
  `<TODO INSERT SCREENSHOT>`
- [ ] Go to [/waypoint/integrations?tiers=official&components=builder%2Cconfig-sourcer%2Cplatform](https://dev-portal-git-ambintegrations-loading-skeleton-hashicorp.vercel.app/waypoint/integrations?tiers=official&components=builder%2Cconfig-sourcer%2Cplatform) (with query parameters)
- [ ] The loading skeleton should match the following shape:
  `<TODO INSERT SCREENSHOT>`

## 💭 Anything else?

- These changes should be considered a skateboard version of loading skeletons that solve an immediate problem on `ProductIntegrationsLanding` pages with query string parameters on page load. The approach taken here is not stable and requires higher maintenance than necessary to keep the loading skeleton shape in sync with the shape of the view. Soon there will be an RFC for an improved Loading Skeleton system, and this has been noted in a `TODO` block comment above the `ProductIntegrationsLanding` definition.

---
  - https://app.asana.com/0/0/1203868487235738


src/views/product-integrations-landing/components/loading-skeleton/integrations-loading-skeleton.module.css
```diff
@@ -0,0 +1,76 @@
...
+	background-size: 75%;
+	border-radius: 2px;
+	opacity: 0.55;
+	width: 100%;
+
+	/* Only enable faster animation if query is supported and value is no-preference */
+	@media (prefers-reduced-motion: no-preference) {
+		--loading-skeleton-animation-duration: 2s;
+	}
+}
```
ashleemboyer, line 28: Slowing both animation durations by `0.5s` after chatting with @braydenlove and @kellyschummer in Slack. 💖

```suggestion
.placeholder {
	--loading-skeleton-animation-duration: 4.5s;

	animation: loading var(--loading-skeleton-animation-duration) infinite;
	background-color: var(--token-color-palette-alpha-100);
	background-image: linear-gradient(
		90deg,
		transparent,
		var(--token-color-palette-alpha-100),
		var(--token-color-palette-alpha-200),
		transparent
	);
	background-position: -300% 0;
	background-repeat: no-repeat;
	background-size: 75%;
	border-radius: 2px;
	opacity: 0.55;
	width: 100%;

	/* Only enable faster animation if query is supported and value is no-preference */
	@media (prefers-reduced-motion: no-preference) {
		--loading-skeleton-animation-duration: 2.5s;
	}
}
```

braydenlove (approved):
Nice work on this.

src/views/product-integrations-landing/index.tsx
```diff
@@ -27,6 +28,36 @@ export default function ProductIntegrationsLanding({
...
+	 * To balance these two needs we show the loading skeleton slightly longer
+	 * than needed to prevent the flash, and we animate the loading skeleton to
+	 * keep users engaged while it's displayed.
+	 */
+	const [isClientLoading, setIsClientLoading] = useState(true)
+	useEffect(() => {
+		setTimeout(() => {
+			setIsClientLoading(false)
+		}, 1500)
+	}, [])
```
BrandonRomano, line 59: I am strongly against using an arbitrary number for a loader, and always running it.  I would much rather use an actual hook which resolves when the loading is actually complete.

There's a few reasons for this:
- Regardless of the number that we pick, there is theoretically a machine that will take longer than the number we pick.
- We're slowing it down more than it needs to be in many cases, which is degrading the experience for users. (I'll hit the a11y point below separately).
- We're actually adding a loader, even when nothing needs to load (when there are no URL parameters, which will be a vast majority of page loads).

## RE: a11y

I am unfamiliar with the a11y concerns with loaders; so you'll have to help me understand the implications regarding flashing content.  I may be interpreting these specs incorrectly here, so please correct me if I am wrong:

> "Flashing" refers to content that can trigger a seizure (if it is more than 3 per second and large and bright enough). This cannot be allowed even for a second or it could cause a seizure. And turning the flash off is also not an option since the seizure could occur faster than most users could turn it off. 
>
> https://www.w3.org/WAI/WCAG21/Understanding/pause-stop-hide.html

Given this, I don't see how this situation meets the "flashing" criteria based off of this definition.

CleanShot 2023-02-28 at 12 22 43@2x

kellyschummer (approved):
🦴 🥇 - Thank you @ashleemboyer!

src/views/product-integrations-landing/index.tsx
```diff
@@ -39,7 +70,11 @@ export default function ProductIntegrationsLanding({
 						heading={`${product.name} Integrations`}
 						description={product.integrationsConfig.description}
 					/>
-					<SearchableIntegrationsList className={s.searchList} />
+					{isClientLoading ? (
+						<LoadingSkeleton />
+					) : (
+						<SearchableIntegrationsList className={s.searchList} />
+					)}
```
BrandonRomano, line 77: I like this approach, very straightforward 👍

src/views/product-integrations-landing/components/loading-skeleton/integrations-loading-skeleton.module.css
```diff
@@ -0,0 +1,76 @@
...
+	background-size: 75%;
+	border-radius: 2px;
+	opacity: 0.55;
+	width: 100%;
+
+	/* Only enable faster animation if query is supported and value is no-preference */
+	@media (prefers-reduced-motion: no-preference) {
+		--loading-skeleton-animation-duration: 2.5s;
+	}
+}
```
BrandonRomano, line 28: In the future we might be able to abstract this to a `SkeletonPart` component or something as part of our future abstraction ideas.  Just a thought!

That being said, I like this straightforward approach for now!

BrandonRomano (changes requested):

src/views/product-integrations-landing/components/loading-skeleton/integrations-loading-skeleton.module.css
ashleemboyer: There are definitely plans to lift this up into a reusable component! I'd like us to wait to do that until we go through an RFC process, which should start in the next 1-2 weeks. 😊 

- [Asana task](https://app.asana.com/0/1202097197789424/1202791375540720/f)
- [In-progress RFC](https://docs.google.com/document/d/18b-n8c-mWXH_YF05-zn9tJcW9voO-R8RcWH-HCDwCXw/edit)

src/views/product-integrations-landing/index.tsx
ashleemboyer: > Regardless of the number that we pick, there is theoretically a machine that will take longer than the number we pick.

To make sure we're on the same page -- this effect is running after hydration, which is after data is loaded server-side. Are you referring to network loading times or JavaScript computation time that happens with the client-side data filtering?

---

> We're actually adding a loader, even when nothing needs to load (when there are no URL parameters, which will be a vast majority of page loads).

We do have to wait for the client to load every time the page is visited, because the server isn't capable of knowing if there are URL parameters or not on the client. We're not able to conditionally render the loading skeleton based on the presence of query string parameters.

---

> We're slowing it down more than it needs to be in many cases, which is degrading the experience for users.

> Given this, I don't see how this situation meets the "flashing" criteria based off of this definition.

Unfortunately the definition you've linked is limited, and so are the ones under [SC 2.3: Seizures and Physical Reactions](https://www.w3.org/WAI/WCAG21/Understanding/seizures-and-physical-reactions). 😕 The summary text of the SC says "Do not design content in a way that is known to cause seizures or physical reactions." but it does not go into depth about physical reactions other than seizures.

[SC 2.3.3: Animation from Interactions](https://www.w3.org/WAI/WCAG21/Understanding/animation-from-interactions.html) talks about vestibular disorders, but only in the scope of animation. The physical reactions mentioned in the latter are also applicable to page flashes (even just one flash), even though the standards don't recognize that.

So the intent of manually extending the render time of the loading skeleton is intended to reduce the one flash as much as possible, to reduce the risk of causing physical reactions to users as much as possible. Animating the loading skeleton helps keeps other users engaged during the brief time we show the loading skeleton -- those who have little to no risk of having a physical reaction to the page flash.

BrandonRomano: > To make sure we're on the same page -- this effect is running after hydration, which is after data is loaded server-side. Are you referring to network loading times or JavaScript computation time that happens with the client-side data filtering?

I believe the client side filtering; presumably if that takes longer than 1500ms (unlikely, I know, but still!) you run into this situation.

> ... even though the standards don't recognize that

I guess this is where I am struggling here. We're committed to following the W3 standard, but what you're suggesting we adhere to is above & beyond the requirements laid out in the standard. 

I don't disagree that we should attempt to reduce something that could cause any negative reaction; but if we're going to expect this as our standard, we need to codify this somewhere (an RFC, etc). I want to make sure that if this is something that we want to follow, it can be followed in a manner that doesn't require you personally catching it / having to provide feedback on. Additionally, because it is more than the specification lays out, and may have implications that design needs to adhere to when designing out interactions, I would like them to weigh in on it. 

Would like to hear other thoughts from the team here though (Maybe @kendallstrautman , @BRKalow)

ashleemboyer: > I don't disagree that we should attempt to reduce something that could cause any negative reaction; but if we're going to expect this as our standard, we need to codify this somewhere (an RFC, etc). I want to make sure that if this is something that we want to follow, it can be followed in a manner that doesn't require you personally catching it / having to provide feedback on. Additionally, because it is more than the specification lays out, and may have implications that design needs to adhere to when designing out interactions, I would like them to weigh in on it.

I plan to include this topic in [the in-progress loading skeletons RFC](https://docs.google.com/document/d/18b-n8c-mWXH_YF05-zn9tJcW9voO-R8RcWH-HCDwCXw/edit) I linked in another comment. Do we want to block this PR on that RFC, or can we come back and make additional changes from where we're at after the RFC reaches resolution?

BrandonRomano: Let's wait for others to chime in, and then we will reassess!

kendallstrautman: > I am strongly against using an arbitrary number for a loader,

I agree with this. Ideally we would connect that loading state with a meaningful dependency, a network request etc. 

> I plan to include this topic in [the in-progress loading skeletons RFC](https://docs.google.com/document/d/18b-n8c-mWXH_YF05-zn9tJcW9voO-R8RcWH-HCDwCXw/edit) I linked in another comment

That would be a good place to discuss our teams expectations around handling 'flashing'. But we have implemented a [time based delay](https://github.com/hashicorp/learn/blob/30c1a4de84e56a2eaab4016dde346d72a61171c8/lib/hooks/progress/use-progress.ts#L27) like this to smooth out transitions before with loading skeletons on learn. So I have seen the flashing issue before and I interpreted is as unacceptable UX since the network often resolves so fast that the skeleton ends up looking like a bug.

kendallstrautman: The loading state should be separate from the animation delay though. So there isn't a possibility of the loading state 'resolving' when the true dependency hasn't yet resolved.

BRKalow: I think it's good we're trying to go above and beyond the success criteria here, but we should be explicit about that in a comment with some resources to provide additional context if that's the approach we want to take.

Another consideration here with the long artificial delay is that the timer won't start until the component is hydrated, and on slower connections / devices this could result in a much longer delay than necessary. For example, if I simulate a 4x CPU slowdown and reload, the page takes ~3s to hydrate and then another 1.5s to display the rendered components. We're effectively optimizing for the best case scenario (a really fast initial load time), which I don't think is an ideal experience for all.

ashleemboyer: > I think it's good we're trying to go above and beyond the success criteria here, but we should be explicit about that in a comment with some resources to provide additional context if that's the approach we want to take

FWIW I had a long block comment above the code this comment thread is attached to (link is on the specific commit):
https://github.com/hashicorp/dev-portal/blob/b697a13dc2d5b36745e920fdb096b70e940a88fa/src/views/product-integrations-landing/index.tsx#L31-L59

> Based on the linked resources above, it feels like we're satisfying the explicit criteria even without the delay. I'm curious to understand better how a single flash can be harmful!

Yes, what we have meets the WCAG criteria. I think about this like we think about component-level vs. page-level accessibility. A component can be accessible on its own, but that doesn't guarantee a page that renders the component is accessible. Meeting these SC within the `LoadingSkeleton` component makes it accessible on its own, but doesn't guarantee it's accessible when placed in a page.

I think it's also important to note that WCAG has shortcomings, which is called out in the introductory parts of the standards documents.

I can collect some more pieces on vestibular issues, but this is a good one to start with: [Accessibility for Vestibular Disorders: How My Temporary Disability Changed My Perspective](https://alistapart.com/article/accessibility-for-vestibular/). There are a wide range of triggers that can cause physical reactions, and peoples' tolerance for those triggers also varies.

At a site level, someone could have a reaction to _this_ flash because they experienced some on the pages before this one. Or they could have a reaction to the next flash they see after this one. And similarly at a browser level, we don't know what sites people view before or after ours. Even though we can't control what other sites do, we can control our unintentional contribution to causing physical reactions. And in this case, we can do it with minimal effort.

> Another consideration here with the long artificial delay is that the timer won't start until the component is hydrated, and on slower connections / devices this could result in a much longer delay than necessary. For example, if I simulate a 4x CPU slowdown and reload, the page takes ~3s to hydrate and then another 1.5s to display the rendered components. We're effectively optimizing for the best case scenario (a really fast initial load time), which I don't think is an ideal experience for all.

We can definitely work to refine our delay value! However, I don't think "an ideal experience for all" is an attainable goal. The balance here centers around measuring harm and inconvenience. Sometimes the process of preventing harm for some users increases inconvenience for others, and I think that's OK. It just needs to be thoroughly considered.

BRKalow: Yeah I saw your comment, definitely helpful! I was thinking more specifically links to resources that helped inform this decision. 👍  For example, how did we land on the 1.5 second delay? Is that based on a suggestion from somewhere?

src/views/product-integrations-landing/components/searchable-integrations-list/index.tsx
```diff
@@ -193,6 +196,54 @@ export default function SearchableIntegrationsList({
...
+		useState(hasQueryParams)
+	useEffect(() => {
+		if (!isLoadingQueryParams) {
+			return
+		}
+
+		if (
+			contextState.integrations.length !==
+			contextState.filteredIntegrations.length
+		) {
```
BrandonRomano, line 233: This check will cause the loading to never resolve, in the event the query parameters filter logic does not filter out any integrations.  A concrete example is that Waypoint integrations are all built-ins right now:

https://dev-portal-git-ambintegrations-loading-skeleton-hashicorp.vercel.app/waypoint/integrations?flags=builtin

src/views/product-integrations-landing/index.tsx
ashleemboyer: Ah, nope, that number doesn't come from anywhere specific.

BrandonRomano (changes requested):
Looks like there's an edge case here to how we are determining if the integrations have been filtered yet.

src/views/product-integrations-landing/index.tsx
BRKalow: Do you think there's a minimum delay that would be acceptable to address concerns you've outlined (thanks for the context btw)? Is that 1.5s?

ashleemboyer:
Leaving this closed as https://github.com/hashicorp/dev-portal/pull/1730 is now obsolete.

PR #1683, opened 2023-02-24
refactor: rethink enableActiveSection for OutlineNav

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Updates the logic to determine whether to `enableActiveSection` functionality in `OutlineNavWithActive`.

## 🤷 Why

The previous logic relied on the not-fully-correct `isDesktop` hook, based on CSS breakpoints. To reduce brittleness and ensure active section logic was not being calculated when `OutlineNav` is not visible, we wanted to revise our approach.

## 🛠️ How

- Adds a `useOnScreen` hook
    - This uses an `intersectionObserver` to monitor whether an element is visible.
- Uses `useOnScreen` to set `enableActiveSection` within `OutlineNavWithActive`
    - When `OutlineNavWithActive` is visible, we `enableActiveSection`. Else, we don't enable it.

## 🧪 Testing

- [ ] Visit a page that uses `OutlineNavWithActive`, such as [/vault/api-docs/auth/aws][/vault/api-docs/auth/aws]
    - The active section highlighting should continue to function as expected
    - To validate that `enableActiveSection` is `false` when the nav is not visible, you could if you want to clone the project locally and add a `console.log` of `enableActiveSection` in the render body of `OutlineNavWithActive`. Then, on a page running locally, resize the browser window and note the correlation between the outline nav being hidden (via CSS) and the changes in `enableActiveSection`.

## 💭 Anything else?

Not at the moment.

[task]: https://app.asana.com/0/1202097197789424/1204024966433307/f
[preview]: https://dev-portal-git-zsrethink-use-active-section-ac-8c3ad1-hashicorp.vercel.app/
[/vault/api-docs/auth/aws]: https://dev-portal-git-zsrethink-use-active-section-ac-8c3ad1-hashicorp.vercel.app/vault/api-docs/auth/aws


ashleemboyer (approved):
Checked out locally, and worked for me! ✅

src/lib/hooks/use-on-screen.ts
```diff
@@ -0,0 +1,36 @@
+import { useEffect, useState, MutableRefObject } from 'react'
+
+/**
+ * Given a `ref`, returns a `boolean` indicating whether the `ref.current`
+ * element is visible on screen.
+ *
+ * From https://usehooks.com/useOnScreen/
+ */
+function useOnScreen(ref: MutableRefObject<HTMLElement>): boolean {
```
alexcarpenter, line 9: 💭 non-blocker: if I am understanding correctly, this is doing a lot of what you get from framer-motion `useInView` hook? https://www.framer.com/motion/use-in-view/ Checking to see if an element is within the viewport?

zchsh: ah true yeah, great catch! i'll try swapping in `useInView` 👍

zchsh: ✅ 🙌 swapped it in, works great! Thanks for the suggestion 💖

alexcarpenter: Not sure if its a 1to1 mapping of `useOnScreen` but was curious.

PR #1684, opened 2023-02-24
feat: adds revalidate all tutorials endpoint

kendallstrautman:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1202097197789424/1203961914300039) 🎟️

## 🗒️ What

Adds a new api endpoint to handle revalidating all tutorials landing, collection and individual tutorial paths. This endpoint will be called after the production content sync is done to bust the cache for all tutorials routes -[ see related PR](https://github.com/hashicorp/tutorials/pull/1175)

After this PR is merged, I will test the revalidation endpoint before merging the update to the tutorials prod sync workflow.


src/views/collection-view/helpers/get-slug.ts
```diff
@@ -22,11 +23,7 @@ export function getTutorialSlug(
 	const [rawProductSlug, collectionFilename] = collectionDbSlug.split('/')
 	const tutorialFilename = splitProductFromFilename(tutorialDbSlug)
 
-	// @TODO genericize this to use 'topic' or 'section' instead of 'product'
-	if (
-		rawProductSlug === 'well-architected-framework' ||
-		rawProductSlug === 'onboarding'
-	) {
+	if (rawProductSlug in SectionOption) {
```
kendallstrautman, line 26: Random cleanup here

src/pages/api/revalidate-all-tutorials.ts
```diff
@@ -0,0 +1,117 @@
...
+			batchRevalidatePromises.push(response.revalidate(path))
+
+			// flush the batch every N paths, or at the end of the loop
+			if (
+				batchRevalidatePromises.length >= BATCH_SIZE ||
+				i >= paths.length - 1
+			) {
+				await Promise.allSettled(batchRevalidatePromises)
+				// delay to ensure we don't overwhelm the server
+				await delay(DELAY_TIME)
```
kendallstrautman, line 62: Wanted to keep this simple and but open to suggestions here! I saw issues when only doing the batching (this is like ~1800 paths? according to my tests) and the delay makes things run great locally.

src/pages/api/revalidate-all-tutorials.ts
```diff
@@ -0,0 +1,117 @@
...
+		]
+
+		if (!paths || paths.length === 0) {
+			response.status(200).end()
+		}
+
+		// Loop over all paths to revalidate in batches
+		// as this endpoint will fire off >1000 revalidation requests
+		let batchRevalidatePromises = []
+		for (let i = 0; i < paths.length; i++) {
```
kendallstrautman, line 49: Referenced the batching implementation in the learn api content sync here.

src/pages/api/revalidate-all-tutorials.ts
BRKalow: FWIW, I don't think we have run into any rate limiting with the docs revalidation endpoint yet, and a few of the products have >500 page that get revalidated in one go without any throttling (locally, it's a no-op so that shouldn't be a concern).

We do have `p-map` in our dependencies already, so that could be a good solution to add some throttling!

src/pages/api/revalidate-all-tutorials.ts
```diff
@@ -0,0 +1,117 @@
...
+			// remove any trailing slash
+			const path = paths[i].replace(/\/$/, '')
+			console.log('[Revalidate]', path)
```
BRKalow, line 52: nit, for consistency:
```suggestion
			console.log('[revalidate]', path)
```

BRKalow (commented):
I think this is a great first pass at on-demand ISR for tutorials! Nice work.

src/pages/api/revalidate-all-tutorials.ts
kendallstrautman: Okay, I'll remove the delay for now and watch after we roll it out!

kendallstrautman: I think p-map only supports concurrency limits but not throttling? I added a note reference both p-map and p-throttle in case we need to add this later on

kendallstrautman:
@BRKalow just removed the delay implementation and left a note for the future on throttling. Lmk if you see any other things to improve ☺️

BRKalow (approved):
:shipit: Let's give it a shot!

zchsh (approved):
🙌  I'm not super familiar with the revalidation process, but from what I am aware of this looks good to me!

If there are any more detailed testing steps that'd be helpful to dive into, happy to do that 🙇

src/pages/api/revalidate-all-tutorials.ts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+async function getCollectionAndTutorialPaths() {
+	const collectionPaths = []
+	const tutorialPaths = []
+	const allCollections = await getAllCollections()
+
+	allCollections.forEach((collection: ClientCollection) => {
+		// build collection paths
+		collectionPaths.push(getCollectionSlug(collection.slug))
```
kendallstrautman, line 99: @nayres this is some code to reference for generating the collection paths!

src/pages/api/revalidate-all-tutorials.ts
```diff
@@ -0,0 +1,118 @@
...
+		]
+
+		if (!paths || paths.length === 0) {
+			response.status(200).end()
+		}
+
+		// Sending 'accepted' status to avoid socket hang-up as the path revaliadation takes a very long time
+		response
+			.status(StatusCodes.ACCEPTED)
+			.send(`Revalidating all tutorials: ${paths.length} paths`)
```
kendallstrautman, line 48: @BRKalow would this be a better response since we're expecting this batch to take a really long time and the client could timeout? I experienced a socket hangup when testing with postman locally and wondered if we should keep the response open while waiting for all of these to settle.

BRKalow: @kendallstrautman I think that's reasonable yeah. The potential risk here is that we don't have visibility into the status of the revalidation, but we can iterate on that. Any idea how long the full revalidation takes?

kendallstrautman: Yeah... I'm unsure the right way to handle this as an API newb. Locally I've seen it take between 5-8 minutes.

kendallstrautman:
Testing the full flow from a tutorials action trigger here https://github.com/hashicorp/tutorials/actions/runs/4325350508/jobs/7551391150

src/pages/api/revalidate/tutorials.ts
```diff
@@ -0,0 +1,115 @@
...
+
+function getTutorialLandingPaths(): string[] {
+	const paths = []
+
+	// build all product tutorial landing paths
+	activeProductSlugs.forEach((productSlug: ProductSlug) => {
+		paths.push(`/${productSlug}/tutorials`)
+	})
+
+	// for waf and onboarding routes that are top level
```
kendallstrautman, line 86: TODO: Do not build /onboarding since it doesn't exist

src/pages/api/revalidate-all-tutorials.ts
kendallstrautman: Okay yeah looks like we get a [socket hangup in the action](https://github.com/hashicorp/tutorials/actions/runs/4325350508/jobs/7551391150) if we wait for everything to resolve. 
Screenshot 2023-03-03 at 8 35 10 AM

BRKalow: 😱

nayres (approved):

zchsh:
Closing this for now, see 🎟 [Tutorials view on-demand ISR](https://app.asana.com/0/1202097197789424/1203961914300039/f) for context.

PR #1685, opened 2023-02-27
refactor: move docs outline logic to server

zchsh:
- chore: extract outline-from-headings util & type
- docs: move comment to correct area
- chore: split out getDocsMdxComponents as prep

## 🔗 Relevant links

- [Asana task](url) 🎟️


src/components/outline-nav/types.ts
```diff
@@ -38,12 +38,3 @@ export type OutlineLinkItem = OutlineLinkProps | OutlineLinkWithNestingProps
 export interface OutlineNavProps {
 	items: OutlineLinkItem[]
 }
-
-/**
- * Heading type, for compatibility with existing conventions.
- */
-export interface OutlineHeading {
```
zchsh, line 45: I've renamed this type `AnchorLinksPluginHeading` and moved it into `lib/docs/outline-items-from-headings.ts`, since I needed this type in a few places!

src/layouts/docs-view-layout/index.tsx
```diff
@@ -3,19 +3,25 @@
...
+	SidebarSidecarLayoutProps,
+} from 'layouts/sidebar-sidecar'
 import { DocsVersionAlertBanner } from './components'
 
 /**
  * Lightweight wrapper around SidebarSidecarWithToc which passes along some docs
  * specific props.
  */
-const DocsViewLayout = (props: SidebarSidecarWithTocProps) => {
+const DocsViewLayout = ({
```
zchsh, line 17: All views that use `DocsViewLayout` must now ensure they're passing already-processed `outlineItems` from `getStaticProps`, not `headings` like before.

src/pages/packer/plugins/[...page].tsx
```diff
@@ -170,23 +171,30 @@ export async function getStaticProps({ params, ...ctx }) {
...
-				description: props.frontMatter.description ?? null,
-			},
-			mdxSource: props.mdxSource,
-			product: productData,
+	const finalProps: DocsViewProps = {
+		layoutProps: {
+			breadcrumbLinks,
+			githubFileUrl: props.githubFileUrl,
+			/**
+			 * TODO: we should likely try to avoid casting here, and instead ensure
```
Comment (zchsh) [inline src/pages/packer/plugins/[...page].tsx:179]: I've tried to fix up `DocsViewProps` related types wherever I can. Usually this is a step towards type safety, but doesn't always get us there. Will be keeping an eye out in the future for opportunities to improve types; I think trying to fix up all related types would balloon the scope of this PR a little too much 🎈

src/views/docs-view/index.tsx
```diff
@@ -7,19 +7,27 @@ import dynamic from 'next/dynamic'
...
+ * Note: layout in frontmatter is not supported yet, this is early stage work.
  * Asana task: https://app.asana.com/0/1202097197789424/1202850056121889/f
+ *
+ * Note: this layout logic is used for non-`/docs` landing pages.
+ * The `/docs` landing pages use `ProductRootDocsPathLanding`.
+ * We could consider this approach, as it implies arbitrary layouts
+ * are supported via frontmatter. We could also consider renaming the
+ * "layout" property to something like "contentLayout", as otherwise
+ * it seems like it could be confused with broader "src/layouts" code.
+ * Task: https://app.asana.com/0/1202097197789424/1204069295311480/f
```
zchsh, line 30: I got confused here, despite the fact that IIRC i think I implemented some of this stuff (😅), so added a task here to try to make the flow here a little more clear in the future.

src/views/product-root-docs-path-landing/server.ts
```diff
@@ -134,19 +132,31 @@ const getStaticProps = async (context: GetStaticPropsContext) => {
...
 			baseName,
 		})
-
-	// TODO: replace any with accurate type
-	const generatedProps = (await generatedGetStaticProps({
+	const getStaticPropsResult = (await generatedGetStaticProps({
 		...context,
 		params: { page: [] },
-	})) as $TSFixMe
+		// TODO: surely there isn't a need to cast here... but there seems to be?
```
zchsh, line 138: 💭 I would love to clean this type up, but got kind of stuck with it... any and all input welcome!

I'm going to try to dig a bit more into NextJS + TypeScript, it seems odd to me that this requires casting since I've set what I _thought_ was the correct return type on the function that returns `generatedGetStaticProps` 🤔 ... but which must not be correct since this isn't working 😅

zchsh: I think I figured this out - I needed to account for `redirect` and `notFound` cases. Have added a little conditional to handle those cases, which conveniently also serves as a type guard.

src/layouts/docs-view-layout/index.tsx
```diff
@@ -3,19 +3,43 @@
...
+			 * DOM directly in `useActiveSection`.
+			 *
+			 * For context, with the current approach of accessing the DOM directly,
+			 * we must ensure that when the content re-renders (updating heading
+			 * DOM elements), the OutlineNav must also re-render (so that
+			 * `useActiveSection` runs setup again and finds the updated DOM elements,
+			 * rather than maintaining stale references).
+			 *
+			 * We took a spike at this in the following commit:
+			 * https://github.com/hashicorp/dev-portal/pull/1685/commits/70a034789006df2f14afae665962c4087f2a6ba1
```
zchsh, line 40: I paired with Bryce on this - I'm not exactly _happy_ with the `slice()` solution here... but at the same time, doing the "lift `useActiveSection`" work, while it may be a lil more robust since it gets to the root of the issue (changing `useActiveSection` so it doesn't access the DOM directly)  is kind of a significant refactor, and felt outside the scope of this PR. I've added a follow-up task at 🎟 [[maintenance] refactor uses of useActiveSection to not access the DOM directly](https://app.asana.com/0/1202097197789424/1204098852315293/f).

kendallstrautman: Wow this is a sticky one, thanks for the explanation in the comment and link to the task 💯

src/pages/packer/plugins/[...page].tsx
Comment (kendallstrautman) [reply src/pages/packer/plugins/[...page].tsx:188]: Seems fair! I know I've run into the menu items type issues before as well.

src/views/docs-view/server.ts
```diff
@@ -31,13 +31,15 @@ import {
...
-import { EnrichedNavItem } from 'components/sidebar/types'
 import { getBackToLink } from './utils/get-back-to-link'
 import { getDeployPreviewLoader } from './utils/get-deploy-preview-loader'
 import { getCustomLayout } from './utils/get-custom-layout'
 import type { DocsViewPropOptions } from './utils/get-root-docs-path-generation-functions'
 import { getStaticPathsFromAnalytics } from 'lib/get-static-paths-from-analytics'
 import { withTiming } from 'lib/with-timing'
+import outlineItemsFromHeadings from 'lib/docs/outline-items-from-headings'
+import { DocsViewProps } from './types'
+import { GetStaticPaths, GetStaticProps, GetStaticPropsResult } from 'next'
```
kendallstrautman, line 42: ⛏️ could we lift this to the top of the file?

src/views/docs-view/server.ts
```diff
@@ -77,7 +79,10 @@ export function getStaticGenerationFunctions<
 	mainBranch?: string
 	navDataPrefix?: string
 	options?: DocsViewPropOptions
-}): ReturnType<typeof _getStaticGenerationFunctions> {
+}): {
+	getStaticPaths: GetStaticPaths
+	getStaticProps: GetStaticProps<DocsViewProps>
```
kendallstrautman, line 84: Nice, this is a good improvement

src/views/docs-view/utils/get-docs-mdx-components.ts
```diff
@@ -0,0 +1,68 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import { ReactNode } from 'react'
+import dynamic from 'next/dynamic'
+import { ProductSlug } from 'types/products'
+import defaultMdxComponents from 'layouts/sidebar-sidecar/utils/_local_platform-docs-mdx'
```
kendallstrautman, line 9: Much better in this file, thank you!

src/views/product-root-docs-path-landing/index.tsx
```diff
@@ -30,42 +34,55 @@ const ProductRootDocsPathLanding = ({
 
 	let mdxSlot: ReactElement
 	if (mdxSource) {
-		const classes = classNames(s[`${product.slug}MDXWrapper`], s.mdxSlotWrapper)
+		const docsMdxComponents = getDocsMdxComponents(product.slug)
+		const classes = classNames(
+			s[`${product.slug}MDXWrapper`],
+			s.mdxSlotWrapper,
+			docsViewStyles.mdxContent
```
kendallstrautman, line 41: Where was this class before?

src/views/product-root-docs-path-landing/server.ts
```diff
@@ -15,12 +15,14 @@ import {
 	GETTING_STARTED_CARD_HEADING_SLUG,
 } from './components/marketing-content'
 import { prepareMarketingBlocks } from './utils/prepare-marketing-blocks'
+import { ProductRootDocsPathLandingProps } from './types'
+import outlineItemsFromHeadings from 'lib/docs/outline-items-from-headings'
+import { DocsViewProps } from 'views/docs-view'
```
kendallstrautman, line 20: ⛏️ Another nit to reorder these imports

kendallstrautman (approved):
👍🏻 LGTM! I tested the preview against main. Thanks for the refactor and cleanup, extra kudos 💯

src/lib/docs/outline-items-from-headings.ts
```diff
@@ -0,0 +1,33 @@
...
+	title: string
+	slug: string
+	level: 1 | 2 | 3 | 4 | 5 | 6
+}
+
+/**
+ * Given a flat array of headings,
+ * Return an array of outline items, for use with OutlineNav components.
+ */
+function outlineItemsFromHeadings(
```
ashleemboyer, line 17: 🙋🏻 I might be missing some context, won't views like `TutorialView` need this functionality as well? Not just docs?

src/pages/packer/plugins/[...page].tsx
Comment (ashleemboyer) [reply src/pages/packer/plugins/[...page].tsx:188]: Yes we've got some `TODO`s building up about this! It's related to the `MenuItem` interface being partially phased out but not fully.

https://github.com/hashicorp/dev-portal/blob/1f3bac6d9a870d0e1f9d7a30e6a5c84af21b97d3/src/views/product-landing/server.ts#L68-L79

We could update the wording in this comment to be more definitive! 😊

src/pages/packer/plugins/index.tsx
```diff
@@ -64,11 +66,20 @@ async function getStaticProps(ctx) {
...
+			 * Task: https://app.asana.com/0/1202097197789424/1202405210286689/f
+			 */
+			nodes: rawNavData as $TSFixMe as NavNode[],
 		})
 
-		// Replace our original navData with our prepared navData
-		staticProps.props.layoutProps.sidebarNavDataLevels[2].menuItems = navData
+		/**
+		 * Replace our original navData with our prepared navData
+		 * TODO: fix up types related to MenuItem[] & EnrichedMenuItem[]
```
ashleemboyer, line 78: 🙋🏻 Did you mean `EnrichedNavItem`? Could you also share your thoughts behind what needs to change about it? It was added as part of the efforts to start removing `MenuItem`, so I think I see it staying as-is and extended if needed.

src/views/docs-view/types.ts
```diff
@@ -40,6 +42,21 @@ export interface DocsViewProps {
 	 * An optional, human-readable name to be rendered where the section is referenced as a versioned project
 	 */
 	projectName?: string
+
+	/**
+	 * Product data. Note we include `currentRootDocsPath` for version switching.
+	 */
+	product: ProductWithCurrentRootDocsPath
```
ashleemboyer, line 49: 🚧 This isn't quite accurate. `DocsView` doesn't use a `product` prop, this is actually for the `CurrentProductProvider` that's initialized in `src/pages/_app.tsx`. It's value is pulled from `pageProps` here:

https://github.com/hashicorp/dev-portal/blob/1f3bac6d9a870d0e1f9d7a30e6a5c84af21b97d3/src/pages/_app.tsx#L97

src/views/product-root-docs-path-landing/server.ts
```diff
@@ -158,27 +172,45 @@ const getStaticProps = async (context: GetStaticPropsContext) => {
 		marketingContentBlocksWithHeadingLevels
 	)
 
+	/**
+	 * Transform sidecarHeadings into outlineItems, and pageHeading
+	 */
+	const firstHeading = sidecarHeadings[0]
+	const pageHeading = { id: firstHeading.id, title: firstHeading.title }
```
ashleemboyer, line 179: 🙋🏻 Did pulling `pageHeading` out like this solve a particular problem? Just curious!

ashleemboyer (changes requested):

src/views/docs-view/types.ts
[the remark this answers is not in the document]
ashleemboyer: Meant to type in the review submission -- this is my only blocking comment!

src/pages/packer/plugins/[...page].tsx
Comment (zchsh) [reply src/pages/packer/plugins/[...page].tsx:188]: Sounds good - I added a link in this comment to 🎟 [Clean up references to MenuItem interface](https://app.asana.com/0/1202097197789424/1202405210286689/f) 👍

src/views/docs-view/server.ts
zchsh: ah yep! I rely on auto-imports in VSCode a little heavily 😅 I wonder if we could add import re-ordering to pre-commit formatting or whatnot 🤔 I might look into that when I have bandwidth!

zchsh: ✅ updated to re-order imports, matched the existing `Third-party` / `HashiCorp` / `Global` / `Local` sections 👍

src/views/product-root-docs-path-landing/index.tsx
zchsh: It was in `docs-view.module.css` before as well - but it was used indirectly because the entire `DocsView` component was rendered here within the `mdxSlot`. I think it could make sense to move the associated styles (prevent `iframe` overflow; remove margin on first item) into `DevDotContent`... but that would affect a bunch of places in the app so I was a little hesitant to start down that rabbit hole 😅 

Made a backlog task with some details at 🎟 [[maintenance] move docsViewStyles.mdxContent onto DevDotContent](https://app.asana.com/0/1202097197789424/1204107433878314/f)

src/views/product-root-docs-path-landing/server.ts
zchsh: ty for catching these - ✅ updated

src/lib/docs/outline-items-from-headings.ts
zchsh: Ah true! I've moved this util into `components/outline-nav/utils` - hopefully that works! Happy to move it again in a subsequent PR if needed.

src/views/docs-view/types.ts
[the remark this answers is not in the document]
zchsh: I can see where you're coming from here - `DocsView` doesn't use the `product` prop, so in a sense it's not accurate to say that it's part of the props interface of `DocsView`.

The rationale is for why I felt it was worth including is:

1. It's very helpful to have our `getStaticProps` return typed. For that case, we do need the `product` prop to be documented. We could create a separate type, but...
2. Anything returned by `getStaticProps` will be available to the view component, so it feels helpful to know what's available. In this case for example, knowing that `product` data is available, we could likely simplify the implementation of `DocsView` by grabbing `product` and using it directly rather than having to call the `useCurrentProduct()` hook at render time. Whether or not we do this (this PR makes no changes so we don't do it), the `product` prop will still be returned from `getStaticProps` and available to `DocsView`, so it feels accurate to me to have it documented in `DocsViewProps`.

In general, it feels helpful to have `getStaticProps` and the view it will pass props to share types. So despite the fact that `DocsView` doesn't use the incoming `product` prop directly, to me it feels worth the inaccuracy in that sense in order to accurately reflect the relationship between the return type of `getStaticProps` and the incoming prop types of `DocsView`.

Happy to reconsider though if there are strong opinions here - would love to hear other thoughts cc @kendallstrautman @BRKalow

src/views/product-root-docs-path-landing/server.ts
zchsh: Yep - broadly it carries forward an existing solution, we need to return `pageHeading` as part of the page props. 

From a more specific style perspective, I added a named `const` here primarily for clarity (it felt helpful to write this out in words, rather than assuming the meaning of `sidecarHeadings[0]` was clear), and also to avoid accessing the same array index twice. 

Hopefully that answers the question 🙇

zchsh:
@ashleemboyer I think this is ready for re-review! Re: the blocking comment on including `product` in `DocsViewProps`, currently this is necessary to prevent type errors when using `DocsViewProps` to define the return value type of the associated `getStaticProps`. I have some related rationale for why we might want to keep the return value type of `getStaticProps.props` and the prop types of the related view linked. Let me know what you think 🙇

src/views/docs-view/types.ts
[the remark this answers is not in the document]
ashleemboyer: No argument with improving the TypeScript and the return types included for `getStaticProps`! I'd just like the block comment above the `product` property to be updated for accuracy. 😊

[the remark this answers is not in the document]
zchsh: Ah cool, I'll update the comment! One sec...

ashleemboyer (approved):

PR #1687, opened 2023-02-27
Updated Integrations Header Card

BrandonRomano:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-brcontent-header-card-hashicorp.vercel.app/waypoint/integrations/hashicorp/docker/latest/components/builder) 🔎
- [Asana task](https://app.asana.com/0/1203792616809419/1204016771909024/f) 🎟️
- [Figma Designs](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=20233%3A245304&t=qCYOFQx1c3Ce2XPh-0)

## TODO

- [x] ~Add copyright blurbs to all files~ (handled by bot)
- [x] Add swingset MDX example for further testing
- [x] Adjust buttons to left-align if there are no links (per Brayden)
- [x] camelCase `GetIntegrationBadges`

## 🗒️ What

Update our Integration Header card component to match the latest designs.

CleanShot 2023-02-27 at 11 19 47@2x

## 🛠️ How

Implements a generic `ContentHeaderCard` component to use on the Integrations landing page.  This was implemented via a view model to keep the content separate from the presentational layer.  This was done for two primary reasons:
1. We could use this `ContentHeaderCard` elsewhere in the future
2. All of the functionality of the Integrations usage is not ready yet (e.g. Install buttons, report etc), but I was able to build out the presentational logic.

In addition to the `ContentHeaderCard`, I also abstracted out the `BadgeList` that appears in the `ContentHeaderCard`, as that will be used in multiple places in the Integrations section of the website.

## 📸 Design Screenshots

CleanShot 2023-02-27 at 10 35 17@2x

## 🧪 Testing

Visit the preview and poke around the different Integrations.


src/components/badge-list/badge-list.module.css
```diff
@@ -1,9 +1,4 @@
-/**
```
kendallstrautman, line 1: We probably don't want to delete this copyright blurb, right?

src/components/badge-list/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+	icon?: React.ReactElement
+	tooltip?: string
+}
+
+interface BadgeListProps {
+	badges: Array<Badge>
+	className?: string
+	size?: BadgeProps['size']
+	type?: BadgeProps['type']
+	color?: BadgeProps['color']
```
kendallstrautman, line 18: Wouldn't this be passed via the badge data?

kendallstrautman: Or I see, you're supporting the assumption that all badges will be the same here.

BrandonRomano: Oh you are reviewing the wrong PR; this is mostly a POC :)

src/components/badge-list/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+				return (
+					<li key={badge.text} className={s.badge}>
+						{badge.tooltip ? (
+							<Tooltip label={badge.tooltip}>
+								<Badge
+									icon={badge.icon}
+									text={badge.text}
+									color={color}
+									type={type}
+									size={size}
```
kendallstrautman, line 40: Might want to defer to the badge data here if it is passed, since these props are optional

src/components/badge-list/index.tsx
BrandonRomano: What you are looking at here is me lifting some logic to very sloppily get this to render!

src/components/badge-list/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export default function BadgeList({
+	badges,
+	className,
+	size = 'medium',
+	color = 'neutral',
+	type = 'filled',
+}: BadgeListProps) {
+	return (
+		<ul className={classNames(s.badgeList, className)}>
```
kendallstrautman, line 29: 💭 One approach I really like with these list wrapper components is for them to just render children. I find the flexibility there making them much easier to reuse. [Here's one example.](https://github.com/hashicorp/dev-portal/blob/76355ca7a76e69d24e6e32d42bc5a046a98e5687/src/components/navigation-header/components/product-page-content/index.tsx#L44-L59) Not necessary to act on, but just sharing as food for thought.

src/components/badge-list/index.tsx
BrandonRomano: OH; I am sorry; I misread this PR context :homer-fade:, you are on the right PR!

BrandonRomano: Correct! This ensures that all badges displayed in a list are of the same size / type / color treatment.

src/components/badge-list/badge-list.module.css
BrandonRomano: Yep I'll have to add it back... I have to add it to all of the new files as well.

src/components/badge-list/index.tsx
BrandonRomano: Not 100% quite sure I follow!

Adding context that might help, if I assume what you are suggesting correctly:

I am passing in the color / type / size as passed at the `badge-list` level here (the individual badge interfaces I have is a reduced interface on the actual Badge component, so you cannot set the color / type / size of the badge directly).

src/components/badge-list/index.tsx
kendallstrautman: Personally, if were lifting this up to be used globally, I think forcing the same size etc should be handled by the badge data generation logic.

src/components/badge-list/index.tsx
kendallstrautman: Another great example of rendering children here https://github.com/hashicorp/dev-portal/blob/main/src/components/cards-grid-list/index.tsx

src/components/content-header-card/content-header-card.module.css
```diff
@@ -0,0 +1,161 @@
+.contentHeaderCard {
+	padding: 0;
+
+	& .cardTop {
```
kendallstrautman, line 4: This could be a personal preference or i'm not sure we've officially discussed this, but for the most part in developer we tend to define component classes at the root level, with some exceptions. I find this makes the CSS easier to read and everything is already scoped to the component. I'd be curious to hear your take though on the benefits of nesting.

src/components/badge-list/index.tsx
BrandonRomano: I suppose the question I would ask is:

> Should it ever be possible to render a list of badges of different types?

Assuming the answer is "no, from a design perspective this should never be allowed":

By allowing the individual badges at this level to handle it, it is _technically possible_ to misuse the badge list component via a consumer.  My motivation for adding it at the list level makes it impossible to mix badge types.

It's a super light opinion here; but just wanted to share my take on the pros of lifting the badge style config props.

src/components/content-header-card/content-header-card.module.css
```diff
@@ -0,0 +1,161 @@
...
+		}
+	}
+
+	& .cardBottom {
+		padding: 12px 16px;
+		background: var(--token-color-surface-faint);
+
+		/* This is the color from --token-surface-base-box-shadow,
+		which the color value from it is not in a variable alone */
+		border-top: 1px solid rgba(101, 106, 118, 0.2);
```
kendallstrautman, line 103: I'd prefer to not deviate from the system here unless we have to. What's the context on not just using box-shadow here?

src/components/content-header-card/content-header-card.module.css
```diff
@@ -0,0 +1,161 @@
...
+		& .linksList {
+			list-style: none;
+			display: flex;
+			padding: 0;
+			margin: 0;
+			& li {
+				&:not(:first-of-type) {
+					margin-left: 12px;
+				}
+				& > * {
```
kendallstrautman, line 125: 💭 I'd rather eject from using `StandaloneLink` than do an override here, agreeing with your note.

src/components/content-header-card/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+	icon?: Exclude<ProductSlug, 'sentinel'>
+	title: string
+	attribution?: string
+	description?: string
+	note?: string
+	badges?: Array<Badge>
+	dropdown?: Dropdown
+	links?: Array<Link>
+	buttons?: Array<Button>
+}
```
kendallstrautman, line 49: 💭  One thought to extract these into a types file

src/components/content-header-card/content-header-card.module.css
BrandonRomano: Hah, I find the nesting easier to read because it follows the same hierarchical shape of the HTML.  🥴 I feel like there's definitely a preference aspect to this, but I do see the value in all aligning.

Materially, I like the flexibility of being able to write scoped logic (e.g.; I can have two `left` classes in my component and contextually style them based off of them being nested in a component).  Additionally I can write styles that are contextual to a class being in a specific context.

src/components/content-header-card/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+		<Card
+			elevation="base"
+			className={classNames(s.contentHeaderCard, className)}
+		>
+			<div className={s.cardTop}>
+				{icon && (
+					<IconTileLogo className={s.icon} size="large" productSlug={icon} />
+				)}
+				<div className={s.content}>
+					<div className={s.upper}>
```
kendallstrautman, line 73: I think it could be nice for this upper and lower sections to render a slot or component instead of being props driven, given the amount of conditional rendering. Or if we want to keep it opinionated, perhaps not elevating it as a global component yet until we have examples of other use-cases.

src/components/content-header-card/content-header-card.module.css
BrandonRomano: This is a box in a box (which already has box shadow), because it is a transparent box shadow the left, bottom, and right are .2 transparency + .2 transparency when using box shadow.

src/layouts/product-integration-layout/index.tsx
```diff
@@ -126,18 +127,102 @@ export default function ProductIntegrationLayout({
...
+				]}
 			/>
+
 			<h1>{title}</h1>
 			<div className={className}>{children}</div>
 		</SidebarSidecarLayout>
 	)
 }
+
+function lastUpdatedString(updatedAtString: string) {
```
kendallstrautman, line 174: Personal preference: This would be nice to extract in a utils file within the integrations layout dir.

src/components/content-header-card/content-header-card.module.css
BrandonRomano: To visualize their positioning.  Red overlaps orange.


CleanShot 2023-02-28 at 16 29 17@2x

src/lib/get-integration-badges.tsx
```diff
@@ -1,95 +1,49 @@
...
-			})}
-		</ul>
-	)
-}
+import { Badge } from 'components/badge-list'
+import { Flag, Integration, Tier } from './integrations-api-client/integration'
 
-export function GetIntegrationTags(
+// Converts an Integration object into an array of Badges to be used by our badge-list
+export function GetIntegrationBadges(
```
kendallstrautman, line 12: ⛏️ nitpick, I think this should be camelCase 
```suggestion
export function getIntegrationBadges(
```

src/layouts/product-integration-layout/index.tsx
BrandonRomano: Hah if we do that, I might want to harden this a bit more 😆 this gets the job done for now for integrations, but is currently probably not 100% accurate (e.g. 5 weeks ago is 1 months ago, regardless of how far into the current month you are).

src/lib/get-integration-badges.tsx
BrandonRomano: 👍 you caught my Go habit of capitalizing thing that are exported

BrandonRomano: (I will do this but need to adjust the imports too, so won't just accept the comment)

src/components/badge-list/badge-list.module.css
BrandonRomano: Actually looks like a bot will come through and add this in after the fact; I'll let that manage it!

braydenlove:
@BrandonRomano the @author typography's baseline isn't aligned. I mocked up a quick fix here but feel free to do what is the most scalable, just trying to give you an idea on spacing. 
image

BrandonRomano:
Oh good catch @braydenlove ; I thought this was to be center aligned so I did it that way.  I'll adjust to align on the baseline.

src/components/badge-list/index.tsx
BrandonRomano: Oh interesting!  This is a good example.

Given this, we probably should move the color to the badge itself then.

Do badge types ever get mixed?  I would imagine the answer is _no_ because they are different stylistic treatments that may clash next to each other, but this example makes me want to ask it out loud explicitly.

src/components/content-header-card/content-header-card.module.css
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	& .cardBottom {
+		padding: 12px 16px;
+		background: var(--token-color-surface-faint);
+
+		/* This is the color from --token-surface-base-box-shadow,
+		which the color value from it is not in a variable alone */
+		border-top: 1px solid rgba(101, 106, 118, 0.2);
+		border-radius: 0 0 6px 6px;
```
ashleemboyer, line 104: 🙋🏻 Are we able to use `overflow: hidden` on the parent element of this one, rather than set the border-radius with a hardcoded value? I think that would enable us to inherit the `border-radius` setting directly from `Card`.

src/components/content-header-card/index.tsx
ashleemboyer: +1 -- I think these should be moved to align with how other components store their types.

src/components/content-header-card/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+											}
+											icon={button.icon}
+											iconPosition="leading"
+											size="small"
+										/>
+									</li>
+								)
+							})}
+						</ul>
+					)}
```
ashleemboyer, line 179: 🙋🏻 Should this support `ButtonLink`s as well? My understanding is Yes based on the conversation had in [the design review doc](https://docs.google.com/document/d/1kqpW1jIJF-h36zrrVPB3AFEf93FYz4wo5rHc0pyxGJ4/edit), relevant to the `Install` element that's shown in designs.

src/components/content-header-card/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+											onClick={(e) => {
+												e.preventDefault()
+												button.onClick()
+											}}
+											text={button.text}
+											color={
+												button.isPrimary && index === 0
+													? 'primary'
+													: 'secondary'
+											}
```
ashleemboyer, line 170: 🙋🏻 Do we plan to put checks in place so that only primary button can be rendered in this list?

src/layouts/product-integration-layout/index.tsx
```diff
@@ -126,18 +127,102 @@ export default function ProductIntegrationLayout({
...
+									}),
+						  }
+						: undefined
+				}
+				links={[
+					{
+						text: 'GitHub',
+						href: integration.subdirectory
+							? `${integration.repo_url}/tree/main${integration.subdirectory}`
+							: integration.repo_url,
```
ashleemboyer, line 162: 🙋🏻 📌 Maybe out of scope for this PR -- how do we know this is the subdirectory URL? It assumes `main` is the branch, but what if the repo the integration comes from doesn't have a `main` branch?

src/layouts/product-integration-layout/index.tsx
```diff
@@ -126,18 +127,102 @@ export default function ProductIntegrationLayout({
...
+						icon: <IconGithub16 />,
+					},
+				]}
 			/>
+
 			<h1>{title}</h1>
```
ashleemboyer, line 168: 🔴 Just noticing this (probably out of scope for this PR) -- This ends up always rendering a second `h1` on the page, which we don't want to do. Can you create a ticket for resolving this? I'm not sure what board to place it on.

src/layouts/product-integration-layout/index.tsx
ashleemboyer: I'd like to see this moved too -- not to `src/lib`, but to a new folder in `src/layouts/product-integration-layout/`. This file is getting pretty long, so it'd help reduce the overwhelm factor when opening this file. 😊

ashleemboyer:
Ugh I forgot to submit a main comment with my others! 🙏🏻 

I left quite a few, so let me know if you'd like to pair on the feedback I left! Also open to chatting via Slack since that'd be a little faster.

src/components/content-header-card/content-header-card.module.css
BrandonRomano: ahhah, thank you! This is much more preferable than what I have.

src/components/badge-list/index.tsx
kendallstrautman: In the interface, the props are defined as optional so they could be `undefined` and then you would be using the badge default anyways. I think we should do something like this so that one could drive with the badge data as an option. 

```suggestion
size={badge.size || size}
```

src/components/badge-list/index.tsx
kendallstrautman: I think we should leave it open ended since the colors are already mixed, I could see a world where the types are mixed. Agree that size is a stretch but hard to anticipate. But as long as we support the [badge data as a default ](https://github.com/hashicorp/dev-portal/pull/1687#discussion_r1120929948), I think it's fine!

src/components/content-header-card/content-header-card.module.css
kendallstrautman: Resurfacing what we chatted about via slack. The expectation is that classes are defined at the root of the file unless the nesting is actually needed, like in the case of the flex positioning. If you have multiple classes named `left`, my suggestion is to adjust the names to be more descriptive.

src/components/badge-list/index.tsx
BrandonRomano: Ah! I see!  That makes it more clear thank you — yeah this should be added!

src/layouts/product-integration-layout/index.tsx
kendallstrautman: Yep, exactly @ashleemboyer I'm not suggesting in the global utils folder but often we will make a local utils file: `src/layouts/product-integration-layout/utils.ts`

src/components/content-header-card/content-header-card.module.css
```diff
@@ -0,0 +1,174 @@
...
+					/* TODO: Rather than override our existing link component, 
+					we should eject or build in support for what we're trying
+					to do here. These styles assume the StandaloneLink styles
+					are part of its contract, but they are not and are subject
+					to change */
+					text-decoration: underline;
+					color: var(--token-color-foreground-primary);
+					&:hover {
+						color: var(--token-color-foreground-strong);
+					}
```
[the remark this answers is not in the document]
kendallstrautman: 💡 Should we adjust standalone link to support this style variation?

src/components/content-header-card/content-header-card.module.css
kendallstrautman: Fwiw this isn't necessarily blocking feedback, if you don't want to refactor this whole file. But I do think we should get aligned for future work.

src/components/badge-list/index.tsx
BrandonRomano: Oh!  I handle this in the destructuring; so this will never be undefined:

https://github.com/hashicorp/dev-portal/pull/1687/files#diff-ddce842efc461cbfc09ed3f97cd1630552d470994031f2425f72c1387e617781R24-R25

src/components/content-header-card/content-header-card.module.css
BrandonRomano: I'm going to defer refactoring this for now; and will take this as feedback for future components I will avoid doing this to align with the team's approach!

src/components/content-header-card/content-header-card.module.css
```diff
@@ -0,0 +1,174 @@
...
+			margin-right: 12px;
+			display: none;
+
+			@media (--dev-dot-tablet-up) {
+				display: block;
+			}
+		}
+
+		& .content {
+			flex-grow: 1;
```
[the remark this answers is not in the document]
BrandonRomano: Without it!

CleanShot 2023-03-03 at 14 48 39@2x

src/components/content-header-card/content-header-card.module.css
```diff
@@ -0,0 +1,174 @@
...
+
+			& .upper {
+				display: flex;
+				gap: 16px;
+				flex-direction: column-reverse;
```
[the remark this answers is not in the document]
BrandonRomano: These are reversed on desktop. Generally, CTA button alignment are aligned in the direction of button list alignment (relative to its container).  e.g. a right aligned button list would have its primary button to the right.

CleanShot 2023-03-03 at 14 49 57@2x

CleanShot 2023-03-03 at 14 50 03@2x

I believe that we additionally would want the tabbing order to match this, as the primary button should go first. The text box of this comment actually illustrates the same behavior:

https://user-images.githubusercontent.com/2105067/222849715-26ef023f-66f3-46db-9008-e447f38222fb.mp4

src/components/content-header-card/content-header-card.module.css
BrandonRomano: Ah, that is a good idea!

src/components/content-header-card/content-header-card.module.css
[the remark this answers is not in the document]
BrandonRomano: OH WHOMP, I added this comment in the wrong place; I will delete & move it.

src/components/content-header-card/content-header-card.module.css
```diff
@@ -0,0 +1,174 @@
...
+
+		& .buttonsList {
+			padding: 0;
+			margin: 0;
+			list-style: none;
+			display: flex;
+			flex-grow: 1;
+			flex-direction: row;
+			@media (--dev-dot-tablet-up) {
+				flex-direction: row-reverse;
```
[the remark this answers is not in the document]
BrandonRomano: These are reversed on desktop. Generally, CTA button alignment are aligned in the direction of button list alignment (relative to its container).  e.g. a right aligned button list would have its primary button to the right.

CleanShot 2023-03-03 at 14 49 57@2x

CleanShot 2023-03-03 at 14 50 03@2x

I believe that we additionally would want the tabbing order to match this, as the primary button should go first. The text box of this comment actually illustrates the same behavior:

https://user-images.githubusercontent.com/2105067/222849715-26ef023f-66f3-46db-9008-e447f38222fb.mp4

src/components/content-header-card/content-header-card.module.css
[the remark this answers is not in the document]
BrandonRomano: The dropdown goes above the integration title/description on mobile!
 
CleanShot 2023-03-03 at 15 07 03@2x

src/components/content-header-card/content-header-card.module.css
BrandonRomano: Going to resolve this, as it won't work 😢

src/components/content-header-card/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+							className={classNames(s.buttonsList, {
+								[s.buttonsOnly]: !hasLinks,
+							})}
+						>
+							{buttons.map((button: Button, index: number) => {
+								return (
+									<li key={button.text}>
+										<Button
+											onClick={(e) => {
+												e.preventDefault()
```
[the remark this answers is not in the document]
BrandonRomano: Nope!  I don't think it's needed; there is no default behavior from this button so it can be removed 👍

src/components/content-header-card/index.tsx
BrandonRomano: I think that you mean:

>  Do we plan to put checks in place so that only **one** primary button can be rendered in this list?

If that's what you intended: I verify that the `index === 0` above.  This allows the user to opt-in to the first button being blue (but not allowing others to opt-in).

src/components/content-header-card/index.tsx
BrandonRomano: So the answer is *yes*, but I'm not adding them in right now, if that's OK!

I think that my plan will be to allow a `onClick` or `url` but not both to be set.  From there we can infer what button type we render.

src/layouts/product-integration-layout/index.tsx
BrandonRomano: This is a great callout!  I will have to think about this.

It's not quite a problem _yet_ but this will be something we have to address.  I'm going to file a ticket.

BrandonRomano: 🎟️ [Edge case: Link to an integration not on the `main` branch](https://app.asana.com/0/1203792616809419/1204110003219358)

src/layouts/product-integration-layout/index.tsx
BrandonRomano: I have a ticket for this one already:

[Remove potential Duplicative H1 from README](https://app.asana.com/0/1203792616809419/1203295231021660)

It has been bugging me for a little bit as well.  I've bumped up the priority on this one, thanks for the callout here.

src/components/content-header-card/content-header-card.module.css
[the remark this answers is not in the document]
ashleemboyer: Ah, got it. Thanks for the context!

We'll need to use a different approach than `flex-direction` to handle this. As-is, the visual order of the content does not match the DOM order.

https://user-images.githubusercontent.com/43934258/223171287-36a9313d-74fa-4119-ac79-02181c52e4f2.mov

src/components/content-header-card/content-header-card.module.css
ashleemboyer: Bummer! Was worth a try 😁

src/components/content-header-card/content-header-card.module.css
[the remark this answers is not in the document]
ashleemboyer: The [Figma designs](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=20364%3A254758&t=7oII6fQE1NXSOf75-11) I'm looking at don't show the button order changing on narrower viewports.

image

The visual order of these elements must also match the DOM order, and on wider viewports they don't match. This needs to be fixed before merging this PR.

https://user-images.githubusercontent.com/43934258/223172986-30e955e4-fd70-49e4-9814-36c83cf1da13.mov

src/components/content-header-card/content-header-card.module.css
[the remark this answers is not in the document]
ashleemboyer: I don't think we need to add an additional variant in this case where just `text-decoration` and `color` are overridden. In both approaches we'd still be using `StandaloneLink`, so they have equal risk in being unintentionally affected by changes to the component. Since `StandaloneLink` is widely used, changes to it would need to be well-tested across references anyways, so using the component here doesn't increase maintenance effort from my perspective.

src/components/content-header-card/index.tsx
ashleemboyer: Whoops, I did mean _one_ primary button! Thanks for clarifying, my eyes overlooked that `index === 0`. 😆

src/components/badge-list/index.tsx
ashleemboyer: @BrandonRomano have you given this feedback further thought?

ashleemboyer (changes requested):
The 2 visual representation vs. DOM order issues need resolved, and then I think this is good on my part!

src/components/badge-list/index.tsx
BrandonRomano: For this PR though; I don’t think I’m going to change it, although I’ve thought about the suggestion!

There are a few things that I want to codify as rules of this list component onto badges
- I want to be able to codify `size` and `type` always being the same at the list level.
- I want a badge to be able to specify its `tooltip`; this is something the consumer would have to wrap in a composition based approach, whereas with a props approach we can hide the internals that behavior.

That being said, I broke out a thread in slack to talk about this pattern from a higher level. Would love to hear your thoughts in there: https://hashicorp.slack.com/archives/C02T80JTNS2/p1677687494767269

src/components/content-header-card/content-header-card.module.css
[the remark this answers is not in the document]
BrandonRomano: I had a conversation with @braydenlove to verify this behavior via DM.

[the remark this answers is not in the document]
ashleemboyer: For the visual order changing -- that sounds good and glad y'all are aligned! We'll just need to make the DOM order match the visual order. It falls under [SC 1.3.2: Meaningful Sequence](https://www.w3.org/WAI/WCAG21/Understanding/meaningful-sequence) which is a Level A SC.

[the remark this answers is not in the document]
braydenlove: Sorry I haven't had a chance to update the designs after we pairing on this @BrandonRomano. After chatting we decided on keeping the primary button on the side the group is floating (if button group is floating right primary is displayed on the right, if button group is floating left primary is displayed on the left). Here are a few scenarios: https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=20364%3A254758&t=KRTf0BF6CQAWZNIQ-1

src/components/content-header-card/content-header-card.module.css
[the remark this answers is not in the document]
BrandonRomano: @ashleemboyer do you have any suggestions on how to accomplish this without flex-direction?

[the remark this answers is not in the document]
ashleemboyer: We can render it twice and use our `g-hide-on-(mobile|tablet|desktop)` helpers on each!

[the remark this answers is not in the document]
BrandonRomano: Ah nice, that is a better idea that what I had in mind (I was thinking we had to do some Javascript logic aware of the width of the screen).

I'll work this in at some point this week!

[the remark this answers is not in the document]
BrandonRomano: Also I'll keep in mind in the future whenever I encounter a `row|column-reverse` situation that it will cause a dom-ordering issue.

src/components/content-header-card/content-header-card.module.css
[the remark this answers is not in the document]
zchsh: When we've taken the "override" approach in the past, it's bitten us pretty badly down the road. Our transition to CSS modules was motivated in part by feeling that components were brittle -- we could no longer change the markup or styling with an acceptable degree of confidence (in a reasonable amount of time) because of having to account for a slew of widely varying "overrides" that had been layered on top of each component.

I'd agree that for a relatively atomic component like `StandaloneLink` this may be less of an issue. I also do agree though that we should avoid these overrides regardless - personally I think the `TODO` here feels accurate, we should either eject or build in support for this theme option.

Stepping back a bit, I am kind of curious as to the design system logic here - it looks like the main change here is we show an underline by default (the color difference compared to our `secondary` version of `StandaloneLink` seems subtle). Is that possibly a divergence from the [HDS Standalone Link](https://helios.hashicorp.design/components/link/standalone)? If so I think it might be worth checking with design on why that change feels necessary.

src/components/content-header-card/content-header-card.module.css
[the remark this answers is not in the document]
zchsh: Would the definition of "meaningful sequence" in [SC 1.3.2: Meaningful Sequence](https://www.w3.org/WAI/WCAG21/Understanding/meaningful-sequence) come into play here?

> A sequence is meaningful if the order of content in the sequence cannot be changed without affecting its meaning.

It seems possible we can change the order of these action buttons without affecting the meaning of the content on the page. As an example, I think this is why the GitHub commenting buttons example that Brandon mentioned are likely acceptable:

CleanShot 2023-03-06 at 16 19 43

It seems like the visual and DOM order there doesn't technically match - I'm reading in a left-to-right language, and visually the buttons appear `Cancel` and then `Comment` left-to-right, but the focus order is `Comment` then `Cancel`. But, I think this is okay because the visual order isn't affecting the meaning. And, I think there is a case to be made that with the way we scan content, having the primary button at the "edge" of the container (but still first in DOM order!) is likely the most all-around intuitive solution.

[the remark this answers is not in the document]
ashleemboyer: I do think we imply some meaning here with 2 visual specifics:

- the right-alignment of the actions group
- the rightmost action having a color that stands out from the others

Something important to consider with the ordering of actions like this whether or not users are aware of all their options.

- Sighted users that don't rely on assistive technology or things like magnification are able to visually scan a page and see all of their options at once.
- Many screen reader users don't have the ability to view a page at-a-glance. They experience a page linearly. So if we place a primary action first in a group of buttons, they may not know what all of their options are.
- Users that can see the focus indicator move as they navigate a page, whether with assistive technology or not, can easily become disoriented, confused, distracted, etc. if the page jumps around in a way they don't expect.

The Understanding docs for [Technique C27: Making the DOM order match the visual order](https://www.w3.org/WAI/WCAG21/Techniques/css/C27) get into that last point more.

> The objective of this technique is to ensure that the order of content in the source code is the same as the visual presentation of the content. The order of content in the source code can be changed by the author to any number of visual presentations with CSS. Each order may be meaningful in itself but may cause confusion for assistive technology users. This could be due to the user switching off the author-specified presentation, by accessing the content directly from the source code (such as with a screen reader), or by interacting with the content with a keyboard. If a blind user, who reads the page with a screen reader that follows the source order, is working with a sighted user who reads the page in visual order, they may be confused when they encounter information in different orders. A user with low vision who uses a screen magnifier in combination with a screen reader may be confused when the reading order appears to skip around on the screen. A keyboard user may have trouble predicting where focus will go next when the source order does not match the visual order.
>
> There may also be situations where the visually presented order is necessary to the overall understanding of the page, and if the source order is presented differently, it may be much more difficult to understand.
>
> When the source order matches the visual order, everyone will read the content and interact with it in the same (correct) order.

BrandonRomano:
Okay @ashleemboyer I adjusted the dom ordering for the button (via https://github.com/hashicorp/dev-portal/pull/1687/commits/497714e93362d78436a0db36c39a694b655702e7) and the version selector (via https://github.com/hashicorp/dev-portal/pull/1687/commits/31fa65e6d2a3406317e25009ce3101bc2e45c62b).

Ready for re-review!

ashleemboyer (approved):

braydenlove:
Looks good @BrandonRomano, nice work.

PR #1702, opened 2023-03-01
Update worker filter link

ZedLi:
## 🔗 Relevant links
[JIRA Issue](https://hashicorp.atlassian.net/browse/ICU-8192)

This is to update one of our links to a new url.


kendallstrautman (approved):

proxied-redirects/www.boundaryproject.io.redirects.js
```diff
@@ -161,7 +161,7 @@ module.exports = [
 	},
 	{
 		source: '/help/admin-ui/worker-filters',
-		destination: '/docs/concepts/filtering/worker-tags',
```
kendallstrautman, line 164: Just making sure you don't want to add a new redirect and instead you want to override this legacy one?

ZedLi: Yup we want to point to this new redirect since the old one won't be used anymore

randallmorey (approved):

randallmorey (approved):

PR #1724, opened 2023-03-08
fix: remove ent release lockstep assumptions

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Updates Enterprise download page logic to more directly fetch the latest `+ent` version number, rather than assuming that the latest version number with an `+ent` string suffix added exists.

## 🤷 Why

To try to improve the reliability of enterprise download pages, so that we don't have build failures like [this one](https://vercel.com/hashicorp/dev-portal/5gngCQ948tHUby7mPLSVgSgZssyh).

## 🛠️ How

- Add a `getIsEnterpriseVersion` helper that uses `semverParse` to determine version data, namely identifying enterprise releases by their `"ent"` build ID.
- Updates `fetch-release-data` and `product-downloads-view/server` to directly fetch the `latestVersion` for enterprise pages, removing the previous assumption that `${latestVersion}+ent` exists.
- Fixes and adjacent issue on Enterprise downloads pages, where due to quirks in enterprise version formatting and pre-release version detection, some pre-releases were still listed in our version dropdown

## 🧪 Testing

- [ ] Visit a standard downloads page, such as [/consul/downloads] (vs [upstream][upstream/consul/downloads])
- [ ] Visit an enterprise downloads page, such as [/consul/downloads/enterprise] (vs [upstream][upstream/consul/downloads/enterprise])

On both the above pages:

- The page should look & behave as they do `upstream`, except...
- The version dropdown on the Enterprise page should now exclude all pre-releases

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/1100423001970639/1204055733223571/f
[preview]: https://dev-portal-git-zsrefine-ent-download-check-hashicorp.vercel.app/
[/consul/downloads]: https://dev-portal-git-zsrefine-ent-download-check-hashicorp.vercel.app/consul/downloads
[/consul/downloads/enterprise]: https://dev-portal-git-zsrefine-ent-download-check-hashicorp.vercel.app/consul/downloads/enterprise
[upstream/consul/downloads]: https://developer.hashicorp.com/consul/downloads
[upstream/consul/downloads/enterprise]: https://developer.hashicorp.com/consul/downloads/enterprise


src/views/product-downloads-view/helpers.ts
```diff
@@ -261,6 +262,50 @@ export function prettyOs(os: string): string {
...
+ * could be brittle if we ever want a build ID that does contain dashes.
+ *
+ * A more robust path forward would be to fix the versioning format at the
+ * source to match the semver spec, which specifies that the build ID should
+ * always be after the pre-release ID.
+ *
+ * Spec: https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions
+ * List of Consul releases: https://releases.hashicorp.com/consul/index.json
+ */
+export function getIsEnterpriseVersion(version: string): boolean | null {
```
zchsh, line 300: I've replaced the previous regex-based approach with `semverParse`. There are some quirks here I thought might be worth noting, that are likely related to the fix in this PR for ensuring pre-releases don't show up in the enterprise downloads page dropdowns.

| Before | After |
| - | - |
| before-ent-pre-releases-shown | after-ent-pre-releases-not-shown |

src/views/product-downloads-view/helpers.ts
```diff
@@ -277,19 +322,22 @@ export const sortAndFilterReleaseVersions = ({
...
+			/**
+			 * Filter in enterprise versions if enterprise mode (build === "ent"),
+			 * or filter out any custom "build" values otherwise.
+			 */
 			if (isEnterpriseMode) {
+				const isEnterpriseVersion = getIsEnterpriseVersion(version)
 				return isEnterpriseVersion
+			} else {
+				const { build } = semverParse(version)
+				return build.length === 0
```
zchsh, line 339: Whoops, just noticed this as well - the regular downloads page in earlier previews on this PR showed some pre-release-ish enterprise versions, since eg `1.11.0+ent-rc` is identified as having `build: [ "ent-rc" ]` and `prerelease: []`, which means it's _not_ identified as a prerelease _or_ as an enterprise version. This updated logic ensures _any_ version string with a build value (eg `x.y.z+build`) will get filtered out of our dropdown.

src/views/product-downloads-view/helpers.ts
kendallstrautman: wow, nice fix!

kendallstrautman (approved):
Followed the testing steps. Everything looks good from my end! Thank you for this! 🙇🏻

BRKalow (approved):
👏

PR #1726, opened 2023-03-08
style: cleanup open-api-docs typography

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksapi-docs-accordion-hashicorp.vercel.app/boundary/api-docs/group-service) 🔎
- [Asana task](https://app.asana.com/0/1199634971449915/1204129860249685) 🎟️
- [Figma](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=12849%3A162751&t=LlSZjMQrZ5lFg6wj-0)

## 🗒️ What

- Updates typography to[ match this spec](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=12849%3A162751&t=LlSZjMQrZ5lFg6wj-0).
- Removes old mktg-global-styles token references. 

### Dark theme screenshot from local dev
Screenshot 2023-03-14 at 1 11 56 PM

## 📸 Design Screenshots

Screenshot 2023-03-14 at 1 10 12 PM

## 🧪 Testing

- Visit [Preview link](https://dev-portal-git-ksapi-docs-accordion-hashicorp.vercel.app/boundary/api-docs), [hcp packer](https://dev-portal-git-ksapi-docs-accordion-hashicorp.vercel.app/hcp/api-docs/packer), and [waypoint api docs,](https://dev-portal-git-ksapi-docs-accordion-hashicorp.vercel.app/waypoint/api-docs) validate that the type colors and styles match the spec for both dark and light mode.
- Ensure the open api page disclosure [accordion on swingset](https://dev-portal-git-ksapi-docs-accordion-hashicorp.vercel.app/swingset/components/openapipage) isn't affected by the new prop for light mode.

## 💭 Anything else?

We will [follow up](https://app.asana.com/0/1199634971449915/1204167616054159) with an override for the accordion background in dark mode to be surface-faint to match the card override.


braydenlove:
@kendallstrautman you did a really nice job on all the fixes here. One quick thing I am considering is trying to avoid the `.card_root` bg color from sneaking in. This thing is really hard to scan and read when it is flipping colors. I know we want to avoid overrides based on context, just trying to think of a good option here. The color feels like it should denote something intentionally but doesn't, ideally one color for all the bgs here like it is on light mode. Happy to figure out how to best solve this.
api-pr

kendallstrautman:
@braydenlove that makes sense. So would you prefer all black or the card color?

braydenlove:
> @braydenlove that makes sense. So would you prefer all black or all the same card color (surface-faint)?

I would prefer (surface-faint) over the darker value!

kendallstrautman:
@braydenlove do you want all accordions to have this surface faint background color in darkmode? Or just keep it on this view for now?

braydenlove:
> @braydenlove do you want all accordions to have this surface faint background color in darkmode? Or just keep it on this view for now?

Let's do all of them to keep consistent with cards (thing that has that color and elevation will do something when I click it). The only place I see it getting funky is on the certifications page but we will figure that out asap.

kendallstrautman:
@braydenlove okay I'd like to do the global accordion update as a separate PR, if that's cool. To contain reviewing. I'll make a new ticket with a note to update these overrides when we make that change.

braydenlove:
Sounds good I will approve this smaller one to keep us moving.

braydenlove (approved):

src/components/open-api-page/partials/operation-object/style.module.css
```diff
@@ -52,7 +57,8 @@
 
 .summary {
 	/* Base typography */
-	composes: g-type-body from global;
+	composes: g-type-body-200 from global;
```
zchsh, line 60: ⛏️ Nit, maybe out of scope - should we update this to like `hds-typography-body-300` or similar? I think it'd have minimal to no difference, but would help root out "mktg" classes (which we might be able to remove once `dot-io` sites are in `hashicorp/web`, I think!)

```suggestion
	composes: hds-typography-body-300 from global;
```

| Type class | Preview |
| - | - |
| `g-type-body-200` | g-type-body-200 |
| `hds-typography-body-300` | hds-typography-body-300 |

zchsh (approved):
🙌 Looks good!

Had one nitpick on a stray `g-type-body` class, but very optional - probably something that could make sense to clean up post-dot-io migration, but on my mind so couldn't help but notice it 😅

PR #1737, opened 2023-03-14
ubuntu: use gpg -o instead of tee for getting gpg key

shoenig:
Second attempt at https://github.com/hashicorp/dev-portal/pull/797

This time using `gpg -o` as suggested by @phinze 

Attaching a screenshot of what happens when users follow the instructiosn as-is. Your terminal turns into an unusable mess. It's an awful UX.

Screenshot from 2023-03-14 11-19-27

### DIFF
--- a/src/views/product-downloads-view/helpers.ts
+++ b/src/views/product-downloads-view/helpers.ts
@@ -43,7 +43,7 @@ export const generateDefaultPackageManagers = (
 		{
 			label: 'Ubuntu/Debian',
 			commands: [
-				`wget -O- https://apt.releases.hashicorp.com/gpg | gpg --dearmor | sudo tee /usr/share/keyrings/hashicorp-archive-keyring.gpg`,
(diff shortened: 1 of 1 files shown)


ashleemboyer:
Hey @shoenig! 😊 These will affect the `/${productSlug}/downloads` pages (e.g. `/waypoint/downloads`, `/vault/downloads`), and not the page shown in your screenshot. I think the page in your screenshot is a Nomad tutorial, and that content would not be affected by your changes here. Are you sure these cross-product changes are what you're looking for?

phinze:
Hello! I can confirm that the usage of `tee` dumps binary output onto the terminal in every place it is used, so this would be a UX improvement everywhere.

Here's an example of a recent [Vagrant bug report](https://github.com/hashicorp/vagrant/issues/13101) from a user who hit this.

zchsh (approved):

PR #1744, opened 2023-03-15
chore: refactor semantic tokens

kendallstrautman:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1199634971449915/1204167618837036) 🎟️

## 🗒️ What

Refactoring the semantic tokens to reference the core palette tokens. 

## 🤷 Why

It's been difficult to track the propagation of token value updates when we were using hex values across the board. This refactor allows us to more easily see how the core palette is being used within semantic tokens. And it also helps us weed out and prevent unintentional random hex tokens that are outside the core palette. 

## 🧪 Testing

- There should be no visible changes, this is a pure refactor. To fully validate, [visit the file](https://github.com/hashicorp/dev-portal/blob/42e7dddf70508f55dd731ad7cda44802ae988a68/src/styles/themes/dark.css) and ensure that the new core palette token references match the hex value of that core palette token. Responsibility to test on this PR leans heavily on design @braydenlove

### DIFF
--- a/src/styles/themes/dark.css
+++ b/src/styles/themes/dark.css
@@ -60,7 +60,7 @@
 	/* focus */
 	--token-color-focus-action-internal: var(--token-color-palette-blue-300);
 	--token-color-focus-action-external: #5990ff; /* this is a special color used only for the focus style, to pass color contrast for a11y purpose */
-	--token-color-focus-critical-internal: #c00005;
+	--token-color-focus-critical-internal: var(--token-color-palette-red-300);
 	--token-color-focus-critical-external: #dd7578; /* this is a special color used only for the focus style, to pass color contrast for a11y purpose */
 
 	/* foreground */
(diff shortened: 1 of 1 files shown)


kendallstrautman:
@braydenlove this is a 1-1 swap of the current state, without those purple and form updates. I'd like to get this change in before those tickets so I can update those other values to use the core palette tokens.

braydenlove:
Yours:
`--token-color-focus-critical-external: #dd7578; `

Semantic:
`--token-color-focus-critical-external: var(--token-color-palette-red-300);`

kendallstrautman:
☝🏻 we met on zoom and that comment is actually speaking to the `critical-internal`

braydenlove (approved):
Double checked the product colors and it looks like we are all set! Thanks 🧛🏼‍♂️

zchsh (approved):
✅ I can't say I individually cross-checked every value... but this looks good to me based on the preview and glancing through the `palette` and update `reference the palette` tokens in side-by-side windows 👍 If an more thorough check would be helpful, let me know! 🙇

PR #1749, opened 2023-03-16
Add enable_home_page_redesign flag

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambadd-home-page-flag-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1201147515801429/1204132608541457/f) 🎟️

## 🗒️ What

- Adds a new `enable_home_page_redesign` flag
- Conditionally renders content in `HomePageView` using the new flag

## 🤷 Why

- This feature flag will make it easy to iteratively work on the newly redesigned home page without having to use a long lived feature branch

## 🧪 Testing

- [ ] Go to [the home page](https://dev-portal-git-ambadd-home-page-flag-hashicorp.vercel.app/) on the preview
- [ ] The page content area should just include the text: "I am the new home page!"

### DIFF
--- a/config/base.json
+++ b/config/base.json
@@ -45,6 +45,7 @@
 		"enable_io_beta_cta": true,
 		"enable_product_docs_search": true,
 		"enable_global_search": true,
-		"enable_dark_mode": true
+		"enable_dark_mode": true,
+		"enable_home_page_redesign": false
 	}
 }
--- a/config/development.json
+++ b/config/development.json
@@ -13,6 +13,7 @@
 		"max_static_paths": 1
 	},
 	"flags": {
-		"enable_global_search": true
+		"enable_global_search": true,
+		"enable_home_page_redesign": true
 	}
 }
--- a/config/preview.json
+++ b/config/preview.json
@@ -4,6 +4,7 @@
 		"max_static_paths": 10
 	},
 	"flags": {
-		"enable_global_search": true
+		"enable_global_search": true,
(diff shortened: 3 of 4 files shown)


kendallstrautman (approved):
LGTM. 

💭 I'm curious to chat through what types of projects are a good match for the feature flag. Since this project is scoped to a single view, I would think it would align more with the assembly branch approach. In this case it doesn't matter much that the previews render an empty view here for the short term, but if it were a more high traffic view, we would need to consider overlap with dark mode, outside contributors, authors etc not being able to see it. In this instance, I guess its just a matter of preference, if it feels easier to rip out the conditional rendering / flag stuff later on. But I'm curious to stoke a conversation for the team as well and get your take!

ashleemboyer:
> LGTM.
> 
> 💭 I'm curious to chat through what types of projects are a good match for the feature flag. Since this project is scoped to a single view, I would think it would align more with the assembly branch approach. In this case it doesn't matter much that the previews render an empty view here for the short term, but if it were a more high traffic view, we would need to consider overlap with dark mode, outside contributors, authors etc not being able to see it. In this instance, I guess its just a matter of preference, if it feels easier to rip out the conditional rendering / flag stuff later on. But I'm curious to stoke a conversation for the team as well and get your take!

@kendallstrautman happy to chat about it! I'm going to start a doc for the team to discuss and document. 😊

PR #1752, opened 2023-03-16
feat: adds `themedImage`, updates product landing graphics

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksupdate-product-landing-graphics-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1203652232454021/1204107075307470) 🎟️
- [Figma](https://www.figma.com/file/v0Hyqvjdf6QJAvRkdIBV24/Dark-Mode?node-id=371-31825&t=nMLcJJqHSvHCpVYT-0) 🎨 

## 🗒️ What

This PR adds a new `ThemedImage` component and utilizes it on the `overviewCta` for all product landing pages. I also plugs in the updated assets. 

Screenshot 2023-03-16 at 1 32 39 PM

Screenshot 2023-03-17 at 5 01 53 PM

## 🛠️ How

[See the RFC](https://docs.google.com/document/d/10im_t58EFNF2VWbE5SUQGjq_IfWLFy2ERJkm9ZCOLE4/edit#heading=h.8s2t1fscdyxc) for full thoughts behind the approach. 

## 🧪 Testing

Visit all product landings and toggle between dark / light to observe the different assets. Verify that there is no flashing when switching states. 

- [boundary](https://dev-portal-git-ksupdate-product-landing-graphics-hashicorp.vercel.app/boundary)
- [consul](https://dev-portal-git-ksupdate-product-landing-graphics-hashicorp.vercel.app/consul)
- [nomad](https://dev-portal-git-ksupdate-product-landing-graphics-hashicorp.vercel.app/nomad)
- [packer](https://dev-portal-git-ksupdate-product-landing-graphics-hashicorp.vercel.app/packer)
- [terraform](https://dev-portal-git-ksupdate-product-landing-graphics-hashicorp.vercel.app/terraform)
- [vagrant](https://dev-portal-git-ksupdate-product-landing-graphics-hashicorp.vercel.app/vagrant)
- [waypoint](https://dev-portal-git-ksupdate-product-landing-graphics-hashicorp.vercel.app/waypoint)
- [vault](https://dev-portal-git-ksupdate-product-landing-graphics-hashicorp.vercel.app/vault)
- [well-architected-framework](https://dev-portal-git-ksupdate-product-landing-graphics-hashicorp.vercel.app/well-architected-framework)
- [hcp](https://dev-portal-git-ksupdate-product-landing-graphics-hashicorp.vercel.app/hcp)


src/styles/themes/global.css
```diff
@@ -0,0 +1,8 @@
+/** 
+* Currently used to hide images based on theme
+* see components/themed-image
+ */
```
kendallstrautman, line 4: This is a new file where we can place global theme related config. I didn't want to put it in the 'dark' theme css file since this should be independent of the theme state.

src/components/themed-image/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * This component can be used in views where we need to
+ * conditionally render an image based on the theme state.
+ *
+ * See styles/themes/global.css for the attribute handler
+ * that uses `display:none` to conditionally render content
+ * based on the theme state with css
+ */
+export default function ThemedImage({ src, alt }: ThemedImageProps) {
```
kendallstrautman, line 11: Feels pretty simple but works well in the browser. Open to feedback!

kendallstrautman:
We need to look into the aspect ratio or padding around these assets, as they currently are much smaller than whats on production.

src/components/themed-image/index.tsx
kendallstrautman: Ideally we would use next image here but id be nervous about making that work for all contexts. I could also localize this to the product landing for now until we have another use case.

src/components/themed-image/index.tsx
```diff
@@ -0,0 +1,31 @@
+import { ThemedImageProps } from './types'
+
+/**
+ * This component can be used in views where we need to
+ * conditionally render an image based on the theme state.
+ *
+ * See styles/themes/global.css for the attribute handler
+ * that uses `display:none` to conditionally render content
+ * based on the theme state with css
+ */
```
kendallstrautman, line 10: Should pass other image props here, classname etc.

src/components/themed-image/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+	/**
+	 * @TODO could add protection here to throw
+	 * when used outside of the theme context. Since the
+	 * entire app is wrapped with the provier when using the
+	 * core devdot layout, it's currently low-risk
+	 * */
+	return (
+		<>
+			{/* When the theme is dark, hide this div */}
+			<div data-hide-on-theme="dark">
```
kendallstrautman, line 21: Should use `GlobalThemeOption` here

kendallstrautman: Also this only allows two themes....should we do a 'show-on-theme' instead?

kendallstrautman: Maybe I don't want the wrapper div 💭 

also this could use Image and accept next/image props

src/components/themed-image/index.tsx
kendallstrautman: I ended up localizing this to the product landing dir for now since this is the only known usecase and I can't anticipate all of the global component's interface for this initial PR.

src/views/product-landing/components/themed-image/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * This component can be used in views where we need to
+ * conditionally render an image based on the theme state.
+ *
+ * See styles/themes/global.css for the attribute handler
+ * that uses `display:none` to conditionally render content
+ * based on the theme state with css
+ */
+export default function ThemedImage({ src, alt }: ThemedImageProps) {
```
kendallstrautman, line 11: [See this comment](https://github.com/hashicorp/dev-portal/pull/1752#discussion_r1140836074), I decided to keep this component local for simplicity on this pass and also I don't know enough about the generic needs of this component yet to lift it up globally. Also this is the only known usecase at the moment.

src/content/boundary/product-landing.json
```diff
@@ -10,7 +10,10 @@
 			"text": "Learn more about Boundary features",
 			"url": "/boundary/docs/overview/what-is-boundary"
 		},
-		"image": "https://www.datocms-assets.com/2885/1662053765-dev-dot-product-landing-what-is-boundary-diagram.png"
+		"image": {
+			"light": "https://www.datocms-assets.com/2885/1679095102-devdot-boundary_lm.png",
```
kendallstrautman, line 14: We're updating both light and dark here because some of the assets were inconsistent.

src/components/themed-image/index.tsx
alexcarpenter: Have we tried using the picture element here? Now that we're on next-13 with the updated next/image, I think this pattern works (haven't fully tested). And removes the need for the CSS hide/show based on theme value.

```tsx
import Image, { type ImageProps } from 'next/image'

interface ThemedImageProps extends ImageProps {
  darkSrc: string
}

function ThemedImage(props: ThemedImageProps) {
  const { darkSrc, ...restProps } = props
  return (
    <picture>
      <source srcSet={darkSrc} media="(prefers-color-scheme: dark)" />
      <Image {...restProps} />
    </picture>
  )
}
```

alexcarpenter: Do we intend to support themes outside of dark and light?

alexcarpenter: ah right right

kendallstrautman: @BRKalow good call, somehow that was on my TODOs but i'll refactor this to use `next/image` and consider the slot pattern you mentioned. If not in this PR, I think a version with slots will be helpful when we lift this up globally.

kendallstrautman: @alexcarpenter that is a super good tip about the `picture` element. i'm gonna look into that even if it doesn't have the flexibility for additional theming. I'd say the likelihood of us creating additional themes is high.

braydenlove (approved):
Looks good!

src/components/themed-image/index.tsx
kendallstrautman: Oh @BRKalow now I remember, the issue with using next-image is that all of these graphics have a different size. I could try to fiddle with the styles where you don't input the size 🤔

BRKalow: Ooh yeah, makes sense. Feel free to punt on this if it feels like too much of a time sink then 👍

kendallstrautman: Okay, after looking into it, they all mostly had the same height, so was able to refactor to use next/image! Let me know what you think @BRKalow

src/views/product-landing/components/themed-image/themed-image.module.css
```diff
@@ -0,0 +1,12 @@
+.root {
+	align-items: center;
+	display: flex;
+	width: 100%;
```
kendallstrautman, line 4: The widths of the images are all different, so we need to use the `fill` approach and set the height here.

kendallstrautman:
💭 I need to check how this functions with the dark mode flag disabled. To make sure both don't render in prod.

BRKalow (approved):
lgtm!

src/layouts/base-new/index.tsx
```diff
@@ -65,15 +64,16 @@ const BaseNewLayout = ({
...
+	 * @TODO remove conditional after dark mode is released
+	 * Right now we are forcing 'light' mode on all pages
+	 */
+	const Layout = CoreDevDotLayoutWithTheme
 	const layoutProps = __config.flags.enable_dark_mode
 		? {
 				theme,
 		  }
-		: []
+		: { theme: GlobalThemeOption.light }
```
kendallstrautman, line 76: @BRKalow another meaningful change I just pushed up here. We need to enable the light theme for the app so these themed images 'work'. Otherwise both images render since the `data-theme` attribute isn't set on the `html` element unless we render the theme provider. There are other ways we could work around it, but this felt the simplest. 

Lmk your thoughts!

BRKalow: This seems like a good blanket fix. Initially I had a thought that we should add logic to `ThemedImage`, but then we'd have to do the same for any component that has similar logic. This seems more reasonable, nice work!

PR #1754, opened 2023-03-16
refactor: avoid overlap bugs in docs version select

zchsh:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 💭 Anything else?

---
  - https://app.asana.com/0/0/1204130762580109


alexcarpenter (commented):
Not fully grasping the limitations with the current setup, but would CSS float be the tool here?

zchsh:
@alexcarpenter it could be! tbh a lot of the challenge is the DOM ordering - I think part of the reason the current absolute-positioning approach was taken is that it would have been non-trivial to get the version selector element (not an MDX element) to exist in the same container as the "hero heading" thing on docs landing pages and/or the plain `<h1 />` on typical docs pages. The absolute positioning kinda side-stepped that a bit (while unfortunately leaving the possibility of overlap in some edge cases)

This one has been a bit backburner-y for me so it's gotten a lil stale 😅 but just before prioritizing the `hashicorp/web` shifts, I had a thought here, which was we should maybe start by always extracting the `<h1 />` from docs content so that we have more flexibility in what we render in and around it. Once that's done, and we have better control over where the heading element gets rendered, using a `float` or `flex` layout should be much more accessible 👍

src/lib/remark-plugins/remark-plugin-remove-h1/index.ts
```diff
@@ -0,0 +1,35 @@
...
+
+import { Heading } from 'mdast'
+import { Plugin, Transformer } from 'unified'
+import { visit, SKIP, Parent, Node } from 'unist-util-visit'
+
+/**
+ * This plugin removes any `<h1 />` element from the `remark` AST
+ * no which it is run.
+ */
+const remarkPluginRemoveH1: Plugin = (): Transformer => {
```
zchsh, line 14: We now render a `DocsPageHeading` component with an `h1` outside the MDX content area for nearly all docs pages (likely short-lived exception: Packer plugins).

We run this plugin to avoid rendering a duplicative `h1` within the content area.

src/views/docs-view/index.tsx
```diff
@@ -43,33 +43,76 @@ const DocsView = ({
...
+	 * The one use case where we don't have this is for Packer plugin docs,
+	 * which uses this `DocsView` component directly but uses `docs-view/server`
+	 * indirectly, with some other modifications, such as adding badges above the
+	 * page heading. Packer plugins MDX processing does _not_ remove the `<h1 />`
+	 * from MDX content, so we do not want to render a duplicative
+	 * `DocsPageHeading` `h1` element.
+	 *
+	 * The Packer plugins use case will fade away after Integrations work,
+	 * at which point we can always safely render <DocsPageHeading />.
+	 * Task: https://app.asana.com/0/1202097197789424/1204412156894157/f
```
zchsh, line 67: I've tried to be thorough here on documenting that Packer plugins is the exception here.

Feels worth revisiting this as soon as Packer plugins are replaced by integrations, so I've made a follow-up task: 🎟 [[cleanup] Remove docs page heading cruft once /packer/plugins is replaced by integrations](https://app.asana.com/0/1202097197789424/1204412156894157/f)

src/views/docs-view/server.ts
```diff
@@ -178,6 +181,12 @@ export function getStaticGenerationFunctions<
 					 * expected to have been run for remote content.
 					 */
 					[anchorLinks, { headings }],
+					/**
+					 * Remove the `<h1 />` from MDX, we'll render this outside
+					 * the MDX content area, integrating it into our layout
+					 * in various ways depending on the specific docs view used.
+					 */
+					remarkPluginRemoveH1,
```
zchsh, line 189: We now _always_ remove `h1` elements from MDX content for docs pages, as all docs page render `DocsPageHeading`, which contains an `h1`.

Note: The one exception is Packer plugins, but [Packer plugins has separate MDX processing](https://github.com/hashicorp/dev-portal/blob/fec683f2273e4df815a479856efe0a0902b04e07/src/views/packer-plugins/server.ts#L186), so we don't need to worry about that here. This is something we will untangle in the near future once Packer integrations launch: 🎟 [[cleanup] Remove docs page heading cruft once /packer/plugins is replaced by integrations](https://app.asana.com/0/1202097197789424/1204412156894157/f)

src/components/docs-version-switcher/docs-version-switcher.module.css
```diff
@@ -3,29 +3,40 @@
...
-		}
+Should we consider a `fullWidth` prop on `DropdownDisclosure`
+rather than these styles?
 
-		& ul {
-			width: 100%;
-			max-width: 100%;
-		}
+Task:
+https://app.asana.com/0/1202097197789424/1204412156894162/f
```
kendallstrautman, line 16: 💡 I like this idea

src/lib/remark-plugins/remark-plugin-remove-h1/index.ts
```diff
@@ -0,0 +1,35 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import { Heading } from 'mdast'
+import { Plugin, Transformer } from 'unified'
+import { visit, SKIP, Parent, Node } from 'unist-util-visit'
+
+/**
+ * This plugin removes any `<h1 />` element from the `remark` AST
+ * no which it is run.
```
kendallstrautman, line 12:
```suggestion
 * on which it is run.
```
⛏️

src/lib/remark-plugins/remark-plugin-remove-h1/index.ts
kendallstrautman: Do we have a content conformance check in the works for this?

kendallstrautman: 💭  I worry if folks use an h1 and then get no authoring feedback about it getting removed.

src/views/docs-view/components/docs-page-heading/docs-page-heading.module.css
```diff
@@ -0,0 +1,29 @@
...
+
+.root {
+	align-items: flex-start;
+	display: flex;
+
+	/* DOM order is version selector, then title. On small viewports, with
+	   wrapping, we have multiple rows so row-reverse has no effect.
+		 On large viewports, where we have space to put everything in one row,
+		 we show the version selector after the landingHero (if used). */
+	flex-direction: row-reverse;
```
kendallstrautman, line 14: ❓ Could be an issue here with visual order not reflecting DOM order? One option we did to work around this in the footer (with the give feedback button) was render two different components and `display:none` for one based on viewport so that the markup matched the visual order.

src/views/docs-view/components/docs-page-heading/index.tsx
```diff
@@ -0,0 +1,42 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import { ReactNode } from 'react'
+import classNames from 'classnames'
+import s from './docs-page-heading.module.css'
+
+function DocsPageHeading({
```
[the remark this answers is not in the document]
kendallstrautman: I like it! 💯

src/lib/remark-plugins/remark-plugin-remove-h1/index.ts
zchsh: I believe there is a check in place to ensure all documents have an `h1` in their content.

Re: removal of `h1`, I think it's more accurate to say "we render it outside the MDX content area, rather than within it" (we happen to use remark plugins to make that happen, the `anchor-links` plugin extracts all headings including the `h1`, the `remove-h1` plugin ensures we don't render the `h1` in both areas). So we're still rendering the `h1` that authors write in MDX, just in a different spot. With that in mind, my take was we don't need to warn authors about it, since it's more an implementation detail than anything else.

kendallstrautman: Okay sweet, after looking over the rest of the code I see now how the `h1`s were being used within the document and forgot that docs pages don't have a proper `title`, so this makes sense!

kendallstrautman (approved):
Followed the testing steps and everything looks good to me!

src/views/docs-view/components/docs-page-heading/docs-page-heading.module.css
zchsh: Great callout! 💯 

This may or may not be an depending on our interpretation of the "meaningful sequence" success criterion and related techniques. There's a bit more detail in [this GitHub comment](https://github.com/hashicorp/dev-portal/pull/1687#discussion_r1127047577). I'll go into even more detail below in case it's helpful.

The TLDR is:
- **I'd prefer to handle the DOM-vs-visual-order as a separate ticket**, as this PR maintains the upstream approach to DOM vs visual ordering for the version select.
- The **WCAG pages themselves also "break"** under what I understand to be our very strict interpretation of the DOM-must-match-visual-order technique, so **I think our current interpretation warrants some discussion**.

---

## Meaningful sequence success criteria have some subtlety in technique

We _could_ take a really direct and strict interpretation of the [Making the DOM order match the visual order](https://www.w3.org/WAI/WCAG21/Techniques/css/C27) technique. In other words, we could be very strict that  "anything visually to the left must be earlier the DOM order, anything visually to the right must be later in the DOM order".

But, I think this is not necessary to meet the success criterion we've committed to, and may in fact compromise the user experience for many users, especially those who use keyboard navigation. [Making the DOM order match the visual order](https://www.w3.org/WAI/WCAG21/Techniques/css/C27) is _one_ technique to meet the [Meaningful Sequence](https://www.w3.org/WAI/WCAG21/Understanding/meaningful-sequence) criterion, but it's not the _only_ technique. As noted on the criterion page:

> Each numbered item in this section represents a technique or combination of techniques that the WCAG Working Group deems sufficient for meeting this Success Criterion. However, **it is not necessary to use these particular techniques**.

As an abstract example from [the WCAG success criterion page on meaningful sequences](https://www.w3.org/WAI/WCAG21/Understanding/meaningful-sequence#examples), it can be okay for the visual presentation to not match the programmatically determined order if it doesn't affect the meaning of the page:

> Example 2: CSS is used to position a navigation bar, the main story on a page, and a side story. The **visual presentation of the sections does not match the programmatically determined order, but the meaning of the page does not depend on the order of the sections**. 

To me, this seems logical. For example, whether the table-of-contents or the page content appears first in "visual order" does not affect the meaning of the page, so we can _choose_ to match that visual order to the DOM order (or not!) based on what we think will produce the ideal experience.

As a concrete example of exactly this case, the WCAG page itself places its table-of-contents-style navigational sidecar visually _to the right_ of the main content area (arguably _later_ in the visual order), while that `aside` element appears _earlier_ in the DOM order:

CleanShot 2023-04-17 at 19 51 44@2x

I believe this is objectively okay with relation to the "Meaningful sequence" success criteria because the meaning of the page does not depend on the order of the `aside` and `main` sections. And it seems subjectively preferable because putting the `aside` _before_ the `main` element in the DOM order arguably makes for a better user experience, as we can keyboard-navigate through the table-of-contents _before_ having to keyboard-navigate through the entire main content. Quick screen-share of tabbing through things to demonstrate:

https://user-images.githubusercontent.com/4624598/232635346-a7509e86-a231-44d1-b068-7eb581f0b430.mp4

My thinking is that we're in a similar situation with the version select and the `h1`. The visual order of the version selector and the page content does not impact meaning, so I don't think we're strictly required to match it to DOM order. And I think we should in fact _avoid_ strictly matching them, as varying the DOM order seems like it could _negatively_ impact the user experience on the page. Encountering the `h1`, _then_ the version select, _then_ the text content meant to directly follow the `h1` seems non-intuitive to me - especially since it would happen on some viewports but not others (more on this below!).

## Viewport-toggled elements seem non-ideal

All that being said, I can _maybe_ imagine a case for wanting to be very strict about visual order matching DOM order, just as some kind of safeguard.

If this is the case, and we establish team consensus to be strict on that technique front rather than on the success criterion front, **I think we need to solve DOM and visual ordering mismatch problems at the design level**, not the implementation level.

I think by toggling elements on "mobile" vs "desktop" in order to match a strict, meaning-impact-agnostic interpretation of the "Meaningful Sequences" success criterion, we risk introducing additional implementation and end-user complexity by effectively varying the DOM order between "mobile" and "desktop", and have elements pop in and out of existence at different viewport sizes.

As an example of why I think toggling "mobile" and "desktop" elements could be non-ideal, consider the WCAG example as context:

> If a blind user, who reads the page with a screen reader that follows the source order, is working with a sighted user who reads the page in visual order, they may be confused when they encounter information in different orders.

In an analogous example, imagine one user is using working with a slightly smaller viewport and another user is working with a slightly larger viewport (or even, imagine a single user has just resized their viewport slightly, and a second ago had been tabbing through the page). For each viewport in isolation, the visual and DOM order may be more "perfectly matched", but the difference in DOM and focus order across arbitrary viewport breakpoints seems likely to cause some confusion as well. The order is arguably a _more meaningful sequence_ if it is semantically consistent at the programmatic document level across viewports.

So I think ideally, assuming we first establish that we need to strictly match DOM and visual order regardless of impact on page meaning, we'd back this up and come up with a design solution that _avoids_ the order "mismatch" issue altogether, rather than trying to "correct" it in implementation.

PR #1755, opened 2023-03-16
style: adjust supported and product icons

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksadjust-supported-icons-hashicorp.vercel.app/packer) 🔎
- [Asana task](https://app.asana.com/0/1203652232454021/1204108582255298) 🎟️
- [Figma file](https://www.figma.com/file/v0Hyqvjdf6QJAvRkdIBV24/Dark-Mode?node-id=433-32260) 🎨 

## 🗒️ What

Updates the product icon and adjusts the nessecary 'supported icons' that render in cards. 

Screenshot 2023-03-20 at 5 13 53 PM
Screenshot 2023-03-20 at 3 49 18 PM

Screenshot 2023-03-20 at 5 17 51 PM

## 🛠️ How

For the product icons, I updated them to import the non color version then added the color via CSS, referencing the brand token color. This makes it more flexible if product colors shift between themes in the future. Currently vault is the only product where the primary brand color changes between themes. 

For the 'supported icons' that render in product landing card links, 'aws' was the only one that needed to be updated between light and dark. Instead of refactoring the entire supported icons component to support theming, I chose to force the aws logo to be monochromatic, and render all white or all black.

More discussion and discovery needs to be had here to decide how we handle logos moving forward. For the company icons, this doesn't feel like the final state, but a pragmatic step forward. 

## 🧪 Testing

- Visit the preview: [packer](https://dev-portal-git-ksadjust-supported-icons-hashicorp.vercel.app/packer), [terraform](https://dev-portal-git-ksadjust-supported-icons-hashicorp.vercel.app/terraform)
- Validate that the card links to external company resources and the product icons on the tutorial cards render as expected.


src/content/supported-icons.tsx
```diff
@@ -35,14 +35,19 @@ import { IconTerminal16 } from '@hashicorp/flight-icons/svg-react/terminal-16'
...
 
+/**
+ * @TODO refactor to remove the HashiCorp product logos from this list
+ * when product logos are needed, component should use `ProductIcon`
+ */
+
 export const SUPPORTED_ICONS = {
 	api: <IconApi16 />,
-	'aws-color': <IconAwsColor16 />,
+	'aws-color': <IconAws16 />,
```
kendallstrautman, line 50:
```suggestion
	'aws': <IconAws16 />,
```
This shouldn't be color. This is what we should connect on.

kendallstrautman: Write a note for context on aws being the outlier here

src/components/product-icon/index.tsx
```diff
@@ -3,36 +3,36 @@
...
+	waypoint: IconWaypoint16,
 }
 
 const ProductIcon = ({ productSlug, ...rest }: ProductIconProps) => {
 	const Icon = productSlugsToIcons[productSlug]
 	if (!Icon) {
 		return null
 	}
-	return <Icon {...rest} />
+	return <Icon {...rest} color={`var(--token-color-${productSlug}-brand)`} />
```
kendallstrautman, line 35: Write a note re: why we need to set the color here.

src/content/supported-icons.tsx
```diff
@@ -35,14 +35,22 @@ import { IconTerminal16 } from '@hashicorp/flight-icons/svg-react/terminal-16'
...
+ * when product logos are needed, component should use `ProductIcon`
+ *
+ * With the introduction of theming, when adding new icons to this list,
+ * consider both dark and light states.
+ */
+
 export const SUPPORTED_ICONS = {
 	api: <IconApi16 />,
-	'aws-color': <IconAwsColor16 />,
+	aws: <IconAws16 />,
```
kendallstrautman, line 53: 'aws' is the outlier here, I decided to remove it from rendering the color version to avoid making all of these supported icons support theming. Until we have a larger or more broad usecase, we will do this until we have a stance on external company icon styles in the app. External company logos will always be troublesome with theming, so I'm proposing that we consider always using just black or white for these icons. More discussion need to be had now.

kendallstrautman: Additionally, this approach doesn't deviate from the existing patterns where generic icons use black or white fills. See example here: 

Screenshot 2023-03-20 at 5 22 17 PM

kendallstrautman: In the future, if the flight library's aws icon allows the 'color' of the wordmark to be inherited by the parent, we could work around this (like how we handle terraform clouds logo for example). But given the legality of logo use, I can see why they wouldn't be able to do this.

kellyschummer (approved):
Approach and Implementation 💯 !!

BRKalow:
Lgtm!

BRKalow (approved):

PR #1759, opened 2023-03-18
style: fix up interactive lab

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksupdate-interactive-lab-hashicorp.vercel.app/waypoint/tutorials/get-started-docker/get-started-docker) 🔎
- [Asana task](https://app.asana.com/0/1203652232454021/1204108584803961) 🎟️
- [Figma](https://www.figma.com/file/v0Hyqvjdf6QJAvRkdIBV24/Dark-Mode?node-id=513-38950&t=f04aO7Y2CxzLHAt4-0) 🎨 

## 🗒️ What

This PR fixes up the interactive lab and updates the terminal graphic. 

## 📸 Design Screenshots
Screenshot 2023-03-17 at 5 26 42 PM

## 🧪 Testing

- [Visit the preview](https://dev-portal-git-ksupdate-interactive-lab-hashicorp.vercel.app/waypoint/tutorials/get-started-docker/get-started-docker)
- Validate that it looks like the[ figma file ](https://www.figma.com/file/v0Hyqvjdf6QJAvRkdIBV24/Dark-Mode?node-id=513-38950&t=f04aO7Y2CxzLHAt4-0)in both dark and light mode 

## 💭 Anything else?

The card dark mode override caused this issue. Something to be aware of, that any dark mode overrides with this current approach mess up the CSS waterfall.


src/components/interactive-lab-callout/interactive-lab-callout.module.css
```diff
@@ -11,11 +11,17 @@
 		display: flex;
 		gap: 32px;
 	}
+
+	@nest html[data-theme='dark'] & {
+		background-color: var(
+			--token-color-foreground-strong
+		); /** Needed to supercede the dark mode card .root dark mode background  */
+	}
```
kendallstrautman, line 19: This is the second time i've seen an issue with scope using the overrides method. The current CSS with this component should work, as it is the same value that we've put here -- it should be the same for dark and light mode. But this is wrapped in a card and therefore the card background color override was rendering instead of the bg color that was set on line 8 in this file. So we have to redefine the background color for dark mode here. 

Another reason to _really_ avoid overrides. When we `@nest html` this supercedes even child styles that are scoped to those components.

kellyschummer (approved):
💯 approved. Thank you @kendallstrautman

src/components/interactive-lab-callout/interactive-lab-callout.module.css
```diff
@@ -11,11 +11,17 @@
...
 }
 
 .title {
 	composes: hds-typography-body-200 from global;
-	color: var(--token-color-surface-primary);
+	color: var(--token-color-surface-faint);
```
ashleemboyer, line 24: From what I'm seeing in the Figma, the title should still be surface/primary (although the body text _is_ surface/faint):
https://www.figma.com/file/v0Hyqvjdf6QJAvRkdIBV24/Dark-Mode?node-id=598-35371&t=fVgpqBU7D48pQX2o-0

CleanShot 2023-03-21 at 21 19 20

src/components/interactive-lab-callout/interactive-lab-callout.module.css
ashleemboyer: Trying to make sure I'm following -- are you saying we should be avoiding `@nest html`?

ashleemboyer (changes requested):
Small color discrepancy I noticed, and a quick question for clarification!

src/components/interactive-lab-callout/interactive-lab-callout.module.css
kendallstrautman: Oh good catch! thank you 🙇🏻

kendallstrautman, line 24:
```suggestion
	color: var(--token-color-surface-primary);
```

src/components/interactive-lab-callout/interactive-lab-callout.module.css
kendallstrautman: I think we should just be aware of potential scope issues when we do use it, given the behavior here. And we for that reason we should limit overrides. 

To clarify what happened here: I had to redefine the background color for this component even though it was already set with that set value since the `Card` (which is a parent here) dark mode background-color override was taking precedent over this component's styles.

ashleemboyer: Can you elaborate on what you mean by `override`? I'm fully understanding the technical CSS issue -- we only need this `@nest` here because there's one in `Card`.

ashleemboyer (approved):

src/components/interactive-lab-callout/interactive-lab-callout.module.css
kendallstrautman: Yes exactly, I understand why the specificity is taking precedent, I'm just noting it as a side-effect of doing theming overrides that we need to keep in mind since we're so used to scoped css modules.

ashleemboyer: Ok sweet we're on the same page then! I thought when you said "The current CSS with this component should work" in your first comment, maybe you were worried about this being a bug. 💟

PR #1761, opened 2023-03-21
feat: localize enterprise alert

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksupdate-enterprise-alert-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1199634971449915/1204169535040910) 🎟️
- [Figma File](https://www.figma.com/file/v0Hyqvjdf6QJAvRkdIBV24/Dark-Mode?node-id=509-33711&t=VIp9mtvLuzzFxhN4-0) 🎨 

## 🗒️ What

This PR adds a new custom `EnterpriseAlert` component to replace the one used [from react-components. ](https://react-components.vercel.app/components/enterprisealert) The code is based off of the react-components implementation. 

Screenshot 2023-03-21 at 2 47 08 PM

Screenshot 2023-03-21 at 2 46 53 PM

Screenshot 2023-03-21 at 2 45 13 PM

Screenshot 2023-03-21 at 2 50 27 PM

Screenshot 2023-03-21 at 2 57 02 PM

## 🤷 Why

The react-components version was not compatible with dark mode. Instead of doing overrides on that legacy component, we redesigned and implemented one for dev dot. 

## 📸 Design Screenshots

Screenshot 2023-03-21 at 2 47 55 PM

## 🧪 Testing

Test the following variants, ensure that they match the design spec in both light and dark mode.

- [Example with long form custom text]( https://dev-portal-git-ksupdate-enterprise-alert-hashicorp.vercel.app/consul/docs/nia/configuration#terraform-cloud-driver)
- [Another product long form example](https://dev-portal-git-ksupdate-enterprise-alert-hashicorp.vercel.app/nomad/docs/configuration#configuration-reload)
- [Inline next to a heading](https://dev-portal-git-ksupdate-enterprise-alert-hashicorp.vercel.app/consul/docs/agent/telemetry#license-expiration)
- [Inline in a table cell](https://dev-portal-git-ksupdate-enterprise-alert-hashicorp.vercel.app/consul/docs/nia/configuration#consul-kv-source-input)

## 💭 Anything else?

See more examples [in content here ](https://hashicorp-mktg.sourcegraph.com/search?q=context:global+file:.mdx+EnterpriseAlert&patternType=standard&case=yes&sm=1&groupBy=repo)


src/components/inline-link/types.ts
```diff
@@ -3,10 +3,12 @@
  * SPDX-License-Identifier: MPL-2.0
  */
 
+import type { ComponentProps } from 'react'
+import NextLink from 'next/link'
 import { LinkProps } from 'components/link'
 import { TextProps } from 'components/text'
 
-export interface InlineLinkProps {
+export interface InlineLinkProps extends ComponentProps<typeof NextLink> {
```
kendallstrautman, line 11: I needed to add this to pass `target`

src/layouts/sidebar-sidecar/utils/_local_platform-docs-mdx/components/enterprise-alert/enterprise-alert.module.css
```diff
@@ -0,0 +1,57 @@
...
+	&.inline {
+		background: none;
+		border: none;
+		padding: 0 7px;
+		display: inline-flex;
+		vertical-align: middle;
+
+		--tag-margin: 0;
+
+		&:first-child {
```
kendallstrautman, line 27: This is for table styles

src/layouts/sidebar-sidecar/utils/_local_platform-docs-mdx/components/enterprise-alert/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+	productSlug: ProductSlug
+	inline?: boolean
+	className?: string
+	children?: ReactNode
+}
+
+//
+// Base components need to be slightly modified
+// to fit our use cases in dev-portal
+//
```
kendallstrautman, line 28:
```suggestion
```

src/layouts/sidebar-sidecar/utils/_local_platform-docs-mdx/components/enterprise-alert/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+				</Text>
+			)}
+		</Element>
+	)
+}
+
+export function DocsEnterpriseAlert(
+	props: Omit<EnterpriseAlertProps, 'productSlug'> & {
+		product?: ProductSlug
+	}
```
kendallstrautman, line 92: Could be a better way to do this. I wanted only the product prop to be optional. The `Partial` utility makes them all optional but that isn't necessarily what I want here.

src/layouts/sidebar-sidecar/utils/_local_platform-docs-mdx/components/enterprise-alert/index.tsx
```diff
@@ -0,0 +1,101 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
```
kendallstrautman, line 4: 💭 wondering if this is the right place for this to live. I also thought of putting it in 'author-primitives'?

braydenlove:
@kendallstrautman awesome job, these look a ton better. I did catch one tiny thing:

When a badge is sitting next to a title can we have it be left aligned when going to its own line? 
image

Otherwise we are good to go!

src/components/inline-link/types.ts
BrandonRomano: Oh I like this NextLink extension!

braydenlove:
> @braydenlove let me see if I can get creative here. This is due to the padding on the element and its difficult to 'know' when things break to two lines and adjust the padding dynamically. Fwiw what you've shared is how the current enterprise alert functions.

Yeah I feel ya, if you can great, if not it still works _imo_

BrandonRomano (approved):
This looks good!  I don't have a strong opinion on the location (RE: your comment), mostly because I don't know where we keep other similar things.

src/components/inline-link/types.ts
ashleemboyer, line 11: We should extend `LinkProps` here since that's the component being interacted with. It already handles extending `ComponentProps<typeof NextLink>` for us! 😊

```suggestion
export interface InlineLinkProps extends LinkProps {
```

Also curious what you think about leveraging the `opensInNewTab` prop instead of using `target="_blank"`?

src/layouts/sidebar-sidecar/utils/_local_platform-docs-mdx/components/enterprise-alert/index.tsx
ashleemboyer: Personally I think we should consolidate `author-primitives` into `dev-dot-content/mdx-components`. I see them like "legacy" components since they were primarily kept separate for the "lift & shift" of .io sites to `dev-portal`.

src/layouts/sidebar-sidecar/utils/_local_platform-docs-mdx/components/enterprise-alert/index.tsx
ashleemboyer: 🙋🏻 Was renaming the property from `productSlug` to `product` intentional?

💭 I think we could simplify to get the same output like this, it works like an overwrite so `Omit` isn't needed:

```suggestion
	props: EnterpriseAlertProps & {
		productSlug?: ProductSlug
	}
```

---

🎨 Optional -- I see value in pulling this `&` logic out and naming it (such as `DocsEnterpriseAlertProps`), so there's less cognitive load required to figure out what's tied to the `props` variable:

```tsx
type DocsEnterpriseAlertProps = EnterpriseAlertProps & {
	productSlug?: ProductSlug
}

export function DocsEnterpriseAlert(props: DocsEnterpriseAlertProps) {
	const currentProduct = useCurrentProduct()
	return (
		<EnterpriseAlert
			productSlug={props.productSlug || currentProduct.slug}
			{...props}
		/>
	)
}
```

src/layouts/sidebar-sidecar/utils/_local_platform-docs-mdx/components/enterprise-alert/enterprise-alert.module.css
```diff
@@ -0,0 +1,57 @@
...
+	/** 
+	* The mdx content anchor styles overwrite the link styles needed 
+	* here, which should inherit body-200 styles
+	*/
+	& > a {
+		font-size: inherit;
+		font-weight: inherit;
+		line-height: inherit;
+	}
+}
```
ashleemboyer, line 47: 🙋🏻 Did you mean to use the `.link` class for this? I see one passed to `InlineLink` in the JSX, but not seeing it used.

src/layouts/sidebar-sidecar/utils/_local_platform-docs-mdx/components/enterprise-alert/enterprise-alert.module.css
```diff
@@ -0,0 +1,57 @@
...
+
+.badge {
+	align-self: baseline;
+	margin-right: var(--tag-margin);
+	margin-bottom: var(--tag-margin);
```
ashleemboyer, line 52: 💭 I _think_ we can replace both of these with `gap` on the parent container, and then the alignment issue Brayden noticed potentially won't be an issue anymore.

src/layouts/sidebar-sidecar/utils/_local_platform-docs-mdx/components/enterprise-alert/enterprise-alert.module.css
ashleemboyer: 🙋🏻 Which tables? I think I'm missing some context!

src/components/inline-link/types.ts
kendallstrautman: Cool, I'll update this to extend `LinkProps`. Using the `target` attribute feels more natural to me but do you have thoughts on the `opensInNewTab` props?

src/layouts/sidebar-sidecar/utils/_local_platform-docs-mdx/components/enterprise-alert/enterprise-alert.module.css
kendallstrautman: Nice! i'll refactor to use gap for the base state but the issue brayden pointed out is still a problem due to the padding + inline-flex when the component is set to inline

ashleemboyer: Ah! I see now that was about the headings 💟

src/components/inline-link/types.ts
ashleemboyer: I do! 😊 Have my thoughts in a reply here in the `Link` component's RFC: https://docs.google.com/document/d/14embplHRyd2zO1A9a7Zws75xZeNZ6yG6WaXr57DnSuQ/edit?disco=AAAAijiamkE

src/layouts/sidebar-sidecar/utils/_local_platform-docs-mdx/components/enterprise-alert/enterprise-alert.module.css
kendallstrautman: See this example where the alert renders [inline in a table cell](https://dev-portal-git-ksupdate-enterprise-alert-hashicorp.vercel.app/consul/docs/nia/configuration#consul-kv-source-input)

src/layouts/sidebar-sidecar/utils/_local_platform-docs-mdx/components/enterprise-alert/index.tsx
kendallstrautman: Yeah the reason I kept `product` here is to keep it backwards compatible with the [original implementation](https://react-components.vercel.app/components/enterprisealert) from react-components so that none of the content breaks.

👌🏻 I can separate out the interface!

src/components/inline-link/types.ts
kendallstrautman: Thanks for linking to that doc, I'm going to retain the use of `target` here for now.

braydenlove (approved):

PR #1780, opened 2023-03-27
fix: don't add index page api docs waypoint

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-kstesting-waypoint-hashicorp.vercel.app/waypoint/api-docs) 🔎
- [Asana task](url) 🎟️

## 🗒️ What

A breadcrumb 'Waypoint' was being added for the index waypoint api-docs page that was a broken link. This fix updates the logic for the breadcrumb and sidebar menu item to not add items for this index page. [Slack Context](https://hashicorp.slack.com/archives/C02T80JTNS2/p1679949878261239)

Screenshot 2023-03-27 at 2 54 27 PM

## 🧪 Testing

- [Visit the preview](https://dev-portal-git-kstesting-waypoint-hashicorp.vercel.app/waypoint/api-docs)
- Validate that 'Waypoint' is no longer in the breadcrumb or sidebar [compared to upstream](https://developer.hashicorp.com/waypoint/api-docs)


src/pages/waypoint/api-docs/[[...page]].tsx
```diff
@@ -118,8 +122,14 @@ export async function getStaticProps({ params }) {
 	// @ts-expect-error `isCurrentPage` is an expected optional field
 	breadcrumbLinks[breadcrumbLinks.length - 1].isCurrentPage = true
 
+	// Remove the 'index' page, whose path is '/waypoint/api-docs'
+	const filteredMenuItems = apiPageProps.navData.filter(
+		(menuItem) => menuItem.path != ''
+	)
+	console.log(apiPageProps.navData)
```
Comment (kendallstrautman) [inline src/pages/waypoint/api-docs/[[...page]].tsx:129]:
```suggestion
	console.log(apiPageProps.navData)
```
```suggestion
```

src/pages/waypoint/api-docs/[[...page]].tsx
```diff
@@ -105,9 +105,13 @@ export async function getStaticProps({ params }) {
 		{ title: 'Waypoint', url: `/waypoint` },
 		{ title: 'API', url: `/waypoint/api-docs` },
 	]
+	// Remove the 'index' page '/waypoint/api-docs' from the breadcrumbs
+	const shouldAddBreadcrumb = !(
+		apiPageProps.operationCategory.slug === productData.slug
```
Comment (kendallstrautman) [inline src/pages/waypoint/api-docs/[[...page]].tsx:110]:
```suggestion
		apiPageProps.operationCategory?.slug === productData.slug
```
Do I need to guard here?

src/pages/waypoint/api-docs/[[...page]].tsx
```diff
@@ -118,8 +122,13 @@ export async function getStaticProps({ params }) {
 	// @ts-expect-error `isCurrentPage` is an expected optional field
 	breadcrumbLinks[breadcrumbLinks.length - 1].isCurrentPage = true
 
+	// Remove the 'index' page, whose path is '/waypoint/api-docs'
```
Comment (kendallstrautman) [inline src/pages/waypoint/api-docs/[[...page]].tsx:125]:
```suggestion
	// Remove the 'index' menu item, whose path is '/waypoint/api-docs'
```

src/pages/waypoint/api-docs/[[...page]].tsx
```diff
@@ -105,9 +105,13 @@ export async function getStaticProps({ params }) {
...
 		{ title: 'API', url: `/waypoint/api-docs` },
 	]
+	// Remove the 'index' page '/waypoint/api-docs' from the breadcrumbs
+	const shouldAddBreadcrumb = !(
+		apiPageProps.operationCategory.slug === productData.slug
+	)
 
 	// Breadcrumbs - Render conditional category
-	if ('operationCategory' in apiPageProps) {
+	if (shouldAddBreadcrumb && 'operationCategory' in apiPageProps) {
```
Comment (kendallstrautman) [inline src/pages/waypoint/api-docs/[[...page]].tsx:114]: @zchsh lmk if i'm missing context. Seems like waypoint is different from boundary api docs in that everything is just in one index file.

src/pages/waypoint/api-docs/[[...page]].tsx
Comment (zchsh) [reply src/pages/waypoint/api-docs/[[...page]].tsx:110]: It _shouldn't_ be needed, all `operationCategory` objects should have a slug 🤔 But also can't imagine this hurting in any ways, so I'd say guard is probably better than leaving it out 👍

src/pages/waypoint/api-docs/[[...page]].tsx
Comment (zchsh) [reply src/pages/waypoint/api-docs/[[...page]].tsx:114]: Yep, this sounds exactly right! 💯 👍

At some point in the near future we'll probably want to refine things a bit so the differences are handled in a more intentional way, rather than being mostly defined by "works with Boundary by default cause that's what we started with, everything else might need exceptions tacked on" ... I feel like that's boxed us into having to make these kinds of fixes 😅 has felt like a decent skateboard but could use some love to become a .... bicycle or whatever 🤷‍♂️ 🤣

zchsh (approved):

PR #1793, opened 2023-03-29
feat: theme-switcher in user dropdown

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-kstheme-switcher-spike-hashicorp.vercel.app/waypoint) 🔎
- [Asana task](https://app.asana.com/0/1203652232454021/1204303889362363) 🎟️
- [Figma](https://www.figma.com/file/v0Hyqvjdf6QJAvRkdIBV24/Dark-Mode?node-id=1166-44816&t=fX2DuVInKTlje6l5-0) 🎨 

## 🗒️ What

This PR implements the `ThemeSwitcher` select component in the user dropdown. _Note_ this is merging into an assembly because we don't want to remove icons from the dropdown navigation menu items until we release dark mode with this switcher. 

Screenshot 2023-03-30 at 3 58 31 PM
Screenshot 2023-03-30 at 3 58 59 PM
Screenshot 2023-03-30 at 3 59 12 PM
Screenshot 2023-03-30 at 3 59 59 PM
Screenshot 2023-03-30 at 3 59 52 PM

## 🛠️ How

Native `select` was used for ease of implementation and in our competitive analysis, uses were varied between dropdown and select. This was a value engineering decision since the effort is much less and the end result for the user is basically the same. See Vercel's example, where they use native select: 

### Vercel's reference
Screenshot 2023-03-30 at 3 47 17 PM

Screenshot 2023-03-30 at 3 47 20 PM

## 📸 Design Screenshots

Screenshot 2023-03-30 at 3 48 55 PM
Screenshot 2023-03-30 at 3 48 58 PM
Screenshot 2023-03-30 at 3 49 02 PM

## 🧪 Testing

Visit a page with [theming enabled:](https://dev-portal-git-kstheme-switcher-spike-hashicorp.vercel.app/waypoint)
- Open the user dropdown in an unauthenticated state
- Switch themes
- Validate that the user dropdown stays open after the theme option has been selected
- Sign in, test the authenticated state

### Keyboard usage
- TAB through the user dropdown menu
- When you focus on the theme select, hit SPACE and use your arrows to navigate through the options. 
- Select an option by hitting ENTER. 
- The user dropdown menu should stay open 
- If you TAB again after selecting a theme option, the user dropdown should close. 
- When focus is on the theme select, and its not open, if you hit ESCAPE, the entire user dropdown should close
- When focus is on the theme select, and it is open, if you hit ESCAPE, the option select menu should close but the whole user dropdown remains open. 

## 💭 Anything else?

This does not add the theme switcher to mobile, that will be handled in a follow up PR. Also, the refactor to the icon placement for the external link with 'Account Settings' will be handled in a follow PR. 

Also, conversation was had around this component being a dropdown versus a select, we landed on a select for the first phase since the implementation is simpler. [See the follow up ticket ](https://app.asana.com/0/1203652232454021/1204284168518969)here to reconsider using the dropdown to fully represent[ the original designs. ](https://www.figma.com/file/v0Hyqvjdf6QJAvRkdIBV24/Dark-Mode?node-id=835-42742&t=fX2DuVInKTlje6l5-0)


src/components/navigation-header/index.tsx
```diff
@@ -64,13 +61,12 @@ const AuthenticationControls = () => {
...
 						: [
 								{
-									icon: <IconSignIn16 />,
 									label: 'Sign in',
 									onClick: () => signIn(),
 								},
 								{
 									href: '/sign-up',
-									icon: <IconUserPlus16 />,
+
```
kendallstrautman, line 69:
```suggestion
```

kendallstrautman:
@ashleemboyer this isn't ready for review yet, but I'll tag you when it is

ashleemboyer:
> @ashleemboyer this isn't ready for review yet, but I'll tag you when it is

@kendallstrautman totally! I see it's still in draft mode. Just wanted to tag myself on it so I get the notification when it's RFR. 😊

src/components/theme-switch/user-dropdown-switcher/index.tsx
```diff
@@ -0,0 +1,20 @@
+import ThemeSwitcher from 'components/theme-switch'
+import {
+	DropdownDisclosureListItem,
+	DropdownDisclosureSeparatorItem,
+} from 'components/dropdown-disclosure'
+import s from './user-dropdown-switcher.module.css'
+
+export default function UserDropdownDisclosureThemeSwitcher() {
```
kendallstrautman, line 8: Composed version for the user dropdown.

src/components/user-dropdown-disclosure/index.tsx
```diff
@@ -67,6 +68,13 @@ const UserDropdownDisclosure = ({
 	listPosition,
 	user,
 }: UserDropdownDisclosureProps) => {
+	const { pathname } = useRouter()
+	// TODO handle elsewhere?
+	const shouldRenderThemeSwitcher =
+		__config.flags.enable_dark_mode &&
+		pathname !== '/' &&
+		pathname !== 'sign-up'
```
kendallstrautman, line 76: 💭 I wonder if we codify this in config or something, the routes that don't support theming.

src/components/user-dropdown-disclosure/index.tsx
```diff
@@ -67,6 +68,13 @@ const UserDropdownDisclosure = ({
 	listPosition,
 	user,
 }: UserDropdownDisclosureProps) => {
+	const { pathname } = useRouter()
+	// TODO handle elsewhere?
```
kendallstrautman, line 72:
```suggestion
```

src/components/theme-switch/user-dropdown-switcher/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+	return (
+		<>
+			<DropdownDisclosureSeparatorItem />
+			<DropdownDisclosureListItem className={s.listItem}>
+				<label className={s.label} htmlFor="theme-switcher">
+					Theme
+				</label>
```
kendallstrautman, line 15: I'm unsure if I should still use this as a `label` since I added an `aria-label` as well. Open to feedback here.

braydenlove:
Super minor thing I noticed, it appears like the descenders on the text are being masked/cut off. (_g_ on light and _y_ on system)
image
image

kendallstrautman:
@braydenlove can I handle that account settings icon placement adjustment in another PR?

kendallstrautman:
💭 I think we may want this in the assembly as well since it would affect the nav without icons in prod which we may not want.

kendallstrautman:
@braydenlove I think I fixed up that type issue. thank you! 🦅 👁️

braydenlove (approved):
Awesome job!

src/components/navigation-header/index.tsx
```diff
@@ -64,13 +61,11 @@ const AuthenticationControls = () => {
 						? getUserMenuItems({ signOut })
 						: [
 								{
-									icon: <IconSignIn16 />,
```
zchsh, line 67: 💭 makes sense re: assembly given the removal of icons... if we wanted to though, could we use `__config.flags.enable_dark_mode` to conditionally show these icons?

Something like:

```jsx
icon: __config.flags.enable_dark_mode ? undefined : <IconSignIn16 />
```

Note sure if that would work, just tossing it out there in case it might 😬

zchsh (approved):
✅ Ran through the testing steps, looks great to me!

💭 Had a thought on using flags vs assembly branch for the icon removal, but def non-blocking.

src/components/theme-switch/user-dropdown-switcher/user-dropdown-switcher.module.css
```diff
@@ -0,0 +1,17 @@
...
+	border-radius: 5px;
+	color: var(--token-color-foreground-primary);
+	grid-gap: 8px;
+	gap: 8px;
+	width: 100%;
+	border: none;
+	padding: 8px 10px 8px 16px;
+	justify-content: space-between;
+	align-items: center;
+}
```
ashleemboyer, line 17: 🎨 `justify-content: space-between` will override `gap`/`grid-gap` if there's more space between than the value for those gaps, so both aren't needed here! Based on spacing I'm seeing in Figma ([link](https://www.figma.com/file/v0Hyqvjdf6QJAvRkdIBV24/Dark-Mode?node-id=1166-44816&t=gB7xuXM6NrfgBgE9-11)), I'd probably define `gap: 12px` and remove `justify-content`. This would have to be done in combination with some other tweaks I noted like the `max-width` comment above.

CleanShot 2023-03-31 at 13 00 05

src/components/user-dropdown-disclosure/index.tsx
ashleemboyer: I think that's an amazing idea! 🙌🏻  I believe it was @BRKalow who did some similar refactoring many months ago, and created `__config.dev_dot.non_product_mobile_menu_routes`. We could name it something similar like `non_themed_routes`.

src/components/user-dropdown-disclosure/index.tsx
```diff
@@ -91,6 +98,9 @@ const UserDropdownDisclosure = ({
 				</>
 			) : null}
 			{items.map(renderItem)}
+			{shouldRenderThemeSwitcher ? (
+				<UserDropdownDisclosureThemeSwitcher />
+			) : null}
```
ashleemboyer, line 103: 🙋🏻 This is just a question about the shift to using an assembly branch -- Is dark mode getting released when that assembly branch is merged? Or when the the config flag is flipped? If it's going to be when the assembly branch is merged, we could probably avoid the conditionals now! 💭

ashleemboyer (commented):
Looking great! 👏🏻 Just a few things that should be quick to update, and happy to pair or discuss in Slack if that'd be helpful. 😊

src/components/navigation-header/index.tsx
kendallstrautman: Good point, In a future PR though I'll need to refactor the placement of the icons so the assembly felt like an easier thing to handle given that we want to ship in a week or so.

src/components/theme-switch/index.tsx
```diff
@@ -26,10 +38,22 @@ export default function ThemeSwitch() {
...
 	return (
-		<select value={theme} onChange={(e) => setTheme(e.target.value)}>
-			<option value="system">System</option>
-			<option value="dark">Dark</option>
-			<option value="light">Light</option>
-		</select>
+		<div className={s.root}>
+			<span className={s.themeIcon}>{ThemeIcons[theme]}</span>
+			<select
+				aria-label="Choose a different theme"
```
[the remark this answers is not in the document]
kendallstrautman: Ah okay thank you! i was worried the visible label wasn't clear enough. But I'll remove this!

src/components/theme-switch/theme-switch.module.css
```diff
@@ -0,0 +1,38 @@
+.root {
+	position: relative;
+	width: 100%;
+	color: var(--token-color-foreground-primary);
+	max-width: 110px;
```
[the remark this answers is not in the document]
kendallstrautman: Hmmm i'll think on this. The width is intentional to align with the designs, but I'll checkout the fit content. I think the issue with that is that these acons aren't _within_ that element so the content width would cut off the icons.

src/components/user-dropdown-disclosure/index.tsx
kendallstrautman: Good point, Well also flip the config in the assembly and merge the assembly for the release. I can remove the conditional here though.

src/components/theme-switch/theme-switch.module.css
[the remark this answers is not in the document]
kendallstrautman: Screenshot 2023-04-03 at 9 19 07 AM
When @braydenlove and I both zoom to 200% on our browsers we don't see the text clip. Could you share how to recreate this?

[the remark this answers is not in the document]
ashleemboyer: @kendallstrautman what I'm referring to is text-only zoom, not browser zoom. The easiest way to test 200% text-only zoom is by setting `font-size: 32px` on the `html` element in devtools.

src/components/theme-switch/theme-switch.module.css
```diff
@@ -0,0 +1,38 @@
...
+	composes: g-focus-ring-from-box-shadow-dark from global;
+	composes: hds-font-weight-medium from global;
+	composes: hds-typography-body-100 from global;
+	appearance: none;
+	background-color: var(--token-color-surface-faint);
+	border-radius: 5px;
+	border: 1px solid var(--token-color-border-strong);
+	color: var(--token-color-foreground-primary);
+	cursor: pointer;
+	padding: 3px 20px 3px 30px;
```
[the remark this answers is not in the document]
kendallstrautman: Okay I think I made all the updates that make it match the spec

src/components/user-dropdown-disclosure/index.tsx
kendallstrautman: I think i'll do this in a follow up PR!

kendallstrautman: https://app.asana.com/0/1203652232454021/1204323512217789

ashleemboyer:
@kendallstrautman taking another look now! 😊

src/components/theme-switcher/user-dropdown-switcher/user-dropdown-switcher.module.css
```diff
@@ -0,0 +1,15 @@
...
+.listItem {
+	display: flex;
+	border-radius: 5px;
+	color: var(--token-color-foreground-primary);
+	width: 100%;
+	border: none;
+	padding: 8px 16px 8px 16px;
+	gap: 12px;
+	align-items: center;
+}
```
ashleemboyer, line 15: 🎨 `border-radius`, `width`, and `border` aren't doing anything for us on this class, so they can be removed.

```suggestion
.listItem {
	display: flex;
	color: var(--token-color-foreground-primary);
	padding: 8px 16px 8px 16px;
	gap: 12px;
	align-items: center;
}
```

src/components/theme-switcher/user-dropdown-switcher/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export default function UserDropdownDisclosureThemeSwitcher() {
+	return (
+		<>
+			<DropdownDisclosureSeparatorItem />
+			<DropdownDisclosureListItem className={s.listItem}>
+				<label className={s.label} htmlFor="theme-switcher">
+					Theme
+				</label>
+				<ThemeSwitcher id="theme-switcher" />
```
ashleemboyer, line 16: 📌 💭 A breadcrumb for subsequent PRs: we'll want to make sure that all each of switchers have a unique ID, since there'll be multiple rendered on a single page.

src/components/theme-switcher/theme-switcher.module.css
```diff
@@ -0,0 +1,37 @@
+.root {
+	position: relative;
+	color: var(--token-color-foreground-primary);
+	width: 100%;
+}
+.select {
```
ashleemboyer, line 6: 🎨 Previous comment like this hasn't been resolved yet but it's not showing in the diff anymore, so just making it again for posterity's sake!

```suggestion
}

.select {
```

ashleemboyer (approved):

src/components/user-dropdown-disclosure/index.tsx
kendallstrautman: Oh yes! somehow I missed this comment this morning. Will update now

src/components/theme-switcher/user-dropdown-switcher/index.tsx
kendallstrautman: i'll make this one for the dropdown nav more specific! good point.

src/components/user-dropdown-disclosure/index.tsx
kendallstrautman: So actually I still need the conditional for the themed route, and I'll leave the config flag just for clarity at the moment.

PR #1803, opened 2023-03-31
Aligning UserDropdownDisclosure focus ring for consistency

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambfix-user-dropdown-focus-ring-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202114367927919/1204304009719193/f) 🎟️

## 🗒️ What

- Adds `activatorClassName` prop to `DropdownDisclosure`
- Adds `activatorClassName` prop to `UserDropdownDisclosure`, inherited directly from `DropdownDisclosure`
- Renames the `userDropdownDisclosure` class to `userDropdownDisclosureActivator`
- Updates the same class to compose `g-focus-ring-from-box-shadow-dark`

## 🤷 Why

- This element was the only one in the nav header that did not compose `g-focus-ring-from-box-shadow-dark` for the focus ring styles

## 🧪 Testing

- [ ] Go to any page in the app
- [ ] Check that the `focus-visible` style of `UserDropdownDisclosure` matches the other elements in the nav header

## 💭 Anything else?

- Not directly related to this PR, but a Dark Mode question for @kendallstrautman & @braydenlove: Does the `--custom-token-focus-ring-action-box-shadow-dark` variable that the `g-focus-ring-from-box-shadow-dark` global class uses need to be swapped at all when the theme switches?

### DIFF
--- a/src/components/navigation-header/navigation-header.module.css
+++ b/src/components/navigation-header/navigation-header.module.css
@@ -99,7 +99,11 @@
 	}
 }
 
-.userDropdownDisclosure button[aria-expanded] {
+.userDropdownDisclosureActivator {
+	/* composition */
+	composes: g-focus-ring-from-box-shadow-dark from global;
+
+	/* properties */
 	background-color: transparent;
 	color: var(--token-color-foreground-high-contrast);
 
--- a/src/components/dropdown-disclosure/index.tsx
+++ b/src/components/dropdown-disclosure/index.tsx
@@ -35,6 +35,7 @@ const handleError = (message: string) => {
 }
 
 const DropdownDisclosure = ({
+	activatorClassName,
 	'aria-label': ariaLabel,
 	children,
 	className,
@@ -89,7 +90,7 @@ const DropdownDisclosure = ({
 		>
 			<DropdownDisclosureActivator
 				aria-label={ariaLabel}
-				className={s.activator}
+				className={classNames(s.activator, activatorClassName)}
 				color={color}
 				hideChevron={hideChevron}
 			>
--- a/src/components/dropdown-disclosure/types.ts
+++ b/src/components/dropdown-disclosure/types.ts
@@ -25,6 +25,12 @@ interface DropdownDisclosureProps
 	extends PickedDropdownDisclosureActivatorProps,
 		PickedDisclosureProps {
 	/**
+	 * An optional `className` that is passed to the internally rendered
+	 * `DropdownDisclosureActivator`.
+	 */
+	activatorClassName?: string
+
+	/**
 	 * Text that should be announced by a screen reader when the activator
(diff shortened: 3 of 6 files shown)


src/components/navigation-header/navigation-header.module.css
```diff
@@ -99,7 +99,11 @@
 	}
 }
 
-.userDropdownDisclosure button[aria-expanded] {
+.userDropdownDisclosureActivator {
+	/* composition */
+	composes: g-focus-ring-from-box-shadow-dark from global;
```
kendallstrautman, line 104: Is there a reason we wouldn't put these focus styles on the activator directly? Do we have a scenario where we need different focus styles?

kendallstrautman:
@ashleemboyer i'm actually unclear on the use of `g-focus-ring-from-box-shadow-dark` versus `g-focus-ring-from-box-shadow`. Whats the historical context on these different classes?

ashleemboyer:
> userDropdownDisclosureActivator

@kendallstrautman here's the context:

- `g-focus-ring-from-box-shadow` was created to be a global helper for applying a consistent custom focus ring to interactive elements
- `g-focus-ring-from-box-shadow` was created as with the same purpose, just for use on elements that are on a dark background (e.g. interactive elements in the dark navigation header)

src/components/navigation-header/navigation-header.module.css
ashleemboyer: To add on to the general context I just left in reply to your other comment -- `DropdownDisclosure` is primarily used on light backgrounds outside of the dark navigation header. In this case, the `UserDropdownDisclosure` is sitting on a dark background, so it needs to have a different focus ring style.

ashleemboyer:
> @ashleemboyer I think you are asking about the specific focus colors for dark mode. If so they (and any values that should be referencing said focus styles) should be the following on dark:
> 
> --token-color-focus-action-internal: blue-300 (#389AFF) --token-color-focus-action-external: #5990FF
> 
> The -focus-action-external value is a custom hex value that doesn't reference one of the semantic ones. That color passes a11y on both backgrounds (AA) per the Graphical Objects and User Interface Components image

@braydenlove did you mean to tag @kendallstrautman in your comment? I was explaining our custom global helper CSS classes that have been around for a long time.

kendallstrautman:
@ashleemboyer are we using these action tokens in the box shadow? If not, I'd say the dark mode translation for the blue should follow the change to the action tokens that Brayden linked to.

kendallstrautman (approved):
Looks good to me! If we need to make a follow up ticket to adjust any colors / tokens for these custom box-shadows, that makes sense 💙

zchsh (approved):
✅ Checked out the preview in dark and in light mode, looks good to me! 👍

PR #1807, opened 2023-04-03
Update the integration header’s title to be to be a span instead of a h1

BrandonRomano:
We do not want this integration header to be the H1 of the page, as it remains consistent beween subpages (with different page titles). In a [slack discussion](https://hashicorp.slack.com/archives/C03RZAF1XS8/p1680542560922859) it was brought up that there are two `h1`'s on the integration pages, due to the Header component having a H1, and also due to the template having a H1.

The header component always refers to the integration, and is not the source of the page title, so I'm updating this component to be driven via a `span`.

I was considering adjusting this to be another header (like a h2), but realized that it does not fall within the header hierarchy at all.

Let's wait on merging this one until we have https://github.com/hashicorp/dev-portal/pull/1805 deployed & in.

### DIFF
--- a/src/components/content-header-card/content-header-card.module.css
+++ b/src/components/content-header-card/content-header-card.module.css
@@ -181,3 +181,7 @@
 		}
 	}
 }
+
+.title {
+	composes: hds-typography-display-300 from global, hds-font-weight-bold from global;
+}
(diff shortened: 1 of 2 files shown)


zchsh (approved):
✅ Seems like a reasonable solution to me!

I did a quick GitHub search to confirm `ContentHeaderCard` is only used for integrations (wasn't sure, figured it'd be worth a double-check in case the `h1` removal could have impact elsewhere).

PR #1811, opened 2023-04-03
feat: theme switcher in footer

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-kstheme-switcher-footer-hashicorp.vercel.app/waypoint) 🔎
- [Asana task](https://app.asana.com/0/1203652232454021/1204303889362369) 🎟️
- [Figma](https://www.figma.com/file/v0Hyqvjdf6QJAvRkdIBV24/Dark-Mode?node-id=815-40857&t=i4cJ02K2WSH382hp-0) 🎨 

## 🗒️ What

This PR adds the theme select in the footer. 
Screenshot 2023-04-03 at 5 37 03 PM
Screenshot 2023-04-03 at 5 37 12 PM
Screenshot 2023-04-03 at 5 42 51 PM

## 🛠️ How

I used the `ThemeSelect` primitive directly in the footer and added and `ariaLabel` prop since there isn't a visual label in the design.

## 📸 Design Screenshots

Screenshot 2023-04-03 at 5 32 05 PM

## 🧪 Testing

- [Visit the preview](https://dev-portal-git-kstheme-switcher-footer-hashicorp.vercel.app/waypoint), checkout the footer
- Ensure that it matches the spec and scales between desktop, tablet and mobile.
- Also ensure that the switcher doesn't render on the non-themed views (the home page). 

## 💭 Anything else?

This PR only addresses the addition of the theme switcher, but not other style updates in the designs related to the consent manager button or other footer updates.


src/components/footer/index.tsx
```diff
@@ -131,17 +135,19 @@ function Footer({
...
 						href={FEEDBACK_FORM_URL}
 						color="secondary"
 						size="small"
 						icon={<IconPencilTool16 />}
 						opensInNewTab={true}
-						className={classNames(s.feedbackButton, s.desktop)}
 					/>
 				</li>
+				<li className={classNames(s.actionDesktop, s.linkListItem)}>
+					<ThemeSelect />
```
kendallstrautman, line 149: Probably need to add an aria label since there isn't a visible label

braydenlove: Good call!

src/components/footer/footer.module.css
```diff
@@ -7,10 +7,10 @@
 
 .root {
 	--padding: 24px;
-	--logo-button-margin-bottom: 8px;
```
kendallstrautman, line 10: Was able to remove this in lieu of `gap`.

src/components/footer/footer.module.css
```diff
@@ -7,10 +7,10 @@
...
 
 	align-items: center;
 	color: var(--token-color-palette-neutral-600);
+	gap: 12px;
```
kendallstrautman, line 13: I sort of eyeballed this, could use some nuance here.

src/components/footer/index.tsx
```diff
@@ -60,6 +62,9 @@ function Footer({
 	openConsentManager,
 	className,
 }: FooterProps): React.ReactElement {
+	const { pathname } = useRouter()
+	const shouldRenderThemeSwitcher = pathname !== '/' && pathname !== 'sign-up'
```
kendallstrautman, line 66: Again, will refactor this in a follow up PR https://app.asana.com/0/1203652232454021/1204323512217789

zchsh (approved):
✅ This looks good to me! Nothing blocking.

---

💭 I have a couple of _very_ nitpicky pieces of feedback on some specs that look just a little off to me. Here's a couple screenshots of the `Give Feedback` and theme dropdown for context, these nitpicks apply across viewports:

| | |
| - | - |
| CleanShot 2023-04-03 at 20 51 02@2x | CleanShot 2023-04-03 at 20 49 05@2x |

- Height difference - the `Give Feedback` button is `30px` tall, where the theme dropdown is `28px`. I'm not sure which is correct from a design perspective (in the specific Figma linked it looks like `28px`?), but the inconsistency stuck out a smidge!
- Icon side difference - the `Give Feedback` button has a `16px` square icon, while the theme dropdown has `12px` square. 

On both fronts, I wasn't even sure there was inconsistency at first, had to double-check, and even then not sure if intentional or not, so IMO non-blocking 🤷  But wanted to flag in case we do have a stronger design opinion on how-whether to establish consistency there 👍

braydenlove:
@zchsh good eyes, you caught the few tiny things I saw as well. Kendall I will get you very specific values tomorrow and we can get this thing merged in the morning. Thanks!!

braydenlove:
Let's remove the icon for give feedback, I don't think it adds much and is visually conflicting with the theme selector in a negative way.

braydenlove:
Screen Shot 2023-04-04 at 9 08 01 AM

Last thing I noticed. Can we get the grid-gap and gap to be 16px instead of 12px? This will keep the spacing a bit more refined throughout the smaller viewports when it starts wrapping. Otherwise this LGTM.

braydenlove (approved):
Looks good, thanks!

src/components/footer/index.tsx
```diff
@@ -131,17 +140,20 @@ function Footer({
...
 						color="secondary"
 						size="small"
-						icon={<IconPencilTool16 />}
 						opensInNewTab={true}
-						className={classNames(s.feedbackButton, s.desktop)}
 					/>
 				</li>
+				{shouldRenderThemeSwitcher ? (
+					<li className={classNames(s.actionDesktop, s.linkListItem)}>
+						<ThemeSelect ariaLabel="Select a theme" />
```
ashleemboyer, line 154: 🚧 Blocking: With this value for `aria-label`, the `<select>`'s Accessible Name becomes "Select a theme". What that means for voice control users, is that they have to speak "Select a theme" in order to activate the element. Since there's no visible label with this text (hence why `aria-label` was reached for), voice control users aren't going to be able to activate this element except by pure luck and guessing that text.

🙋🏻 A few questions for my own context gathering:

- Did we explore adding a visible label for this switcher?
- Did we explore placing a label "inside" the switcher, maybe something like `Theme: ${option.label}`?

ashleemboyer (changes requested):

src/components/footer/index.tsx
kendallstrautman: @ashleemboyer is there a way to communicate the label to voice control users without visibly showing the label?

ashleemboyer: Additional context will help me make a recommendation here 😊

kendallstrautman: Screenshot 2023-04-04 at 11 39 55 AM
We use the same approach on our install version switcher

braydenlove: We wanted to be lighter on the footer treatment as:

1. The footer is the secondary place to change theme (need to validate this hypothesis)
2. There is a ton of content in the footer, adding a label creates some complications with the footer designs especially when we get into the smaller viewports.

kendallstrautman: Screenshot 2023-04-04 at 12 12 59 PM
@ashleemboyer could you share why its okay to not have a visual label here but we need to refactor to add one in this case?

kendallstrautman: This feedback wasn't in the initial design review and it feels a late to fully change up the designs to support this. Could we potentially merge this iteration in and then follow up to address the label issue when we have space to explore the right design solution.

kendallstrautman: @ashleemboyer I updated the `aria-label` to read `theme` and I added a ticket in this aria label improvements task here to follow up on this. https://app.asana.com/0/1202097197789424/1204334637222649/f

Since this requires a design dependency, we would like to move forward with this implementation as is and address the above concerns separately in a design exercise.

ashleemboyer (changes requested):
Going to re-mark this as not approved on my end, pending the conversation I moved over to Slack for easier discussion! 🙏🏻

ashleemboyer:
@kendallstrautman after this branch finishes rebuilding, I'll give this PR another review!

ashleemboyer (approved):
Approving based on our Slack discussions and PR #1820 that was reviewed, approved, and merged into this branch! 👍🏻

And for future reference + posterity, here's a little screen recording on how I went about testing this a little more quickly, without having to enable speech recognition software.

https://user-images.githubusercontent.com/43934258/230233374-9267e19c-2f29-436f-a234-32329ea34e27.mp4

ashleemboyer (changes requested):
There was a potential bug called out in Slack based on the changes I made in the sibling PR, I'm going to block this merging for a sec to investigate! It'll be easier for me to debug here than in the assembly. 🙏🏻

ashleemboyer (approved):
Figured it out, so re-approving! ✅

The issue identified is that `UserDropdownDisclosureThemeSwitcher` isn't associating its `label` with `ThemeSwitcher` correctly. Have created a task on the Dark Mode board for it!

kendallstrautman:
Closing in favor of https://github.com/hashicorp/dev-portal/pull/1831

PR #1820, opened 2023-04-05
[ThemeSelect] Auto generating aria-label when no id

ashleemboyer:
## 🗒️ What

- Branching off of #1811, updates `ThemeSelect` to auto generate the rendered `<select>`'s `aria-label`
- Also lifts up a `THEME_CONFIGS` constant that stores each theme's `icon`, `label`, and `text`

## 🤷 Why

- The `aria-label` originally proposed in #1811 poses an issue for speech recognition users: the text is not visible, so not intuitive for these users to know how to interact with the element
- We can't totally omit `aria-label` in this case because `<select>` elements do not set their Accessible Name based on their content. This contrasts how `DropdownDisclosure` works, which uses a `<button>` under the hood.
- These changes emulate how `DropdownDisclosure` would behave if we used it here, which is that the Accessible Name becomes the text of the currently selected theme.

## 🧪 Testing

- [ ] Go to a themed page on the preview URL
- [ ] Scroll to the footer of the site
- [ ] Inspect the `ThemeSelect` in your browser devtools
- [ ] Open the Accessibility tab to view the accessibility tree
- [ ] Confirm the Accessible Name for the element matches the text of the currently selected theme

## 💭 Anything else?

- Worth noting this backlog task regarding `DropdownDisclosure` for posterity: [Reconsider updating the ThemeSwitcher to use a dropdown instead of native select](https://app.asana.com/0/1203652232454021/1204284168518969/f)

### DIFF
--- a/src/components/footer/index.tsx
+++ b/src/components/footer/index.tsx
@@ -82,7 +82,7 @@ function Footer({
 					opensInNewTab={true}
 					className={s.feedbackButton}
 				/>
-				{shouldRenderThemeSwitcher ? <ThemeSelect ariaLabel="theme" /> : null}
+				{shouldRenderThemeSwitcher ? <ThemeSelect /> : null}
 			</span>
 			<ul className={s.links}>
 				{FOOTER_ITEMS.map((item: FooterItem, index: number) => {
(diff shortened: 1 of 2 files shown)


kendallstrautman (approved):
LGTM, thank you for this.

braydenlove (approved):

PR #1821, opened 2023-04-05
chore: removes dark mode flag

kendallstrautman:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1199634971449915/1204323590906663) 🎟️

## 🗒️ What

Removes all use of the dark mode flag. This is what will 'ship' the features when we merge the assembly and also does some preemptive cleanup since we moved to the assembly branch.

## 🧪 Testing

- Validate that all the theme switching UI renders as expected.


src/layouts/base-new/index.tsx
```diff
@@ -95,8 +84,8 @@ const BaseNewLayout = ({
 					hideOnMobile
 				/>
 			)}
-			<Layout {...layoutProps}>
-				<div className={s.root} data-layout="base-new">
+			<CoreDevDotLayoutWithTheme theme={theme}>
+				<div className={s.root} data-CoreDevDotLayoutWithTheme="base-new">
```
zchsh, line 88: ⛏️ What does `data-layout` (or now `data-CoreDevDotLayoutWithTheme`) do? I took a peek in code and couldn't find anything... (searched for both `data-layout` and `dataset` use, and found neither).

I have a feeling we _might_ be able to drop this 🤔 

```suggestion
				<div className={s.root}>
```

zchsh (approved):
✅ 🙌 Nice, I dig the preemptive cleanup 💯 

⛏️ Had one lil nitpick about a potentially-not-needed `data-` attribute, but not really sure about it and doesn't seem like it'll cause any harm either way, so didn't feel blocking.

src/layouts/base-new/index.tsx
kendallstrautman: Oh whoops I think that was a find and replace error from me. I'll look into data-layout though to validate whether we still need that. Thanks for the catch!

kendallstrautman: I updated it back to the original but im a bit unsure of removing in this PR since I don't have the context on what it used to be used for!

PR #1830, opened 2023-04-06
Remove Override from Integration Header

BrandonRomano:
An issue cropped up which was hot-fixed via https://github.com/hashicorp/dev-portal/pull/1825, which addressed an issue where the styles were no longer targeting the right element (leading to the links in this header component to be incorrect).

This PR removes the Override from the integrations header. After some discussion with Design, we landed on the fact that we are OK not including the permanent underline, in favor of using the StandaloneLink as designed.

[🔍 Preview](https://dev-portal-git-brintegration-header-remove-override-hashicorp.vercel.app/waypoint/integrations/hashicorp/aws-ami). 

> There is an outstanding bug right now, unrelated to this PR, where underlines are displaying the incorrect color, which is being worked on currently.

### DIFF
--- a/src/components/content-header-card/content-header-card.module.css
+++ b/src/components/content-header-card/content-header-card.module.css
@@ -127,16 +127,18 @@
 				&:not(:first-of-type) {
 					margin-left: 12px;
 				}
-				& > * {
-					/* TODO: Rather than override our existing link component, 
-					we should eject or build in support for what we're trying
-					to do here. These styles assume the StandaloneLink styles
-					are part of its contract, but they are not and are subject
-					to change */
-					text-decoration: underline;
-					color: var(--token-color-foreground-primary);
-					&:hover {
-						color: var(--token-color-foreground-strong);
+				& > a {
+					& > * {
+						/* TODO: Rather than override our existing link component,
+						we should eject or build in support for what we're trying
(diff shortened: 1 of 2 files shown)


zchsh (commented):
💭 Not sure if this is within the scope of this PR, but there's what I think is an upstream bug in `StandaloneLink`:



- `text-decoration: underline` is being set on the `<a />` `StandaloneLink` element...
- _but_ the `color-${color}` class is only set on the `StandaloneLinkContents` element...

... so the result is the underline doesn't match the `secondary` color scheme:

CleanShot 2023-04-06 at 14 55 25@2x

I think one way to fix this would be to add the `color` class the the `<a />` element in `StandaloneLink` ([here](https://github.com/hashicorp/dev-portal/blob/main/src/components/standalone-link/index.tsx#L56)):

```ts
const classes = classNames(s.root, className, s[`color-${color}`])
```

zchsh (approved):
✅ Approving cause:
1. The issue mentioned above will be handled in a separate PR, which we could wait on before merging this
2. I don't think we _need_ to wait to merge this, it's such a small difference so would be just fine to merge as-is (and said separate PR will deliver the underline color fix)

BrandonRomano:
I vote we just wait until the underline is fixed in the separate PR; no need to rush this one.

PR #1831, opened 2023-04-06
feat(footer): theme switcher with label

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-kstheme-switcher-footer-v2-hashicorp.vercel.app/waypoint) 🔎
- [Asana task](url) 🎟️
- [Figma](https://www.figma.com/file/v0Hyqvjdf6QJAvRkdIBV24/Dark-Mode?node-id=1321-47189&t=5zgruva2fsHwF9pv-0) 🎨 

## 🗒️ What

A new iteration from #1811 where there is a visible label with the theme select. 

Screenshot 2023-04-06 at 11 39 48 AM
Screenshot 2023-04-06 at 11 39 43 AM
Screenshot 2023-04-06 at 11 39 38 AM
Screenshot 2023-04-06 at 11 38 48 AM

## 🤷 Why

To make the component accessible for both voice control and keyboard users. 

## 📸 Design Screenshots
Screenshot 2023-04-06 at 11 43 28 AM

## 🧪 Testing

- [Visit the preview](https://dev-portal-git-kstheme-switcher-footer-v2-hashicorp.vercel.app/waypoint)
- Validate that the theme switcher renders as expected in both dark and light mode. 
- Test with Voice Over, validate that the label `Theme` is read.


src/components/footer/footer.module.css
```diff
@@ -6,22 +6,31 @@
...
+	gap: 24px;
 	display: flex;
 	flex-wrap: wrap;
-	justify-content: space-between;
-	padding: var(--padding);
+	padding: 24px;
 	position: relative;
+	flex-direction: column;
+
+	&.row {
```
kendallstrautman, line 17: For the home page since no switcher renders

src/components/footer/index.tsx
```diff
@@ -126,22 +142,9 @@ function Footer({
...
 					return (
 						// eslint-disable-next-line react/no-array-index-key
-						<li className={s.linkListItem} key={index}>
-							{innerElement}
-						</li>
+						<li key={index}>{innerElement}</li>
 					)
 				})}
-				<li>
-					<ButtonLink
```
kendallstrautman, line 135: Since the switcher stays in the same place between mobile and desktop now we can remove this duplicate markup and the visual order matches the focus order always.

braydenlove (approved):
Looks good, thanks.

zchsh (approved):
✅ 🙌 Followed testing steps, looks great 👍

src/components/footer/index.tsx
```diff
@@ -107,13 +121,15 @@ function Footer({
...
+
 							<a
 								className={s.linkAction}
 								href={item.href}
 								target={item.opensInNewTab ? '_blank' : undefined}
 								rel={item.opensInNewTab ? 'noreferrer' : undefined}
 							>
 								{textElement}
+								{item.opensInNewTab ? <IconExternalLink16 /> : null}
 							</a>
```
ashleemboyer, line 133: 📌 I think we probably want to leverage `components/Link` here since it handles making sure "opens in new tab" is communicated to screen reader users as well?

src/components/theme-switcher/index.tsx
```diff
@@ -8,23 +8,42 @@ import { IconCaret16 } from '@hashicorp/flight-icons/svg-react/caret-16'
 import { IconMonitor16 } from '@hashicorp/flight-icons/svg-react/monitor-16'
 import { IconMoon16 } from '@hashicorp/flight-icons/svg-react/moon-16'
 import { IconSun16 } from '@hashicorp/flight-icons/svg-react/sun-16'
-import { useState, useEffect } from 'react'
+import classNames from 'classnames'
+import { useState, useEffect, ReactElement } from 'react'
```
ashleemboyer, line 12: 🎨 Should we adjust import order while we're here?

src/components/theme-switcher/index.tsx
```diff
@@ -58,11 +79,24 @@ export default function ThemeSelect({ id }: ThemeSelectProps) {
...
 		<>
-			<label className={s.label} htmlFor={id}>
+			<label
+				className={classNames(
+					visuallyHideLabel ? 'g-screen-reader-only' : undefined,
+
+					s.label
+				)}
+				htmlFor={id}
+			>
```
ashleemboyer, line 99: 🙋🏻 Can we remove `visuallyHideLabel` for now since it's not used? I don't think we want to make it available for use unless absolutely necessary.

src/components/mobile-menu-container/index.tsx
```diff
@@ -17,7 +17,7 @@ import Button from 'components/button'
 import ButtonLink from 'components/button-link'
 
 // Local imports
-import { ThemeSelectWithLabel } from 'components/theme-switcher'
+import ThemeSelectWithLabel from 'components/theme-switcher'
```
ashleemboyer, line 20: 📌 Can we align the default exported component and file name as follow-up when this is stable?

ashleemboyer (approved):
This looks great! 🙌🏻 So glad we were able to land on a visible label. 💖

Checked the Accessible Name of all 3 locations of the switcher while testing and they're consistent and correctly associated with their labels ✅

src/components/theme-switcher/index.tsx
kendallstrautman: What order were you thinking? I always think 'external' at the top.

src/components/theme-switcher/index.tsx
kendallstrautman: Yeah sure!

src/components/mobile-menu-container/index.tsx
kendallstrautman: Sure, I keep waffling on the name 😆  Do you have thoughts on what this should be called?

src/components/theme-switcher/index.tsx
ashleemboyer: Yeah same! And then within "external" we've commonly done React, then Next, then totally external (like classnames), then HashiCorp (like flight)

src/components/mobile-menu-container/index.tsx
ashleemboyer: I think `theme-switcher` + `ThemeSwitcher` makes sense! We've used similar names for context-switch-y type components before, and if they're all in Swingset makes them slightly easier to search by "switcher"

kendallstrautman: Sweet, makes sense!

src/components/theme-switcher/index.tsx
kendallstrautman: ah okay, I'll reorder.

src/components/footer/index.tsx
kendallstrautman: Just updated!

PR #1823, opened 2023-04-06
[Home Page Redesign] Styling FeaturedSearchCard

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambhome-page-search-card-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/0/1204200670480305/f) 🎟️

## 🗒️ What

- Updates `src/lib/hooks/usePrefersReducedMotion.js` to no longer use the deprecated `addListener` function
- Breaks the `FeaturedContentGrid` file into several new ones (without functionality changes)
- Adds functionality & styling to `SearchFeaturedCard`

## 📸 Design Screenshots

[Figma link](https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/Developer?node-id=20706-250693&t=GfrwDOE8y8zM5YU5-11)

<details>
  <summary>Expand for screenshot</summary>
  
</details>

## 🧪 Testing

Each of the following should be tested in the following browsers:

- [ ] Chrome
- [ ] Safari
- [ ] Firefox

### Auto scrolling

- [ ] When reduced motion is turned on, the search card should not auto scroll
- [ ] When reduced motion is turned off, the search card should auto scroll and repeat through the options
- [ ] When hovering over the scrollable container, auto scroll should permanently disable
- [ ] When moving keyboard focus into the scrollable container, auto scroll should permanently disable

### Focus/centering

- [ ] When using `TAB`/`SHIFT`+`TAB` through the buttons, the focused button should become centered
- [ ] Bringing focus to the first or last button should center them, if they're not already centered

### Other tests

- [ ] The Accessible Name for each button should match the suggested search text (excludes "Explore")
- [ ] Clicking the non-centered buttons to the direct left and right of the centered button should center the clicked button
- [ ] Clicking the centered button should open the `CommandBar` with results matching the button's suggested term

## 💭 Anything else?

Only 3 files need closer review, others are just copy/pasted code from their old file:

- `src/views/homepage/components/featured-content-grid/search-featured-card.tsx`
- `src/views/homepage/components/featured-content-grid/search-featured-card.module.css`
- `src/views/homepage/img/search-card-graphic.svg`

PR #1832, opened 2023-04-06
[StandaloneLinkContents] Fixing how color prop is handled

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambfix-standalone-link-contents-hashicorp.vercel.app/) 🔎

## 🗒️ What

- Updates `StandaloneLinkContents`'s `.contents` class to inherit it's `color` when nested inside of `StandaloneLink`
- Adds `DEFAULT_COLOR_VARIANT` since it's needed in 2 places in the file now
- Simplifies the classname associated with `StandaloneLink`'s `color` prop (easier to maintain 2 references to)
- Fixes the color classes to only apply a color change on active/hover if the element it's applied to has an `href` property

## 🧪 Testing

- [ ] Go to [the `StandaloneLink` Swingset page](https://dev-portal-git-ambfix-standalone-link-contents-hashicorp.vercel.app/swingset/components/standalonelink)
- [ ] The `StandaloneLink` examples should behave as expected when hovered
- [ ] The `StandaloneLink` examples should behave as expected when active
- [ ] The `StandaloneLinkContents` examples should behave as expected when hovered
- [ ] The `StandaloneLinkContents` examples should behave as expected when active

### DIFF
--- a/src/components/standalone-link/index.tsx
+++ b/src/components/standalone-link/index.tsx
@@ -3,7 +3,6 @@
  * SPDX-License-Identifier: MPL-2.0
  */
 
-import { ReactElement } from 'react'
 import classNames from 'classnames'
 import Link from 'components/link'
 import {
@@ -12,21 +11,18 @@ import {
 } from './types'
 import s from './standalone-link.module.css'
 
+const DEFAULT_COLOR_VARIANT = 'primary'
+
 const StandaloneLinkContents = ({
 	className,
-	color,
+	color = DEFAULT_COLOR_VARIANT,
 	icon,
 	iconPosition,
 	size,
 	text,
 	textClassName,
 }: StandaloneLinkContentsProps) => {
-	const containerClasses = classNames(
-		s.contents,
-		s[`color-${color}`],
-		s[size],
-		className
-	)
+	const containerClasses = classNames(s.contents, s[size], s[color], className)
 	const textClasses = classNames(s.text, textClassName)
 
 	return (
@@ -41,8 +37,8 @@ const StandaloneLinkContents = ({
 const StandaloneLink = ({
 	ariaLabel,
 	className,
+	color = DEFAULT_COLOR_VARIANT,
 	'data-heap-track': dataHeapTrack,
-	color = 'primary',
 	download,
 	href,
 	icon,
@@ -52,8 +48,8 @@ const StandaloneLink = ({
 	size = 'medium',
 	text,
 	textClassName,
-}: StandaloneLinkProps): ReactElement => {
-	const classes = classNames(s.root, className)
+}: StandaloneLinkProps) => {
+	const classes = classNames(s.root, s[color], className)
 	const rel = opensInNewTab ? 'noreferrer noopener' : undefined
 
 	return (
@@ -68,7 +64,6 @@ const StandaloneLink = ({
 			opensInNewTab={opensInNewTab}
 		>
 			<StandaloneLinkContents
-				color={color}
 				icon={icon}
 				iconPosition={iconPosition}
 				size={size}
--- a/src/components/standalone-link/index.tsx
+++ b/src/components/standalone-link/index.tsx
@@ -5,6 +5,7 @@
 
 import classNames from 'classnames'
 import Link from 'components/link'
+import { ToastColor, developmentToast } from 'components/toast'
 import {
 	type StandaloneLinkContentsProps,
 	type StandaloneLinkProps,
@@ -15,14 +16,29 @@ const DEFAULT_COLOR_VARIANT = 'primary'
 
 const StandaloneLinkContents = ({
 	className,
-	color = DEFAULT_COLOR_VARIANT,
+	color,
 	icon,
 	iconPosition,
+	inheritColor = false,
 	size,
 	text,
 	textClassName,
 }: StandaloneLinkContentsProps) => {
-	const containerClasses = classNames(s.contents, s[size], s[color], className)
+	if (color && inheritColor) {
+		developmentToast({
+			color: ToastColor.warning,
+			title: 'Warning in `StandaloneLinkContents`',
+			description:
+				'`StandaloneLinkContents` does not accept both `color` and `inheritColor`; `inheritColor` takes precedence.',
+		})
+	}
+
+	const containerClasses = classNames(
+		s.standaloneLinkContents,
+		s[size],
+		!inheritColor && s[color ?? DEFAULT_COLOR_VARIANT],
+		className
+	)
 	const textClasses = classNames(s.text, textClassName)
 
 	return (
@@ -49,7 +65,7 @@ const StandaloneLink = ({
 	text,
 	textClassName,
 }: StandaloneLinkProps) => {
-	const classes = classNames(s.root, s[color], className)
+	const classes = classNames(s.standaloneLink, s[color], className)
 	const rel = opensInNewTab ? 'noreferrer noopener' : undefined
 
 	return (
@@ -66,6 +82,7 @@ const StandaloneLink = ({
 			<StandaloneLinkContents
 				icon={icon}
 				iconPosition={iconPosition}
+				inheritColor
 				size={size}
 				text={text}
 				textClassName={textClassName}
--- a/src/components/standalone-link/standalone-link.module.css
+++ b/src/components/standalone-link/standalone-link.module.css
@@ -15,10 +15,7 @@
 	text-decoration: none;
 	width: fit-content;
 
-	&:active {
-		text-decoration: underline;
-	}
-
+	&:active,
 	&:hover {
 		text-decoration: underline;
 	}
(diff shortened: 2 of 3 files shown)


ashleemboyer:
🟡 We'll probably have to wait to merge this until https://github.com/hashicorp/dev-portal/pull/1830 is reviewed and merged?

src/components/standalone-link/standalone-link.module.css
```diff
@@ -32,6 +29,14 @@
 		margin: -5px;
 		padding: 5px;
 	}
+
+	/*
+  When rendered inside of `StandaloneLink`, prevent `StandaloneLinkContents`
+  from setting its own color.
+  */
+	& .contents {
+		color: inherit !important;
```
zchsh, line 38: 🚧 If possible, I think it'd be preferable to avoid the `!important` here... easier said than done though 😅 

One way to do this might be pass in an optional "inheritColor" boolean prop to `StandaloneLinkContents`. When that prop is set to `true`, the `color` classes would not be applied.

src/components/standalone-link/standalone-link.module.css
```diff
@@ -58,22 +63,25 @@
...
 
-	&:active {
+	/* only change color on active for <a href> */
+	&[href]:active {
 		color: var(--token-color-foreground-action-active);
 	}
 
-	&:hover {
+	/* only change color on active for <a href> */
+	&[href]:hover {
```
zchsh, line 75: 💭 For clarity here, this means the `:hover` color change effect will no longer apply to `StandaloneLinkContents`, but the underline-on-hover will still apply, right? I think we might want to retain the `:hover` for now given that's the behaviour upstream 🤔

zchsh (changes requested):

zchsh:
> 🟡 We'll probably have to wait to merge this until https://github.com/hashicorp/dev-portal/pull/1830 is reviewed and merged?

@ashleemboyer I think it might be other way around - I think we want to merge this PR first, so that the override removal in #1830 doesn't affect the underline color.

src/components/standalone-link/standalone-link.module.css
ashleemboyer: I like this idea! I can't remember if I already tried it last night or not; maybe it was just something similar I tried that TypeScript yelled at me about 😅

src/components/standalone-link/standalone-link.module.css
ashleemboyer: Nope! The underline-on-hover only happens in the `.root` class, which is only applied to `StandaloneLink`'s `<a>`. that's not super clear from just reading the CSS though. I'm going to rename some classes!

The purpose of this `href` specificity check is to make it so that `StandaloneLinkContents` does not change color on hover. This is because the component itself is just plain text, and not interactive. If color _does_ need to change on hover/active of an element `StandaloneLinkContents` is inside of, it can easily be layered on top of the base CSS for `StandaloneLinkContents`.

Since `StandaloneLinkContents` is not used anywhere on its own but the Swingset page, this separation shouldn't have any downstream affects.

ashleemboyer:
> > 🟡 We'll probably have to wait to merge this until #1830 is reviewed and merged?
> 
> @ashleemboyer I think it might be other way around - I think we want to merge this PR first, so that the override removal in #1830 doesn't affect the underline color.

@zchsh ohhhhh I understand what you were calling out over there now! Yes, agreed now! We can merge this one first, then that one. 👍🏻

---

I believe the changes I've added in fd4c4b069936c1978761f1a0e8b42af2f4033aa7 resolve both of your comments now!

- Added `inheritColor`, with a fancy `developmentToast` in the case that both `color` and `inheritColor` are passed (an ambiguous case that may be hard to debug without that information)
- Removed the `!important` part, and conditionally apply the color helper class in `StandaloneLinkContents` based on `inheritColor`
- Renamed `.root` to `.standaloneLink`, and `.contents` to `.standaloneLinkContents`
- Updated the `&[href]:active`/`&[href]:hover` selectors to be clearer with `&.standaloneLink:active`/`&.standaloneLink:hover` (I didn't think this would work yesterday, but it did! 🎉)

What do you think?

src/components/standalone-link/index.tsx
```diff
@@ -68,9 +80,9 @@ const StandaloneLink = ({
 			opensInNewTab={opensInNewTab}
 		>
 			<StandaloneLinkContents
-				color={color}
 				icon={icon}
 				iconPosition={iconPosition}
+				inheritColor
```
BrandonRomano, line 85: It is taking me a little while to understand this `inheritColor` behavior...

It appears that on the `StandaloneLinkContents`, setting `inheritColor` simply just disallows the usage of the `color` attribute.  Are there any other behaviors here that I am missing?

It also appears that the only place that we would ever use `inheritColor=true` is on this line here (to inherit the color as set from the `StandaloneLink`).

If those is correct, could we simply just remove this `inheritColor` prop and simply just _not_ pass a color to the `StandaloneLinkContents`?

ashleemboyer: @BrandonRomano this section of code already doesn't pass `color` down to `StandaloneLinkContents`. Because `StandaloneLinkContents` has a `color` prop, and by default applies the helper classes based on that prop, `StandaloneLink` needs a way to disable that behavior so it can handle the color setting _for_ `StandaloneLinkContents`.

Here's an example of how `StandaloneLinkContents` will be used on the home page:
https://github.com/hashicorp/dev-portal/pull/1823/files#diff-4cefbd1b2095426710643ee7c5e1f98fa0ebda085a4ce03b49e76b175b5f0f71R67-R72

I basically need all of `StandaloneLink`'s behavior and styling, aside from a link being rendered and color changing on active/hover. So I need `color`, `icon`, `iconPosition`, and `text` handling here, and in the other featured cards in this section of the page those props and `size` will be useful.

src/components/standalone-link/standalone-link.module.css
BrandonRomano: I spun up this branch here locally and was poking around here, and here's a (very sloppy) commit which illustrates my exploration here.  I believe the changes I make (which remove the inherit prop) I believe still yields a similar result.

https://github.com/hashicorp/dev-portal/commit/f3437e3180a7da744382117cbf738eb1d4dc6af1

I'll have to noodle on your approach more tomorrow.

ashleemboyer: @BrandonRomano unfortunately it doesn't yield the same result, because the active/hover color style changes only apply to `.standaloneLink`. So when both `.standaloneLink` and `.standaloneLinkContents` have the same color class applied, `.standaloneLink`'s color changes on active/hover, but `.standaloneLinkContents`'s color stays the same and overwrites the color `.standaloneLink` would otherwise pass down for `.standaloneLinkContents` to inherit.

Here's how I tested:

- [ ] Go to the `StandaloneLink` Swingset page for your branch: https://dev-portal-git-brstandalone-link-diff-hashicorp.vercel.app/swingset/components/standalonelink
- [ ] Scroll down to one of the `StandaloneLink` examples
- [ ] Inspect the `a.standaloneLink` element of one of the examples
- [ ] Note what the `color` is set to

CleanShot 2023-04-06 at 21 32 36@2x

- [ ] Inspect the `div.standaloneLinkContents` element directly inside of the `a.standaloneLink` element
- [ ] Note what the `color` is set to

CleanShot 2023-04-06 at 21 34 11@2x

- [ ] Inspect the `a.standaloneLink` element again
- [ ] Force the `:hover` state on the element using devtools
- [ ] Note what the `color` is now set to (it's correctly updating based on the color classname)

CleanShot 2023-04-06 at 21 35 23@2x

- [ ] With the `:hover` state still forced on `a.standaloneLink`, inspect the `div.standaloneLinkContents` element directly inside of the `a.standaloneLink` element
- [ ] Note what the `color` is set to (it's still the base color, and not matching `a.standaloneLink`)

CleanShot 2023-04-06 at 21 40 10@2x

BrandonRomano: Thank you for walking me through this here!

src/components/standalone-link/index.tsx
zchsh: @BrandonRomano @ashleemboyer If I'm following Brandon's suggestion, I think an alternative to the `inheritColor` prop could be to _remove_ the `DEFAULT_COLOR_VARIANT` fallback from `StandaloneLinkContents`. Then we could not pass `color` when rendering `StandaloneLinkContents` in `StandaloneLink`. l think that's a pretty good idea, and could remove some props complexity 👍

I think the tradeoff would be it'd then be maybe easier to "misuse" `StandaloneLinkContents` - it'd be possible to render that component without a `color` prop, resulting in unexpected but difficult-to-warn-against appearance.

Personally happy with either solution 🌻

zchsh (approved):

PR #1833, opened 2023-04-07
refactor: non-themed routes to config

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksthemed-product-icon-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1199634971449915/1204323512217789) 🎟️

## 🗒️ What

So we can control the non-themed paths from one spot. 

## 🛠️ How

I added a utility for checking whether the current path is _themed_ and used that where we render theming UI. 

## 🧪 Testing

- [Visit the preview](https://dev-portal-git-ksthemed-product-icon-hashicorp.vercel.app/)
- Validate that the themed ui (user nav dropdown, mobile nav, & footer) shows up on all routes except the homepage and the sign up path

### DIFF
--- a/src/lib/isThemedPath.ts
+++ b/src/lib/isThemedPath.ts
@@ -1,7 +1,3 @@
 export default function isThemedPath(currentPath: string): boolean {
-	return (
-		__config.dev_dot.non_themed_paths.findIndex(
-			(nonThemedPath: string) => nonThemedPath === currentPath
-		) === -1
-	)
+	return !__config.dev_dot.non_themed_paths.includes(currentPath)
 }
--- a/src/lib/isThemedPath.ts
+++ b/src/lib/isThemedPath.ts
@@ -1,0 +1,7 @@
+export default function isThemedPath(currentPath: string): boolean {
+	return (
+		__config.dev_dot.non_themed_paths.findIndex(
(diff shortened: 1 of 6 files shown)


src/lib/isThemedPath.ts
```diff
@@ -0,0 +1,7 @@
+export default function isThemedPath(currentPath: string): boolean {
+	return (
+		__config.dev_dot.non_themed_paths.findIndex(
+			(nonThemedPath: string) => nonThemedPath === currentPath
+		) === -1
```
zchsh, line 5: ⛏️ Very small nitpick: we could use [Array.includes()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) here if we want, might be a little cleaner:

```suggestion
		__config.dev_dot.non_themed_paths.includes(currentPath)
```

zchsh (approved):
✅ 🙌 I dig this, followed testing steps, seeing no theme switcher on Home Page, and seeing it on all other pages I visited 💯 

⛏️ Had the nit-pickiest of nit-picks on using `Array.includes`, but not at all blocking, definitely feel free to ignore it 👍

PR #1843, opened 2023-04-14
Add pre-release info on version alert banner

nayres:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-napre-release-version-alert-hashicorp.vercel.app) 🔎
- [Asana task](https://app.asana.com/0/1203347061500809/1203374761617819) 🎟️

## 🗒️ What

This PR changes the message on the `VersionAlertBanner` when a pre-released or non-stable version is selected. 

**Pre-release Version Message:**  You are viewing pre-released documentation for version __
**Stable Version Message (unchanged):**  You are viewing documentation for version __

## 🤷 Why

This will get allow the user to know that the docs version they are currently looking at is not a stable version

## 🛠️ How

To achieve this I passed `versions` from `DocsViewProps` (`views > docs-view > server.ts > getStaticProps()`) through `DocsView`, into `DocsViewLayout`, ultimately into `DocsVersionAlertBanner`. 

From there, I find the `VersionSelectItem` that matches the version in the URL, and pass 'releaseStage' into a optional prop that I added to `VersionAlertBanner`. 

In the `VersionAlertBanner` component I check if the passed in `releaseStage` prop is `!== stable`, and re-assign the default message to a pre-release specific message.

Handling the logic this way allows for a non-breaking change for `VersionAlertBanner`.

### DIFF
--- a/src/components/version-alert-banner/index.tsx
+++ b/src/components/version-alert-banner/index.tsx
@@ -26,14 +26,11 @@ export default function VersionAlertBanner({
 	releaseStage?: string
 	latestVersionUrl: string
 }) {
-	function determineVersionMessage() {
-		if (typeof releaseStage !== 'undefined' && releaseStage !== 'stable') {
-			return 'You are viewing pre-released documentation for version'
-		}
-		return 'You are viewing documentation for version'
-	}
+	let versionMessage = 'You are viewing documentation for version'
 
-	const versionMessage = determineVersionMessage()
+	if (typeof releaseStage !== 'undefined' && releaseStage !== 'stable') {
+		versionMessage = 'You are viewing pre-released documentation for version'
+	}
 
 	return (
 		<PageAlert
--- a/src/components/version-alert-banner/index.tsx
+++ b/src/components/version-alert-banner/index.tsx
@@ -29,7 +29,7 @@ export default function VersionAlertBanner({
 	let versionMessage = 'You are viewing documentation for version'
 
 	if (typeof releaseStage !== 'undefined' && releaseStage !== 'stable') {
-		versionMessage = 'You are viewing pre-released documentation for version'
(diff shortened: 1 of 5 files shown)


src/components/version-alert-banner/index.tsx
```diff
@@ -19,17 +19,25 @@ import s from './version-alert-banner.module.css'
...
 	latestVersionUrl,
 }: {
 	currentVersion: string
+	releaseStage?: string
 	latestVersionUrl: string
 }) {
+	let versionMessage = 'You are viewing documentation for version'
+
+	if (typeof releaseStage !== 'undefined' && releaseStage !== 'stable') {
+		versionMessage = 'You are viewing pre-released documentation for version'
```
BRKalow, line 32: 💭 I think this would be correct:
```suggestion
		versionMessage = 'You are viewing pre-release documentation for version'
```

I'm wondering if we want to include the `releaseStage` in the messaging here, or if `pre-release` is enough. What do you think? 🤔

thiskevinwang:
Looking pretty good to me!

In case this is helpful, here is an option for local testing, in lieu of backend testing ### data

CleanShot 2023-04-14 at 12 19 40@2x

src/components/version-alert-banner/index.tsx
thiskevinwang: Can we use the wording from the [RFC](https://docs.google.com/document/d/1yPrfcQCaEZyyLii8hv-OmoYScqpg94_OAfaFXymo7Y4/edit#)?

> You are viewing documentation for pre-release version {version}. **View latest version.**

CleanShot 2023-04-14 at 12 27 43@2x

thiskevinwang (approved):

BRKalow (approved):
lgtm! 🚢

PR #1846, opened 2023-04-18
feat: basic integrations markdown anchor links

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Implements sanitize-friendly heading anchor links for user content on integrations `readme` and `component` views.

## 🛠️ How

- Adds `remark-plugin-anchor-link-data`, and uses it for integration markdown.
    - Note that we intentionally do not use the existing `@hashicorp/remark-plugins` `anchor-links` plugin since 1) it adds extraneous "code list item" anchor links which are not intended to be added at this time, and 2) it injects `<a />` elements directly, which we have to manually undo at render time.
- Adds and uses `use-user-content-anchor-links`, which makes it so that links like `#user-content-some-hash` are treated like `#some-hash`. This is necessary as `rehype-sanitize` automatically adds the `user-content-` prefix to integrations content during sanitization.
- Related updates and refactors to align with the above functionality, mainly in `MdxHeadings` and `MdxHeadingPermalink`.

## 📸 Design Screenshots

| Before | After |
| - | - | 
| before | after |

## 🧪 Testing

- [ ] Visit a `readme` page, such as [/waypoint/integrations/hashicorp/aws-ami]
    - **Headings should show permalinks** on hover
    - **Permalinks should function as expected**. Test clicking then scrolling away, re-clicking the same permalink repeatedly between scrolls, and alternating between different permalinks.
    - **Navigating directly to #hash URLs**, and similarly reloading the page, should jump to the correct section. For example, [/waypoint/integrations/hashicorp/aws-ami#components] should jump the page down shortly after load.
- [ ] Visit a `component` page, such as [/waypoint/integrations/hashicorp/aws-ami/latest/components/builder/aws-ami-builder]
    - **Headings should show permalinks** on hover
    - **Permalinks should function as expected**. Test clicking then scrolling away, re-clicking the same permalink repeatedly between scrolls, and alternating between different permalinks.
    - **Navigating directly to #hash URLs**, and similarly reloading the page, should jump to the correct section. For example, [/waypoint/integrations/hashicorp/aws-ami/latest/components/builder/aws-ami-builder#interface] should jump the page down shortly after load.

> **Note**: since page content tends to be short, the target heading will likely not be at the top of your viewport on the example pages. This is a persistent issue we keep running into. Further details are in 🎟️ [[table of contents outline nav sidecar jump links] Address non-highlighting of last items](https://app.asana.com/0/1202097197789424/1202975829609765/f).

### 🧪 Testing ESM modules

The upgraded `rehype-sanitize` plugin used in this PR is ESM-only. To continue running tests on work we've done using the updated plugin, this PR grabs the [dynamic Jest config approach from `@hashicorp/web-platform-packages`](https://github.com/hashicorp/web-platform-packages/blob/kw.improve-integrations-hcl-error/jest.config.js), as well as the related [`test:esm` & `test:cjs` commands](https://github.com/hashicorp/web-platform-packages/blob/kw.improve-integrations-hcl-error/package.json#L46-L50).

### 📃 Sidecar anchor-links sold separately

This PR adds permalink elements in the user content area, but does _not_ add those links to the sidecar. Intent is to handle that in a separate task and PR, 🎟 [Incorporate the top-level permalink elements into sidecar](https://app.asana.com/0/0/1204198106778101/f) and [feat: sidecar links for integrations markdown #1848](https://github.com/hashicorp/dev-portal/pull/1848) respectively.

[task]: https://app.asana.com/0/1202097197789424/1204218375412307/f
[preview]: https://dev-portal-git-zsintegrations-md-anchor-links-hashicorp.vercel.app
[/waypoint/integrations/hashicorp/aws-ami]: https://dev-portal-git-zsintegrations-md-anchor-links-hashicorp.vercel.app/waypoint/integrations/hashicorp/aws-ami
[/waypoint/integrations/hashicorp/aws-ami#components]: https://dev-portal-git-zsintegrations-md-anchor-links-hashicorp.vercel.app/waypoint/integrations/hashicorp/aws-ami#components
[/waypoint/integrations/hashicorp/aws-ami/latest/components/builder/aws-ami-builder]: https://dev-portal-git-zsintegrations-md-anchor-links-hashicorp.vercel.app/waypoint/integrations/hashicorp/aws-ami/latest/components/builder/aws-ami-builder
[/waypoint/integrations/hashicorp/aws-ami/latest/components/builder/aws-ami-builder#interface]: https://dev-portal-git-zsintegrations-md-anchor-links-hashicorp.vercel.app/waypoint/integrations/hashicorp/aws-ami/latest/components/builder/aws-ami-builder#interface


jest.config.mjs
```diff
@@ -0,0 +1,12 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+/** @type {import('@jest/types').Config} */
+const config = {
```
zchsh, line 7: This is a stopgap solution to get `rehype-stringify` tests to work. Will expand more in the [Slack thread](https://hashicorp.slack.com/archives/C02T80JTNS2/p1682028713094729) 👍

jest.config.ts
```diff
@@ -15,7 +15,10 @@ const config: Config.InitialOptions = {
 		'!**/*.d.ts',
 		'!**/node_modules/**',
 	],
-	testPathIgnorePatterns: ['<rootDir>/src/__tests__/e2e'],
+	testPathIgnorePatterns: [
+		'<rootDir>/src/__tests__/e2e',
+		'rehype-sanitize.test.ts',
```
zchsh, line 20: The updated version of `rehype-sanitize` is ESM only... and I couldn't seem to find a way to get Jest to work with it. There probably is a way, would love any input here.

See also [Slack thread](https://hashicorp.slack.com/archives/C02T80JTNS2/p1682028713094729) for potential alternatives.

package.json
```diff
@@ -30,9 +30,10 @@
 		"test:e2e": "playwright test --project default",
 		"test:e2e:linkRewrites:loadData": "hc-tools ./scripts/docs-content-link-rewrites/generate-page-paths-for-repo-and-base-path.ts",
 		"test:e2e:linkRewrites": "npx playwright test --project docs-content-link-rewrites",
-		"test:watch": "jest --watch",
-		"test": "jest",
-		"test:ci": "jest --runInBand",
+		"test:watch": "jest --config=jest.config.ts --watch",
+		"test": "jest --config=jest.config.ts",
+		"test:esm": "NODE_OPTIONS=--experimental-vm-modules jest --config=jest.config.mjs",
+		"test:ci": "jest --config=jest.config.ts --runInBand && npm run test:esm -- --runInBand",
```
zchsh, line 36: 🚧 I think I should clean this up somehow before merging - for now "it works". See [Slack thread](https://hashicorp.slack.com/archives/C02T80JTNS2/p1682028713094729) for deets.

package.json
```diff
@@ -198,7 +200,7 @@
 		"react-instantsearch-hooks-web": "^6.33.0",
 		"react-intersection-observer": "^8.33.1",
 		"react-player": "^2.9.0",
-		"rehype-sanitize": "^4.0.0",
+		"rehype-sanitize": "^5.0.1",
```
zchsh, line 203: I've upgraded to the latest `rehype-sanitize`, as it allows for regex-based filtering of properties. This lets us allow `data-text-content` properties needed for permalink generation, while ensuring those properties contain only alphanumeric characters.

I've also added `hast-util-sanitize` as a dependency, as we do import the `defaultSchema` we use for sanitization from that package directly.

src/components/dev-dot-content/mdx-components/mdx-heading-permalink/index.tsx
```diff
@@ -3,20 +3,31 @@
...
-export default function MdxHeadingPermalink(props) {
-	const { className, level, href, 'aria-label': ariaLabel } = props
+interface MdxHeadingPermalinkProps {
+	className?: string
+	level: 1 | 2 | 3 | 4 | 5 | 6
+	href: string
+	ariaLabel: string
+}
 
+export default function MdxHeadingPermalink({
```
zchsh, line 18: I've updated `MdxHeadingPermalink` to use `Link`. This ensures more efficient client-side navigation, and ensures the interactions we expect with `next/router`. The `next/router` bit is important for monitoring the URL for `#hash` changes, which is a key part of our `#user-content-` hash workaround.

I also added types here for completeness.

src/components/dev-dot-content/mdx-components/mdx-headings/index.tsx
```diff
@@ -48,30 +49,120 @@ const HEADING_LEVELS_TO_PROPS: Record<
...
+ * the incoming `className`, `level`, `size`, and `weight`, and we
+ * intentionally omit the incoming `data-text-content`.
  */
 export function makeMdxHeadingElement(level: HeadingProps['level']) {
 	const fixedClassName = classNames(s.heading, s[`h${level}`])
 	const { size, weight } = HEADING_LEVELS_TO_PROPS[level]
 
-	return function MdxHeading(props) {
-		const { children, ...rest } = props
+	return function MdxHeading({
```
zchsh, line 59: I've updated this component to support the new `remark-plugin-anchor-link-data`, which is also introduced in this PR, for use with Integrations. There are several reasons we've wanted to move away from the "inject elements" approach in our older `anchor-links` plugin. As well, we don't seem to have any need for "code list item permalinks" in Integrations, which are bundled inseparably in our older `anchor-links` plugin.

These updates ensure the `MdxHeading` component works with _either_ the old `anchor-links` or newly introduced `remark-plugin-anchor-link-data` plugins. We have a related task with more detail, at 🎟 [Write RFC for refactoring anchor-links plugin's anchor element injection](https://app.asana.com/0/1100423001970639/1202560954440296/f).

src/lib/hooks/use-user-content-anchor-links.ts
```diff
@@ -0,0 +1,103 @@
...
+ * Despite this `id` sanitization, we want users to share and use hash URLs
+ * without the `user-content-` prefix. On the same note, for our own convenience
+ * in implementation, we want to be able to create links to `#hashes` without
+ * worrying about the effects of sanitization.
+ *
+ * This solution is adapted for NextJS based on the `rehype-sanitize` example:
+ * https://github.com/rehypejs/rehype-sanitize/blob/main/readme.md#example-headings-dom-clobbering
+ * (If link breaks, search `scrollIntoView` on the `rehype-sanitize` readme).
+ */
+function useUserContentAnchorLinks() {
```
zchsh, line 21: Have tried to go heavy on the comments here, since dealing with anchor links in sanitized content isn't something we've done before in this repo. The TLDR is there's an example implementation at <https://github.com/rehypejs/rehype-sanitize/blob/main/readme.md#example-headings-dom-clobbering> which informed this work.

src/lib/remark-plugins/rehype-sanitize/__tests__/rehype-sanitize.test.ts
```diff
@@ -5,10 +5,13 @@
 
 import fs from 'fs'
 import path from 'path'
-import type { Root, RootContent, Node } from 'hast'
```
zchsh, line 8: 🚧 Had to update this to not really be Typescript to get tests to work in the only way I could find. There's probably a better way.

src/lib/remark-plugins/rehype-sanitize/index.ts
```diff
@@ -19,20 +19,38 @@ const schema = deepmerge(defaultSchema, {
...
+			 *
+			 * We use config here to ensure that only alphanumeric characters are
+			 * present. Any `data-text-content` attribute with any other characters
+			 * will be removed completely.
+			 *
+			 * In the future, we could safely expand this to 'data*' to avoid having
+			 * to maintain and reason about specific use cases.
+			 * Ref: https://github.com/syntax-tree/hast-util-sanitize#attributes
+			 */
+			['data-text-content', /[\w\-\s]+/],
```
zchsh, line 51: Tried to be thorough here with comments.

In short the upgraded `rehype-sanitize` dependency lets us only allow `data-text-content` properties if they have strictly alphanumeric values. For example, if our plugin code for any reason failed to remove some potentially dangerous characters, or if a user decided to try to add them in, `rehype-sanitize` is still the last thing we run, so would filter those entire properties out.

Due to the test complexity issues I encountered (I think related to `rehype-sanitize` being ESM-only), might need to downgrade again if we can't find a good solution for tests, and find a different solution here.

src/lib/remark-plugins/remark-plugin-anchor-link-data/guarantee-unique-slug.ts
```diff
@@ -0,0 +1,43 @@
...
+import slugify from 'slugify'
+
+/**
+ * Given a string to be made into a slug,
+ * and given an array of existing slugs with which our new slug cannot collide,
+ *
+ * Return a slug version of the provided string which is guaranteed
+ * to not collide with any strings in the existing slugs array.
+ */
+function guaranteeUniqueSlug(string: string, existingSlugs: string[]): string {
```
zchsh, line 15: Lil util used for `anchor-links`, similar to https://github.com/hashicorp/remark-plugins/blob/main/generate_slug.js, but different in a couple ways.

First, it better handles edge cases where there are duplicate-like slugs occur in content. As a contrived example, imagine a document contains first the heading `## Step`, then a duplicate heading `## Step`, and finally another heading `## Step 1`. The previous `generate_slug` would fail to avoid conflicts here, generating the slugs `step`, `step-1`, and `step-1`, with a conflict between the latter slugs.  The revised `guaranteeUniqueSlug` generates the slugs `step`, `step-1`, and `step-1-1`.

Second, unlike `generate_slug`, it does not mutate the `links` argument. Instead, it takes an array of `existingSlugs`. This is intended to allow us to more easily decouple the heading anchor link generation from the "code list item" anchor link generation, as those plugins will no longer have to be in the same execution scope.

src/lib/remark-plugins/remark-plugin-anchor-link-data/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+ * It requires an `anchorLinks` array argument, to which it
+ * appends an `{ id, level, title }` object for each heading element
+ * it traverses.
+ *
+ * Note: this is a revised version of our older `anchor-links` from
+ * `@hashicorp/remark-plugins`. This revised version avoids injecting elements,
+ * instead only adding the necessary data for us to render anchor links using
+ * MDX components.
+ */
+const remarkPluginAnchorLinkData: Plugin<
```
zchsh, line 35: Again, going heavy on the comments here. This relates closely to 🎟 [Write RFC for refactoring anchor-links plugin's anchor element injection](https://app.asana.com/0/1100423001970639/1202560954440296/f) - while this implementation serves the scope of this PR for now, I think it could be useful in helping us meet our refactoring goals mentioned in the task.

src/lib/serialize-integration-markdown.ts
```diff
@@ -3,27 +3,40 @@
...
+// Types
+import type { AnchorLinkItem } from 'lib/remark-plugins/remark-plugin-anchor-link-data'
 
 export default async function serializeIntegrationMarkdown(
-	markdown: string
+	markdown: string,
+	anchorLinks: AnchorLinkItem[] = []
 ): Promise<MDXRemoteSerializeResult> {
+	// TODO: export types from `next-mdx-remote` v3
+	const SERIALIZE_OPTIONS: Parameters<typeof serialize>[1] = {
```
zchsh, line 22: Moved the options into the function body here, as we need to pass `anchorLinks` to the anchor links data plugin in order to get the data we need added to that array.

src/views/product-integration/component-view/components/variable-group-list/index.tsx
```diff
@@ -105,6 +105,7 @@ export function VariableGroupList({
 										</span>
 									)}
 									<MdxHeadingPermalink
+										ariaLabel={`${variable.key} permalink`}
```
zchsh, line 108: Adjacent required change, this should have been here before, but didn't catch it till I added types for `MdxHeadingPermalink` 👍

src/views/product-integration/component-view/index.tsx
```diff
@@ -9,6 +9,7 @@ import type { OutlineNavProps } from 'components/outline-nav'
 import { OutlineNavWithActive } from 'components/outline-nav/components'
 import VersionAlertBanner from 'components/version-alert-banner'
 import ProductIntegrationLayout from 'layouts/product-integration-layout'
+import useUserContentAnchorLinks from 'lib/hooks/use-user-content-anchor-links'
```
zchsh, line 12: Have added a call to `useUserContentAnchorLinks` on the `component-view` and `readme-view`. Figured this makes sense for now since our use cases for user content are relatively isolated. In the future, if we see this popping up on a lot of views, it might make sense to elevate it to say our core layout or something.

src/lib/remark-plugins/rehype-sanitize/__tests__/rehype-sanitize.test.ts
zchsh: ✅ Updated, now back to TypeScript thanks to Kevin's pointer on the dynamic jest config in `web-platform-packages` 🙌

package.json
```diff
@@ -31,8 +31,10 @@
 		"test:e2e:linkRewrites:loadData": "hc-tools ./scripts/docs-content-link-rewrites/generate-page-paths-for-repo-and-base-path.ts",
 		"test:e2e:linkRewrites": "npx playwright test --project docs-content-link-rewrites",
 		"test:watch": "jest --watch",
-		"test": "jest",
-		"test:ci": "jest --runInBand",
+		"test:cjs": "jest",
+		"test:esm": "TEST_ESM=true NODE_OPTIONS=--experimental-vm-modules jest",
```
zchsh, line 35: The updated `rehype-sanitize` is ESM-only. To run tests on it, we use a separate command, with our Jest config updated to toggle its "mode" based on `TEST_ESM`. This is cribbed from the approach in `web-platform-packages` (thanks Kevin for pointing me to that! 🙌 )

package.json
zchsh: ✅ Cleaned up using suggestions from Slack thread.

jest.config.ts
zchsh: Updated to use a "dynamic config" like in `web-platform-packages` 👍

jest.config.mjs
zchsh: ✅ have updated this - we no longer need separate Jest configs, this PR now uses the "dynamic config" approach from `web-platform-packages`.

src/views/product-integration/component-view/index.tsx
BrandonRomano: I think this makes sense here right now; I agree with your decision!

BrandonRomano (approved):
🤯 Wow this is great.

I don't really have any major comments here.  I think it would be beneficial if one additional person gives this a once-over as well.  I get the general gist of how this is working (thanks to your great comments), but perhaps one of the platform folks can also give this a check too.

jest.config.ts
```diff
@@ -6,6 +6,51 @@
 import { Config } from '@jest/types'
 import makeConfig from '@hashicorp/platform-configs/jest/config.js'
 
+/**
+ * We use a dynamic jest config in order to be able to run both
+ * ESM and CommonJS tests using the same config, with separate test commands.
+ *
+ * This approach was taken from `@hashicorp/web-platform-packages`.
+ */
+const isRunningInEsmMode = !!process.env.TEST_ESM
```
kendallstrautman, line 15: 🧠

src/lib/remark-plugins/remark-plugin-anchor-link-data/guarantee-unique-slug.ts
```diff
@@ -0,0 +1,43 @@
...
+	 * increment our uniqueIndex suffix and check again.
+	 *
+	 * Note: there are potentially more efficient ways of doing this,
+	 * for example by matching existing <slug>-<index> values, finding
+	 * the largest `index` value by parsing integers from all matches,
+	 * and finally incrementing that `index` value. However, the efficiency
+	 * gains in this case do not seem worth the potential error-prone-ness
+	 * and relative complexity of using regex and parsing integers.
+	 */
+	while (existingSlugs.indexOf(uniqueSlug) !== -1) {
```
kendallstrautman, line 35: ✨ I rarely see a while loop. I like it 👏🏻

kendallstrautman (approved):
Followed the testing steps and everything seems to work as expected! 💯

jest.config.ts
BRKalow, line 1: Another option here is to use `transformIgnorePatterns`, like we do in `next-mdx-remote`: https://github.com/hashicorp/next-mdx-remote/blob/main/jest.config.js#L8-L10

```
transformIgnorePatterns: [
    '/node_modules/(?!(rehype-sanitize))',
  ],
```

What this does is force jest to transform the `rehype-sanitize` package from ESM to CJS, instead of running jest with native ESM. I think what you have here is reasonable. If we continue to run into issues and have to continue adding to the `ESM_TEST_DIRS` list, we can revisit the approach.

I like the ESM/CJS split as it forces us to use native ESM incrementally, instead of hiding it behind transforming to CJS.

BRKalow (approved):
The jest config changes LGTM! Nice work

PR #1848, opened 2023-04-19
feat: sidecar links for integrations markdown

zchsh:
## 🔗 Relevant links

- [Asana task](url) 🎟️


src/views/product-integration/component-view/server.ts
```diff
@@ -218,6 +218,14 @@ async function getStaticProps({
 		releaseComponent
 	)
 
+	/**
+	 * Serialize the README, extracting anchor links as we do
+	 */
+	const anchorLinks = []
+	const serializedREADME = releaseComponent.readme
+		? await serializeIntegrationMarkdown(releaseComponent.readme, anchorLinks)
```
BrandonRomano, line 226: I know that this is code beyond this PR itself, but it looks like `anchorLinks` is being passed in and gets mutated.  It took me a little while to grok this which is giving me a little bit of pause that it may not be clear what is happening here. Perhaps this is a pretty standard way that remark plugins behave, and it's _expected_ behavior, but I am always a little suspicious of the mutation of pass-by-reference parameters, so wanted to say that out loud.

My understanding of this here is that it looks like [SerializeIntegrationMarkdown](https://github.com/hashicorp/dev-portal/blob/main/src/lib/serialize-integration-markdown.ts#L17) has the remark plugin [remarkPluginAnchorLinkData](https://github.com/hashicorp/dev-portal/blob/main/src/lib/remark-plugins/remark-plugin-anchor-link-data/index.tsx#L35) which [mutates the anchorLinks param](https://github.com/hashicorp/dev-portal/blob/main/src/lib/remark-plugins/remark-plugin-anchor-link-data/index.tsx#L61) that's passed in.

I think for this side-effect, we should document it a little more clearly that the `anchorLinks` array is mutated; if I have the correct understanding!

BrandonRomano (changes requested):
It was unclear to me how `anchorLinks` was getting set without having to dig in really deep.

I don't think you need to adjust the strategy if it's standard behavior for remark plugins, but I think we need to improve the documentation of this side-effect!

src/views/product-integration/component-view/server.ts
zchsh: Ah great call - yep, your understanding is spot one 👍 `serializeIntegrationsMarkdown` mutates `anchorLinks` when it's provided.

I think this is relatively standard with how `remark` plugins behave (or at least, that's how our existing `anchor-links` plugin has always done it!) - possibly it relates to not having an easy way to grab a return value from a plugin.

We could fix this up so we only do the pass-by-reference thing at the plugin call level. So we'd still pass `anchorLinks` by reference _within_ `serializeIntegrationsMarkdown` (and great call on adding a comment to clarify the pass-by-reference / mutation situation)... but we could have something like this where `serializeIntegrationsMarkdown` is called:

```ts
const { serializedREADME, anchorLinks } = releaseComponent.readme
    ? await serializeIntegrationMarkdown(releaseComponent.readme)
    // If we don't have a README to serialized, return empty values
    : { serializedREADME: undefined, anchorLinks: [] }
```

That would change up the return value even in cases where the `anchorLinks` arg is not used, of course, eg:

```
// Contrived example not using `anchorLinks` for some reason
const { serializedREADME } = await serializeIntegrationMarkdown(releaseComponent.readme)
```

Let me know what you think!

zchsh: ☝️ Updated to do this! Also added an explicit comment on `anchorLinks` being mutated by `remarkPluginAnchorLinkData` within `serializeIntegrationsMarkdown`. Hopefully it's a good direction, easy enough to revert if not 👍

src/lib/serialize-integration-markdown.ts
```diff
@@ -15,10 +15,14 @@ import remarkPluginAnchorLinkData from 'lib/remark-plugins/remark-plugin-anchor-
...
 
 export default async function serializeIntegrationMarkdown(
-	markdown: string,
-	anchorLinks: AnchorLinkItem[] = []
-): Promise<MDXRemoteSerializeResult> {
+	markdown: string
+): Promise<{
+	serializeResult: MDXRemoteSerializeResult
+	anchorLinks: AnchorLinkItem[]
+}> {
```
BrandonRomano, line 22: Oh I am a HUGE fan of this change.  This quells all of my concerns.

BrandonRomano (approved):

PR #1851, opened 2023-04-20
Packer Integrations

BrandonRomano:
Introduces the Integrations section for Packer.

[🔍 Preview](https://dev-portal-git-brpacker-plugins-hashicorp.vercel.app/packer/integrations)

[📝 Document detailing further migrations](https://whimsical.com/migrating-packer-plugins-F5foVbcDMsaTXJSaWV35ZV)


build-libs/integration-packer-redirects.js
```diff
@@ -0,0 +1,895 @@
...
+		case TYPE_DATA_SOURCE:
+			return 'datasources'
+		default:
+			return ''
+	}
+}
+
+/**
+ * An array of legacy Packer plugins that were previously in the
+ * Packer Plugin library that are now in the Integrations library.
```
thiskevinwang, line 25: If there's a migration guide/document that we can link to in this comment, that would be great.

It's not immediately clear if this list is expected to grow or if it's stable, or it needs to change in tandem with some external work.

thiskevinwang (approved):

build-libs/integration-packer-redirects.js
zchsh: ☝️ related question: this is related to https://github.com/hashicorp/packer/blob/main/website/data/plugins-manifest.json right? Essentially redirecting the plugins detailed in that manifest to their new integration locations?

zchsh (approved):
✅ Preview looks good to me!

Given we're now redirecting `/packer/plugins` to `/packer/integrations`, should we follow up this PR with one to remove `Plugins` for the Packer nav? (as below)

CleanShot 2023-08-31 at 15 18 27@2x

BrandonRomano:
@zchsh - we're actually going to keep those packer plugin links _for now_ basically until we fully decomission that secion of the website.  I detail it in the video that I talked over this PR, but basically we want to keep those "plugins" links for now so users who are used to navigating to those places get appropriately redirected.

build-libs/integration-packer-redirects.js
```diff
@@ -0,0 +1,895 @@
...
+ *
+ * Redirects are generated from this array.
+ */
```
BrandonRomano, line 28:
```suggestion
 * https://whimsical.com/migrating-packer-plugins-F5foVbcDMsaTXJSaWV35ZV
 */
```

PR #1862, opened 2023-04-26
[Home Page Redesign] Implement search card auto scroll with transition on `transform: translateX()` property

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ambhp-safari-autoscroll-hashicorp.vercel.app/) 🔎

## 🗒️ What

- Replaces the `scrollBy` autoscroll implementation with programmatically setting the `transform: translateX()` property in JS

## 🤷 Why

- Autoscroll was not working consistently in macOs or iOS Safari

## 🧪 Testing

- [ ] The "Search with ease" card autoscroll should work smoothly in:
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

## 💭 Anything else?

- My first attempt at this tried to animate against the `left` CSS property. When testing it in Safari, it was jittery/laggy. Shout out to @BRKalow for a resource about this and the suggestion to use `translate: transformX()` instead! The resource is: [Stick to Compositor-Only Properties and Manage Layer Count](https://web.dev/stick-to-compositor-only-properties-and-manage-layer-count/).

---
  - https://app.asana.com/0/0/1204484002819134

### DIFF
--- a/src/views/homepage/components/featured-content-grid/search-featured-card.module.css
+++ b/src/views/homepage/components/featured-content-grid/search-featured-card.module.css
@@ -4,6 +4,8 @@
  */
 
 .root {
+	--auto-scroll-transition-length: 400ms;
+
 	background-color: var(--token-color-foreground-strong);
 	background-position: center top;
 	background-repeat: no-repeat;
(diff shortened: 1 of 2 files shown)


zchsh (approved):
✅ Tested in FireFox, Safari, and Chrome! Looked good in all browsers 🙌

PR #1910, opened 2023-05-11
fix: avoid static api-docs render in preview repos

zchsh:
This PR avoids trying to statically render `/hcp/api-docs/packer` in all "content preview" repositories, as determined by our `isDeployPreview()` utility.


BRKalow (approved):
🙏

PR #1923, opened 2023-05-16
feat: add counts to search result tabs

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Updates the search results interface to show result counts for each content type.

## 🤷 Why

- To make it more clear that results are distributed across content type tabs
- To better signal that when one content type tab has zero results, other tabs may have some results

## 🛠️ How

- Sets up a `hitCount` context provider to store information about result counts across all content types
- Adds support for a `headingSlot` in `Tabs`, so that we can render a count element
    - Previously only plain text and icons were supported in Tab headings
- Implements a `TabHeadingWithCount` component
- Puts `hitCount` data into tab headings using `TabHeadingWithCount`

## 📸 Design Screenshots

| Before | After |
| - | - |
| before | after |

## 🧪 Testing

- [ ] Visit any page with search enabled, such as [the home page][preview]
    - The search experience should continue to function as expected
    - Result counts should be shown in each tab heading
    - Note: on smaller viewports, result counts are not shown in the `<select />` option text

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/1204333060736722/1204604641329011/f
[preview]: https://dev-portal-git-zssearch-tab-result-counts-hashicorp.vercel.app/


emilypersson1 (approved):
looking good, thanks!

src/components/command-bar/commands/search/components/custom-hits-container/index.tsx
```diff
@@ -21,8 +23,22 @@ const CustomHitsContainer = ({
 	noResultsSlot,
 	type,
 }: CustomHitsContainerProps) => {
+	const [hitCounts, setHitCounts] = useHitsContext()
```
kendallstrautman, line 26: Is there some max hits in algolia for docs or tutorials? I noticed the numbers 20 (docs) & 25 (tutorials) coming up a lot when interacting with it. If there's a max hits setting, they should probably be consistent? Perhaps it was coincidence, but curious if you came across this.

src/components/command-bar/commands/search/components/dialog-body/index.tsx
```diff
@@ -152,7 +155,17 @@ const SearchCommandBarDialogBodyContent = ({
 					const { heading, icon, content } =
 						tabsBySearchableContentType[contentType]
 					return (
-						<Tab heading={heading} icon={icon} key={contentType}>
+						<Tab
+							heading={heading}
```
kendallstrautman, line 159: Do we still need to pass heading if the heading slot is being used?

src/components/command-bar/commands/search/helpers/hit-counts-provider.tsx
```diff
@@ -0,0 +1,30 @@
+import { SearchableContentType } from 'contexts'
+import { ReactNode, createContext, useContext, useState } from 'react'
+
+const SearchHitsContext = createContext([])
```
kendallstrautman, line 4: ❓ we have stored other contexts in `/src/contexts`, I'm not super opinionated but just pointing it out in case it feels better for all of our providers to live in one place.

src/components/tabs/hooks/use-tab-items.ts
```diff
@@ -160,8 +160,16 @@ function validateTabChildren(
...
-		return { content: children, group, icon, label: heading }
+		const { children, renderContent, group, heading, headingSlot, icon } =
+			childTab.props
+		return {
+			content: children,
+			renderContent,
+			group,
+			icon,
+			label: heading,
+			labelSlot: headingSlot,
```
kendallstrautman, line 171: ❓ should we check for either `label` or `labelSlot`, but not both? Definitely low prio, unless we use the `label` as well even when using `labelSlot`.

kendallstrautman (approved):
Followed the testing steps and everything works as described. I had a few non-blocking comments. The only question worth getting clarity on is the max hits config?

src/components/command-bar/commands/search/components/dialog-body/index.tsx
zchsh: Unfortunately yes, as `headingSlot` doesn't work for the `<select />`-based Tabs interface used on mobile. I'll add a note in the `types` file for `heading` to make this more clear 👍

src/components/command-bar/commands/search/components/custom-hits-container/index.tsx
zchsh: Ah great call-out, I didn't notice this! 🤔 I've added a task to our search board to investigate: 🎟 [Look into "max hits", might be different between Docs and Tutorials](https://app.asana.com/0/1204333060736722/1204613586844721/f)

I think it could make sense to handle the potential config updates separately from this PR, but open to taking a different approach 🙇‍♂️

src/components/tabs/hooks/use-tab-items.ts
zchsh: ✅ I've added comments to `types` files to clarify the relationship between `label` and `labelSlot`.

src/components/command-bar/commands/search/components/dialog-body/index.tsx
zchsh: ✅ I've added comments to `types` files to clarify the relationship between `heading` and `headingSlot` (which ultimately become `label` and `labelSlot`, bit of a historical naming conflict I think).

src/components/command-bar/commands/search/helpers/hit-counts-provider.tsx
zchsh: Ah yeah great call I forgot about that!

I think in this case the provider feels component-specific, say similar to the tab nesting styles provider, or tab groups provider. I think a single home for _all_ contexts could get unwieldy... where a provider is only used by a component and its partials, I think it could make sense to keep it co-located with whatever scope it's intended to apply to (sometimes `component`-specific, sometimes `view`-specific, etc). Or to put it another way, I think `src/contexts` could make a lot of sense for any contexts we use in `_app` 🤔

zchsh:
@kendallstrautman For posterity:
- For the "max hits config" question, I've made a separate task for now: 🎟 [Look into "max hits", might be different between Docs and Tutorials](https://app.asana.com/0/1204333060736722/1204613586844721/f). Will look into it, feels like it'll be just as relevant but _harder to catch_ when we have a unified list 🧐 thanks for calling this out! 💖 
- For the `heading` vs `headingSlot`, have added some comments to clarify that the `headingSlot` is an optional prop, used for `button`-based tab selection on wider viewports when present, while `heading` remains a required prop, used for `select`-based tab selection in all cases, and also for `button`-based tab selection if `headingSlot` is not provided. This is definitely making me think both about the naming, and seems like another data point in favour of a potential to refactor from the `select` approach to something that could support more dynamic content, as in 🎟️ [[tabs] replace mobile select element with something better](https://app.asana.com/0/1202097197789424/1202133172981709/f)

src/components/command-bar/commands/search/components/custom-hits-container/index.tsx
zchsh: Okay, poked into this, it is indeed set to `20` for docs and `25` for tutorials. Have updated so they're both `25`. Will check in with product on this as well though, so leaving 🎟 [Look into "max hits", might be different between Docs and Tutorials](https://app.asana.com/0/1204333060736722/1204613586844721/f) open for now 👍

zchsh:
@heitzke I've update this PR to use the shared `<Badge />` component, for that slightly more "square" look on the count numbers in each content type tab heading. Let me know if that looks alright 🙇‍♂️

heitzke (approved):
Looks good to me!

PR #1928, opened 2023-05-16
feat: adds variants dropdown disclosure

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksadds-varaints-dropdown-disclosure-hashicorp.vercel.app) 🔎
- [Asana task](https://app.asana.com/0/1204333057896641/1204615849802262) 🎟️
- [Design file 🎨 ](https://www.figma.com/file/NAFjIPkbwjyY8jPl6T9nlP/Tutorial-Variants?type=design&node-id=238-39328&t=J7vcx4Y3jEZjOfta-0)

🚧 This PR is merging into an assembly 🚧 

## 🗒️ What

This PR adds a variant dropdown disclosure selector to the sidecar slot for the tutorial view. This component goes away at tablet, that is expected. We will add additional variants UI in future PRs.

Screenshot 2023-05-16 at 3 58 28 PM
Screenshot 2023-05-16 at 4 19 47 PM
Screenshot 2023-05-16 at 4 19 37 PM
Screenshot 2023-05-16 at 4 34 07 PM
Screenshot 2023-05-16 at 4 34 10 PM

## 📸 Design Screenshots
Screenshot 2023-05-16 at 3 59 56 PM
Screenshot 2023-05-16 at 4 01 09 PM

## 🧪 Testing

- Visit a route with a [variant query param](https://dev-portal-git-ksadds-varaints-dropdown-disclosure-hashicorp.vercel.app/vault/tutorials/getting-started-ui/getting-started-policies-ui?variants=deployment:hcp) defined
- Open the dropdown, validate that the options show
- Use voice over to test again. Note the question on the activator aria label in the code comments.
- Use keyboard navigation to test, it should behave the exact same as all other dropdown disclosure components, no changes there.
- Add in more text to the variant name to validate the overflow state
- Validate that plain tutorial [paths without variants](https://dev-portal-git-ksadds-varaints-dropdown-disclosure-hashicorp.vercel.app/consul/tutorials/kubernetes/kubernetes-secure-agents) render as normal

## 💭 Anything else?
This PR uses `_tmp` data since the backend changes are still underway. I've tried to mark that explicitly with `@TODO` comments and `_tmp` prefixes.

🔴 _Note_ This PR hard codes the active variant option, so clicked a new variant option in the list will change the URL but the active option in the dropdown won't be reflected. That will happen when we hook up to real data. 

❓ I have [an open question ](https://github.com/hashicorp/dev-portal/pull/1928#issuecomment-1550495638)on how we handle the 'active' list item, whether it should be removed from the list like we do with version selector or we want to add an 'active' state.


src/views/tutorial-view/components/variants-dropdown-disclosure/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+	// @TODO hook this into useVariants hook once data is wired
+	const activeOption = variant.options[0]
+
+	return (
+		<div className={s.root}>
+			<Text weight="semibold" size={100} className={s.label}>
+				{variant.name}
+			</Text>
+			<DropdownDisclosure
+				aria-label={`${variant.name}: ${activeOption.name}`}
```
kendallstrautman, line 29: Would love feedback on the right thing to do here. We have a visual label but currently with dropdown disclosure there isn't a way to associate the visual label with the list of dropdown disclosure items. I think that could be a good candidate for a refactor later on but feels outside the scope of this PR, so looking for the best case here, which could perhaps be using no aria label and just having the button text there, but then we lose the extra context of the parent variant name as the visual label.

src/views/collection-view/helpers/index.ts
```diff
@@ -6,3 +6,4 @@
 export * from './format-sidebar-sections'
 export * from './get-slug'
 export * from './generate-hcp-sidebar'
+export * from './generate-collection-sidebar-nav-data'
```
kendallstrautman, line 9: Just a little cleanup here for the imports in the tutorial view file.

src/views/tutorial-view/components/variants-dropdown-disclosure/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+import Text from 'components/text'
+import {
+	getVariantParam,
+	getVariantPath,
+	TutorialVariantOption,
+} from 'views/tutorial-view/utils/variants'
+import { VariantsDropdownDisclosureProps } from './types'
+import s from './variants-dropdown-disclosure.module.css'
+
+// @TODO add aria-labelledBy
```
kendallstrautman, line 14:
```suggestion
```

src/views/tutorial-view/index.tsx
```diff
@@ -43,16 +46,19 @@ import {
 import MDX_COMPONENTS from './utils/mdx-components'
 import { formatTutorialToMenuItem, generateCanonicalUrl } from './utils'
 import getVideoUrl from './utils/get-video-url'
+import { useProgressToast } from './utils/use-progress-toast'
```
kendallstrautman, line 49: Just reordering imports here

src/views/tutorial-view/index.tsx
```diff
@@ -248,7 +254,16 @@ function TutorialView({
 					sidebarNavDataLevels={sidebarNavDataLevels as any}
 					showScrollProgress={true}
 					AlternateSidebar={TutorialsSidebar}
-					sidecarSlot={<OutlineNavWithActive items={outlineItems} />}
+					sidecarSlot={
+						<>
+							{metadata.variant ? (
+								<VariantsDropdownDisclosure
+									variant={variantData[0] as TutorialVariant}
```
kendallstrautman, line 261: This will be the active variant once we get data wired up.

src/views/tutorial-view/components/variants-dropdown-disclosure/variants-dropdown-disclosure.module.css
```diff
@@ -0,0 +1,8 @@
+.root {
```
kendallstrautman, line 1: These styles don't account for super long variant option names at the moment. I'm leaning towards making a fix for that if it seems like the real data will have option names that long. I could be persuaded otherwise though.

src/layouts/sidebar-sidecar/components/sidecar-scroll-container/sidecar-scroll-container.module.css
```diff
@@ -77,6 +77,7 @@
 
 .scrollContainer {
 	overflow-y: auto;
+	overflow-x: hidden;
```
kendallstrautman, line 80: We need this to avoid the horizontal scrollbar if the dropdown disclosure active item text is really long. Since the sidecar items are truncated I don't believe this affects anything there but please let me know if there are concerns with this.

src/views/tutorial-view/components/variants-dropdown-disclosure/variants-dropdown-disclosure.module.css
```diff
@@ -0,0 +1,25 @@
...
+}
+
+.dropdownDisclosure {
+	/** 
+	* 🏴‍☠️ FLAG
+	* Reaching into the component here to
+	* override styles. This would be better to
+	* update the actual dropdown disclosure list width, which
+	* is currently set to `max-content`.
+	*/
```
kendallstrautman, line 17: I chose to not make the update on dropdown disclosure list directly in this PR because that would really increase the scope of review. I did a quick test and updating from `max-content` seemed fine but I just didn't want to balloon the changes here, so this could be a quick follow along clean up item 🧼

kendallstrautman:
@braydenlove do you want to communicate a 'current' state in the dropdown disclosure for the selected variant? Or do you want to remove the active option from the dropdown list?

kaitlynnefuery (approved):
I went through the tests and it all checks out. Approving this with the expectation that you'll want other eyes on your area-label parent variant question. I like what you have in the PR as is, but am not opposed to your alternate option to just have button text.

src/layouts/sidebar-sidecar/components/sidecar-scroll-container/sidecar-scroll-container.module.css
zchsh: 💭 Tested this out and I think we may be able to get away with some other solution - looks like the scroll is being caused by the "disclosure activator" `<button />` `:before` pseudo-element, which I think is a shadow or something? I think we could set `overflow: hidden` on `variants-dropdown-disclosure_root` to scope the overflow fix to this use case 👍

https://github.com/hashicorp/dev-portal/assets/4624598/1596d851-b5ff-48f7-bfe3-b68d8b195e26

That being said, could also see the case for hiding all horizontal overflow in sidecar as done here... but worried it could hide a future quirk or bug we might otherwise notice 🤔 So probably either one is fine!

src/views/tutorial-view/components/variants-dropdown-disclosure/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+} from 'views/tutorial-view/utils/variants'
+import { VariantsDropdownDisclosureProps } from './types'
+import s from './variants-dropdown-disclosure.module.css'
+
+export function VariantsDropdownDisclosure({
+	variant,
+}: VariantsDropdownDisclosureProps) {
+	const { asPath } = useRouter()
+	// @TODO hook this into useVariants hook once data is wired
+	const activeOption = variant.options[0]
```
zchsh, line 19: 🙌 ah sweet, i was wondering about the active state reflecting the URL, will look out for that in a later PR 👍

src/views/tutorial-view/components/variants-dropdown-disclosure/index.tsx
zchsh: 👍 I'm not sure, but seems like a possible case for adding support for an `aria-labelledby` prop to `DropdownDisclosure`? +1 on a separate PR for that, if we want. I think the `aria-label` here makes sense. Only thing I can think of is adding a comment here if we're relatively sure we wanna revisit it 🍞

src/views/tutorial-view/components/variants-dropdown-disclosure/variants-dropdown-disclosure.module.css
zchsh: 🙌 Thanks for calling this out - this is ringing bells and I've seen something similar at least once - in [version-selector](https://github.com/hashicorp/dev-portal/blob/main/src/components/version-switcher/version-switcher.module.css) work recently, at least.

We have a follow-up task to try to resolve the duplicate "reach-in-ish" CSS at some point: [[discuss] Consider fullWidth prop for DropdownDisclosure to avoid style overrides](https://app.asana.com/0/1202097197789424/1204412156894162/f). Feels like this is another data point there 👍 Heck maybe this `fit-content` fix would be an alternative to the more complex CSS overrides we have in `version-selector` 🤔

zchsh (approved):
✅ Looks great, followed testing steps, everything worked as expected 👍

No strong opinion on the `aria-label` question - what you have seems reasonable to me, and on board with a potential follow-up to explore what might be ideal 🚂

src/layouts/sidebar-sidecar/components/sidecar-scroll-container/sidecar-scroll-container.module.css
kendallstrautman: Ooo I'll check out this suggestion! I'm definitely down for minimizing the radius of this change. TY

src/views/tutorial-view/components/variants-dropdown-disclosure/variants-dropdown-disclosure.module.css
kendallstrautman: I can add a note about this case to that ticket! Good to hear you've ran into this as well.

src/views/tutorial-view/components/variants-dropdown-disclosure/index.tsx
kendallstrautman: Yeah I agree, adding support for `aria-labelledby` would be the best end result here. I can make a follow up ticket and add a breadcrumb 🍞

braydenlove: I agree with the approach here ^

src/layouts/sidebar-sidecar/components/sidecar-scroll-container/sidecar-scroll-container.module.css
kendallstrautman: Hmmm @zchsh if I set overflow hidden to the root, then the dropdown items aren't visible 🤔

kendallstrautman: I'm going to roll with this for now but we should keep an eye out for any side effects. I've clicked around to see a bunch of sidecars and haven't noticed an issue yet.

braydenlove (approved):

PR #1931, opened 2023-05-17
feat: variants button list

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksvariants-button-list-hashicorp.vercel.app/vault/tutorials/getting-started-ui/getting-started-policies-ui?variants=deployment:hcp) 🔎
- [Asana task](https://app.asana.com/0/1204333057896641/1204615849802274) 🎟️
- [Figma File 🎨 ](https://www.figma.com/file/NAFjIPkbwjyY8jPl6T9nlP/Tutorial-Variants?type=design&node-id=165-30497&t=XEhCsJPFalEhRqQK-0)

🚧 Merging into an Assembly 🚧 

## 🗒️ What

This PR adds the variants button link list below the tutorial metadata.

Screenshot 2023-05-17 at 10 31 58 AM

## 🛠️ How

Adds a new `VariantList` component within the tutorial metadata.

## 📸 Design Screenshots

Screenshot 2023-05-17 at 9 00 06 AM

## 🧪 Testing

- [Visit the preview](https://dev-portal-git-ksvariants-button-list-hashicorp.vercel.app/vault/tutorials/getting-started-ui/getting-started-policies-ui?variants=deployment:hcp), validate that the list matches the design spec
- Validate that it goes away <tablet sizes, we will add the mobile dropdown disclosure in a separate PR. 
- Testing at 200% text zoom
Screenshot 2023-05-17 at 11 34 54 AM

- Test VO, ensure the label for the list is read when focusing on the first item along with noting the current list item.
Screenshot 2023-05-17 at 12 46 27 PM

## 💭 Anything else?

- This isn't hooked up to real data yet so the interactivity is limited. The PR hard codes the active variant option as the first one, this won't change despite shifting URLs.

### DIFF
--- a/src/components/tutorial-meta/components/variant-list/index.tsx
+++ b/src/components/tutorial-meta/components/variant-list/index.tsx
@@ -33,7 +33,10 @@ export default function VariantList({ variant }: { variant: TutorialVariant }) {
 								aria-current={activeVariantOption.slug === option.slug}
 							>
 								<ButtonLink
-									className={classNames(!isActiveOption && s.linkBorderColor)}
+									className={classNames(
+										s.linkOverrides,
+										!isActiveOption && s.secondaryLinkOverrides
+									)}
 									size="small"
 									color={isActiveOption ? 'primary' : 'secondary'}
 									text={option.name}
--- a/src/components/tutorial-meta/components/variant-list/index.tsx
+++ b/src/components/tutorial-meta/components/variant-list/index.tsx
@@ -11,7 +11,7 @@ import {
 import { useRouter } from 'next/router'
 import s from './variant-list.module.css'
 
-export default function VariantList({ variant }: { variant: TutorialVariant }) {
+export function VariantList({ variant }: { variant: TutorialVariant }) {
 	// @TODO hook this into a useVariants hook
 	const activeVariantOption = variant.options[0]
 	const VARIANT_LIST_ID = 'variant-list-label'
--- a/src/components/tutorial-meta/components/variant-list/index.tsx
+++ b/src/components/tutorial-meta/components/variant-list/index.tsx
@@ -1,5 +1,6 @@
 // @TODO uncomment when enabling cookies
 // import Cookies from 'js-cookie'
+import Link from 'next/link'
 import classNames from 'classnames'
 import {
 	TutorialVariant,
@@ -7,7 +8,6 @@ import {
 	getVariantParam,
 	getVariantPath,
 } from 'views/tutorial-view/utils/variants'
-import ButtonLink from 'components/button-link'
 import { useRouter } from 'next/router'
 import s from './variant-list.module.css'
 
@@ -32,14 +32,11 @@ export default function VariantList({ variant }: { variant: TutorialVariant }) {
 								key={option.slug}
 								aria-current={activeVariantOption.slug === option.slug}
 							>
-								<ButtonLink
+								<Link
 									className={classNames(
-										s.linkOverrides,
-										!isActiveOption && s.secondaryLinkOverrides
+										s.link,
+										isActiveOption ? s.activeLink : s.inActiveLink
 									)}
-									size="small"
-									color={isActiveOption ? 'primary' : 'secondary'}
-									text={option.name}
 									href={getVariantPath(asPath, variantParam)}
 									onClick={() => {
 										// @TODO add this in when we have real data to check against
@@ -49,7 +46,9 @@ export default function VariantList({ variant }: { variant: TutorialVariant }) {
 										// 	Cookies.set(variant.id, variantParam)
 										// }
 									}}
-								/>
+								>
+									{option.name}
+								</Link>
 							</li>
 						)
 					})}
--- a/src/components/tutorial-meta/components/variant-list/index.tsx
+++ b/src/components/tutorial-meta/components/variant-list/index.tsx
@@ -1,6 +1,6 @@
 // @TODO uncomment when enabling cookies
 // import Cookies from 'js-cookie'
-import Link from 'next/link'
+import Link from 'components/link'
 import classNames from 'classnames'
 import {
 	TutorialVariant,
@@ -28,16 +28,14 @@ export default function VariantList({ variant }: { variant: TutorialVariant }) {
 						const variantParam = getVariantParam(variant.slug, option.slug)
 						const isActiveOption = activeVariantOption.slug === option.slug
 						return (
-							<li
-								key={option.slug}
-								aria-current={activeVariantOption.slug === option.slug}
-							>
+							<li key={option.slug}>
 								<Link
 									className={classNames(
 										s.link,
 										isActiveOption ? s.activeLink : s.inActiveLink
 									)}
 									href={getVariantPath(asPath, variantParam)}
+									aria-current={isActiveOption ? 'page' : false}
 									onClick={() => {
 										// @TODO add this in when we have real data to check against
 										// const variantCookie = Cookies.get(variant.slug)
--- a/src/components/tutorial-meta/components/variant-list/index.tsx
+++ b/src/components/tutorial-meta/components/variant-list/index.tsx
@@ -1,5 +1,6 @@
 // @TODO uncomment when enabling cookies
 // import Cookies from 'js-cookie'
+import classNames from 'classnames'
 import {
 	TutorialVariant,
 	TutorialVariantOption,
@@ -25,18 +26,16 @@ export default function VariantList({ variant }: { variant: TutorialVariant }) {
 				<ul aria-labelledby={VARIANT_LIST_ID} className={s.list}>
 					{variant.options.map((option: TutorialVariantOption) => {
 						const variantParam = getVariantParam(variant.slug, option.slug)
+						const isActiveOption = activeVariantOption.slug === option.slug
 						return (
 							<li
 								key={option.slug}
 								aria-current={activeVariantOption.slug === option.slug}
 							>
 								<ButtonLink
+									className={classNames(!isActiveOption && s.linkBorderColor)}
 									size="small"
-									color={
-										activeVariantOption.slug === option.slug
-											? 'primary'
-											: 'secondary'
-									}
+									color={isActiveOption ? 'primary' : 'secondary'}
 									text={option.name}
 									href={getVariantPath(asPath, variantParam)}
 									onClick={() => {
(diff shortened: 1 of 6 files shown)


src/components/tutorial-meta/components/variant-list/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+	TutorialVariant,
+	TutorialVariantOption,
+	getVariantParam,
+	getVariantPath,
+} from 'views/tutorial-view/utils/variants'
+import ButtonLink from 'components/button-link'
+import { useRouter } from 'next/router'
+import s from './variant-list.module.css'
+
+export default function VariantList({ variant }: { variant: TutorialVariant }) {
```
kendallstrautman, line 13: We may get all of the variant data from the hook and not props, so this could change when we wire up real data.

src/views/tutorial-view/utils/variants/_tmp-variants-data.json
```diff
@@ -8,7 +8,12 @@
 				"slug": "hcp",
 				"name": "Hashicorp Cloud Platform (HCP)"
 			},
-			{ "slug": "self-managed", "name": "Self Managed" }
+			{ "slug": "self-managed", "name": "Self Managed" },
+			{ "slug": "self-managed", "name": "Koba" },
+			{ "slug": "self-managed", "name": "Ubuntu" },
+			{ "slug": "self-managed", "name": "Frank Herbert" },
+			{ "slug": "self-managed", "name": "Calamansi" },
+			{ "slug": "self-managed", "name": "Lavender" }
```
kendallstrautman, line 16: TODO - remove extra data before merging

src/components/tutorial-meta/components/variant-list/index.tsx
```diff
@@ -0,0 +1,59 @@
+// @TODO uncomment when enabling cookies
+// import Cookies from 'js-cookie'
+import Link from 'next/link'
```
ashleemboyer, line 3: 🙋🏻 Should this be `component/link`?

src/components/tutorial-meta/components/variant-list/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+	const VARIANT_LIST_ID = 'variant-list-label'
+	const { asPath } = useRouter()
+
+	return (
+		<div className={s.root}>
+			<label id={VARIANT_LIST_ID} className={s.label}>
+				{variant.name}
+			</label>
+			<nav>
+				<ul aria-labelledby={VARIANT_LIST_ID} className={s.list}>
```
ashleemboyer, line 26: 🙌🏻 Nice! `aria-labelledby`!

src/components/tutorial-meta/components/variant-list/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+					{variant.options.map((option: TutorialVariantOption) => {
+						const variantParam = getVariantParam(variant.slug, option.slug)
+						const isActiveOption = activeVariantOption.slug === option.slug
+						return (
+							<li
+								key={option.slug}
+								aria-current={activeVariantOption.slug === option.slug}
```
ashleemboyer, line 33: 🚧 2 small things here:

- `aria-current` should be on the `<a>` elements rather than the `<li>` ones
- the value we want to set for the current element is `"page"` rather than a boolean since these are navigation links

More on `aria-current` for reference here: https://www.w3.org/TR/wai-aria-1.1/#aria-current

ashleemboyer:
Non-blocking UX question -- Will there be visual affordance (beyond color) added that enables users to distinguish between the current variant from the other variant options?

zchsh (commented):
✅ Ran through testing steps, didn't find anything other than the things Ashlee already noted 👍 

- 🚧 `component/link` vs `next/link` - I for one have probably forgotten this at some point, future thought but I wonder if a lint rule would be helpful 🤔
- 🚧 `aria-current="page"` on the `<a />` element
- 🚧 `hds-typgraphy-body-200` creates a lil design nit

zchsh:
@ashleemboyer Re: visual affordance other than color, I believe that's what the inversion of the button styling is intended to cover - with the current variant option styled light text on dark, and other options styled dark text on light, I think the contrasting appearance acts as a distinguishing factor even in grayscale 🌞

src/components/tutorial-meta/components/variant-list/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+										// @TODO add this in when we have real data to check against
+										// const variantCookie = Cookies.get(variant.slug)
+										// // if it exists and its not already set with the same value
+										// if (!variantCookie || variantCookie !== variantParam) {
+										// 	Cookies.set(variant.id, variantParam)
+										// }
+									}}
+								>
+									{option.name}
+								</Link>
```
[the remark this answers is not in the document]
kendallstrautman: No the styles are quite different so we created a unique case here to avoid CSS overrides. I started out with using `ButtonLink` and ran into the style issues and chose to eject.

src/components/tutorial-meta/components/variant-list/index.tsx
kendallstrautman: Ah okay, thank you!

ashleemboyer:
> @ashleemboyer Re: visual affordance other than color, I believe that's what the inversion of the button styling is intended to cover - with the current variant option styled light text on dark, and other options styled dark text on light, I think the contrasting appearance acts as a distinguishing factor even in grayscale 🌞

@zchsh Oops, apologies! I should have linked to the WCAG criteria that I had in my head but wasn't saying out loud.

[Success Criterion 1.4.1 Use of Color](https://www.w3.org/TR/WCAG21/#use-of-color) is a Level A SC, and says:

> Color is not used as the only visual means of conveying information, indicating an action, prompting a response, or distinguishing a visual element.

src/components/tutorial-meta/components/variant-list/index.tsx
kendallstrautman: Forgot about this component! I'll update

braydenlove:
> Non-blocking UX question -- Will there be visual affordance (beyond color) added that enables users to distinguish between the current variant from the other variant options?

Awesome question, I did research and documented it here: https://www.figma.com/file/NAFjIPkbwjyY8jPl6T9nlP/Tutorial-Variants?type=design&node-id=177%3A31207&t=eKV5emNSpFC80rM7-1.

kendallstrautman:
@ashleemboyer @braydenlove goes into this during the design review video and [here are some figma notes](https://www.figma.com/file/NAFjIPkbwjyY8jPl6T9nlP/Tutorial-Variants?type=design&node-id=177-31207&t=owDJd9L9oxLB25dQ-0) on it but [to my understanding](https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html), so long as the contrast between the two option passes 3:1, the colors differ in luminance enough to create a contrast that provides the distinction. 

Screenshot 2023-05-17 at 12 39 03 PM

ashleemboyer:
> > Non-blocking UX question -- Will there be visual affordance (beyond color) added that enables users to distinguish between the current variant from the other variant options?
> 
> Awesome question, I did research and documented it here: https://www.figma.com/file/NAFjIPkbwjyY8jPl6T9nlP/Tutorial-Variants?type=design&node-id=177%3A31207&t=eKV5emNSpFC80rM7-1.
> 
> > Note
> > If content is conveyed through the use of colors that differ not only in their hue, but that also have a significant difference in lightness, then this counts as an additional visual distinction, as long as the difference in relative luminance between the colors leads to a contrast ratio of 3:1 or greater. For example, a light green and a dark red differ both by color (hue) and by lightness, so they would pass if the contrast ratio is at least 3:1. Similarly, if content is distinguished by inverting an element's foreground and background colors, this would pass (again, assuming that the foreground and background colors have a sufficient contrast). However, if content relies on the user's ability to accurately perceive or differentiate a particular color an additional visual indicator will be required regardless of the contrast ratio between those colors. For example, knowing whether an outline is green for valid or red for invalid.
> 
> Source: https://www.w3.org/WAI/WCAG21/Understanding/use-of-color
> 
> Edit: Sorry for the jinx!

@braydenlove thank you for the break down & having it documented! 💖

src/components/tutorial-meta/components/variant-list/variant-list.module.css
```diff
@@ -0,0 +1,67 @@
...
+	padding: 6px 12px;
+	box-shadow: var(--token-elevation-low-box-shadow);
+}
+
+.activeLink {
+	background-color: var(--token-color-foreground-action);
+	border-color: var(--token-color-foreground-action-hover);
+	color: var(--token-color-foreground-high-contrast);
+	cursor: not-allowed;
+	box-shadow: none;
```
kendallstrautman, line 52: We need a lighter color for the box-shadow here, working to get that info from @braydenlove now.

src/components/tutorial-meta/components/variant-list/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+				<ul aria-labelledby={VARIANT_LIST_ID} className={s.list}>
+					{variant.options.map((option: TutorialVariantOption) => {
+						const variantParam = getVariantParam(variant.slug, option.slug)
+						const isActiveOption = activeVariantOption.slug === option.slug
+						return (
+							<li key={option.slug}>
+								<Link
+									className={classNames(
+										s.link,
+										isActiveOption ? s.activeLink : s.inActiveLink
```
ashleemboyer, line 35: 🙋🏻 Thoughts on hooking the `activeLink` styles directly into the state of the link via `[aria-current="page"]` selector? Have done this in the past in a few components to ease maintenance of accessibility requirements, inspired by [Using CSS to Enforce Accessibility](https://adrianroselli.com/2021/06/using-css-to-enforce-accessibility.html).

ashleemboyer (approved):

zchsh (approved):

src/components/tutorial-meta/components/variant-list/index.tsx
kendallstrautman: That's a good idea!

kendallstrautman: Refactored in https://github.com/hashicorp/dev-portal/pull/1931/commits/04030ec10beb11bc119b572409a77f31b06a5a9d 💖

braydenlove (approved):
G O T E A M !

PR #1936, opened 2023-05-17
feat: search tab hints in empty state

zchsh:
## 🔗 Relevant links

- [Asana task](url) 🎟️

### DIFF
--- a/src/components/command-bar/commands/search/components/dialog-body/index.tsx
+++ b/src/components/command-bar/commands/search/components/dialog-body/index.tsx
@@ -36,6 +36,7 @@ import {
 	TabHeadingWithCount,
 } from '../'
 import s from './search-command-bar-dialog-body.module.css'
+import { NoResultsMessageProps } from '../no-results-message'
 
 // TODO(brkalow): We might consider lazy-loading the search client & the insights library
 const appId = process.env.NEXT_PUBLIC_ALGOLIA_APP_ID
@@ -48,7 +49,14 @@ const PRODUCT_SLUGS_WITH_INTEGRATIONS =
 interface SearchableContentTypeTab {
 	heading: TabProps['heading']
 	icon: TabProps['icon']
-	content: TabProps['children']
+	/**
+	 * TODO: maybe noResultsSlot within each
+	 * <Type>TabContents component would make more sense?
+	 */
+	renderContent: ({
+		tabData,
+		currentTabIndex,
+	}: NoResultsMessageProps) => TabProps['children']
 }
 
 const SearchCommandBarDialogBodyContent = ({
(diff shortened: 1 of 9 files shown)


src/components/command-bar/commands/search/components/dialog-body/index.tsx
```diff
@@ -69,47 +61,25 @@ const SearchCommandBarDialogBodyContent = ({
 		return generateSuggestedPages(currentProductTag?.id as ProductSlug)
 	}, [currentProductTag])
 
-	/**
-	 * Generate an object used to render all of the Tab elements to preselect the
-	 * Tab for the CurrentContentType.
-	 */
-	const tabsBySearchableContentType = useMemo<
```
zchsh, line 76: I've split this out to a separate file, which now exports `tabContentByType` object very similar to this `tabsBySearchableContentType`.

One notable difference is `tabContentByType` uses a `renderContent` function rather than a `content` slot, so that we can pass in the changing-at-render-time `currentProductTag` and `suggestedPages`. And `renderContent` also lets us pass in the revised `noResultsMessageSlot`, where the real substance of this PR happens.

src/components/command-bar/commands/search/components/dialog-body/index.tsx
```diff
@@ -166,7 +136,20 @@ const SearchCommandBarDialogBodyContent = ({
 							icon={icon}
 							key={contentType}
 						>
-							{content}
+							{renderContent({
+								currentProductTag,
+								suggestedPages,
+								noResultsMessageSlot: (
```
zchsh, line 142: Previously the `<NoResultsMessage />` was rendered in the same way within each `<[ContentType]Contents />` component. We now need to pass some data into the no results message, to call out which tabs have results as the user searches.

Rather than thread that data through each content-type component separately, I switched to a slot interface. You'll see this `noResultsMessageSlot` change in each of the `<[ContentType]Contents />` components as well.

src/components/command-bar/commands/search/components/no-results-message/index.tsx
```diff
@@ -3,14 +3,56 @@
...
+		icon: ReactElement<React.JSX.IntrinsicElements['svg']>
+	}[]
+	currentTabHeading: string
+}
+
+/**
+ * Renders a "no results" message, that optionally directs visitors
+ * to other content-type tabs with results, if such results are present.
+ */
+function NoResultsMessage({
```
zchsh, line 25: I've updated the `NoResultsMessage` component to render information on which other tabs have results. If no other tabs have results, we have a simpler "no matches" message. 

I've included content type icons here to hopefully add a bit of visual callback to the tabs themselves - cc @heitzke, let me know what you think 🙇‍♂️ Easy to remove if we'd prefer to omit them.

src/components/command-bar/commands/search/components/no-results-message/index.tsx
```diff
@@ -3,14 +3,56 @@
...
+	tabsWithResults,
+	currentTabHeading,
+}: NoResultsMessageProps) {
+	const { currentInputValue } = useCommandBar()
```
zchsh, line 29: We use `currentInputValue` from the command bar context to show the current query, so we can say "no results for X query" rather than just "no results". I think the intent here is to help users see things like typos, by making the query text a little more visible when there are no results.

src/components/command-bar/commands/search/components/dialog-body/index.tsx
kendallstrautman: I like it! The slots seem to be a great pattern.

kendallstrautman (approved):
Works great in the preview and code is v clean 💯 🧼 👍🏻

heitzke:
Nice! Can we change the copy to drop the "Sorry", make the second paragraph body-200 and maybe vertically center the copy in the modal body? I'm not sure how it'll look centered up but feels a little odd floating there at the top within all that space.

heitzke (approved):
This all looks good to me and thanks for checking into the vertical centering. It feels... not worth doing if it's a hassle and has bigger implications as it's strictly just a visual nicety imo.

PR #1937, opened 2023-05-17
[wip] feat: adds variants context

kendallstrautman:
## 🔗 Relevant links

- [Asana task](url) 🎟️


src/lib/learn-client/api/tutorial/formatting.ts
```diff
@@ -128,3 +141,16 @@ export function formatHandsOnLab({
...
 	return handsOnLab
 }
+
+function formatVariant(variant: ApiTutorialVariant): TutorialVariant {
+	const { options, ...rest } = variant
+	return {
+		...rest,
+		options: options.map(
+			({ display_order, ...rest }: ApiTutorialVariantOption) => ({
+				displayOrder: display_order,
```
kendallstrautman, line 151: We have a convention in this client of manipulating api data to camelCase

src/views/tutorial-view/utils/variants/types.ts
```diff
@@ -1,12 +1,12 @@
...
 
-export interface TutorialVariantOption {
-	id: string
-	name: string
-	slug: string
+export interface TutorialVariant extends ClientTutorialVariant {
+	activeOption: TutorialVariantOption
 }
+
+export type TutorialVariantOption = ClientTutorialVariantOption
```
kendallstrautman, line 12: I plan to adjust these type names so they dont duplicate the api client types in a follow up PR. It would just cause too much conflict with already open UI work to do so now.

src/views/tutorial-view/index.tsx
```diff
@@ -57,9 +56,6 @@ import {
 } from './components'
 import s from './tutorial-view.module.css'
 
-// @TODO temporary stub of variant data
-import variantData from './utils/variants/_tmp-variants-data.json'
```
kendallstrautman, line 61: Will remove the stub data in another PR to avoid conflicts with open UI work.

src/views/tutorial-view/utils/variants/context.tsx
```diff
@@ -0,0 +1,52 @@
...
+}
+
+const VariantContext = createContext({ activeVariant: null })
+VariantContext.displayName = 'VariantContext'
+
+export default function VariantProvider({
+	children,
+	variant,
+}: VariantProviderProps) {
+	const [activeVariant, setActiveVariant] = useState<TutorialVariant>(variant)
```
kendallstrautman, line 34: 💭 might just rename this to `variant` & `setVariant`, since `activeOption` is a prop and that feels repetitive.

src/views/tutorial-view/utils/variants/context.tsx
```diff
@@ -0,0 +1,52 @@
...
+	activeVariant?: TutorialVariant
+	setActiveVariant?: Dispatch<SetStateAction<TutorialVariant>>
+}
+
+interface VariantProviderProps {
+	children: ReactNode
+	variant?: TutorialVariant
+}
+
+export function useVariants(): VariantContextValue {
```
kendallstrautman, line 23:
```suggestion
export function useVariant(): VariantContextValue {
```
Maybe not make this plural to make it clear that we don't support that yet.

zchsh: 👍 I could see the case for either! I'd lean slightly towards `useVariant`, but def not blocking 👍

src/views/tutorial-view/utils/variants/context.tsx
zchsh: 👆 I dig that!

zchsh (approved):
✅ Looks great - got things working locally, but without the UI syncing, not sure if that adds any confidence but it was fun 👍

Also watched the Loomer (🪦), thanks for that 🙌 

💭 No strong opinions on the renames you mentioned in comments.

kendallstrautman:
@zchsh sorry I updated the testing steps with a blurb about updating the UI to the hook too late 🤭 But we can just test heavy when I do hook in the UI 🤞🏻

PR #1942, opened 2023-05-18
chore: remove opt-in redirects

BRKalow:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Removes opt-in redirect logic. This was only applicable to the .io domains when they were being served from `dev-portal`. Now that they have been moved, the logic here is unnecessary.


thiskevinwang (approved):
yes

PR #1945, opened 2023-05-18
chore: remove product modifier from zendesk crawlers 

kaitlynnefuery:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🤷 Why

This PR is a follow up to https://github.com/hashicorp/dev-portal/pull/1915 and removes `--[product]` from each Zendesk meta tag. Documentation showed nothing saying the `name` had to be `zd-site verification` but after deploying, the tags weren't working. I moved the product name to it's own data attribute instead. Hoping this approach works while also signaling which product the meta crawler is for for future us and for the requestor [here](https://app.asana.com/0/1202801212949828/1204197006767451/f).

## 💭 Anything else?

These Zendesk crawler tags are for the sandbox instance. There is a ticket already created to handle the change to production crawlers once this implementation is verified.

### DIFF
--- a/src/components/head-metadata/index.tsx
+++ b/src/components/head-metadata/index.tsx
@@ -90,49 +90,58 @@ export default function HeadMetadata(props: HeadMetadataProps) {
 				 */}
 
 				<meta
-					name="zd-site-verification--hcp"
+					name="zd-site-verification"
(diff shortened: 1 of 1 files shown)


BRKalow (approved):
lgtm!

PR #1953, opened 2023-05-22
feat: enable variant cookies

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksenable-variant-cookies-hashicorp.vercel.app/waypoint/tutorials/get-started-nomad-cloud/get-started-nomad-cloud) 🔎
- [Asana task](https://app.asana.com/0/1199634971449915/1204638008370801) 🎟️

## 🗒️ What

Enabled cookies on the tutorial UI interactions to the `onClick` handler.

## 🧪 Testing

- Visit this preview
- Click one of the variant ui buttons
- Validate that the cookie is set in your browser
Screenshot 2023-05-22 at 4 48 57 PM

- Test the sidebar dropdown disclosure and mobile dropdown disclosure as well

## 💭 Anything else?

Using the cookies to handle state management will occur in a follow up PR via rewrites.

### DIFF
--- a/src/views/tutorial-view/utils/variants/index.ts
+++ b/src/views/tutorial-view/utils/variants/index.ts
@@ -26,13 +26,16 @@ export function getVariantParam(
 	return `${slug}:${optionSlug}`
 }
 
-export function handleVariantCookie(slug: string, optionSlug: string) {
-	const key = `variant-${slug}`
-	const value = Cookies.get(key)
+export function handleVariantCookie(id: string, optionId: string) {
+	const key = 'variants'
+	const allVariants = Cookies.get(key)
+	const variantsObj = allVariants ? JSON.parse(allVariants) : {}
+	const optionValue = variantsObj[id]
 
-	// if it exists and its not already set with the same value
-	if (!value || value !== optionSlug) {
-		Cookies.set(`variant-${slug}`, optionSlug)
+	// if it doesnt exist, set it, if not check that its not already set with the same value
+	if (!optionValue || optionValue !== optionId) {
+		variantsObj[id] = optionId
+		Cookies.set(key, JSON.stringify(variantsObj))
 	}
 }
 
--- a/src/views/tutorial-view/utils/variants/index.ts
+++ b/src/views/tutorial-view/utils/variants/index.ts
@@ -1,3 +1,4 @@
+import Cookies from 'js-cookie'
 import {
 	TutorialVariant as ClientTutorialVariant,
 	TutorialVariantOption as ClientTutorialVariantOption,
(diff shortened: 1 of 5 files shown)


src/views/tutorial-view/utils/variants/index.ts
```diff
@@ -25,6 +26,16 @@ export function getVariantParam(
 	return `${slug}:${optionSlug}`
 }
 
+export function handleVariantCookie(slug: string, optionSlug: string) {
+	const key = `variant-${slug}`
```
BRKalow, line 30: ❓ can we store all of the variant data in a single cookie? It'll give us more control in the event we end up changing a variant slug, and keeps the number of cookies we're setting in check. We'd store a stringified object like: `{ [slug]: optionSlug }`

kendallstrautman: I could also use the uuid as the key to work around the slug issue, but good point on the single cookie, will look into that.

kendallstrautman: Updated in https://github.com/hashicorp/dev-portal/pull/1953/commits/eef1fe7b7acce97dc8eb8312b6bc8fbae30dbafb! Still on the fence re: using the uuid, worried that might make debugging more difficult, but is a better long term identifier.

BRKalow: Yeah, fair. Starting with a slug seems reasonable for now

src/views/tutorial-view/utils/variants/index.ts
```diff
@@ -25,6 +26,19 @@ export function getVariantParam(
 	return `${slug}:${optionSlug}`
 }
 
+export function handleVariantCookie(id: string, optionId: string) {
+	const key = 'variants'
+	const allVariants = Cookies.get(key)
+	const variantsObj = allVariants ? JSON.parse(allVariants) : {}
```
BRKalow, line 32: It might be a good idea to wrap the `JSON.parse()` in a try/catch, just to guard against any unexpected behavior if for some reason the cookie has been modified by something else. If it fails to parse we would just assign to `{}` in that case.

src/views/tutorial-view/utils/variants/index.ts
kendallstrautman: [Looping in this slack convo for shared context,](https://hashicorp.slack.com/archives/C050X5P9WFJ/p1684876960651469) Brandon agrees that slug is the right way to go.

Screenshot 2023-05-23 at 2 32 08 PM

BRKalow (commented):
legitness 🍪  😋 

🚢

PR #1955, opened 2023-05-23
chore: tweak tabs to allow vertical centering

zchsh:
This PR targets #1936, building off of it to center the contents of the "no results" tab.

Due to how our `Tabs` component is constructed, this currently requires a few changes:
1. Removal of a wrapper `<div />` within `Tabs`
    - As far as I can tell this was in place so that `TabNestingProvider` could wrap all tabs content
    - Using a `<div />` here seems to limits layout possibilities
    - Have switched to `React.Fragment`
2. Styling of `.tabsWrapper > [role="tabpanel"]`
    - This feels a little bit like dangerously reaching in to the `Tabs` component...
    - Maybe this would be better as a layout prop on `Tabs`? Or maybe `Tabs` should accept a `tabPanelClass`? Open to ideas here, adding the `[role="tabpanel"]` for now to get styles in place for discussion.

## 🎨 Screenshots

| Before | After |
| - | - |
| docs-before | docs-after |
| tutorials-before | tutorials-after |
| integrations-beofre | integrations-after |

## 💭 Anything else?

While simple on the surface, this is unfortunately a relatively complex change to make as it will involve changing some small things about how our global `Tabs` component works. So, split this out from #1936 in order to be able to better test and VQA these changes, to help confirm they're not having unexpected effects on other `Tabs` uses.

### DIFF
--- a/src/components/command-bar/commands/search/components/custom-hits-container/custom-hits-container.module.css
+++ b/src/components/command-bar/commands/search/components/custom-hits-container/custom-hits-container.module.css
@@ -7,3 +7,9 @@
 	margin-bottom: 12px;
 	margin-top: 12px;
 }
+
+.noResultsSlot {
+	display: flex;
+	flex-direction: column;
+	flex: 1 0 0;
+}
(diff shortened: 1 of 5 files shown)


zchsh:
Closing this PR, messaged with @heitzke and aligned on this being a relatively high-effort tweak 👍

PR #1957, opened 2023-05-24
[Tutorials Front Door] ContentCardLink, CollectionContentCardLink, CertificationContentCardLink

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-tfd-ambcollection-card-hashicorp.vercel.app/swingset/components/tutorialslandingview) 🔎
- Asana tasks 🎟️
  - [CollectionContentCardLink](https://app.asana.com/0/0/1204660968035226/f)
  - [CertificationContentCardLink](https://app.asana.com/0/0/1204660968035227/f)

## 🗒️ What

- Adds `TutorialsLandingView` component under `src/components`
- Adds 3 subcomponents of `TutorialsLandingView`:
  - `ContentCardLink`
  - `CollectionContentCardLink`
  - `CertificationContentCardLink`
- Adds product graphics for `CollectionContentCardLink` and `CertificationContentCardLink`
- Adds a swingset page for previewing the components with their product graphics

## 🤷 Why

- `TutorialsLandingView` is under `src/components` so that we can have a dedicated Swinget page for testing

## 📸 Design Screenshots

[Figma link](https://www.figma.com/file/mmbCubXM9sOQqvGG0b8o0D/Tutorial_LP?type=design&node-id=519-26012&t=zidhDWZL6umha1KR-11)

CleanShot 2023-05-26 at 11 23 53@2x

## 🧪 Testing

- [ ] Visit [the new Swingset page](https://dev-portal-git-tfd-ambcollection-card-hashicorp.vercel.app/swingset/components/tutorialslandingview) on the preview
- [ ] Verify that the `CollectionContentCardLink`s are rendering the correct elements & graphics
- [ ] Verify that the `CertificationContentCardLink` are rendering the correct elements & graphics
- [ ] Change the theme to dark mode by setting `data-theme="dark"` on the `<html>` element
- [ ] Verify that the cards are still readable as expected (Swingset text won't be)

## 💭 Anything else?

- The spacing values in the CSS aren't polished and may be changing in Figma. @kellyschummer and I will refine those more closely during VQA & polish next week.


src/components/tutorials-landing-view/tutorials-landing-content-card-link/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+	backgroundImageUrl?: string
+	badges: ContentCardLinkBadge[]
+	description: string
+	eyebrowParts: ContentCardLinkEyebrowPart[]
+	headerImageUrl?: string
+	href: string
+	title: string
+}
+
+const TutorialsLandingContentCardLink = ({
```
BrandonRomano, line 25: Oh one final thought; I don't think it's fully needed to prefix all of your components with `TutorialsLanding`; I think that is provided by its placement in the `tutorials-landing-view`!  I would personally like to avoid the added verbosity of that.  Not a super strong opinion!

src/components/tutorials-landing-view/tutorials-landing-content-card-link/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+interface ContentCardLinkBadge {
+	icon: $TSFixMe
+	label: string
+}
+
+interface ContentCardLinkProps {
+	backgroundImageUrl?: string
+	badges: ContentCardLinkBadge[]
+	description: string
+	eyebrowParts: ContentCardLinkEyebrowPart[]
```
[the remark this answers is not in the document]
ashleemboyer: > 💭 What's the point of this type alias? Not sure I understand the value of that here.

`ContentCardLinkEyebrowPart` is used alongside `eyebrowParts.map()` in the markup here: https://github.com/hashicorp/dev-portal/blob/7efd634ae890f16dac22f9793c4c9b172fa98d58/src/components/tutorials-landing-view/content-card-link/index.tsx#L71

So if the type of `eyebrowParts` changes, from a `string` to an object for example, then there's one less thing in the markup to update.

> 💭 Should this be optional in the type interface: eyebrowParts?: ...

Yep! The types weren't polished yet at the time of your review; I'll make sure to add more `TODO`s to make that clearer in my next PRs!

src/components/tutorials-landing-view/tutorials-landing-content-card-link/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+const TutorialsLandingContentCardLink = ({
+	backgroundImageUrl,
+	badges,
+	description,
+	eyebrowParts,
+	headerImageUrl,
+	href,
+	title,
+}: ContentCardLinkProps) => {
+	const uniqueId = useId()
```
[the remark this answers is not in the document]
ashleemboyer: It’s used to ensure unique React `key` properties in 2 places:

- [`eyebrowParts`](https://github.com/hashicorp/dev-portal/blob/7efd634ae890f16dac22f9793c4c9b172fa98d58/src/components/tutorials-landing-view/content-card-link/index.tsx#L75)
- [`badges`](https://github.com/hashicorp/dev-portal/blob/7efd634ae890f16dac22f9793c4c9b172fa98d58/src/components/tutorials-landing-view/content-card-link/index.tsx#L91)

src/components/tutorials-landing-view/tutorials-landing-content-card-link/index.tsx
ashleemboyer: I normally do this name scoping/prefixing for a more efficient search and search+replace experience, but have removed!

src/components/tutorials-landing-view/tutorials-landing-content-card-link/tutorials-landing-content-card-link.module.css
```diff
@@ -0,0 +1,82 @@
...
+	align-items: center;
+	display: flex;
+	gap: 6px;
+	list-style: none;
+	margin: 0;
+	padding: 0;
+}
+
+.badgeListItem {
+	z-index: 1;
```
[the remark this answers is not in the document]
ashleemboyer: Yep!

src/components/tutorials-landing-view/types.ts
```diff
@@ -0,0 +1,42 @@
+import { ProductSlug } from 'types/products'
+import { type BadgeProps } from 'components/badge'
+
+interface ContentCardLinkBadge {
+	icon: BadgeProps['icon']
+	label: BadgeProps['ariaLabel']
+}
+
+type ContentCardLinkEyebrowPart = string
```
BrandonRomano, line 9: Reviving my question so it can be part of my review:

> 💭 What's the point of this type alias? Not sure I understand the value of that here.

> `ContentCardLinkEyebrowPart` is used alongside `eyebrowParts.map()` in the markup... So if the type of `eyebrowParts` changes, from a `string` to an object for example, then there's one less thing in the markup to update.
> https://github.com/hashicorp/dev-portal/pull/1957#discussion_r1206942333

⛏️ Warning - nitpick ahead:

I largely don't have strong opinions about _this_ occurrence, because it is small and I am confident we won't run into any issues. If we extrapolate this to _every time a primitive type is referenced in MDX, we should defensively alias it in the event it changes in the future_, I would disagree with this approach.

Largely, when a type changes, I **want** my editor telling me that the type is incorrect in usage (e.g a map) because I likely have other things to address around it.

Following this concrete example, in the event we re-alias your `ContentCardLinkEyebrowPart` to a more complex object, the following code below won't highlight anything as an issue, and it would happily render your complex object `{eyebrowPart}` in object form.

https://github.com/hashicorp/dev-portal/pull/1957/files#diff-a4bde531afbc58b87b80cb9dc416ebaaef1c951f3a039e6f037cb70cfd569d26R70-R79

src/components/tutorials-landing-view/collection-content-card-link/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+}
+
+const CollectionContentCardLink = ({
+	badges,
+	description,
+	href,
+	productSlug,
+	title,
+	tutorialCount,
+}: CollectionContentCardLinkProps) => {
```
BrandonRomano, line 30: I think I recommend here just passing this a collection object directly, as this would make the ergonomics of this component super simple, and reduce the burden on the consumer.  For example, right now the consumer of this component needs to know the logic of what badges go into this component, whereas we can actually deterministically coerce a collection object into those badges.

While View Model components decouple the presentation from the domain, Domain Model components can decouple the domain from the presentation.

src/components/tutorials-landing-view/content-card-link/tutorials-landing-content-card-link.module.css
```diff
@@ -0,0 +1,123 @@
...
+	&.hasLightBackgroundImage {
+		color: #0d0e12;
+
+		& .description {
+			color: inherit;
+		}
+	}
+
+	&.hasDarkBackgroundImage {
+		color: #efeff1;
```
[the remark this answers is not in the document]
ashleemboyer: We do, but they're flipped in dark mode. I plan to clean this up when I get into putting these in the page where I can realistically test dark mode (no switcher on this page). 😊

src/components/tutorials-landing-view/index.tsx
```diff
@@ -0,0 +1,16 @@
...
+import CertificationContentCardLink from './certification-content-card-link'
+import CollectionContentCardLink from './collection-content-card-link'
+import ContentCardLink from './content-card-link'
+
+const TutorialsLandingView = () => null
+
+export default Object.assign(TutorialsLandingView, {
+	CertificationContentCardLink,
+	CollectionContentCardLink,
+	ContentCardLink,
```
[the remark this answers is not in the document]
ashleemboyer: Yes, this approach makes it possible to have a Swingset page to test these components in before they're placed in the page. Swingset doesn't support files that aren't in the `src/components/` folder. If we feel strongly about excluding it from Swingset after testing, we can also remove it from the assembly branch before releasing the page!

src/components/tutorials-landing-view/collection-content-card-link/index.tsx
ashleemboyer: In this case I chose not to pass `collection` so that we can do all of our data shaping server-side rather than having the view or its subcomponents do it client-side. If a collection object had all the properties we need for rendering these cards directly accessible, without needing any `map`s or boolean comparisons, then I would choose to pass that object. But because we need to `map` over all the tutorials for a collection to get the unique set of badges, I think it's better for us to do that shaping in the same place we're fetching data and passing it into the view in the shape it needs.

src/components/tutorials-landing-view/content-card-link/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+	backgroundImageColor = 'light',
+	backgroundImageUrl,
+	badges,
+	description,
+	eyebrowParts,
+	headerImageUrl,
+	href,
+	title,
+}: ContentCardLinkProps) => {
+	const uniqueId = useId()
```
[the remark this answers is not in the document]
ashleemboyer: FWIW, I was trying to make it unique in the context of the page, across product sections, not within the component itself. But have updated in https://github.com/hashicorp/dev-portal/pull/1957/commits/39dda6179cd96fd73aa7d005209dd20a53e5a6c6!

src/components/tutorials-landing-view/certification-content-card-link/index.tsx
```diff
@@ -0,0 +1,40 @@
+import { type CertificationContentCardLinkProps } from '../types'
+import TutorialsLandingContentCardLink from '../content-card-link'
```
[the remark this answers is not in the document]
ashleemboyer: Ooop you found some stragglers! 😆 Updated in: https://github.com/hashicorp/dev-portal/pull/1957/commits/528cd406b5e8222a3678fa933837b71591777e9d

src/components/tutorials-landing-view/types.ts
ashleemboyer: So when I change the type of `ContentCardLinkEyebrowPart` to an object, here's the TS error I see:

image

> If we extrapolate this to _every time a primitive type is referenced in MDX, we should defensively alias it in the event it changes in the future_, I would disagree with this approach.

I'd probably disagree with a blanket statement as well, but this isn't meant to be a global proposal! Just felt comfortable in this case. 😊

src/components/tutorials-landing-view/collection-content-card-link/index.tsx
BrandonRomano: >  I think it's better for us to do that shaping in the same place we're fetching data

I may possibly have objections, but I'm looking to understand your perspective further / understand the motivation here 

> If a collection object had all the properties we need for rendering these cards directly accessible, without needing any maps or boolean comparisons, then I would choose to pass that object.

Is this related to a performance concern?  What's the downside if you had to do the mapping / doing the comparisons on the client side?

src/components/tutorials-landing-view/types.ts
BrandonRomano: Cool cool thanks for hearing me out 😄 ; no need to change this here!

src/components/tutorials-landing-view/collection-content-card-link/index.tsx
ashleemboyer: Performance & separation of concerns -- it's not necessary for the component to do this mapping every time it renders, and it if it doesn't _have_ to be concerned about the source data shape it doesn't feel super necessary here.

BrandonRomano: Got it, so performance is something that could influence my thoughts here, but only _after_ we have run into a performance issue.  Something to definitely keep in mind in the event this comes up.

Regarding separation of concerns:

I guess we would have to first define what the concern of this layer is. In my perspective, a domain model component's _sole_ responsibility is to _map the domain to the presentation_.  I'm curious if you have a different definition of what the concern of this layer is responsible for -- that would further help me understand your perspective here!

With the approach that we are currently taking, it makes me not understand the responsibility of this layer. If we plan on mapping the `collection` to this interface at the data fetching layer, we could just also just map it to the View Model's shape.  With that approach, at least all of the business mapping logic would be in once place.

My biggest concern is actually though the data fetching layer _also_ be responsible for mapping the content to its presentational shape. In the future, if we wanted to adjust the card to also display something else, or we wanted to update the business logic to render a new badge, we would need to update the data fetching layer.

ashleemboyer: @BrandonRomano have updated `CollectionContentCardLink` to just accept 2 props now: `productSlug` and `collection`. We've not consistently broken things up into this many layers and also try to minimize the amount of data passed from the server to the client (hence going ahead and shaping it if we're manipulating it anyways), but we don't have enough time to dig in right now anyways so I've made the changes. Let me know what you think!

src/components/tutorials-landing-view/certification-content-card-link/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+	},
+}
+
+const CertificationContentCardLink = ({
+	productSlug,
+	description,
+	href,
+	title,
+}: CertificationContentCardLinkProps) => {
+	const { url, lightOrDark } = PRODUCT_SLUGS_TO_BACKGROUND_IMAGES[productSlug]
```
BrandonRomano, line 28: 💭 We'll run into a runtime error if we try to use a product not in the `PRODUCT_SLUGS_TO_BACKGROUND_IMAGES` array, but we won't run into a type check error.  No need to address, just a thought I wanted to voice.

src/components/tutorials-landing-view/collection-content-card-link/index.tsx
```diff
@@ -0,0 +1,61 @@
+import { ProductUsed, TutorialLite } from 'lib/learn-client/types'
+import { getCollectionSlug } from 'views/collection-view/helpers'
+import { type CollectionContentCardLinkProps } from '../types'
+import ContentCardLink from '../content-card-link'
+import { BADGE_ICON_MAP, PRODUCT_SLUGS_TO_HEADER_IMAGES } from './constants'
+
+const CollectionContentCardLink = ({
+	productSlug,
```
BrandonRomano, line 8: We could calculate this from the first segment of the collection slug!

```ts
import { normalizeSlugForDevDot } from 'lib/tutorials/normalize-product-like-slug'
const productSlug = normalizeSlugForDevDot(collection.slug.split('/')[0])
```

src/components/tutorials-landing-view/content-card-link/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+	const hasDarkBackgroundImage =
+		hasBackgroundImage && backgroundImageColor === 'dark'
+
+	if (hasHeaderImage && hasBackgroundImage) {
+		developmentToast({
+			color: ToastColor.critical,
+			title: 'Error in `ContentCardLink`',
+			description:
+				'`ContentCardLink` does not accept both `headerImageUrl` and `backgroundImageUrl`. Only provide one.',
+		})
```
BrandonRomano, line 37: ✨ Super helpful toast!

BrandonRomano (approved):
Had a few final notes but nothing that will hold up my approval. 

I definitely recommend calculating `productSlug` from the collection, which will further simplify the components usage & also prevent any accidental misattribution of a collection to a product it doesn't belong to!

src/components/tutorials-landing-view/collection-content-card-link/index.tsx
ashleemboyer: Updated in:

- https://github.com/hashicorp/dev-portal/pull/1957/commits/c6a89204356847b55f0daa69b5dcb4b2933510a4
- https://github.com/hashicorp/dev-portal/pull/1957/commits/ed3313de84f8db3a19906b30eefd058825f17fe9

src/components/tutorials-landing-view/collection-content-card-link/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+}: CollectionContentCardLinkProps) => {
+	const title = collection.name
+	const description = collection.description
+	const headerImageUrl = PRODUCT_SLUGS_TO_HEADER_IMAGES[productSlug]
+	const href = getCollectionSlug(collection.slug)
+
+	const tutorialCount = collection.tutorials.length
+	const eyebrowParts = [
+		'Learning path',
+		`${tutorialCount} tutorial${tutorialCount > 1 ? 's' : ''}`,
```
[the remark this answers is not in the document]
ashleemboyer: Updated in: https://github.com/hashicorp/dev-portal/pull/1957/commits/811b43d19953fac4c806cbb6e8e5c59295fb3c83

src/components/tutorials-landing-view/certification-content-card-link/index.tsx
ashleemboyer: I'm probably going to rework the types on here once we plug in the data, so I'll make sure this is top of mind in that work.

PR #1973, opened 2023-05-31
[Tutorials Front Door] Add ProductSection subcomponent

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-tfd-ambproduct-section-hashicorp.vercel.app/tutorials) 🔎
- [Asana task](https://app.asana.com/0/0/1204660968035223/f) 🎟️

## 🗒️ What

- Adds `ProductSection` subcomponent under `TutorialsLandingView`
- Updates `CertificationContentCardLinkProps` to just accept a `certification` prop
- Decouples certification content card images from product slugs; couples them to program slugs defined under `src/content/`
- Adds `certificationProgram` property under each product object in `src/content/tutorials-landing.json`
  - Ideally, we'd be able to read data from the content files. But the properties in those files aren't defined consistently across programs in either language format or character length.
- Updates `getStaticProps` for this page to just pass through fetched data (rather than shape it for the UI components)

## 📸 Design Screenshots

[Figma link](https://www.figma.com/file/mmbCubXM9sOQqvGG0b8o0D/Tutorial_LP?type=design&node-id=221-18790&t=P7ccBpGAtbDkI5JI-11)

CleanShot 2023-05-31 at 19 13 32

## 🧪 Testing

- [ ] Go to [/tutorials](https://dev-portal-git-tfd-ambproduct-section-hashicorp.vercel.app/tutorials) on the preview URL
- [ ] Verify that the elements are the correct size, weight, etc.
- [ ] Verify that the elements are placed correctly in relation to one another
- [ ] Verify that the text colors and images are correct in dark & light themes

## 💭 Anything else?

- CSS spacing & images will be polished alongside VQA work
- The way this component sits on the page will be handled in a separate PR
- The geometric background SVGs will be handled in the separate layout PR


src/views/tutorials-landing/components/product-section/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+import s from './product-section.module.css'
+import CollectionContentCardLink from 'components/tutorials-landing-view/collection-content-card-link'
+import CertificationContentCardLink from 'components/tutorials-landing-view/certification-content-card-link'
+
+const ProductSection = ({
+	certification,
+	featuredCollections,
+	featuredUseCases,
+	product,
+}: $TSFixMe) => {
```
BrandonRomano, line 13: I think my feedback here is largely that we should first implement this component with a view model first, rather than passing the domain models in.

A very rough example of this component with a View Model approach, to illustrate:

```typescript
interface ProductSectionProps {
	id: string
	logo: ReactElement
	description: string
	topLinks: Array<StandaloneLinkProps>
	cards: Array<ReactElement>
	extraLinks?: ExtraLinks
}

interface ExtraLinks {
	title: string
	links: Array<StandaloneLinkProps>
}
```

BrandonRomano (changes requested):
Just commenting on the interface first, as that's what I expected the first pass of this PR to be. I want to get alignment on this before I review any further.

src/views/tutorials-landing/components/product-section/index.tsx
ashleemboyer: @BrandonRomano Since `ProductSection` is a composition of many base UI components and it's only needed on this page in one form, I'm not convinced we need to abstract the extra view model layer in this case. Can you share a little more about value you think this abstraction will give us?

BrandonRomano: So, largely the motivation for wanting to implement the view model has to do with future proofing.  Future proofing in both the "this may be used somewhere else in some other context", and also in the "Our content desires might be adjusted / extended in the future".

I'll provide a few examples for the _later_, because this page evolving over time is much more of a real thing.
- We may in the future want to have cards that don't just link to collections.  Maybe the "Collections of Collections" idea manifests in the future, and we want to render a different DomainModel implementation in the cards list.
- I could very easily imagine the links of "Featured use cases" in the future we want to leverage linking to something else than use cases.
- We may simply just want to render this presentation in a section below to communicate a use-case rather than a product.

These scenarios are not destined to happen, but in explicitly decoupling the _domain_ (e.g. collection) from the presentation, these possibilities become easier to support, as the View Layer would happy render any future domain that our future business needs will require.

ashleemboyer: @BrandonRomano thank you for the elaboration! Going to continue chatting via Slack to save a little time 😊

ashleemboyer: Sharing an update from Slack for posterity: we will make non-release-blocking Asana tasks for following up on abstractions

src/components/tutorials-landing-view/certification-content-card-link/index.tsx
```diff
@@ -1,38 +1,40 @@
...
 	},
-	vault: {
-		url: vaultGraphic,
+	'security-automation': {
+		url: securityAutomationGraphic,
 		lightOrDark: 'light',
 	},
 }
 
 const CertificationContentCardLink = ({
```
BrandonRomano, line 22: I would consider these `CertificationContentCardLink` changes out of scope of a well-scoped `ProductSection` PR.  No need to remove these / break this up, as I know you are moving fast but just something to be mindful of!

src/pages/tutorials/index.tsx
```diff
@@ -1,68 +1,45 @@
...
+import { stripUndefinedProperties } from 'lib/strip-undefined-props'
 import { getCollections } from 'lib/learn-client/api/collection'
 import TutorialsLandingView from 'views/tutorials-landing'
 
 const getStaticProps = async () => {
+	// Build up list of collection slugs to fetch
 	const collectionSlugsToFetch = new Set<string>()
-
-	Object.values(pageContent).forEach(({ featuredCollectionSlugs }) => {
+	Object.values(pageData).forEach(({ featuredCollectionSlugs }) => {
```
BrandonRomano, line 13: I largely think the way we have structured our JSON & this `Object.values()` treatment is a bit of an anti-pattern.

It makes the `tutorials-landing.json` object subject to an really restrictive interface where every field within the top-level object must be of the same type.  In the event we wanted to move more data into that object (for example, the "Better Together" collections), we couldn't without breaking these references here.  I think it's reasonable to imagine authors would want control of that as well.

I think the right way to structure that data would to have an array of products, ex:

```json
{
 "products": [
    {
      "slug": "terraform",
      ...
    },
  ]
}
```

I know this isn't in the context of this PR (and was decided in your spike), but it is something that is making me itch.  If you wanted to file a ticket / defer that's OK.

ashleemboyer:
Quick note -- just pushed up two small commits!

- [Handled merge conflicts](https://github.com/hashicorp/dev-portal/pull/1973/commits/c87e8ecfd1ecef15a5ac0810cd4e1a0fdff37500)
- [Fixed the grid column width](https://github.com/hashicorp/dev-portal/pull/1973/commits/269e03e70fda159b2646d796deb94faac69ff37f)

BrandonRomano (changes requested):
Looks like there's a width issue going on (the only thing blocking 🚧) that we should probably address before merging!

CleanShot 2023-05-31 at 16 45 47@2x

ashleemboyer:
@BrandonRomano I think we commented at the same time! Grid width issue should now be resolved. 😊

src/pages/tutorials/index.tsx
ashleemboyer: Since adding `CrossProductSection` I'm actually planning to ask about this! When it was decided what was authorable & non-authorable on the page, I believe we decided that the `CrossProductSection` content was totally non-authorable. I can't remember why, but for context that's why in that PR I didn't add a new property under `tutorials-landing.json` for specifying the collection slugs for that section. The JSON file is meant for authorable content and the `constants.tsx` file is meant for non-authorable content. All this to say -- we're on the same page!

BrandonRomano (approved):
As long as the visual bug is addressed nothing holding this up from approval!

src/components/tutorials-landing-view/certification-content-card-link/index.tsx
ashleemboyer: The addition of `ProductSection` is what spurred these changes, so in an ideal situation would you have preferred to see the `CertificationContentCardLink` changes in a PR that came before this one?

PR #1982, opened 2023-06-01
Statically build all variant pages for Tutorials Preview

BrandonRomano:
Updates the `getTutorialPagePaths` function to also return all variant paths.  This is needed for our tutorials preview to statically build all variant paths.

### DIFF
--- a/src/views/tutorial-view/server.ts
+++ b/src/views/tutorial-view/server.ts
@@ -33,6 +33,7 @@ import outlineItemsFromHeadings from 'components/outline-nav/utils/outline-items
 import {
 	TutorialVariantOption,
 	getTutorialViewVariantData,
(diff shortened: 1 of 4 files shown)


src/views/tutorial-view/server.ts
```diff
@@ -208,9 +212,29 @@ export async function getTutorialPagePaths(): Promise<TutorialPagePaths[]> {
...
+									tutorialSlug: [
+										collectionSlug,
+										tutorialSlug,
+										variantParam,
+									] as [string, string, string],
+								},
+							})
+						}
+					)
+				}
```
BRKalow, line 235: 💡 👍

BRKalow (approved):
makes sense to me! Good fix

PR #1983, opened 2023-06-01
[🚧 WIP] [Tutorials Front Door] Responsiveness

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-tfd-ambresponsiveness-hashicorp.vercel.app/tutorials) 🔎
- [Asana task](https://app.asana.com/0/1204535818156240/1204731013909386) 🎟️


src/views/tutorials-landing/tutorials-landing.module.css
```diff
@@ -1,48 +1,106 @@
-.root {
-	--padding-left-right: 56px;
-	--text-max-width: 1240px;
+.variables {
```
BrandonRomano, line 1: Not positive where to attribute this, but there are a few breakpoints where the cards are awkwardly tall:

CleanShot 2023-06-05 at 18 01 40@2x
CleanShot 2023-06-05 at 18 01 12@2x
CleanShot 2023-06-05 at 17 58 29@2x

BrandonRomano (approved):
I think under normal circumstances I would reject this PR, specifically to zoom in and have a discussion on the Mobile / Desktop bifurcation approach we have taken, but I know you're in a crunch to get this out the door, so we can table that conversation for after launch.

src/views/tutorials-landing/components/product-section/index.tsx
```diff
@@ -107,4 +181,29 @@ const ProductSection = ({
...
+				featuredCollections={featuredCollections}
+				featuredUseCases={featuredUseCases}
+			/>
+			<NonMobileProductSection
+				certification={certification}
+				product={product}
+				featuredCollections={featuredCollections}
+				featuredUseCases={featuredUseCases}
+			/>
+		</div>
```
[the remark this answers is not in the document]
ashleemboyer: > I think there could have been a more elegant way to handle this where we conditionally add the featuredUseCase to the primary array only for Desktop, and only eject that on mobile.

If you're referring to conditionally rendering parts using JavaScript to determine the viewport size, unfortunately that will cause hydration mismatch errors. This is why I've used the `.g-hide-on-*` classes and rendered some things twice.

> I think this is a scenario where it might have been helpful to have a conversation about this approach before diving into it; I presume it will be prohibitively expensive to undo this, so I won't reject and we can address this later, if we decide that's the right thing to do.

I wish this project had allowed more time for these discussions. 💟

src/views/tutorials-landing/tutorials-landing.module.css
ashleemboyer: To note -- these things will be refined in vqa/polish PRs!

PR #1986, opened 2023-06-02
feat: stub in All tab for unified search behind flag

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Implements the `All` tab with unified search results across `docs`, `tutorial`, and `integration` content types.

### 🛠️ Summary of `Files changed`

The vast majority of the work in this PR is within the `src/components/command-bar/commands/search/unified-search/components` directory. Within that directory:

- `unified-hit` - Builds a 🚧 **placeholder** 🚧 hit card component that works across content types
    - 🙈 This is _not_ really intended for review, rather intended to validate the other work in this PR.
    - Intent is to kind of ignore this work in this cycle of review, it'll be implemented properly in collaboration with design in a subsequent PR.
- `unified-hits-container` - Builds a `UnifiedHitsContainer` component
    - This component hooks into Algolia results with `useHits` and displays a `Tab`-based interface that filters those incoming hits by content type
- `dialog-body` - Implements `UnifiedSearchCommandBarDialogBody`
    - Replaces s placeholder introduced in #1985, so this updated implementation should now appear wherever the `enable_unified_search` flag is `true`
    - This component connects to the `staging_DEVDOT_omni` index, and uses the above `UnifiedHitsContainer` component to render hit results from that index

Each of these components have associated `helper` functions split out. The intent there was to start to establish a clearer separation between data massaging and presentation.

> **Note**: all search results, whether in the `All` or content-specific tabs, are now sourced from the [`staging_DEVDOT_omni`](https://dashboard.algolia.com/apps/YY0FFNI7MF/explorer/browse/staging_DEVDOT_omni) Algolia index. We plan to mix in the `prod_DEVDOT_omni` index once we've finalized configuration and encoded it in `mktg-terraform`, and once we've ensured the `prod_DEVDOT_omni` index has all content currently live on the site and has workflows that are flowing as expected.

## 📸 Design Screenshots

| Before | After |
| - | - |
| before | after |

## 🧪 Testing

- [ ] Visit the [preview]
    - Open the search modal, search should function as expected
    - The `All` tab should show results from all other tabs in a single list, ranked by relevance
    - Counts on each tab should be correct
    - Each tab should contain the correct content type results
    - Each result card should link to the correct URL for the result

### Skewed ranking by content-type on low-specificity queries

> **Note**: this is not something I think we should address in the scope of this PR, but wanted to make a note of it regardless!

A bit hard to explain briefly, but different content types have different searchable facets, and we only return a max of `25` results, so results counts can be weird with low-specificity queries that _should_ match "equally" across content types, but may not due to how each search object is structured. 

As an example, if your search query is `vault` you'll see `25` tutorial results and `0` results in other tabs. This seems to be because `tutorial` results for Vault more often include `Vault` as the first word in the `page_title`, which is at the top of our list for ranking.

As another example, if your search query is `waypoint` you'll see `23` docs results, `2` tutorial results, and `0` integrations results. Again, this likely has to do with 1) the lack of specificity and 2) differences in how `page_title` values are authored.

One potential solution might be to use exhaustive results. But, the UI could get kinda wild (very very long lists of results). This might be worth exploring? See https://www.algolia.com/doc/guides/managing-results/refine-results/sorting/in-depth/relevant-sort/ for details.

Another solution might be to implement separate `"filters"` for each Tab, via separate `<Configure />` components, similar to the existing multi-index implementation. However, with this approach, the tab counts won't add up, and this may be equally confusing. In the `vault` query example above, the `All` tab would still show `25` `tutorial` results (those are the unified results, merged by relevance!), and all other tabs would also show `25` results each, which might be odd (shouldn't the `All` tab then show `75` results?). 

Or, this may be a bit of a limitation of an intuitive "All" tab interface. Moving to a fully unified results list ("All" as the _only_ tab) could mitigate the confusion here, as could treating content `type` as a _facet_ rather than a tab -- in which case we'd actually use it as a facet when searching Algolia (as we do with products!), rather than a post-search filter on the fully unified list of results as we do in this implementation.

[task]: https://app.asana.com/0/1204333060736722/1204604641329009/f
[preview]: https://dev-portal-git-zsadd-unified-search-results-hashicorp.vercel.app


config/base.json
```diff
@@ -10,7 +10,8 @@
 		},
 		"algolia": {
 			"docsIndexName": "product_DEVDOT",
-			"tutorialsIndexName": "prod_LEARN"
+			"tutorialsIndexName": "prod_LEARN",
+			"unifiedIndexName": "staging_DEVDOT_omni"
```
zchsh, line 14: I've added the `algolia.unifiedIndexName` property here to power unified search. I've set it to `staging_DEVDOT_omni` in all `env` _for now_. Intent is to add in `prod_DEVDOT_omni` later.

config/preview.json
```diff
@@ -2,5 +2,8 @@
 	"extends": "base",
 	"io_sites": {
 		"max_static_paths": 10
+	},
+	"flags": {
+		"enable_unified_search": true
```
zchsh, line 7: Toggling "unified search" on in preview to make review easier. Have also toggled this on in `development.json`.

🚧 Intent is to toggle these both `preview` and `development` to `false` again before merging to `main`. (If we forget, no big stress cause `production` is `false` either way, but don't want other previews to have a half-baked search experience).

config/production.json
```diff
@@ -14,5 +14,8 @@
 			"idp_url": "https://auth.idp.hashicorp.com"
 		},
 		"product_slugs_with_integrations": ["vault", "waypoint", "nomad"]
+	},
+	"flags": {
+		"enable_unified_search": false
```
zchsh, line 19: This is redundant with the same setting in `base.json`, but felt worth adding as a safeguard to ensure half-finished unified search work never appears in production.

src/components/command-bar/commands/search/components/no-results-message/index.tsx
```diff
@@ -4,14 +4,14 @@
...
-import { SearchableContentType } from 'contexts'
+import { CurrentContentType } from 'contexts'
 import Text from 'components/text'
 import { useCommandBar } from 'components/command-bar'
 import s from './no-results-message.module.css'
 
 interface NoResultsMessageProps {
 	tabsWithResults: {
-		type: SearchableContentType
+		type: CurrentContentType
```
zchsh, line 14: Updated this as the `NoResultsMessage` now needs to support the `global` content type, which is part of the `CurrentContentType` but not `SearchableContentType`.

src/components/command-bar/commands/search/unified-search/components/dialog-body/index.tsx
```diff
@@ -1,9 +1,92 @@
...
+ * If we have an input value in the command bar, we render search results
+ * from Algolia, through the `UnifiedHitsContainer` component. We apply
+ * an initial filter for the current product context where applicable.
+ *
+ * If we don't have an input value, we render suggested pages, which are
+ * tailored to the current product context where applicable.
+ *
+ * This component also tracks recent searches through `useRecentSearches`.
+ */
+export function UnifiedSearchCommandBarDialogBody() {
```
zchsh, line 49: The `UnifiedSearchCommandBarDialogBody` component does the work of setting up the Algolia index components, managing recent searches, connecting to the CommandBar context, managing the `currentProductTag` part of that same context, and applying the related product-based filters when we're in a specific "product area" of the site (ie any URL that starts with `/<product-slug`, I think).

For the related `helpers`, I've leaned heavily into comments on each function to explain what's going on, and relative to the previous implementation have tried to split things out a bit to set the stage for further iteration and refinement.

src/components/command-bar/commands/search/unified-search/components/unified-hit/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+import s from './unified-hit.module.css'
+
+/**
+ * TODO: this is a placeholder component for a "hit" card component.
+ * It can likely be ignored for now.
+ *
+ * We'll replace this with something properly usable in a future pass
+ * to the "All" tab work for unified search.
+ */
+export function UnifiedHit({ hit }: { hit: Hit }) {
```
zchsh, line 18: Generally: Intent is to ignore the `unified-hit` work for now.

This component is the "ugly first draft" part of this PR. It's chonky and lengthy and can mostly be ignored I think - intent is to just get _something_ rendering that shows everything else is working, or to put it another way, intent is to refine the "hit" cards in a later PR rather than within this PR.

src/components/command-bar/commands/search/unified-search/components/unified-hit/helpers/build-url-path.ts
```diff
@@ -0,0 +1,48 @@
...
+// Types
+import type { Hit } from 'instantsearch.js'
+
+/**
+ * Builds a URL path to an arbitrary hit from our unified `<env>_DEVDOT_omni`
+ * Algolia indices.
+ *
+ * TODO: consider baking a `urlPath` property into each Algolia object?
+ */
+export function buildUrlPath(searchHit: Hit): string {
```
zchsh, line 12: Noteworthy: I've been considering the value of baking the `urlPath` for each item into the Algolia search object. So far leaning towards _not_ doing that. I think the effort of refining the search card will confirm that it's feasible to build a highlighted `urlPath` rather than bake it in, but if not (or if the "construct URL from disparate ranked properties" approach has weird ranking implications), might mosey back to the back-end and reconsider.

src/components/command-bar/commands/search/unified-search/components/unified-hits-container/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+import s from './unified-hits-container.module.css'
+
+/**
+ * Render search results from our unified index into content-type tabs.
+ *
+ * Note: this component needs to be used within an `InstantSearch` container
+ * imported from 'react-instantsearch-hooks-web'. That container provides
+ * the context from which `rawHits` are pulled.
+ */
+export function UnifiedHitsContainer({
```
zchsh, line 32: As with the `dialog-body` component in this PR, leaning heavily into splitting out helper functions and trying to document everything rather than going into a super heavily engineered refactor of the broader `CommandBar` etc.

Of note in this component, I tried to split out the "mostly business" logic of setting up the `tabData` vs the "mostly presentation" logic of rendering the tabs. Very open to iterating on this, in this PR and beyond 🙇‍♂️

braydenlove:
Seeing a mismatch in number of items here (24 instead of 25) could be a simple explanation but I wanted to flag it!

src/components/command-bar/commands/search/unified-search/components/unified-hit/helpers/build-url-path.ts
BrandonRomano: I am slightly on the fence as it sounds like you are on this one, but I think it's probably the right move to not hard code the URL into the indexing!

kendallstrautman (commented):
Screenshot 2023-06-15 at 9 44 22 AM
I know the UI is WIP but just calling out the view on dark mode!

src/components/command-bar/commands/search/unified-search/components/dialog-body/helpers/get-algolia-product-filter-string.ts
```diff
@@ -0,0 +1,32 @@
...
+		/**
+		 * The edition:hcp only applies to `tutorials` records, which will
+		 * never have products:hcp, but we can't apply complex filters
+		 * via the Algolia filters API parameter, so we keep it simple here.
+		 * Ref: https://www.algolia.com/doc/api-reference/api-parameters/filters/
+		 *
+		 * TODO: look into how tutorials library handles this? Does it use
+		 * the filters parameter?
+		 */
+		filterString = `products:${productSlug} OR edition:${productSlug}`
```
kendallstrautman, line 28: Looks like we do a [similar thing here](https://github.com/hashicorp/dev-portal/blob/5d49306ec4a582efa590f1fee0f81030cfb08c01/src/components/command-bar/commands/search/components/tutorials-tab-contents/index.tsx#L20)

Should we be specific about `edition:hcp` here? Since none of the other products use that?

src/components/command-bar/commands/search/unified-search/components/dialog-body/helpers/use-command-bar-product-tag.ts
```diff
@@ -0,0 +1,68 @@
...
+	const currentProductTag = useMemo(
+		() =>
+			getCurrentProductTag({
+				currentProduct,
+				currentTags,
+			}),
+		[currentProduct, currentTags]
+	)
+
+	return currentProductTag
```
kendallstrautman, line 67: 💭 do we plan to unify this logic later on with the _non-unified-search_ `dialog-body`? [Perhaps a breadcrumb ](https://github.com/hashicorp/dev-portal/blob/5d49306ec4a582efa590f1fee0f81030cfb08c01/src/components/command-bar/commands/search/components/dialog-body/index.tsx#L161-L210)to the other file would be good.

src/components/command-bar/commands/search/unified-search/components/unified-hit/helpers/build-url-path.ts
```diff
@@ -0,0 +1,48 @@
...
+			product: { slug: product_slug } as $TSFixMe,
+			organization: { slug: organization_slug } as $TSFixMe,
+			slug,
+		} as $TSFixMe)
+	} else {
+		/**
+		 * Something's gone wrong, this should never happen in our indexing.
+		 * Link to the home page as insurance. And ideally, should log
+		 * an error here.
+		 * TODO: log an error maybe?
```
kendallstrautman, line 44: Should we just add the error log? 😆

src/components/command-bar/commands/search/unified-search/components/unified-hit/unified-hit.module.css
```diff
@@ -0,0 +1,9 @@
+.withHighlight {
+	color: black;
```
kendallstrautman, line 2: Total nit: but this is probably the value we want to make it work with theming. Totally cool if you want to adjust when actually doing the UI later on 👌🏻 
```suggestion
	color: var(--token-color-foreground-primary);
```

src/components/command-bar/commands/search/unified-search/components/unified-hits-container/helpers/filter-unified-search-hits.ts
```diff
@@ -0,0 +1,48 @@
...
+ * incoming value has no corresponding `type` (such as for `global`).
+ */
+const algoliaContentTypeMap: Record<
+	CurrentContentType,
+	AlgoliaContentType | null
+> = {
+	integrations: 'integration',
+	docs: 'docs',
+	tutorials: 'tutorial',
+	global: null,
```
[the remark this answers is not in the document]
kendallstrautman: Oh I see now in the other file, global represents this 'all' or unified search. nvm 😄

src/components/command-bar/commands/search/unified-search/components/dialog-body/helpers/use-command-bar-product-tag.ts
kendallstrautman: Or is the plan to fully delete the old non-unified search once this is ready?

src/components/command-bar/commands/search/unified-search/components/unified-hits-container/helpers/get-should-render-integrations-tab.ts
```diff
@@ -0,0 +1,31 @@
...
+	currentProductSlug: ProductSlug
+): boolean {
+	let shouldRenderIntegrationsTab
+	if (PRODUCT_SLUGS_WITH_INTEGRATIONS.length <= 0) {
+		// If no products have integrations, do not render it
+		shouldRenderIntegrationsTab = false
+	} else if (currentProductSlug) {
+		// If there is a product tag, render it if the product has integrations
+		const productHasIntegrations =
+			PRODUCT_SLUGS_WITH_INTEGRATIONS.includes(currentProductSlug)
```
kendallstrautman, line 24: Re: [the issue that](https://github.com/hashicorp/dev-portal/pull/1986#issuecomment-1589666460)  @braydenlove brought up. It looks like `packer` isn't in the config integration slugs array so that's likely the source of the issue? 
Screenshot 2023-06-15 at 10 06 04 AM

src/components/command-bar/commands/search/unified-search/components/unified-hits-container/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+						>
+							{hitCount > 0 ? (
+								<>
+									<div id={resultsLabelId} className="g-screen-reader-only">
+										{type} search results
+									</div>
+									<div className={s.commandBarListWrapper}>
+										<CommandBarList ariaLabelledBy={resultsLabelId}>
+											{/* TODO: would be great to have a more detailed type 
+											    here for `hit`, but for now, not over-engineering. */}
```
kendallstrautman, line 81: 👍🏻

kendallstrautman (approved):
Looks great! A few random comments but nothing blocking 💯 I think rebuilding it from scratch is smart instead of refactoring the current dialog body.

src/components/command-bar/commands/search/unified-search/components/unified-hits-container/helpers/get-should-render-integrations-tab.ts
kendallstrautman: ☝🏻 not necessary to solve now, just trying to find a source of the unexpected behavior.

zchsh: ah yeah thanks for this - I believe the issue boils down to:
- The Algolia index will contain all the latest integrations records, and return those if they're relevant
- The Dev Dot UI needs to only show integrations results for products that have explicitly been toggled on in `dev-portal` `config`

I think one potential solution here will be to try to apply an additional filter when fetching from Algolia, based on which products have integrations activated... so that even if a matching record for an integration result _exists_ in Algolia, we'll avoid showing it in the UI but also still fetch the full "page" of relevant records from Algolia... rather than say filter them out _after_ making the query, which would kinda work but reduce the number of results shown, perhaps significantly in some cases, and even for queries that might have many more relevant results... anyways 😅  🤔 I think I might need facets here or something... going to explore in a follow up PR!

src/components/command-bar/commands/search/unified-search/components/dialog-body/helpers/use-command-bar-product-tag.ts
zchsh: Plan is the latter, delete the non-unified search! ☺️

zchsh:
Started diving into the Integrations issue here (thanks for flagging that y'all!), and thinking that handling in a separate PR likely makes the most sense, as 1) we've got this behind a flag so not shipping to prod at all, and 2) the potential solution seems to have a decent enough scope to merit a separate PR.

With that in mind, going to drop the config flags here and merge this in! 🙇

src/components/command-bar/commands/search/unified-search/components/dialog-body/helpers/get-algolia-product-filter-string.ts
zchsh: Ah great call! Will update this to only do `OR edition:${productSlug}` if `productSlug === 'hcp'` 👍

src/components/command-bar/commands/search/unified-search/components/unified-hit/helpers/build-url-path.ts
zchsh: ah yeah, probably 😅 I added a `console.error` for now, and expanded the comment to mention wanting to maybe log this remotely in some way, so these messages end up beyond the browser, maybe even alerting us wheen they happen.

src/components/command-bar/commands/search/unified-search/components/unified-hit/unified-hit.module.css
zchsh: I dig it - made some updates so the results card work in dark mode 👍

src/components/command-bar/commands/search/unified-search/components/unified-hit/helpers/build-url-path.ts
kendallstrautman: Sounds good! The console.errors should go to datadog I think?

src/components/command-bar/commands/search/unified-search/components/unified-hits-container/helpers/get-should-render-integrations-tab.ts
kendallstrautman: Makes sense! Sounds good to me.

PR #1994, opened 2023-06-06
[Tutorials Front Door] ProductSection background svgs

ashleemboyer:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-tfd-ambproduct-section-svgs-hashicorp.vercel.app/tutorials) 🔎
- [Asana task](https://app.asana.com/0/1204535818156240/1204713803064097) 🎟️

## 🗒️ What

Adds background svgs to product sections.


BrandonRomano (commented):
I think this approach is generally good in terms of the way this is structured.  The only thing I would call out that I would like to avoid is always rendering left/right SVGs and hitting only one with display:none.  We can likely figure out a way to lift that logic so we can determine that in JS.

I would actually potentially recommend making this `productSectionWrapper` a component at this point, so you can encapsulate some of this logic (including the ProductSectionBackgroundSvg in it as a subcomponent).

```tsx
<ProductSectionWrapper productSlug="terraform" svgSide="left">
  <ProductSection...></ProductSection>
</ProductSectionWrapper>
```

ashleemboyer:
> I think this approach is generally good in terms of the way this is structured. The only thing I would call out that I would like to avoid is always rendering left/right SVGs and hitting only one with display:none. We can likely figure out a way to lift that logic so we can determine that in JS.
> 
> I would actually potentially recommend making this `productSectionWrapper` a component at this point, so you can encapsulate some of this logic (including the ProductSectionBackgroundSvg in it as a subcomponent).
> 
> ```tsx
> <ProductSectionWrapper productSlug="terraform" svgSide="left">
>   <ProductSection...></ProductSection>
> </ProductSectionWrapper>
> ```

@BrandonRomano asking for clarity -- does this mean you rather see the SVG left/right placement handled by the `.map` calls instead of the `:nth-child(even|odd)` CSS selectors?

BrandonRomano:
> @BrandonRomano asking for clarity -- does this mean you rather see the SVG left/right placement handled by the .map calls instead of the :nth-child(even|odd) CSS selectors?

Yeah, something like that!

kendallstrautman (approved):

BrandonRomano (approved):

PR #2009, opened 2023-06-16
feat: unified search hit component

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Builds out the `UnifiedHit` component, so that search results look nice!

## 📸 Design Screenshots

| Before | After |
| - | - |
| before | after |

## 🧪 Testing

- [ ] Visit the [preview]
    - Search result items should look as expected
    - Note the "meta" area does not yet match design, see `Anything else` for details

## 💭 Anything else?

In our current search experience, we show three pieces of information: title, description, and products.

In our revised designs, we retain these three pieces of information, and have floated the concept of display all possible matched text strings, such as URL paths, "headings", and "code list items", in a "meta" area at the lower part of each result list item. The addition of "headings" and "code list items" may require slightly more nuanced logic to execute as expected, and this logic is dependent on having matches highlighted in our URL paths, which is in turn dependent on back-end changes.

With this in mind, in this PR, we've taken a first cut at the "meta" area, including the URL path only. This is intended to be something we can ship and iterate on further.

| Design | Implementation |
| - | - | 
| design | implementation |

Refining the "meta" area, such as adding more matched information and highlighting URL paths or breadcrumbs, will come in later work.

[task]: https://app.asana.com/0/1204333060736722/1204875724611004/f
[preview]: https://dev-portal-git-zsunified-search-hit-component-hashicorp.vercel.app


src/components/command-bar/commands/search/unified-search/components/link-region/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+			 * "Perhaps the worst thing you can do for a block link is to wrap
+			 * everything in the <a href>"
+			 *
+			 * https://adrianroselli.com/2020/02/block-links-cards-clickable-regions-etc.html
+			 *
+			 * Re: `span` with &nbsp; — Safari will only focus on links
+			 * that have content. This markup allows focus to behave as
+			 * expected while retaining the desired sibling
+			 * 'empty link' structure
+			 */}
```
thiskevinwang, line 54: Do you think this is worth lifting into `Link` and enabling via a prop like `allowEmptyFocus`, or even conditionally checking if there are _no `children`_  ?

src/components/command-bar/commands/search/unified-search/components/link-region/types.ts
```diff
@@ -0,0 +1,34 @@
...
+import { LinkProps } from 'components/link'
+import { ReactNode } from 'react'
+
+type InheritedLinkProps = Pick<LinkProps, 'href' | 'onClick' | 'opensInNewTab'>
+
+export interface LinkRegionProps extends InheritedLinkProps {
+	/**
+	 * Children to render in the linked region.
+	 */
+	children: ReactNode
```
thiskevinwang, line 15: 💭 you can get out-of-the-box `children` typing with `PropsWithChildren` from react!

2 usages:
- either `MyInterface extends PropsWithChildren`
- or `PropsWithChildren<MyInterface>`

src/components/command-bar/commands/search/unified-search/content.tsx
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Basic heading and icon content for each tab, which we'll build on.
+ */
+export const tabContentByType: Record<
+	UnifiedSearchableContentType,
+	{
+		heading: string
+		icon: ReactElement<React.JSX.IntrinsicElements['svg']>
+	}
+> = {
```
thiskevinwang, line 19: 💭 What do you think about replacing the the type annotation (`: Record<...`) here with `as const`, (`export const tabContentType = {...} as const`)?

Thought: This frees us up from manual type annotation and lets us lean into automatic type inference instead!

thiskevinwang (approved):
LGTM — Left some of stylistic & compile-time comments!

Also, the preview looks great ✨ 🔍 !

braydenlove:
Added feedback here: https://docs.google.com/document/d/1DdZdPHjgN2t-c7TLF8_kp4OaTnaiNxIo65RPkbDGE1I/edit

zchsh:
Sweet, thanks for the design feedback @braydenlove  🙇‍♂️ I dig the doc format! 🙌 

Here's my plan to address it (in progress but figured worth sharing for context):

## Unified Search changes

These are the feedback items that currently fit "behind" the unified search flag. We'll implement them now, but they won't show up in production until we launch the unified search "All" tab.

- **"Hit" list item styling** - I'll update the mobile styling for list items and tweak the "dot" position in this PR
- **Suggested pages styling** (aka navigational `<li >`)- I'll split this out to a separate PR. On the implementation side I think we call these "suggested pages", I'll open a PR like `feat: revise suggested pages design`.
- **Empty state styling** - I'll split this out to a separate PR, something like `feat: revise empty state design`.

## Global changes

These fixes are, at an implementation level at least, outside the scope of what gets switched out by the "unified search" feature flag. These fixes feel like something we can ship right away, independent of unified search! If the separate PRs look good, we can merge them right away, without waiting on unified search.

If we want to only ship these fixes at the same time we ship unified search, we'll need to rethink our feature flag. This is a little extra work but feasible if it's a must-have 👍


- **Command bar modal tweaks** - I'll fix up the command bar body `background-color`, command bar input `font-weight`, and command bar header area `padding` in a separate PR. I've got this work started in [dev-portal#2019 - chore: tweaks to search modal](https://github.com/hashicorp/dev-portal/pull/2019).

- **Tab badge count tweaks** - I'll update Tabs to use `BadgeCount` in [dev-portal#2020 - feat: tab count tweak for search](https://github.com/hashicorp/dev-portal/pull/2020). That PR takes a first cut at implementing `BadgeCount` from HDS, since it's not something we had previously.

---

Happy to adjust the approach here if anything feels off - overall intent is to keep each PR relatively tightly scoped, so that we can focus in on VQA for each item 💖

zchsh:
@braydenlove Okay, with the above message in mind, I think this PR is ready for re-review! To summarize:

"Hit" list items feedback have been addressed in this PR (and screenshots are in the comment above!): 

- ✅📱"Hit" mobile styling has been tweaked (we can iterate on this further in a separate PR, if we want)
- ✅`•` "Hit" list item "dot" has been tweaked (moved it 1px as suggested)

Other feedback has been split into other PRs:

- https://github.com/hashicorp/dev-portal/pull/2024
- https://github.com/hashicorp/dev-portal/pull/2025
- https://github.com/hashicorp/dev-portal/pull/2019
- https://github.com/hashicorp/dev-portal/pull/2020

I should have these other PRs ready for review later today 👍

Let me know how things look on your end 🙇‍♂️

src/components/command-bar/commands/search/unified-search/content.tsx
zchsh: If I'm following, i think one downside of `export const tabContentType = {...} as const` might be less type-backed guarantee that `tabContentByType` contains keys for each and every `UnifiedSearchableContentType`.

With the explicit `Record<UnifiedSearchableContentType, ...>` type in place, in addition to ensuring a consistent shape between each value, we get type warnings if `tabContentByType` has missing properties or unknown properties:

| Missing property | Unknown property |
| - | - | 
| missing-property | unknown-property |

That being said, definitely open to revisiting this if there's another way to get this kind of check through more lightweight type annotations 🙇‍♂️ Let me know what you think!

zchsh:
@thiskevinwang Forgot to follow up on your earlier feedback! 💖 Updated now:

 - I've lifted `LinkRegion` into `src/components`, and made 🎟 [Use LinkRegion in CardLink](https://app.asana.com/0/1202097197789424/1204909667042721/f) as a follow-up task to update `CardLink` to use it as well.
- I've updated `link-region/types.ts` to use `PropsWithChildren`
- I've added a note on the `Record<UnifiedSearchableContentType, ...>` type for `tabContentByType` - I think the explicit type gives us some benefits that are worth the extra weight, but open to other ideas! 🙇‍♂️

src/components/command-bar/commands/search/unified-search/components/link-region/types.ts
zchsh: ✅ great tip, TIL! 🙌 updated to do the `PropsWithChildren<MyInterface>` - type approach!

PR #2010, opened 2023-06-19
feat: improve unified search content type results

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

This PR refactors the "unified" search result tabs to ensure that relevant results are shown for all content types.

## Why

Upstream, we execute a single search query against Algolia for the "all" tab, and then use client-side logic to split "all" results into separate content type tabs.

While efficient, and perhaps intuitive-feeling since the "all" tab truly represents the sum of results in all other tabs, this approach comes with a very bad downside. For generic search terms that return _many_ results, the split of results across content-types is difficult if not impossible to control.

For context, for performance and experience reasons, we do not seem to want "exhaustive" search. Rather than return _every_ record that matches a given query, we return only the most relevant records up to some limit we set in Algolia configuration.

As a concrete example, imagine the limit is `100` record. The query `vault` will return a full page of results in the "all" tab, so `100` results. But when split by content type, there is no reason to expect `~ 33` results in each of the separate content tabs. Since results are ranked by relevance, _regardless of content type_, some content types by nature of their authoring practices and metadata structure will "clobber" results from other content types for particular queries. Here are two screenshots of this exact behaviour happening:

| `vault` upstream | `waypoint` upstream |
| - | - |
| vault-before | waypoint-before |

## 🛠️ How

In this PR, instead of executing a single query to Algolia, we execute four separate queries - one for the "all" tab, and one for each of the three separate content types.

This ensures that if there are any remotely relevant results in a particular content type, they'll show up in that content type tab.

> **Note**: The "all" tab will still show all results fully merged and ranked by relevance. It will still show the same contents as it did previously, with sometimes skewed results for generic terms. The only way to mitigate this while maintaining a unified index is to refine our ranking criteria and authoring practices.

## 📸 Design Screenshots

| `vault` this PR | `waypoint` this PR |
| - | - |
| vault-after | waypoint-after |

## 🧪 Testing

- [ ] Visit the [preview]
    - Unified search results should function as they do upstream
    - Generic queries such as `vault` and `waypoint` should return a relatively balanced number of results across content type tabs

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/1204333060736722/1204857859426859/f
[preview]: https://dev-portal-git-zsunified-search-separate-filters-hashicorp.vercel.app/


src/components/command-bar/commands/search/unified-search/components/unified-hits-container/helpers/gather-search-tabs-content.tsx
```diff
@@ -81,8 +84,12 @@ function getOtherTabsWithResults(
 		})
 }
 
-export function gatherSearchTabsContent(
-	rawHits: Hit[],
+/**
+ * Transform unified search results into tab data that the
+ * `UnifiedHitsContainer` component can render.
+ */
+export function gatherSearchTabsData(
```
zchsh, line 91: Slight refactor here to match some keys to make things easier, and to allow `gather-search-tabs-content` to accept a unified search results object rather than a single array of raw hits.

src/components/command-bar/commands/search/unified-search/components/unified-hits-container/index.tsx
```diff
@@ -14,58 +11,36 @@ import {
...
 import s from './unified-hits-container.module.css'
 
 /**
  * Render search results from our unified index into content-type tabs.
- *
- * Note: this component needs to be used within an `InstantSearch` container
- * imported from 'react-instantsearch-hooks-web'. That container provides
- * the context from which `rawHits` are pulled.
  */
 export function UnifiedHitsContainer({
```
zchsh, line 23: With `tabsData` logic now outside of this component, it's become much more purely presentational, taking the prepped `tabsData` and rendering the tabbed search results.

src/components/command-bar/commands/search/unified-search/components/unified-hits-container/index.tsx
```diff
@@ -14,58 +11,36 @@ import {
...
-
 					return (
 						<Tab
 							heading={heading}
 							headingSlot={
-								<TabHeadingWithCount heading={heading} count={hitCount} />
+								<TabHeadingWithCount
+									heading={heading}
+									count={type === 'global' ? undefined : hitCount}
+								/>
```
zchsh, line 43: We hide the tab count for the `global` tab - it doesn't "add up" with the other tabs, so felt more confusing than helpful to include it.

src/components/command-bar/commands/search/unified-search/types.ts
```diff
@@ -0,0 +1,16 @@
+import type { Hit } from 'instantsearch.js'
+
+// TODO: set up an enum for this
+export type UnifiedSearchableContentType =
```
zchsh, line 3: Added this types file to hold some types I found myself re-using in a couple spots.

braydenlove (approved):
Nice!

zchsh:
@BrandonRomano Re: the 4 queries, they're 4 queries all on the same unified index! So no new indexes needed 👍

As an adjacent note:  in theory, and hopefully in practice in the near future, we should be able to transition `/tutorials/library` to the new unified index as well. Once that is done (and Sentinel is migrated to Dev Dot) I _believe_ we'll then be _only_ using the unified index, which will open the door to making some significant simplifications and reducing maintenance load for our search indexing workflows

emilypersson1 (approved):
thanks Zach! this approach looks great to me 🙌

PR #2021, opened 2023-06-23
chore: remove `App.getInitialProps`

thiskevinwang:
# What

This removes our usage of the legacy `getInitialProps` API
- The Next.js docs mark this as a legacy API. https://nextjs.org/docs/pages/api-reference/functions/get-initial-props

See PR diff comments for a more detailed explanation.

---
  - https://app.asana.com/0/0/1204865849100226

### DIFF
--- a/src/pages/_app.tsx
+++ b/src/pages/_app.tsx
@@ -69,8 +69,7 @@ addGlobalLinkHandler((destinationUrl: string) => {
 export default function App({
 	Component,
 	pageProps: { session, ...pageProps },
-	host,
-}: CustomAppProps & Awaited<ReturnType<(typeof App)['getInitialProps']>>) {
+}: CustomAppProps) {
 	const flagBag = useFlags()
 	useAnchorLinkAnalytics()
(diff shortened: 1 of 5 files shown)


src/pages/_app.tsx
```diff
@@ -143,26 +142,3 @@ export default function App({
...
-	let host
-	if (ctx.req) {
-		host = ctx.req.headers.host
-	} else {
-		host = window.location.host
-	}
-
-	return {
-		pageProps,
-		host,
```
thiskevinwang, line 166: The `host` value was only used during development by `getDeployedUrl` so it's low stakes to remove.

`Component.getInitialProps` is not ever defined to `pageProps` is always an empty `{}`

...therefore, `App.getInitialProps` is safe to remove entirely.

zchsh (approved):
🙌 🧹 🍠  sweeet this feels great to get rid of! good call 💯 

Code changes look good to me, and checked out the preview & Sentinel-mode preview as well. All looked the same! 🎸 💥

PR #2025, opened 2023-06-26
feat(unified-search): revise empty state design

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Revises the unified search empty state to match designs.

## 📸 Design Screenshots

| Design | Implementation |
| - | - |
| design | implementation |

## 🧪 Testing

- [ ] Visit the [preview]
    - Open the search modal, and type a query that yields zero results
    - The empty state should match [designs]

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/1204333060736722/1204910642466499/f
[preview]: https://dev-portal-git-zsunified-search-revise-empty-state-hashicorp.vercel.app/
[designs]: https://www.figma.com/file/O3gJUgXVTuPQn6GpjcZO9T/Search?type=design&node-id=218-15512&mode=dev

### DIFF
--- a/src/components/command-bar/commands/search/unified-search/components/unified-hits-container/index.tsx
+++ b/src/components/command-bar/commands/search/unified-search/components/unified-hits-container/index.tsx
@@ -3,18 +3,17 @@ import Tabs, { Tab } from 'components/tabs'
 import { CommandBarDivider } from 'components/command-bar/components'
 import { CommandBarList } from 'components/command-bar/components'
 // Shared search
-import {
-	SuggestedPage,
-	SuggestedPages,
-	TabHeadingWithCount,
-} from '../../../components'
+import { TabHeadingWithCount } from '../../../components'
 // Unified search
 import NoResultsMessage from '../no-results-message'
+import SuggestedPages from '../suggested-pages'
+
 import { UnifiedHit } from '../unified-hit'
 import { getUnifiedHitProps } from '../unified-hit/helpers'
 // Types
 import type { Hit } from 'instantsearch.js'
 import type { UnifiedSearchTabContent } from './helpers'
+import type { SuggestedPageProps } from '../suggested-pages/types'
 // Styles
 import s from './unified-hits-container.module.css'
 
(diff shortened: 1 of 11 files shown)


src/components/command-bar/commands/search/unified-search/components/unified-hits-container/index.tsx
```diff
@@ -64,14 +64,14 @@ export function UnifiedHitsContainer({
 									</div>
 								</>
 							) : (
-								<>
+								<div className={s.noResultsWrapper}>
```
BRKalow, line 67: nit: what do you think about `NoResultsMessage` accepting `className` and passing the class down to its root element? That way we avoid an additional wrapping element here.

zchsh: I can definitely see the case! Initially I reached for the wrapper cause 1) i have come to hate `margin` and sometimes just wanna do all spacing with `padding` (😭) and 2) `className` feels like it makes it easier to reach for "CSS override" type stuff... Then realized maybe that's something we can avoid through practice rather blocking it off completely 👍

So I went in to try to revise this to remove the extra `<div />`, but had trouble:
- There's a [CSS rule in Tabs that nukes `margin-top` on `:first-child`](https://github.com/hashicorp/dev-portal/blob/zs.unified-search-revise-empty-state/src/components/tabs/tabs.module.css#L80) with a `!important` 😬 (cruft from pre-MDX-component days I think?)
- Setting `padding` on `noResultsWrapper` as a separate wrapper works, but setting it on the same root element as `NoResultsMessage` means we undo the component's desired internal `padding`

With that in mind, left it as-is for now, and made a task 🎟 [Remove !important margin-top setting on Tabs tabpanel :first-child](https://app.asana.com/0/1202097197789424/1204918499533960/f) for us to fix up the root issue here in the future.

zchsh:
@braydenlove This PR revises the empty state for unified search! ☺️

src/components/command-bar/commands/search/unified-search/components/unified-hits-container/index.tsx
BRKalow: Dang 🥲 well thanks for trying! 🙏

BRKalow (approved):
✨

braydenlove (approved):
Awesome job, 🚢

PR #2027, opened 2023-06-26
_app layout: remove unused data prop

thiskevinwang:
# Description

Remove unused `data` prop, reducing dependency on `.layout` pattern.

### DIFF
--- a/src/pages/_app.tsx
+++ b/src/pages/_app.tsx
@@ -121,7 +121,7 @@ export default function App({
 													}
 													strict={process.env.NODE_ENV === 'development'}
 												>
-													<Layout {...allLayoutProps} data={allLayoutProps}>
+													<Layout {...allLayoutProps}>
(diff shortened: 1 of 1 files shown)


zchsh (approved):
✅ Confirmed this change shouldn't break anything:

1. `data` prop was [previously used for Dato-connected `io` layouts](https://github.com/hashicorp/dev-portal/blob/89a908d8bed84f36786674ff19dc2b4c5e2a93a6/src/layouts/_proxied-dot-io/vault/index.tsx#L34), which is why it existed in the first place. All Dato-connected `io` layouts have been removed (Sentinel's still here, but not connected to Dato)
2. Searching `<Layout />` in `_app` is either `EmptyLayout` or `Component.layout`. A search in the project for `.layout` yields a relatively small number of results (even without a more clever regex), and confirmed none of them assign a component with a `data` prop 👍

PR #2007, opened 2023-06-15
feat: adds theming homepage

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksadds-theming-homepage-hashicorp.vercel.app) 🔎
- [Asana task](https://app.asana.com/0/1204737346923690/1204737631862868) 🎟️
- [Figma File](https://www.figma.com/file/v0Hyqvjdf6QJAvRkdIBV24/Dark-Mode?type=design&node-id=1678-56758&t=ss97Qw0GIBxK8LXL-0) 🎨 

## 🗒️ What

Adds theming support to the homepage.

Screenshot 2023-06-20 at 4 27 18 PM

## 🛠️ How

Since we want to keep the homepage 'dark' in both dark and light themes, in dark mode we've done some overrides to switch tokens to the values that will retain the page styles. 

## 📸 Design Screenshots

Screenshot 2023-06-20 at 4 28 25 PM

## 🧪 Testing

- [Visit the homepage](https://dev-portal-git-ksadds-theming-homepage-hashicorp.vercel.app), toggle between themes
- Validate that the dark theme [matches designs](https://www.figma.com/file/v0Hyqvjdf6QJAvRkdIBV24/Dark-Mode?type=design&node-id=1678-56758&t=ss97Qw0GIBxK8LXL-0) and light theme design is retained

PR #2028, opened 2023-06-26
_app/layouts: remove `.theme` usage

thiskevinwang:
# Description

This continues off https://github.com/hashicorp/dev-portal/pull/2027 and cleans up no longer used `.theme` code after #2007 

> **Note**: See https://github.com/hashicorp/dev-portal/pull/2007/files#r1242886411 for the last identified case of `<PageComponent>.theme`

### DIFF
--- a/src/views/homepage/index.tsx
+++ b/src/views/homepage/index.tsx
@@ -7,7 +7,6 @@
 import { ReactElement } from 'react'
 
 // Global imports
-import { GlobalThemeOption } from 'styles/themes/types'
 import BaseNewLayout from 'layouts/base-new'
 
 // Local imports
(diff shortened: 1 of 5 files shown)


src/views/homepage/index.tsx
```diff
@@ -7,7 +7,6 @@
 import { ReactElement } from 'react'
 
 // Global imports
-import { GlobalThemeOption } from 'styles/themes/types'
```
thiskevinwang, line 10: Cleans up a missed dead import ([ref](https://github.com/hashicorp/dev-portal/pull/2007/files#diff-9ae2c4a76d60f41ee367a30dc362054b7b3121f18abdaa0213a6ada51e60f994L10))

zchsh (approved):
🔪 🎉 🥲 nice!

To help confirm this won't cause any issues - I searched `.theme` in the project, excluded `*.css` files, and the results list was small and had no `PageComponent.theme` assignments 👍

Removal of `forcedTheme` prop from `ThemeProvider` also seems very reasonable!

PR #2030, opened 2023-06-27
feat: adds themed image support

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksadds-themed-image-support-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1199634971449915/1204334498378837) 🎟️

## 🗒️ What

|  dark  | light  |  
|---|---|
|  Screenshot 2023-06-23 at 10 45 40 AM |  Screenshot 2023-06-23 at 10 45 47 AM |   

This PR adds support for conditionally hiding images based on theme states by appending a hash to the md source content:
```md
![dark theme](/img/themed/dark-test.png#hide-on-light)
```
The `rewriteStaticAssets` plugin was adjusted to support this hash, and a `getTheme` function is used within the `Image` component to determine whether the `data-hide-on-theme` attribute should be added. 

## 🧪 Testing

- [Visit the preview](https://dev-portal-git-ksadds-themed-image-support-hashicorp.vercel.app/well-architected-framework/operational-excellence/operational-excellence-introduction) where a stub image with the hash is in use via  [staging](https://github.com/hashicorp/tutorials/blob/staging/content/tutorials/well-architected-framework/operational-excellence-introduction.mdx#L15-L16). 
- A[nd another](https://dev-portal-git-ksadds-themed-image-support-hashicorp.vercel.app/well-architected-framework/com/cloud-operating-model) that has plain images aswell
- Switch themes, validate that the asset changes
- Check the markup, validate that two images are there with the `data-hide-on-theme` attribute set. 
- Validate that un-themed images are unaffected in tutorials content, docs content etc.

## 💭 Anything else?

In follow up work, I'll add a remark plugin that automatically detects width / height for images, I chose to separate that out from this work.

PR #2034, opened 2023-06-27
feat: adds image width / height plugin

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksadds-image-width-height-plugin-hashicorp.vercel.app) 🔎
- [Asana task](https://app.asana.com/0/1199634971449915/1204921235104813) 🎟️

## 🗒️ What

This PR adds a plugin that automatically calculates the width / height of assets from the mktg-content-api for tutorials content. This work is related to #2030, but could be applied to other long-form content sources. This allows us to use `next/image` to optimize images. 

Screenshot 2023-06-28 at 8 10 21 AM

## 🧪 Testing

- Go to any tutorials page that has images. Validate that the images are using `next/image` by checking that the `img` tag has a `srcset`. 
- [Check a page](https://dev-portal-git-ksadds-image-width-height-plugin-hashicorp.vercel.app/well-architected-framework/com/cloud-operating-model) with themed image
- [Check a page](https://dev-portal-git-ksadds-image-width-height-plugin-hashicorp.vercel.app/consul/tutorials/kubernetes/kubernetes-admin-partitions) with a no mktg-content-api asset url
- [Check a docs](https://dev-portal-git-ksadds-image-width-height-plugin-hashicorp.vercel.app/vault/docs/what-is-vault) page with images

## 💭 Anything else?

Mid-way through making this, I realized a version of this was already being done with docs images via this plugin: https://github.com/hashicorp/mktg-content-workflows/blob/main/shared/transforms/rewrite-site-static-assets.ts. Someday we should unify all these remark plugins 💭


src/lib/remark-plugins/remark-image-dimensions/index.ts
```diff
@@ -0,0 +1,62 @@
...
+		url.searchParams.append('height', dimensions.height)
+		return url.toString()
+	}
+}
+
+async function getImageDimensions(src: string) {
+	try {
+		return await probe(src)
+	} catch (e) {
+		if (e.statusCode === 404) {
```
kendallstrautman, line 52: Need to make sure there aren't other errors to handle

src/lib/remark-plugins/remark-image-dimensions/index.ts
```diff
@@ -0,0 +1,62 @@
...
+import { Plugin, Transformer } from 'unified'
+import { visit } from 'unist-util-visit'
+import probe from 'probe-image-size'
+
+export const remarkPluginInjectImageDimensions: Plugin = (): Transformer => {
+	return async function transformer(tree: Root) {
+		const imageNodesForDimensions = []
+		visit(tree, 'image', (node: Image) => {
+			// only transform nodes that we can fetch size on
+			if (node.url.startsWith('http')) {
```
kendallstrautman, line 11: I actually need to scope this down to just the content api urls. `process.env.ASSET_API_ENDPOINT`?

src/lib/remark-plugins/remark-image-dimensions/__tests__/index.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+		expect(img).not.toHaveAttribute('width')
+		expect(img).not.toHaveAttribute('height')
+		expect(img).toHaveAttribute('src', imgSrc)
+	})
+
+	it('does not rewrite if width/ height already defined', async () => {
+		const src = `https://content.hashicorp.com/api/assets/img/themed/test-placeholder.png?width=700&height=700`
+		const url = await getUrlWithDimensions(src)
+
```
thiskevinwang, line 46: (Just a thought): I think a helpful assertion here would be something like: 

```
expect(probe).not.toHaveBeenCalled()
```

since this is a clear indicator of "does not rewrite".

Though I wouldn't bother too much with this one as assertions on mock calls are super tedious to get correct 😅

src/lib/remark-plugins/remark-image-dimensions/index.ts
```diff
@@ -0,0 +1,71 @@
...
+		 * where async isn't supported. Taken from suggestion in this issue
+		 *  https://github.com/syntax-tree/unist-util-visit-parents/issues/8#issuecomment-1413405543
+		 */
+		for (const node of imageNodesForDimensions) {
+			const url = await getUrlWithDimensions(node.url)
+
+			if (url) {
+				node.url = url
+			}
+		}
```
thiskevinwang, line 39: (Build optimization nit): Consider using `await Promise.all()` here so we can avoid waiting on sequential image "probe" calls. 

Something like... (pseudocode, not guaranteed to work)

```ts
const promises = imageNodesForDimensions.map(async (node) => {
  const url = await getUrlWithDimensions(node.url)
  if (url) {
    node.url = url
  }
})
await Promise.all(promises)
```

Rationale: This is because remarkPlugins get called during build time for any statically generated pages (ie. docs n tutorials) so if the remarkPlugins slow down, there will be perf implications at **build** and **regeneration** time

src/lib/remark-plugins/remark-image-dimensions/index.ts
```diff
@@ -0,0 +1,71 @@
...
+		if (dimensions) {
+			url.searchParams.append('width', dimensions.width)
+			url.searchParams.append('height', dimensions.height)
+			return url.toString()
+		}
+	} catch (e) {
+		/**
+		 * Catching all errors here to prevent urls from breaking
+		 * if not found or formatted improperly
+		 */
```
thiskevinwang, line 64: 👏 👍

thiskevinwang (changes requested):
Looks pretty good to me!

- requesting a change to clean up an unused test mock
- left a suggestion in the remark plugin to replace `for-loop` with `Promise.all` to reduce build-time perf hits

thiskevinwang (approved):
Approving this! (since i'm heading out to PTO for a few and want to keep you unblocked)

The points from the [previous review](https://github.com/hashicorp/dev-portal/pull/2034#pullrequestreview-1503864105) are non-critical, but do consider incorporating/addressing them 🙏

src/components/image/index.test.tsx
```diff
@@ -6,6 +6,15 @@
...
 import Image from 'components/image'
 
+const probeDimensions = { width: '500', height: '300' }
+
+// Mock the external call to get dimensions
+jest.mock('probe-image-size', () => {
+	return jest.fn(() => {
+		return probeDimensions
+	})
+})
```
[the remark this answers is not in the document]
kendallstrautman: Oh thank you! I think at one point I had the remark plugin also imported in here, will remove 🔪

src/components/image/index.test.tsx
```diff
@@ -33,4 +42,19 @@ describe('image component', () => {
...
+		const { getByAltText } = render(
+			<Image
+				src="https://content.hashicorp.com/api/assets/img/themed/test-placeholder.png?width=500&height=300"
+				alt={alt}
+			/>
+		)
+		const image = getByAltText(alt)
+
+		expect(image).toHaveAttribute('width', probeDimensions.width)
+		expect(image).toHaveAttribute('height', probeDimensions.height)
```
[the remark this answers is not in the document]
kendallstrautman: Yep! that makes sense to me, will do.

kendallstrautman, line 49:
```suggestion
		expect(image).toHaveAttribute('width', '500')
		expect(image).toHaveAttribute('height', '300')
```

src/components/image/index.test.tsx
```diff
@@ -33,4 +42,19 @@ describe('image component', () => {
...
+		expect(image).toHaveAttribute('srcset')
```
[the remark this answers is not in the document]
kendallstrautman: Yeah basically since we don't add `srcSet` on the plain `img`, do you think there's a better way to connect it with `NextImage`?

src/lib/remark-plugins/remark-image-dimensions/__tests__/index.test.tsx
kendallstrautman: Ah yeah I think I tried that and it wasn't behaving as expected so I got frustrated and did this 😅 haha

src/lib/remark-plugins/remark-image-dimensions/index.ts
kendallstrautman: Yes, great suggestion 💯

src/views/tutorial-view/utils/serialize-content.ts
```diff
@@ -44,6 +45,7 @@ export async function serializeContent(tutorial: ClientTutorial): Promise<{
 				[anchorLinks, { headings }],
 				paragraphCustomAlerts,
 				rewriteStaticAssetsPlugin,
+				remarkPluginInjectImageDimensions,
```
kendallstrautman, line 48: Add note here re: order / dependency

PR #2037, opened 2023-06-28
chore: update tutorials library to use unified index

zchsh:
- feat: stub in content type filter
- chore: rm content type filter, not useful yet
- chore: use simpler filter option

## 🔗 Relevant links

- [Asana task](url) 🎟️


config/development.json
```diff
@@ -4,10 +4,7 @@
 		"max_static_paths": 1
 	},
 	"dev_dot": {
-		"max_static_paths": 1,
-		"algolia": {
-			"tutorialsIndexName": "staging_LEARN"
-		}
```
zchsh, line 10: Removes "staging" `algolia` settings from `development.json`, so that we're always seeing data from the production index.

This is a bit of an adjacent change, rationale is that we'll _usually_ want to see our production search index when running the site locally, and will likely only every want to switch to `staging_DEVDOT_omni` if we're making search-related changes.

src/pages/tutorials/library.tsx
```diff
@@ -38,7 +38,8 @@ const DEFAULT_SLUGS = [
 ]
 
 const appId = process.env.NEXT_PUBLIC_ALGOLIA_APP_ID
-const searchClient = algoliasearch(appId, 'bf27a047ba263cba01ee9b4081965a1a')
+const apiKey = process.env.NEXT_PUBLIC_ALGOLIA_SEARCH_ONLY_API_KEY
+const searchClient = algoliasearch(appId, apiKey)
```
zchsh, line 42: This update was needed since the inline read-only Algolia API key here didn't have permission to access `prod_DEVDOT_omni`.

config/development.json
kendallstrautman: 💭 Makes me wonder if we have docs related to our search indices with this info. Like specifically doing local search development on the staging index.

kendallstrautman (approved):
Played around with it, seems to function as expected! Thank you 😄 Do we have a follow up task to tear down the learn specific algolia stuff?

config/development.json
zchsh: ooh yeah these would be a great addition 👍 made a task for myself to add some README notes in https://github.com/hashicorp/dev-portal/blob/main/docs/README.md : 🎟 Add notes on Algolia config to DevDot README

PR #2053, opened 2023-07-07
check UA in middleware

thiskevinwang:
# Description

🚫  :trollface:

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -4,7 +4,7 @@
  */
 
 import { NextResponse } from 'next/server'
-import type { NextFetchEvent, NextRequest } from 'next/server'
+import { type NextFetchEvent, type NextRequest, userAgent } from 'next/server'
 import redirects from 'data/_redirects.generated.json'
 import variantRewrites from '.generated/tutorial-variant-map.json'
 import setGeoCookie from '@hashicorp/platform-edge-utils/lib/set-geo-cookie'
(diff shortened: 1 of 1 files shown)


src/middleware.ts
```diff
@@ -48,6 +48,14 @@ function setHappyKitCookie(
 export async function middleware(req: NextRequest, ev: NextFetchEvent) {
 	const { geo } = req
 
+	// 07/07/2023 — Simple UA check
+	const { ua } = userAgent(req)
```
dstaley, line 52: Does `userAgent()` have a `ua` property? I'm not seeing it in [the docs](https://nextjs.org/docs/pages/api-reference/functions/next-server#useragent).

thiskevinwang: Yep, looks like the docs are inaccurate.

`.ua` is available in TS types, and you can also see it logged [here](https://vercel.com/hashicorp/dev-portal/2P21a73o8F1tMjUsQSCnwDxiPBZH/logs?page=1&timeline=past30Minutes&startDate=1688753652196&endDate=1688755452196&selectedLogId=168875544920475747622191926&selectedLogTimestamp=1688755449204&forceEndDate=1688755449204), from the first commit — [479ec0a](https://github.com/hashicorp/dev-portal/pull/2053/commits/479ec0a20310dac2ca07b0870343ce33ee94c39e)

CleanShot 2023-07-07 at 14 45 07@2x

dstaley (approved):

PR #2055, opened 2023-07-07
Add e2e test for tutorials content correctness

thiskevinwang:
# Description

This adds e2e tests to ensure tutorials content correctness.

_Why?_ — During next.js upgrades fro m`13.0.5` to `13.4.x`, incorrect tutorials have been rendered at certain paths. This test aims to catch those cases.

### DIFF
--- a/src/__tests__/e2e/correct-content.spec.ts
+++ b/src/__tests__/e2e/correct-content.spec.ts
@@ -11,15 +11,6 @@ import { test, expect } from '@playwright/test'
 
 // this currently contains the first collection for each product and the first 2 tutorials
 const table = [
-	['/cloud/tutorials/networking', 'HashiCorp Virtual Network'],
-	[
-		'/cloud/tutorials/networking/amazon-peering-hcp',
-		'Peering an AWS VPC with HashiCorp Cloud Platform (HCP)',
-	],
-	[
-		'/cloud/tutorials/networking/azure-peering-hcp',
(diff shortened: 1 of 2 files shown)


src/__tests__/e2e/correct-content.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+import { test, expect } from '@playwright/test'
+
+/**
+ * List of dev-portal paths and expected H1's.
+ */
+const table = [
+	['/hcp/tutorials/networking', 'HashiCorp Virtual Network'],
+	['/vault/tutorials/operations/configure-vault', 'Configure Vault'],
+	['/consul/tutorials/get-started-hcp/hcp-gs-deploy', 'Deploy HCP Consul'],
+]
```
thiskevinwang, line 15: TODO: add more path+h1 pairs here.

I think having a hard-coded list here should be reasonable for the near future.

Depending on test slowness, I think 100+ tutorials should be a reasonably wide net that'll give us confidence during the next Next.js upgrade.

TODO: also add docs routes.

kendallstrautman: 💭 Definitely fine for now while we troubleshoot but if we want this test to always run in the future, it should be driven off real api data in case headings / filenames for tutorials chance.

kendallstrautman (approved):

PR #2074, opened 2023-07-13
chore: remove contentType pageComponent property

thiskevinwang:
# Description

Preview: https://dev-portal-3eus23edp-hashicorp.vercel.app/

- remove `pageComponent.contentType`
- remove CurrentContentType contenxt
- remove old search code 

---
  - https://app.asana.com/0/0/1204921002518584


src/types/_app.ts
```diff
@@ -6,16 +6,12 @@
...
-import { CurrentContentType } from 'contexts'
-import { GlobalThemeOption } from 'styles/themes/types'
 
 type CustomPageComponent<
 	PageProps = Record<string, unknown>,
 	InitialProps = PageProps
 > = NextPage<PageProps, InitialProps> & {
-	contentType?: CurrentContentType
 	layout?: FC
-	theme?: GlobalThemeOption
```
thiskevinwang, line 18: `theme` removal is some unrelated cleanup

zchsh (approved):
🙌 This looks great! Search seems to work exactly the same on the deploy preview, and also pulled down locally and tested as well.

🎶 To add a bit of context for posterity, the `contentType` removal definitely qualifies as search-related cruft - my understanding is it was previously used only for the "default search tab selection" before the "All" tab existed 👍 Thanks for this cleanup work 💖 🙇‍♂️

PR #2081, opened 2023-07-17
style: capitalize breadcrumb items

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-kstitle-case-breadcrumb-items-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1201987349274776/1202184785235854/f) 🎟️

## 🗒️ What

Capitalizes all breadcrumb items.

## 🧪 Testing

- [Visit this path](https://dev-portal-git-kstitle-case-breadcrumb-items-hashicorp.vercel.app/waypoint/commands/deploy), ensure the items are all capitalized.


braydenlove (approved):

zchsh (approved):
✅ Given the intent of this PR, I think this looks good!

Had a thought about potential alternatives, but could also see the case for going with this simpler fix, even with potential tradeoffs 👍 

---

## Thoughts on alternatives to `capitalize`

💭  Bit of a hot take, but one potential downside of making this change via client-side CSS is that we will likely see mismatches of this data in different parts of the UI. For example, with this change in place, the sidebar and breadcrumb area are now kind of mismatched, in my opinion:

| Before | After |
| - | - | 
| before | after |

As well, in some cases I think CSS-based capitalization may create some unintended capitalizations, see for example `Waypoint.Hcl` in the breadcrumb bar in this screenshot:

side-effect-example

I think there might be a couple of alternatives we could explore:

1. For docs (and tutorials I think!), we could change this content at the authoring level, and it should be reflected accurately in the breadcrumb bar. If the breadcrumb should reflect the page title, and the desired page title is for example `Deploy` and not `deploy`, I think it could make sense to change this in the source content ([this file](https://github.com/hashicorp/waypoint/blob/16bf13f56d19e4e62167517b8295543650ab011d/website/data/commands-nav-data.json#L7), I think).

2. For other content types, I think we sometimes have a shortcoming on the implementation side where we kind of have only the URL to work off of when building the breadcrumb bar... We could look into updating other content types to have full page titles built for the breadcrumb bar server-side at build time. This would be a good deal more effort than the CSS-based approach in this PR, but might be preferable in terms of reflecting authored page titles directly. If we have specific cases where the breadcrumb bar capitalization seems off, those could be helpful for prioritizing specific areas of the site!

kendallstrautman:
@zchsh definitely agreed and very good points. Marking this as a draft again and connecting with Brayden to see if we truly do want to make this change.

braydenlove:
After sleeping on it I think we should make it author-able and shouldn't force anything via CSS. I will get with Emily to figure out a plan on how to discuss this with the edu team. Ideally everything should have the same casing across the page (title, <li>, breadcrumb etc).

kendallstrautman:
Sounds good 👍🏻  I'll close this out then.

PR #2088, opened 2023-07-18
refactor: compose over config for render of mobile menu

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

This PR refactors `BaseLayout` to accept a `mobileMenuSlot` prop.

While not strictly necessary, the intent here encompasses a few ideas:

- Remove the need for global `non_product_mobile_menu_routes` config
     - This config has caused unexpected issues in the past, and we've noted from a DX perspective that it feels a little odd to have to modify global config in order to add any new route with a generic mobile menu
- Move towards a more composable mobile menu pattern
     - I think we'll want to remove the need for app-level context for open-closed state, see 🎟 [[discuss] Refactor MobileMenu to remove context, use portal instead of z-index](https://app.asana.com/0/1202097197789424/1205087147421651/f) for some details
- Set up a props API for `BaseLayout` that aligns with the above patterns
- Set the stage to build a layout for `OpenApiDocsView` that aligns with the above patterns
    - Note `"aligns"` - under the hood we're still using app-level context for now, the idea is that `OpenApiDocsView` will at least not be an _additional_ layout that relies on this context.

> **Note**: this PR also renames `BaseNewLayout` to `BaseLayout`. The `New` in the name was a historical artifact that is no longer relevant, as we've long since remove the non-`New` version of this same layout.

## 🧪 Testing

Visit all pages that previously relied on `non_product_mobile_menu_routes` global config, and confirm they still render as expected, both on desktop and on mobile viewports, where a generic mobile menu should be present:

- [ ] [/ (home page)][preview] - home page
- [ ] [/tutorials] - tutorials landing page
- [ ] [/_error] - error page, any 404 route also works
- [ ] [/certifications] - certifications landing
- [ ] [/certifications/{slug}][/certifications/infrastructure-automation] - individual certifications view

Visit another page that did not rely on `non_product_mobile_menu_routes` global config, but still renders a mobile menu. This is basically every other page not listed above. Examples include:

- [ ] [/hcp/api-docs/vault-secrets]
- [ ] [/waypoint/docs]

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/1204678746647847/1205011719572372/f
[preview]: https://dev-portal-git-zsmobile-menu-slot-hashicorp.vercel.app/
[/tutorials]: https://dev-portal-git-zsmobile-menu-slot-hashicorp.vercel.app/tutorials
[/_error]: https://dev-portal-git-zsmobile-menu-slot-hashicorp.vercel.app/_error
[/certifications]: https://dev-portal-git-zsmobile-menu-slot-hashicorp.vercel.app/certifications
[/certifications/infrastructure-automation]: https://dev-portal-git-zsmobile-menu-slot-hashicorp.vercel.app/certifications/infrastructure-automation
[/waypoint/docs]: https://dev-portal-git-zsmobile-menu-slot-hashicorp.vercel.app/waypoint/docs
[/hcp/api-docs/vault-secrets]: https://dev-portal-git-zsmobile-menu-slot-hashicorp.vercel.app/hcp/api-docs/vault-secrets


config/base.json
```diff
@@ -23,13 +23,6 @@
 		"canonical_base_url": "https://developer.hashicorp.com",
 		"max_static_paths": 10,
 		"revalidate": 360,
-		"non_product_mobile_menu_routes": [
-			"/",
-			"/tutorials",
-			"/_error",
-			"/certifications",
-			"/certifications/[slug]"
-		],
```
thiskevinwang, line 32: > This config has caused unexpected issues in the past, and we've noted from a DX perspective that it feels a little odd to have to modify global config in order to add any new route with a generic mobile menu

💯

src/layouts/base-layout/index.tsx
```diff
@@ -4,7 +4,6 @@
  */
 
 // Third-party imports
-import { useRouter } from 'next/router'
```
thiskevinwang, line 7: Excellent removal of the `pages` router — 1 less compatibility-thing to account for w/ AppRouter

src/layouts/base-layout/index.tsx
```diff
@@ -13,68 +12,44 @@ import usePageviewAnalytics from '@hashicorp/platform-analytics'
...
- * use here.
- */
-const BaseNewLayout = ({
+const BaseLayout = ({
 	children,
 	showFooterTopBorder = false,
-}: BaseNewLayoutProps) => {
-	const router = useRouter()
-
+	mobileMenuSlot,
```
thiskevinwang, line 45: (idea)
What if we removed the `BaseLayoutGenericMobileMenu` wrapper (which feels little like an unnecessary _wrapper-with-default-mobileMenuSlot-prop_) and did either...
1. set a default prop: `mobileMenuSlot={<NonProductPageMobileMenu />}`
2. or require `BaseLayout` consumers to explicitly specify the `mobileMenuSlot`. Acknowledging this is less DRY in practice, but this less-DRY-ness is realistically not that big of a problem.
  - I'm thinking of this as "A little duplication > an incorrect abstraction"

src/layouts/base-layout/index.tsx
```diff
@@ -89,7 +64,7 @@ const BaseNewLayout = ({
 						<NavigationHeader />
 					</div>
 					<div className={s.contentArea}>
-						{shouldShowMobileMenu ? <NonProductPageMobileMenu /> : null}
+						{mobileMenuSlot}
```
thiskevinwang, line 67: 😄 💭  Having the menu element as `props.mobileMenuSlot` feels like a nice precursor to 
https://nextjs.org/docs/app/building-your-application/routing/parallel-routes

src/layouts/base-layout/types.ts
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+import { ReactNode } from 'react'
+import { Products as HashiCorpProduct } from '@hashicorp/platform-product-meta'
+
+interface BaseLayoutProps {
+	/**
+	 * Content to render within the layout.
+	 */
+	children: ReactNode
```
thiskevinwang, line 13: Nit: We should move away from manually typing `children` ourselves.

React has a `PropsWithChildren` type helper which types it for us:

Here's the signature:
```ts
    type PropsWithChildren<P = unknown> = P & { children?: ReactNode | undefined };
```

---

You can use it two ways:

```ts
interface BaseLayoutProps extends PropsWithChildren {
  ...
}

// or

type MyFullyTypedProps = PropsWithChildren<BaseLayoutProps>
```

thiskevinwang (approved):
Code looks good, and the removal of `non_product_mobile_menu_routes` is a great move. No blockers from me!

Question: So at this point, is it accurate that no pages are directly using `BaseLayout`? It looks like it's either through `SidebarSidecarLayout` or `BaseLayoutGenericMobileMenu`

src/layouts/base-layout/index.tsx
zchsh: ☝️ Ooh yeah I dig this direction! Updated to remove `<BaseLayoutGenericMobileMenu>` - replaced all such uses with `<BaseLayout mobileMenuSlot={<MobileMenuLevelsGeneric />}>` 👍

The above change doesn't quite meet `1.` or `2.` exactly - definitely looking forward to improving this further post-API-docs work! Got a task at 🎟 [[discuss] Refactor MobileMenu to remove context, use portal instead of z-index](https://app.asana.com/0/1202097197789424/1205087147421651/f) where I added your notes here cause I think making `mobileMenuSlot` required seems like a great directly 👍

src/layouts/base-layout/types.ts
zchsh: I dig it, TIL! 💖 Updated ✅

zchsh:
> Question: So at this point, is it accurate that no pages are directly using `BaseLayout`? It looks like it's either through `SidebarSidecarLayout` or `BaseLayoutGenericMobileMenu`

@thiskevinwang It was before some recent changes - but I just got rid of `BaseLayoutGenericMobileMenu` so now pages are doing like `<BaseLayout mobileMenuSlot={<MobileMenuLevelsGeneric />}>` rather than the previous `<BaseLayoutGenericMobileMenu>` 👍

PR #2090, opened 2023-07-18
feat: stub OpenAPI docs mobile menu

zchsh:
- feat: add mobile-menu-levels component
- feat: add open-api-docs-mobile-menu-levels
- feat: use stubbed OpenAPI docs menu levels

## 🔗 Relevant links

- [Asana task](url) 🎟️


src/components/mobile-menu-levels/level-components/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+				<SidebarNavMenuItem item={item} key={index} />
+			))}
+		</ul>
+	)
+}
+
+/**
+ * Return mobile menu level data for the shared `main` menu pane.
+ */
+export function mobileMenuLevelMain(): MobileMenuLevelData {
```
zchsh, line 34: The `mobileMenuLevelMain` and `mobileMenuLevelProduct` are intended to be re-usable chunks of content we consistently in mobile menus. The analogous abstraction that currently exists is how we use `generateTopLevelSubNavItems` and `generateProductLandingSidebarMenuItems`, but with our `Sidebar` component (which actually renders both literal "sidebar" contents, and mobile menu panes).

src/pages/hcp/api-docs/vault-secrets/[[...page]].tsx
```diff
@@ -123,7 +123,7 @@ export const getStaticProps: GetStaticProps<
 	 * 🚩 If the flag is enabled, use the revised template
 	 */
 	if (USE_REVISED_TEMPLATE) {
-		return await getOpenApiDocsStaticProps({ params })
+		return await getOpenApiDocsStaticProps(PRODUCT_SLUG, { params })
```
Comment (zchsh) [inline src/pages/hcp/api-docs/vault-secrets/[[...page]].tsx:126]: We pass `PRODUCT_SLUG` here since we need to generate product-specific mobile menu panes to render in the view.

src/views/open-api-docs-view/components/open-api-docs-mobile-menu-levels/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+				mobileMenuLevelMain(),
+				mobileMenuLevelProduct(productData),
+				{
+					levelButtonText: 'Previous',
+					content: (
+						<div>
+							{/* API docs mobile menu contents */}
+							<div style={{ border: '1px solid magenta' }}>
+								PLACEHOLDER for OpenApiDocsView mobile menu contents
+							</div>
```
zchsh, line 33: This is where the navigational "sidebar"-like content of the `OpenApiDocsView` mobile menu will go.

src/components/mobile-menu-levels/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+import s from './mobile-menu-levels.module.css'
+
+/**
+ * This component allows the consumer to render a multi-pane mobile
+ * menu without being required to use `SidebarSidecarLayout`
+ * or the associated `sidebar-nav-data` React context.
+ *
+ * Intent is to solidify this into an alternative pattern
+ * for composing mobile menu contents.
+ */
```
kendallstrautman, line 21: 💯

src/components/mobile-menu-levels/types.ts
```diff
@@ -0,0 +1,6 @@
+import { ReactNode } from 'react'
+
+export interface MobileMenuLevelData {
+	levelButtonText: string
+	content: ReactNode
+}
```
kendallstrautman, line 6: ❓ Just a general question: do you still like defining types like this in separate files? I know its been the convention, but curious to hear your opinion.

src/components/mobile-menu-levels/level-components/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+		levelButtonText: `${productData.name} Home`,
+		content: (
+			<ul className={s.listResetStyles}>
+				{[
+					{
+						title: productData.name,
+						fullPath: `/${productData.slug}`,
+						theme: productData.slug,
+					},
+					...generateProductLandingSidebarMenuItems(productData),
```
kendallstrautman, line 69: What are your thoughts on this component being a pure 'view' and only rendering the sidebar data as opposed to also generating it?

kendallstrautman (approved):
Overall looks good to me! Just one question in particular about coupling the menu item generation with the presentational component, but not blocking

src/components/mobile-menu-levels/types.ts
zchsh: Great question! Yeah this was more habit than anything else 👍 I'd definitely be on board with being less prescriptive about a separate `types.ts` for each and every component. Sometimes it feels useful... but at the same time, I think when types for the props API of a component get complex enough that they feel useful to put in a separate file, that's probably a sign we should break down the component a bit or something 😅

src/components/mobile-menu-levels/level-components/index.tsx
zchsh: I'd be very on board! The initial intent here with `mobile-menu-level` and especially `level-components` was to make it _possible_ to not use "sidebar-nav-data` context to generate multi-level mobile menus, while keeping things consistent with existing mobile menus.

I made a task 🎟 [Refactor mobile menu "level-components" to decouple business logic & presentation](https://app.asana.com/0/1202097197789424/1205093670087688/f) and added some comments to this file to get that context written down! 😅 

In the short term, I took a lil baby step towards splitting out the logic! Let me know if that looks alright for now 🙇

src/components/mobile-menu-levels/level-components/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+			</>
+		),
+	}
+}
+
+/**
+ * Return mobile menu level data for a common `product` menu pane.
+ */
+export function mobileMenuLevelProduct(
+	productData: ProductData
```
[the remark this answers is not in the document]
zchsh: I dig this! I'm thinking that `MobileMenuLevels` will probably be a more flexible component we could use, and for those use cases we could pass whatever `MobileMenuLevelData` we want (much like we do for the "API docs" level in this PR).

In other words, `level-components` are kind of a stopgap for consistency with the existing API docs mobile menu, and def not required to be used with `<MobileMenuLevels />`!

PR #2091, opened 2023-07-19
update Nomad v1.6.0 download box

jrasell:


zchsh (approved):
✅ 💖  Thanks for this update! [The preview](https://dev-portal-git-nomad-160-download-box-hashicorp.vercel.app/nomad/downloads) looks great to me!

💭 Optional, but I think that we can delete `src/pages/nomad/downloads/style.module.css` since it's not longer used.

lgfa29:
> 💭 Optional, but I think that we can delete `src/pages/nomad/downloads/style.module.css` since it's not longer used.

Ah thanks! Normally we would keep this file because, historically, we added and removed this box during the Beta and RC period. But that has become a bit cumbersome to manage, so we will stop doing it in the future, so removing the file sounds like a good thing to do now.

PR #2092, opened 2023-07-19
feat: parse and validate OpenAPI docs data

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Adds schema loading, parsing, and validation to the revised `OpenApiDocsView`.

## 🛠️ How

- Installs [oas-normalize](https://github.com/readmeio/oas-normalize) to handle schema validation
    - This also allows us to up-convert `2.0`-versioned schemas to `3.0`-versioned schemas. The intent is to be able to work with `2.0` or `3.0` versioned schemas in a consistent way. In turn this should make our code a little more future-proof, while keeping types narrow as we know our schema will match OpenAPI 3.0 structure.

## 🧪 Testing

    - The page should show a blob of validated OpenAPI data, in a placeholder component
    - The validated OpenAPI data should have `"openapi": "3.0.0"` schema versioning, even though the source file has [swagger 2.0 schema versioning](https://github.com/hashicorp/hcp-specs/blob/main/specs/cloud-vault-secrets/preview/2023-06-13/hcp.swagger.json)
    - The validated OpenAPI data should be de-referenced, it should not contain any `$ref`. You can `⌘ + F` on the page to verify this.

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/1204678746647847/1205079140949226/f
[preview]: https://dev-portal-git-zsload-validate-open-api-docs-schema-hashicorp.vercel.app/
[/hcp/api-docs/vault-secrets]: https://dev-portal-git-zsload-validate-open-api-docs-schema-hashicorp.vercel.app/hcp/api-docs/vault-secrets


config/preview.json
```diff
@@ -4,6 +4,6 @@
 		"max_static_paths": 10
 	},
 	"flags": {
-		"enable_hcp_vault_secrets_api_docs_revision": false
+		"enable_hcp_vault_secrets_api_docs_revision": true
```
zchsh, line 7: Flags set to make review easier.

package.json
```diff
@@ -176,6 +176,7 @@
 		"next-query-params": "^4.1.0",
 		"next-sitemap": "^3.1.21",
 		"next-themes": "^0.2.1",
+		"oas-normalize": "^8.4.1",
```
zchsh, line 179: We use `oas-normalize` for schema parsing, validation, and de-referencing.

src/pages/hcp/api-docs/vault-secrets/[[...page]].tsx
```diff
@@ -118,23 +122,28 @@ export const getStaticPaths: GetStaticPaths<ApiDocsParams> = async (ctx) => {
 export const getStaticProps: GetStaticProps<
 	ApiDocsViewProps | OpenApiDocsViewProps,
 	ApiDocsParams
-> = async ({ params }: { params: ApiDocsParams }) => {
+> = async ({ params }: GetStaticPropsContext<ApiDocsParams>) => {
+	// Fetch all version data, based on remote `stable` & `preview` subfolders
+	const versionData = await fetchCloudApiVersionData(GITHUB_SOURCE_DIRECTORY)
```
Comment (zchsh) [inline src/pages/hcp/api-docs/vault-secrets/[[...page]].tsx:127]: Switched to fetch version data before the "is the new template flag on" conditional, since I think we'll just be re-using this `fetchCloudApiVersionData` code for now (until we start to tackle 🎟️  [[discuss] Plan out content API for OpenAPI specs used on Dev Dot to avoid loading from GitHub directly](https://app.asana.com/0/1100423001970639/1205041193652635/f)).

src/views/open-api-docs-view/utils/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './find-latest-stable-version'
+export * from './parse-and-validate-schema'
+export * from './sort-date-version-data'
```
thiskevinwang, line 3: How do you feel about explicit exports in this "barrel export" file, like:

```diff
- export * from './find-latest-stable-version'
+ export { findLatestStableVersion } from './find-latest-stable-version'
```

IME, while the `export *` works, the moment you have to debug something, the ambiguity behind it quickly becomes a headache / feels like a footgun

package.json
kendallstrautman: This tool seems so cool! 😄

src/views/open-api-docs-view/utils/find-latest-stable-version.ts
```diff
@@ -0,0 +1,29 @@
...
+import { sortDateVersionData } from './sort-date-version-data'
+
+/**
+ * Given an array of version data,
+ * Return the latest `stable` version.
+ *
+ * Note: only supports date-based version formats, for example "2023-01-15".
+ * We'd need to update the sort logic in order to support other formats.
+ */
+function findLatestStableVersion(
```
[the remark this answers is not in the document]
kendallstrautman: I like it! Seems like this approach worked really well with the omni search also.

src/views/open-api-docs-view/utils/parse-and-validate-schema.ts
```diff
@@ -0,0 +1,41 @@
...
+ * may receive as input here. We use the `oas-normalize` package to
+ * convert the input to the latest version, and then dereference the
+ * schema to resolve any `$ref` references.
+ *
+ * TODO: not sure if we're actually handling circular references here.
+ * This is something we'll likely need to look into for Waypoint API docs.
+ * Task: https://app.asana.com/0/1202097197789424/1203989531295664/f
+ *
+ * The input OpenAPI specification can be in any of the following versions:
+ * - OpenAPI 2.0 (formerly known as Swagger)
```
kendallstrautman, line 18: Oh interesting, are they moving away from that title, 'Swagger'?

kendallstrautman: Read the rest of your comment below, makes sense

src/views/open-api-docs-view/utils/parse-and-validate-schema.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+export async function parseAndValidateOpenApiSchema(
+	fileString: string
+): Promise<OpenAPIV3.Document> {
+	// Parse and validate the file string, and up-convert to OpenAPI 3.0
+	const rawSchemaJson = await new OASNormalize(fileString).validate({
+		convertToLatest: true,
+	})
+	// Dereference the schema to resolve any `$ref` references
+	const schemaJson = await new OASNormalize(rawSchemaJson).deref()
```
kendallstrautman, line 37: For my understanding, could you expand on what this means? why we want to do it?

kendallstrautman (approved):
LGTM according to the testing steps!

src/views/open-api-docs-view/utils/index.ts
zchsh: ah nice call! I've updated to explicitly named `export { thing }` in this file, and will do this going forward 👍

src/layouts/sidebar-layout/sidebar-layout.module.css
```diff
@@ -31,5 +33,8 @@
 }
 
 .mainArea {
-	flex: 1 1 0;
+	flex-grow: 1;
+	flex-shrink: 1;
+	min-width: 0;
+	overflow: hidden;
```
zchsh, line 39: Missed some "too-wide-content-in-main-area" scenarios in the initial `SidebarLayout` work, correcting it here! 👍

src/views/open-api-docs-view/utils/parse-and-validate-schema.ts
zchsh: Ah yeah great question! 🙌 I expanded this comment to add some details on why I think it makes sense to de-reference at this point (immediately after we read the schema) rather than at some other point.

I am a _little_ wary of whether this "de-reference before anything else" approach will be feasible if we have to deal with circular references (which the Waypoint server OpenAPI schema seems to have)... but thinking we can cross that bridge when we get to it 👍

src/views/open-api-docs-view/types.ts
```diff
@@ -1,4 +1,17 @@
...
+/**
+ * A type to describe versioned API docs source files.
+ */
+export interface OpenApiDocsVersionData {
+	// A unique id for this version, used to construct URL paths for example
+	versionId: string
+	// The release stage of this version of the API docs
+	releaseStage?: string // typically 'stable' | 'preview'
+	// The schema file we'll load and render into the page for this version
+	targetFile: GithubFile | string
```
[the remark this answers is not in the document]
zchsh: I dig it! ✅ updated!

PR #2102, opened 2023-07-25
feat: styles operation header

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksoperation-header-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1204678746647847/1205079140949241/f) 🎟️

## 🗒️ What

Styles out the operation header component. Updates the server file to truncate the path.

## 📸 Design Screenshots

Screenshot 2023-07-25 at 12 01 08 PM

## 🧪 Testing

- [Visit the page](https://dev-portal-git-ksoperation-header-hashicorp.vercel.app/hcp/api-docs/vault-secrets)
- Validate that is [matches the designs](https://www.figma.com/file/wjoC3ZVsA1kDuPsFLjEFuY/API-Template?type=design&node-id=526-34648&mode=design&t=T8TGJZC7PM2lxZrj-0)

## 💭 Anything else?

Note there aren't mobile specific designs yet for this component. So this reflects the desktop spec, although shouldn't be 'broken' at mobile. Mostly just missing some global layout padding which should be handled elsewhere.


config/preview.json
```diff
@@ -4,6 +4,6 @@
 		"max_static_paths": 10
 	},
 	"flags": {
-		"enable_hcp_vault_secrets_api_docs_revision": false
+		"enable_hcp_vault_secrets_api_docs_revision": true
```
kendallstrautman, line 7: TODO - don't merge until this is flipped back to `false`

src/views/open-api-docs-view/utils/get-operation-props.ts
```diff
@@ -59,6 +63,11 @@ export function getOperationProps(
 			operationObjects.push({
 				operationId: operation.operationId,
 				slug: slugify(snakeCase(operation.operationId), { lower: true }),
+				type,
+				path: {
+					full: path,
+					truncated: truncateHcpOperationPath(path),
+				},
```
kendallstrautman, line 70: @zchsh Wasn't sure here how you wanted to handle adding in the props. I figured it'd be nice to align the component props with the `OperationProps`. Let me know!

braydenlove:
This looks great @kendallstrautman. Can we make sure the `h4` is `foreground-strong`? Currently it appears to be #000000. Otherwise this is good to go.

src/views/open-api-docs-view/components/open-api-operations/index.tsx
```diff
@@ -37,7 +38,13 @@ export function OpenApiOperations({
 							key={operation.operationId}
 							style={{ border: '1px solid magenta', padding: '2px' }}
 						>
-							<OperationHeader operation={operation} />
+							<OperationHeader
+								className={s.header}
+								slug={operation.slug}
+								headerText={operation.operationId}
+								method={operation.type}
+								path={operation.path.truncated}
```
zchsh, line 46: ⛏️ Okay this is a weird one, and I think is also a problem upstream, but long `path` and also long `headerText` strings can cause overflow and wreak havok on the layout on small viewports, cause they tend to not have any spaces or word-break-able characters:

| overflow issue |
| - |
| overflow-issue |

A "quick fix" could be `overflow-wrap: break-word`, I think 🤔 A more involved fix might be to get `<wbr>` elements interspersed into `headerText` and `path` somehow, which might yield a more polished result:

| `overflow-wrap` | `<wbr>` |
| - | - |
| overflow-wrap-anywhere | injected-word-breaks |

I think we'll actually want `overflow-wrap: break-word` even if we go for the more involved `<wbr>` addition, since we still won't be able to guarantee we've added `<wbr>` in enough spots... so that's something at least! I'd definitely say we could roll with that CSS only fix for now 👍

---

... that being said but I can't help jumping to the conclusion that we might want to pursue the `<wbr>` approach, so I noodled on that over in a branch off this PR:

- https://github.com/hashicorp/dev-portal/pull/2103

src/views/open-api-docs-view/components/operation-header/operation-header.module.css
```diff
@@ -0,0 +1,22 @@
+.wrapper {
+	display: flex;
+	flex-direction: column;
+	gap: 8px;
+}
+
+.heading {
+	composes: g-offset-scroll-margin-top from global;
+	composes: hds-typography-display-400 from global;
+	font-weight: var(--token-typography-font-weight-bold);
```
zchsh, line 10: Quick fix for overflow issues on mobile:

```suggestion
	font-weight: var(--token-typography-font-weight-bold);
	overflow-wrap: break-word;
```

src/views/open-api-docs-view/components/operation-header/operation-header.module.css
```diff
@@ -0,0 +1,22 @@
...
+
+.methodAndPath {
+	display: flex;
+	align-items: center;
+	gap: 8px;
+}
+
+.path {
+	composes: hds-typography-code-300 from global;
+	color: var(--token-color-foreground-primary);
```
zchsh, line 21: Quick fix for overflow issues on mobile:

```suggestion
	color: var(--token-color-foreground-primary);
	overflow-wrap: break-word;
```

zchsh (approved):
✅ ✨ Looking great!

🚧 One almost-blocking thing is the `overflow-wrap` issue. I added some suggestions to add `overflow-wrap: break-word`, so approving pending those additions!

💭 Very optionally, if you wanna add `<wbr >` inject to this PR you could grab or merge in the work from #2103 , but could very much see the case to make that a separate PR entirely rather than roll it in here 👍

src/views/open-api-docs-view/components/open-api-operations/index.tsx
kendallstrautman: I love your exploration with the word-breaks and the idea that we could life that into a higher utility. I'll push up the CSS only fix for now so we can play around more with the work break approach.

kendallstrautman:
@braydenlove updated that color! Thanks for catching that.

PR #2103, opened 2023-07-25
feat: better word-breaks in operation header

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

In `OpenApiDocsView`, implements intentional word-break areas for operation header text and paths.

## 📸 Design Screenshots

| Before | After |
| - | - |
| creategh-before | creategh-after |
| listsecrets-before | listsecrets-after |

## 🧪 Testing

    - Operation names and paths in the "operation header" area should break in more logical places when space is constrained

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/1204678746647847/1205271541668365/f
[preview]: https://dev-portal-git-zsexplore-wbr-hashicorp.vercel.app/
[/hcp/api-docs/vault-secrets]: https://dev-portal-git-zsexplore-wbr-hashicorp.vercel.app/hcp/api-docs/vault-secrets


kendallstrautman:
🔖 we perhaps could use this for the[ tables discussion.](https://hashicorp.slack.com/archives/C01KCU4HDPY/p1690386491445219)

src/views/open-api-docs-view/components/operation-header/index.tsx
```diff
@@ -15,9 +17,10 @@ import s from './operation-header.module.css'
 
 interface OperationHeaderProps {
 	slug: OperationProps['slug']
-	headerText: OperationProps['operationId']
+	headingSlot: ReactNode
+	headingAriaLabel: string
```
zchsh, line 21: Had to split out a separate `ariaLabel` here, for the heading permalink, as we can't render the now-with-word-breaks content into an `ariaLabel` prop 👍

src/views/open-api-docs-view/components/open-api-operations/index.tsx
```diff
@@ -36,9 +40,17 @@ export function OpenApiOperations({
 									<OperationHeader
 										className={s.header}
 										slug={operation.slug}
-										headerText={operation.operationId}
+										headingAriaLabel={operation.operationId}
+										headingSlot={addWordBreaks(
```
zchsh, line 44: `addWordBreaks` and `splitOnCapitalLetters` helpers are where the magic kinda happens, though hopefully the function names make it feel like the very straightforward non-magic that it actually is!

zchsh:
Oooh wait, just realized I can use zero-width spaces instead of `<wbr />`. Simplifies a lot. Going to update to do that now...

src/views/open-api-docs-view/components/open-api-operations/index.tsx
```diff
@@ -29,21 +33,31 @@ export function OpenApiOperations({
...
 					return (
 						<div key={operation.operationId}>
 							<OperationSections
 								headerSlot={
 									<OperationHeader
 										className={s.header}
 										slug={operation.slug}
-										headerText={operation.operationId}
+										headingAriaLabel={operation.operationId}
+										headingSlot={operationIdWithBreaks}
```
thiskevinwang: This is a slight tangent, but do we have access to a more human-friendly value, rather than `operationId`? 

The sibling `summary` field is what should accomplish this for us.
- https://swagger.io/specification#operation-object

And if HCP specs are missing this field, I think we should have some say in tasking them with populating the field.
- https://github.com/hashicorp/hcp-specs/blob/main/specs/cloud-vault-secrets/preview/2023-06-13/hcp.swagger.json#L23C9-L23C9

zchsh: Great call, very on board with this! Made a separate task for this: 🎟️ [Use operation "summary" rather than "operationId" as more human-readable operation title](https://app.asana.com/0/1204678746647847/1205316467934575/f). Will try to get this in earlier rather than later to make it clear we're going more strongly in the direction of "follow the spec" 👍

zchsh: What am i saying i'm already got `summary` available, will just do it in this PR!

thiskevinwang (approved):
LGTM! 

This is likely a non-issue, but sharing it up — I usually expect double clicks to highlight a continuous word.

With this approach, double clicking highlights only a single word as seen in this screenshot
CleanShot 2023-08-23 at 14 07 37@2x

zchsh:
@thiskevinwang Ah great catch! Yeah not sure how we can address that 🤔 I did a bit of testing and it looks like `<wbr />` maintains the same word boundaries as without the wrapping hints... but in either case, a triple-click seems like it can select across word boundaries:

https://github.com/hashicorp/dev-portal/assets/4624598/9e5e5465-9dff-45de-a074-65015eb32e97

Thinking we could either:
1. Leave as-is, zero-width space is easier to handle since it's unicode rather than HTML, so no `dangerouslySetInnerHtml` or client-side insertion of `<wbr />`
2. Switch to `<wbr />` (as explored in earlier commits), though now with the good reason to get better word-selection behaviour

I think there's additional context though - these long-word cases are kinda happening cause we're using `operation.operationId` rather than a more human-readable `operation.summary` (thanks for pointing this out on earlier PRs! 💖).

So personally I lean towards `1` (leave as-is with zero-width space) for the relatively simplicity, and cause hopefully we'll avoid this issue entirely once we can collab with authors to get `operation.summary` values in place. Let me know what you think! 🙇‍♂️

PR #2108, opened 2023-07-31
Command Bar clean up

thiskevinwang:
This PR does 2 main things:
1. refactors `CommandBarDialog` to a more composable API
   ```diff
   - <CommandBarDialog isOpen={state.isOpen} onDismiss={toggleIsOpen}>
   + <Command.Dialog isOpen={state.isOpen} onDismiss={toggleIsOpen}>
   +   <Command.Header />
   +   <Command.Body />
   +   <Command.Footer />
   + </Command.Dialog>
   ```
    - Rationale:
      - Massage API towards [`cmdk`](https://github.com/pacocoursey/cmdk); Note: this makes no guarantees of adopting  [`cmdk`](https://github.com/pacocoursey/cmdk)
      - Reduce obscurity of hiding sub components within `CommandBarDialog`
      - I'd like to make it easy to integrate the incoming chat-UI/UX into the existing command bar component.

1. deletes scattered `types.ts` files and colocates their code with accompanying component files, for easier readability & less file jumping
    - Rationale:
      - the interface declarations are small to the point where colocation with component code reduces the need to jump between files greatly


src/components/command-bar/components/activator/index.tsx
```diff
@@ -2,14 +2,18 @@
...
 import { useCommandBar } from 'components/command-bar'
 import { CmdCtrlIcon, KIcon } from 'components/command-bar/components'
 import Text from 'components/text'
-import { CommandBarActivatorProps } from './types'
 import s from './command-bar-activator.module.css'
 
+interface CommandBarActivatorProps {
+	leadingIcon: ReactElement
+	visualLabel: string
+}
```
BrandonRomano, line 15: I am a big fan of co-locating the types here.  Low hanging fruit to reduce to total number of files you have to navigate to understand this component!

src/components/command-bar/index.tsx
```diff
@@ -171,7 +172,11 @@ const CommandBarProvider = ({ children }: CommandBarProviderProps) => {
 		<SearchHitsProvider>
 			<CommandBarContext.Provider value={contextValue}>
 				{children}
-				<CommandBarDialog isOpen={state.isOpen} onDismiss={toggleIsOpen} />
+				<Command.Dialog isOpen={state.isOpen} onDismiss={toggleIsOpen}>
+					<Command.Header />
+					<Command.Body />
+					<Command.Footer />
+				</Command.Dialog>
```
BrandonRomano, line 179: This feels better!

BrandonRomano (approved):
I'm overwhelmingly in favor of this change!

src/components/command-bar/components/activator/index.tsx
kendallstrautman: I agree, I prefer co-locating types for simple interfaces like this.

src/components/command-bar/index.tsx
```diff
@@ -157,6 +157,7 @@ const CommandBarProvider = ({ children }: CommandBarProviderProps) => {
 			setCurrentCommand,
 			setCurrentInputValue,
 			toggleIsOpen,
+			instructionsElementId: 'footer-keyboard-instructions',
```
kendallstrautman, line 160: 💡 we could use [`useId` ](https://react.dev/reference/react/useId) or update this to not mention the footer as the descriptive element may change.

src/components/command-bar/components/index.ts
```diff
@@ -3,10 +3,10 @@
...
-export * from './k-icon'
-export * from './list-item'
-export * from './list'
+export { CommandBarActivator } from './activator'
+export { CmdCtrlIcon } from './cmd-ctrl-icon'
+export { default as Command } from './dialog'
+export { CommandBarDivider } from './divider'
+export { KIcon } from './k-icon'
+export { CommandBarButtonListItem, CommandBarLinkListItem } from './list-item'
+export { CommandBarList } from './list'
```
kendallstrautman, line 12: 👍🏻

kendallstrautman (approved):
LGTM as well 🚀

src/components/command-bar/index.tsx
kendallstrautman: Should we called this `labelledById` or something? The name confuses me a bit. I wouldn't know straight away what this is used for unless I looked at the interface directly with the code comment.

thiskevinwang: 1. I'll leave this `'footer-keyboard-instructions'` hard-coded value in this PR. We can easily change this post-PR if we need to!
2. I thought about `labelledById` a bit... but I think `instructionsElementId` (the previous variable name) is actually reasonable! In my head, the naming pans out to: _"the footer div is the `instructionsElementId` and the dialog is labelled by `instructionsElementId`"_

PR #2124, opened 2023-08-14
feat: tweak operation column layout for alignment

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

This PR tweaks the layout of operation items on OpenAPI docs views,for better alignment.

## 📸 Design Screenshots

| Before | After | 
| - | - |
| before | after |

## 🧪 Testing

    - The operation URL path code block should be aligned with the operation header
    - The `sticky` behaviour of the code block should continue to function as expected
    - Layout should look as expected across viewports
    - The DOM order of sections should still be `header → example → details`

## 💭 Anything else?

At some point we discussed wanting to hide the "examples" section on smaller viewports. If we still want to do that, that's something we could likely incorporate into this PR pretty easily! 👍

[task]: https://app.asana.com/0/1204678746647847/1205271541668354/f
[preview]: https://dev-portal-git-zsopenapi-adjust-column-align-hashicorp.vercel.app/
[/hcp/api-docs/vault-secrets]: https://dev-portal-git-zsopenapi-adjust-column-align-hashicorp.vercel.app/hcp/api-docs/vault-secrets

### DIFF
--- a/config/preview.json
+++ b/config/preview.json
@@ -4,6 +4,6 @@
 		"max_static_paths": 10
 	},
 	"flags": {
-		"enable_hcp_vault_secrets_api_docs_revision": false
+		"enable_hcp_vault_secrets_api_docs_revision": true
 	}
 }
--- a/src/views/open-api-docs-view/components/operation-sections/operation-sections.module.css
+++ b/src/views/open-api-docs-view/components/operation-sections/operation-sections.module.css
@@ -4,24 +4,41 @@
  */
 
 .root {
-	display: flex;
-	flex-direction: row-reverse;
-	flex-wrap: wrap;
+	/* stylelint-disable-next-line property-no-unknown */
+	container-type: inline-size;
+}
+
+.gridLayout {
+	display: grid;
 	gap: 24px;
+	grid-template-areas: 'header' 'examples' 'details';
+	grid-template-columns: minmax(0, 1fr);
+
+	/* Note: 744px = 360px per column, plus 24px grid gap.
+	   Calc & CSS vars are not supported in container queries. */
+
+	/* stylelint-disable-next-line at-rule-no-unknown */
+	@container (min-width: 744px) {
+		grid-template-areas: 'header examples' 'details examples';
+		grid-template-columns: minmax(0, 1fr) minmax(0, 1fr);
+	}
+}
+
+.headerSlot {
+	grid-area: header;
 }
 
 .detailsSlot {
-	min-width: min(100%, 360px);
 	display: flex;
 	flex-direction: column;
-	flex: 1 1 0;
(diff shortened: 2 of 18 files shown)


config/preview.json
```diff
@@ -4,6 +4,6 @@
 		"max_static_paths": 10
 	},
 	"flags": {
-		"enable_hcp_vault_secrets_api_docs_revision": false
+		"enable_hcp_vault_secrets_api_docs_revision": true
```
zchsh, line 7: 🚧 TODO: revert flags before merging

src/views/open-api-docs-view/components/operation-sections/operation-sections.module.css
```diff
@@ -4,24 +4,41 @@
...
 .root {
-	display: flex;
-	flex-direction: row-reverse;
-	flex-wrap: wrap;
+	/* stylelint-disable-next-line property-no-unknown */
+	container-type: inline-size;
+}
+
+.gridLayout {
+	display: grid;
```
zchsh, line 12: TLDR; is i've switched from a `flex` layout to a `grid` layout with a container query! ✨ I had to add some `stylelint` exceptions cause container CSS didn't seem to be recognized yet by linting.

kendallstrautman: Whoa container query?! ✨

kendallstrautman (approved):
💖

src/views/open-api-docs-view/components/operation-sections/operation-sections.module.css
zchsh: It's a new direction for sure! I am so used to thinking container queries will never be a thing that I have not kept up at all with browser support, but as [Dylan mentioned in Slack](https://hashicorp.slack.com/archives/C031LA4KS4S/p1689708868902629):

> It's been five months since the last major browser engine added support, so I don't think we need to worry about polyfills or anything. I believe our official support matrix is the last two versions of evergreen browsers.

zchsh:
@braydenlove Ah great catch! I've got this updated:

- `SidebarLayout` now has zero opinion on spacing between the sidebar and main area - this makes it a little easier to get consistent left-right padding in place for main-area content, while also allowing main area content to vary in its max-width.
- `OpenApiView` now has `24px` left-right padding when the sidebar is not shown, and `48px` left-right padding when the sidebar is shown

Let me know if that looks alright! 🙇‍♂️

zchsh:
Gonna merge this in an address any design VQA items during a final VQA 👍 (have chatted with Brayden in Slack to confirm this approach!)

PR #2126, opened 2023-08-15
feat: render open-api info.description as mdx

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

This PR updates the "Overview" section of the OpenAPI view to render the `description` from the spec file as MDX.

## 🤷 Why

The OpenAPI spec [explicitly mentions CommonMark support](https://swagger.io/specification/#info-object) for the `description` field on `schema.info`. By rendering the `description` as MDX, we'll enable authors to write descriptive content for a given API spec file within the spec file itself. This has multiple perceived benefits:
- All OpenAPI view content is managed in a single place, the spec file itself
- Markdown is used for authoring, rather than the less author-familiar JSX format
- When a spec file becomes a versioned document, the descriptive content is included in that versioning

## 📸 Design Screenshots

> **Note**: this PR makes updates that mean the overview description now matches designs _less_ closely than it did previously. This is an intentional compromise, and a kind of side-effect of using our **shared long-form MDX components**. We could in theory build a net-new bespoke set of MDX components for use on this page specifically, but it seems like this might not be something we want to dive into immediately.

| Before | After |
| - | - |
| before | after |

## 🧪 Testing

    - The overview content should appear as expected. Note that it does **not** match designs, but rather uses our shared `DevDotContent` component, so matches the design spec for long-form content on other pages.

## 💭 Anything else?

Other than the note on balancing "match the design" with "use existing MDX system", no other thoughts at the moment!

[task]: https://app.asana.com/0/1204678746647847/1205271542994753/f
[figma]: https://www.figma.com/file/wjoC3ZVsA1kDuPsFLjEFuY/API-Template?type=design&node-id=526%3A33909&mode=design&t=W3IsmHRYtcFoOBSy-1
[preview]: https://dev-portal-git-zsopenapi-description-as-mdx-hashicorp.vercel.app/
[/hcp/api-docs/vault-secrets]: https://dev-portal-git-zsopenapi-description-as-mdx-hashicorp.vercel.app/hcp/api-docs/vault-secrets


src/views/open-api-docs-view/components/open-api-overview/components/description-mdx/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+	p: MdxP,
+	pre: MdxPre,
+	table: MdxTable,
+	ul: MdxUnorderedList,
+}
+
+/**
+ * TODO: write description
+ */
+export function DescriptionMdx({
```
zchsh, line 50: Uses all the same MDX component primitives as `DevDotContent`, but without support for additional custom components, as I think we want our spec file `description` values to stay fully portable and CommonMark-compliant!

src/views/open-api-docs-view/index.tsx
```diff
@@ -49,9 +51,16 @@ function OpenApiDocsView({
...
-					description={description}
+					contentSlot={
+						descriptionMdx ? (
+							<DescriptionMdx mdxRemoteProps={descriptionMdx} />
+						) : null
+					}
 				/>
-				<OpenApiOperations operationGroups={operationGroups} />
+				<div className={s.operationsSection}>
+					<h2 className={s.operationsHeading}>Operations</h2>
```
zchsh, line 61: I added this `Operations` heading to ensure that the `h3` for each operation doesn't end up semantically nested under any of the provided `description` markdown content.

Note: in designs this heading is `Services`, I think `Operations` is more technically correct since unlike previous API docs (such as Boundary's), HCP API docs can't be organized into services since `HCP Vault Secrets` is itself the service.

braydenlove:
@zchsh I think that is super pragmatic and I am aligned with what you posted above. Could we remove the `border-bottom: 1px solid` on the Operations `<h2>`? It feels out of place now that we have removed the other `<hr>`s.

zchsh:
@braydenlove re: description approach, awesome! 👯 re: `border-bottom`, definitely - removed! ✅

kendallstrautman (approved):
👍🏻 Looks great. Are we worried about enforcing consistency with the headings etc? For example do we want them all to use 'overview'? Probably something we can informally police.

zchsh:
@kendallstrautman Ah yeah great question re: heading enforcement!

For now, I figure we at least won't expect specific headings - eg we have another task to "link to top of page" (🎟 [Update top-level sidebar link to act as "scroll to top" link](https://app.asana.com/0/1204678746647847/1205211659139665/f)) and for that I think we can add an `id` on the `h1` and link to that, rather than link to a heading within the MDX content.

Not sure if that helps - regardless, def agree that at some point we should establish a content strategy for the OpenAPI `schema.info.description`, give we'll be rendering it here pretty visibly 👍

braydenlove (approved):

PR #2127, opened 2023-08-15
feat: deeper anchor-links for open-api view

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Adds visible anchor link elements alongside many elements on OpenAPI view pages.

## 🤷 Why

To make it easier for visitors to link directly to specific sections of OpenAPI reference documentation.

## 🛠️ How

- Adds a `ContentWithPermalink` component to capture our common pattern of permalink use
- Updates server-side functions to get a unique `slug` value for all target elements
- Updates target elements to use `ContentWithPermalink` in combination with the newly available `slug`

### Before

> No anchor link elements, even on linkable operation headings

https://github.com/hashicorp/dev-portal/assets/4624598/0432e308-4bf6-44fb-9060-a5f6ae08bbd9

### After

> Operations, property sections and groups, and properties themselves are all linkable and display anchor link elements

https://github.com/hashicorp/dev-portal/assets/4624598/bd7123bc-e021-40a7-8c41-547454705f87

## 🧪 Testing

    - The following components should display working anchor links on `:hover`...
    - Operation headings, such as `ListApps`
    - Property sections, namely `Request` and `Response`
    - Property groups, such as `Path` and `Query` (under `Request`) or `200` and `Default` (under `Reponse`)
    - Top-level properties
    - Nested properties at arbitrary depths

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/1204678746647847/1205211654360778/f
[preview]: https://dev-portal-git-zsopenapi-deeper-anchor-links-hashicorp.vercel.app/
[/hcp/api-docs/vault-secrets]: https://dev-portal-git-zsopenapi-deeper-anchor-links-hashicorp.vercel.app/hcp/api-docs/vault-secrets


src/views/open-api-docs-view/utils/get-operation-props.ts
```diff
@@ -36,22 +36,38 @@ export async function getOperationProps(
...
-			)
-			const responseData = await getResponseData(operation.responses)
+			const requestDataSlug = `${operationSlug}_request`
+			const requestData = {
+				heading: { text: 'Request', slug: requestDataSlug },
+				noGroupsMessage: 'No request data.',
+				groups: await getRequestData(
+					parameters,
+					operation.requestBody,
+					requestDataSlug
```
zchsh, line 55: A big part of the work in this PR was passing through `parentSlug` arguments to allow everything in the hierarchy to have a unique slug. Each slug is then structured like `<parentSlug><slug>`.

Above the level of nested properties, we separate `<parentSlug>` and `<slug>` with an `_` character, as `operationId` values are `slugified` and will never contain `_`.

At the property level, we separate `<parentSlug>` and `<slug>` with a `.` character, as property keys are never expected to contain `.` characters that cause collisions.

To be more concrete, we have guaranteed uniqueness at the `operationId` level, as well as at each level where we need slugs, so we progressively build off of that:
    - Operation headings, such as `ListApps`, use the `operationId` (as they did previously)
    - Property sections use the above as the `parentSlug`
        - Property sections are guaranteed unique, we have only two - `request` and `response`
    - Property groups use the above as the `parentSlug`
        - Property groups are guaranteed unique
        - Under `request`, we have `path`, `query`, and `body`
        - Under `response`, we have `default`, plus response codes like `200`, `404`, `500`, etc
    - Top-level properties use the above as the `parentSlug`
        - Properties are unique within an object
    - Nested properties at arbitrary depths use their parent property
        - Again, properties are unique within an object

All of the work done in `src/views/open-api-docs-view/utils` files is to achieve the above implementation of slug uniqueness 👍

zchsh:
@braydenlove This PR adds anchor links that are visible-on-hover to operation titles, property sections and groups, and all properties themselves.

Let me know what you think! 🙇‍♂️

src/views/open-api-docs-view/components/content-with-permalink/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+ * `ariaLabel` is required to ensure the permalink element has meaningful
+ * text for screen readers. For simple use cases where `children` is a
+ * string, the `ariaLabel` can be passed the same value as `children`.
+ * For more complex use cases, the consumer should determine what text
+ * would be appropriate to pass to `ariaLabel`.
+ *
+ * Note that this component does _not_ handle placing the provided `id`
+ * in the DOM. It requires the consumer to place the `id` on an appropriate
+ * element, typically an element rendered in the provided `children`.
+ */
```
kendallstrautman, line 20: 👍🏻

src/views/open-api-docs-view/utils/get-operation-props.ts
kendallstrautman: This is really clever, I was wondering how you were going to guarantee uniqueness. Transferring some of this context to a code comment would be helpful!

kendallstrautman (approved):
Works beautifully. Really cool solution to the uniqueness problem.

PR #2128, opened 2023-08-15
Developer AI

thiskevinwang:
# Description

This introduces "Developer AI". 
This work comes in largely 3 chunks:
1. The new chatbox UI & the streaming processing logic/state handling
2. The new chat server endpoints
3. Moving around existing code to make the UI _fit_

CleanShot 2023-09-08 at 10 13 05@2x

### To test

> [!IMPORTANT]
> To gain access to this feature, you need to complete the following steps:

- You need to be "opted in". To opt-in:
   - log in to https://dev-portal-git-kwai-hashicorp.vercel.app/profile/account
   - grab your "HashiCorp ID" and send it to @thiskevinwang 
- Once opted in, you should see a Toast pop up, notifying you how/where to discover the feature.

### Checklist
- [x] Ensure production has `EXPERIMENTAL_CHAT_API_BASE_URL` env var

---
Builds on work from
- https://github.com/hashicorp/dev-portal/pull/2148
- https://github.com/hashicorp/dev-portal/pull/2151
- #2155
- https://github.com/hashicorp/dev-portal/pull/2157
- https://github.com/hashicorp/dev-portal/pull/2161
- https://github.com/hashicorp/dev-portal/pull/2163
---
  - https://app.asana.com/0/0/1204865849100145


src/components/chatbox/ai-feature-toast.tsx
```diff
@@ -0,0 +1,28 @@
...
+				icon: <IconWand24 />,
+				title: 'Welcome to the Developer AI closed beta',
+				description: 'Try it out in our cmd+K menu!',
+				autoDismiss: 10000,
+			})
+			setShown(true)
+		}
+	}, [session?.meta.isAIEnabled, shown])
+	return null
+}
```
thiskevinwang, line 28: ### What?

Renders the following:
CleanShot 2023-09-08 at 11 27 52@2x

src/components/chatbox/chatbox.tsx
```diff
@@ -0,0 +1,550 @@
+import React, { useEffect, useState, useRef } from 'react'
```
thiskevinwang, line 1: ### What?

Renders the following:

src/components/chatbox/message.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+import Button from 'components/button'
+import Text from 'components/text'
+import FeedbackForm from 'components/feedback-form'
+import IconTile from 'components/icon-tile'
+import useAuthentication from 'hooks/use-authentication'
+
+import s from './message.module.css'
+
+export const UserMessage = ({
```
thiskevinwang, line 24: Renders:
CleanShot 2023-09-08 at 11 30 19@2x

src/components/chatbox/message.tsx
```diff
@@ -0,0 +1,251 @@
...
+				className={cn(s.message_content, s.message_user_input)}
+			>
+				{text}
+			</Text>
+			<div className={cn(s.message_gutter)}></div>
+		</div>
+	)
+}
+
+export const AssistantMessage = ({
```
thiskevinwang, line 48: Renders:

CleanShot 2023-09-08 at 11 30 39@2x

src/components/chatbox/message.tsx
```diff
@@ -0,0 +1,251 @@
...
+				</div>
+			</div>
+
+			<div className={cn(s.message_gutter)}></div>
+		</div>
+	)
+}
+
+// TODO(kevinwang): error styling.
+export const ApplicationMessage = ({ text }: { text: string }) => {
```
thiskevinwang, line 235: Renders: 
CleanShot 2023-09-08 at 11 31 28@2x

src/components/command-bar/commands/search/unified-search/components/dialog-body/chat.module.css
```diff
@@ -0,0 +1,23 @@
+.ask {
```
thiskevinwang, line 1: For styling 
CleanShot 2023-09-08 at 11 32 23@2x

src/components/feedback-form/types.ts
```diff
@@ -29,6 +29,7 @@ interface FeedbackQuestionText extends FeedbackQuestionBase {
 	icon?: ReactElement<React.JSX.IntrinsicElements['svg']>
 	buttonText: string
 	nextQuestion?: string
+	placeholder?: string
```
thiskevinwang, line 32: This change is lifted up into https://github.com/hashicorp/dev-portal/pull/2163

src/components/feedback-form/components/question/index.tsx
```diff
@@ -82,7 +82,7 @@ const Question: React.FC<QuestionProps> = ({
 								s.textArea,
 								inputHasEntry ? s.visited : null
 							)}
-							placeholder="Your feedback..."
+							placeholder={placeholder || 'Your feedback...'}
```
thiskevinwang, line 85: This change is lifted up into https://github.com/hashicorp/dev-portal/pull/2163

src/components/chatbox/chatbox.tsx
```diff
@@ -0,0 +1,550 @@
...
+			}
+		},
+	})
+
+	// A stream reader
+	const [reader, setReader] =
+		useState<ReadableStreamDefaultReader<Uint8Array> | null>(null)
+	// Function to stop the stream
+	const stopStream = () => {
+		reader?.cancel()
```
dstaley, line 138: should this be a callback since it relies on the value of a `useState` hook?

src/components/chatbox/chatbox.tsx
```diff
@@ -0,0 +1,550 @@
...
+			;(async () => {
+				// Reset streamed text
+				setStreamedText('')
+				setIsReading(true)
+
+				// This is our contract with the backend
+				const messageFormat = z.object({
+					content: z.string(),
+				})
+				const sleep = (ms: number) => new Promise((r) => setTimeout(r, ms))
```
dstaley, line 163: can you move this out of the useEffect since it doesn't need to capture any values? also can you rename `r` to `resolve`? Took me a second to realize that's what it was.

src/components/chatbox/chatbox.tsx
```diff
@@ -0,0 +1,550 @@
...
+
+		// append user message to list
+		appendMessage({ type: 'user', text: task, image: user.image })
+	}
+
+	// Throttle this value to enable uninterrupted smooth scrolling
+	const throttledText = useThrottle(streamedText, 400)
+
+	// scroll text response to bottom as text streams in from the backend
+	const textContentRef = useRef<HTMLDivElement>(null)
```
dstaley, line 311: `overflow-anchor` might be a more elegant solution here as it respects the user's ability to scroll, where as this will cause the view to scroll to the bottom even if the user scrolled up. https://css-tricks.com/books/greatest-css-tricks/pin-scrolling-to-bottom/

src/components/chatbox/chatbox.tsx
```diff
@@ -0,0 +1,550 @@
...
+									/>
+								)
+							}
+							// Application message; likely an error
+							case 'application': {
+								return (
+									<ApplicationMessage key={`${e.type}-${i}`} text={e.text} />
+								)
+							}
+						}
```
dstaley, line 379: Can you abstract this to a single component?

src/components/chatbox/chatbox.tsx
```diff
@@ -0,0 +1,550 @@
...
+					</div>
+
+					<Text
+						asElement="span"
+						className={s.disclaimer}
+						/* Display/100/Regular */
+						size={100}
+						weight="regular"
+					>
+						AI Disclaimer: HashiCorp AI may produce inaccurate information and
```
dstaley, line 451: I personally love this, but do we need to adjust the language before merging this?

src/components/chatbox/message.tsx
```diff
@@ -0,0 +1,251 @@
...
+				Authorization: `Bearer ${accessToken}`,
+			},
+			body: JSON.stringify({
+				messageId,
+				conversationId,
+				rating,
+				text,
+			}),
+		})
+		return response
```
dstaley, line 84: would it be possible to re-update the UI after getting the response? For example, resetting the feedback button state if the request fails?

src/components/chatbox/message.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+		if (el) {
+			// https://www.nikouusitalo.com/blog/why-isnt-clipboard-write-copying-my-richtext-html/
+			const clipboardItem = new ClipboardItem({
+				// @ts-expect-error - innerText is available
+				'text/plain': new Blob([el.innerText], { type: 'text/plain' }),
+				'text/html': new Blob([el.outerHTML], { type: 'text/html' }),
+			})
+
+			navigator.clipboard.write([clipboardItem])
```
dstaley, line 114: This can (and often does) fail. It returns a Promise, so we should only call `setIsCopied` if the Promise resolves. At this point I wouldn't worry about communicating a failure.

src/components/chatbox/message.tsx
```diff
@@ -0,0 +1,251 @@
...
+								) : (
+									<IconClipboard24 height={12} width={12} />
+								)
+							}
+							aria-label="Copy to clipboard"
+							text={isCopied ? 'Copied' : 'Copy'}
+							onClick={handleCopy}
+						/>
+
+						<Button
```
dstaley, line 166: Can you move these out to their own components so we can make sure they're always visually aligned?

src/pages/api/chat/route.ts
```diff
@@ -0,0 +1,153 @@
...
+					return new Response(res.body, {
+						headers: {
+							'Content-Type': 'text/event-stream; charset=utf-8',
+							'x-message-id': messageId,
+							'x-conversation-id': conversationId,
+						},
+					})
+				} else {
+					// additional Headers
+					const addl = {}
```
dstaley, line 80:
```suggestion
					const additionalHeaders = {}
```

dstaley (changes requested):
For something so complex, this is a surprisingly readable implementation! Great work! I just noticed a few spots where a bit more abstraction would be useful (primarily to just clean up the JSX and make it a bit easier to understand).

src/components/chatbox/ai-feature-toast.tsx
```diff
@@ -0,0 +1,61 @@
...
+		if (Cookies.get(COOKIE_HAS_SEEN_AI_TOAST)) {
+			return
+		}
+
+		if (session?.meta.isAIEnabled) {
+			toast({
+				color: ToastColor.highlight,
+				icon: <IconWand24 />,
+				title: 'Welcome to the Developer AI closed beta',
+				description: 'Try it out in our cmd+K menu!',
```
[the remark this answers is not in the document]
thiskevinwang: @dstaley (cross-platform-ignorance here) - what would you expect instead of ( <kbd>Cmd</kbd> + <kbd>K</kbd> ) on Windows, and Linux?

---
> [!IMPORTANT]
> TODO(@thiskevinwang) - work with @kellyschummer on this copy

src/components/chatbox/chatbox.tsx
```diff
@@ -0,0 +1,550 @@
...
+// https://helios.hashicorp.design/icons/library
+import { IconArrowDownCircle16 } from '@hashicorp/flight-icons/svg-react/arrow-down-circle-16'
+import { IconBulb16 } from '@hashicorp/flight-icons/svg-react/bulb-16'
+import { IconDiscussionCircle16 } from '@hashicorp/flight-icons/svg-react/discussion-circle-16'
+import { IconLoading24 } from '@hashicorp/flight-icons/svg-react/loading-24'
+import { IconSend24 } from '@hashicorp/flight-icons/svg-react/send-24'
+import { IconStopCircle24 } from '@hashicorp/flight-icons/svg-react/stop-circle-24'
+import { IconUser16 } from '@hashicorp/flight-icons/svg-react/user-16'
+import { IconWand24 } from '@hashicorp/flight-icons/svg-react/wand-24'
+// ms is a transient dep
```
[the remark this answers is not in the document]
thiskevinwang: 🤷 💭  `npm ls ms`

<details>
  <summary>Output</summary>

```console
dev-portal@ /Users/kevin/repos/dev-portal
├─┬ @hashicorp/platform-analytics@0.12.0
│ └─┬ @segment/analytics-next@1.51.7
│   └─┬ @segment/tsub@1.0.1
│     └─┬ @stdlib/math-base-special-ldexp@0.0.5
│       └─┬ @stdlib/constants-float64-exponent-bias@0.0.8
│         └─┬ @stdlib/utils-library-manifest@0.0.8
│           └─┬ debug@2.6.9
│             └── ms@2.0.0
├─┬ @hashicorp/platform-cli@2.7.2
│ ├─┬ @typescript-eslint/eslint-plugin@5.54.0
│ │ ├─┬ @typescript-eslint/type-utils@5.54.0
│ │ │ └─┬ debug@4.3.4
│ │ │   └── ms@2.1.2
│ │ └─┬ debug@4.3.4
│ │   └── ms@2.1.2
│ ├─┬ @typescript-eslint/parser@5.54.0
│ │ ├─┬ @typescript-eslint/typescript-estree@5.54.0
│ │ │ └─┬ debug@4.3.4
│ │ │   └── ms@2.1.2
│ │ └─┬ debug@4.3.4
│ │   └── ms@2.1.2
│ └─┬ eslint-config-next@13.2.1
│   ├─┬ eslint-import-resolver-node@0.3.7
│   │ └─┬ debug@3.2.7
│   │   └── ms@2.1.3 deduped
│   ├─┬ eslint-import-resolver-typescript@3.5.3
│   │ └─┬ debug@4.3.4
│   │   └── ms@2.1.2
│   └─┬ eslint-plugin-import@2.27.5
│     ├─┬ debug@3.2.7
│     │ └── ms@2.1.3 deduped
│     └─┬ eslint-module-utils@2.7.4
│       └─┬ debug@3.2.7
│         └── ms@2.1.3 deduped
├─┬ @hashicorp/platform-nextjs-plugin@4.2.0
│ └─┬ @hashicorp/next-optimized-images@0.2.0
│   └─┬ file-loader@3.0.1
│     └─┬ webpack@4.46.0
│       └─┬ micromatch@3.1.10
│         ├─┬ extglob@2.0.4
│         │ └─┬ expand-brackets@2.1.4
│         │   └─┬ debug@2.6.9
│         │     └── ms@2.0.0
│         └─┬ snapdragon@0.8.2
│           └─┬ debug@2.6.9
│             └── ms@2.0.0
├─┬ @hashicorp/sentinel-embedded@0.0.14
│ └─┬ node-flex-serve@0.0.2
│   ├─┬ @shellscape/koa-static@4.0.5
│   │ ├─┬ @shellscape/koa-send@4.1.3
│   │ │ └─┬ debug@2.6.9
│   │ │   └── ms@2.0.0
│   │ └─┬ debug@2.6.9
│   │   └── ms@2.0.0
│   ├─┬ debug@3.2.7
│   │ └── ms@2.1.3 deduped
│   └─┬ koa-router@7.4.0
│     └─┬ debug@3.2.7
│       └── ms@2.1.3 deduped
├─┬ @vercel/fetch@6.2.0
│ ├─┬ @vercel/fetch-retry@5.1.3
│ │ └─┬ debug@4.3.3
│ │   └── ms@2.1.2
│ ├─┬ agentkeepalive@4.2.1
│ │ └─┬ humanize-ms@1.2.1
│ │   └── ms@2.1.3 deduped
│ └─┬ debug@4.3.3
│   └── ms@2.1.2
├─┬ jsonwebtoken@8.5.1
│ └── ms@2.1.3
├─┬ next-remote-watch@2.0.0
│ └─┬ express@4.17.1
│   ├─┬ body-parser@1.19.0
│   │ └─┬ debug@2.6.9
│   │   └── ms@2.0.0
│   ├─┬ debug@2.6.9
│   │ └── ms@2.0.0
│   ├─┬ finalhandler@1.1.2
│   │ └─┬ debug@2.6.9
│   │   └── ms@2.0.0
│   └─┬ send@0.17.1
│     ├─┬ debug@2.6.9
│     │ └── ms@2.0.0
│     └── ms@2.1.1
├─┬ nock@13.2.4
│ └─┬ debug@4.3.2
│   └── ms@2.1.2
└─┬ probe-image-size@7.2.3
  ├─┬ needle@2.9.1
  │ └─┬ debug@3.2.7
  │   └── ms@2.1.3 deduped
  └─┬ stream-parser@0.3.1
    └─┬ debug@2.6.9
      └── ms@2.0.0
```

</details>

src/components/chatbox/ai-feature-toast.tsx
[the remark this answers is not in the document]
dstaley: Either `Cmd/Ctrl-K` or `Command/Control-K`.

src/components/chatbox/chatbox.tsx
```diff
@@ -0,0 +1,550 @@
...
+	// The backend id of a conversation
+	const [conversationId, setConversationId] = useState('')
+	// The backend id of the most recently returned message
+	const [messageId, setMessageId] = useState('')
+
+	// Is the stream being read?
+	const [isReading, setIsReading] = useState(false)
+	// The streamed-in text
+	const [streamedText, setStreamedText] = useState('')
+	// Error text
```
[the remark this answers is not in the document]
thiskevinwang: "A little goes a long way" + "naming things is hard" = short supportive comments

src/components/chatbox/chatbox.tsx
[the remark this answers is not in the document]
dstaley: If we're going to use it, go ahead and declare it as a dependency. Wouldn't want it to disappear if we removed it from another package (also using dependencies like this is incompatible with tools like bun and pnpm since they don't hoist dependencies).

src/components/chatbox/chatbox.tsx
thiskevinwang: :trollface: Unfortunately, we _do_ have to adjust this 

---

> [!IMPORTANT]
> @thiskevinwang work with @kellyschummer on this copy

src/components/chatbox/chatbox.tsx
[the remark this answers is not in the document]
thiskevinwang: done in https://github.com/hashicorp/dev-portal/pull/2128/commits/26f772db305c514e3aefe0321109f48b7a41ef17

src/components/chatbox/chatbox.tsx
thiskevinwang: Do you mean, "should `stopStream` be memoized with `useCallback`" ?

(Assuming I understood your Q correctly...) I think it's technically fine as-is. Adding `useCallback` doesn't change functionality here.

src/components/chatbox/chatbox.tsx
thiskevinwang: done in 2827e4706

src/components/chatbox/chatbox.tsx
thiskevinwang: Ah, I like the idea of a CSS-only-solution way better, but there's one caveat that I think is a dealbreaker / requires additional JS, making this not that much more elegant.

> _In order for scroll anchoring to work, the page must be scrolled at least once to begin with._

src/components/chatbox/chatbox.tsx
thiskevinwang: Hmm, maybe? Though in this case it _feels more simple_ to keep the 3 components separate. The variance in props-interface doesn't feel right to mash into one component.

WDYT?

src/components/chatbox/message.tsx
thiskevinwang: Yes, very possible and very easy — I opted to do nothing (ie. just do an optimistic update and call it a day) because the 👍/👎-request are so non-critical that it's not a big deal which ever way we go.

I'll leave this as-is, but leave the door open to super easy change in the future.

src/components/chatbox/chatbox.module.css
```diff
@@ -0,0 +1,191 @@
...
+		flex-direction: column;
+		gap: 20px;
+		justify-content: center;
+
+		&.left {
+			& h3 {
+				color: var(--token-color-foreground-strong);
+			}
+			& p {
+				color: var(--token-color-foreground-faint);
```
kendallstrautman, line 70: ⛏️ In general, I prefer to use class names instead of elements selectors so that styles don't break if the markup changes.

src/components/chatbox/chatbox.tsx
```diff
@@ -0,0 +1,551 @@
...
+
+import { useMutation } from '@tanstack/react-query'
+
+import useAuthentication from 'hooks/use-authentication'
+import IconTile from 'components/icon-tile'
+import Button from 'components/button'
+import Heading from 'components/heading'
+import Text from 'components/text'
+
+import cn from 'classnames'
```
kendallstrautman, line 23: ⛏️ another small nitpick, could you organize the imports ?

src/components/chatbox/chatbox.tsx
```diff
@@ -0,0 +1,551 @@
...
+import {
+	streamToAsyncIterable,
+	mergeRefs,
+	useScrollBarVisible,
+	useThrottle,
+} from './utils'
+
+const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms))
+
+const useAI = () => {
```
kendallstrautman, line 36: Might be helpful to split some of these hooks / helpers into their own file.

src/components/chatbox/chatbox.tsx
kendallstrautman: Maybe the suggestion was more like can we abstract this into a single component that still has the switch statement, just to make it easier to read the conditional?

src/components/chatbox/chatbox.tsx
kendallstrautman: Also can you use `classNames` just for consistency sake with the rest of the repo?

src/components/chatbox/chatbox.tsx
```diff
@@ -0,0 +1,551 @@
...
+
+					<div className={s.question}>
+						<div className={s.question_icon}>
+							{isLoading ? (
+								<IconLoading24 className={'loading'} />
+							) : (
+								<IconWand24
+									style={{
+										color: 'var(--token-color-foreground-highlight-on-surface)',
+									}}
```
kendallstrautman, line 403: Can we use CSS here instead of the inline style?

src/components/chatbox/chatbox.tsx
```diff
@@ -0,0 +1,551 @@
...
+								return (
+									<ApplicationMessage key={`${e.type}-${i}`} text={e.text} />
+								)
+							}
+						}
+					})}
+				</div>
+			)}
+
+			<form onSubmit={handleSubmit} ref={formRef}>
```
kendallstrautman, line 385: It'd be nice if this was separated into its own component as well.

src/components/chatbox/chatbox.tsx
```diff
@@ -0,0 +1,551 @@
...
+						AI Disclaimer: HashiCorp AI may produce inaccurate information and
+						cause your computer to implode. Use at your own risk.
+					</Text>
+				</div>
+			</form>
+		</div>
+	)
+}
+
+const WelcomeMessage = () => {
```
kendallstrautman, line 461: It would clean up the stylesheet / limit the length if this file if this were in its own file?

src/components/chatbox/message.module.css
```diff
@@ -0,0 +1,94 @@
+.message {
+	display: flex;
+	flex-direction: row;
+	gap: 12px;
+	padding: 20px 16px;
+	background: var(--token-color-palette-neutral-0);
+}
+
+.message_user {
```
kendallstrautman, line 9: We've typically use camel case for classnames, I'd prefer to maintain this consistency unless there's a good reason to deviate.

src/components/chatbox/message.module.css
```diff
@@ -0,0 +1,94 @@
...
+
+.rating_like {
+	border: 1px solid var(--token-color-border-success) !important;
+	background: var(--token-color-surface-success) !important;
+	color: var(--token-color-foreground-success) !important;
+}
+.rating_dislike {
+	border: 1px solid var(--token-color-border-critical) !important;
+	background: var(--token-color-surface-critical) !important;
+	color: var(--token-color-foreground-critical) !important;
```
kendallstrautman, line 83: Could you expand on these?

src/components/chatbox/message.tsx
```diff
@@ -0,0 +1,251 @@
+import { useState, useEffect } from 'react'
+import cn from 'classnames'
```
kendallstrautman, line 2: ⛏️ nitpick, can you use `classNames` for consistency?

src/components/chatbox/ai-feature-toast.tsx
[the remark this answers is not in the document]
thiskevinwang: Tentative result:
- Request from design in https://app.asana.com/0/0/1205523765153660/f
- Requires a [trivial] type change on `ToastDisplayProps.description` from `string` to `ReactNode`
- Reuses the same SVGs in our top search bar.

CleanShot 2023-09-21 at 09 43 38@2x

CleanShot 2023-09-21 at 09 43 26@2x

src/components/chatbox/chatbox.tsx
thiskevinwang: Here's the abstraction.

```tsx
<MessageList messages={messageList} />
```

And a visual to show the change in readability

CleanShot 2023-09-21 at 09 56 02@2x

thiskevinwang: Abstracted and improved in a043230c0

src/components/chatbox/message.tsx
thiskevinwang: Oh TIL! Addressed 7f41e22ed
- handleCopy is now async
- the caught error is ignored

src/components/chatbox/message.tsx
thiskevinwang: @dstaley can you clarify the request here?

- What should be moved?
- What should be visually aligned? (The code itself? UI elements?)

src/components/chatbox/chatbox.module.css
thiskevinwang: Removed in 71cceec18

src/components/chatbox/chatbox.tsx
thiskevinwang: updated in 47cfb1c13

src/components/chatbox/chatbox.tsx
thiskevinwang: Split out `WelcomeMessage` in c9b9105ab

src/components/chatbox/message.module.css
thiskevinwang: Camel cased! c398f888f

src/components/chatbox/message.module.css
braydenlove: @kellyschummer could we simply do a back arrow if you want to change your selection instead of this red/green button pattern?

src/components/chatbox/ai-feature-toast.tsx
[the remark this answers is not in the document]
thiskevinwang: Fixed with 1705b03a1

src/components/chatbox/message.tsx
dstaley: Sorry, my message was definitely not well worded! These two buttons are almost visually identical, and they should remain so since they're the same action, just inverse of one another. In instances like this, I like to refactor them out into their own component so that I don't run into a situation where I change the prop on one but not the other. That being said this is certainly just a suggestion, and something that isn't a huge deal since they're right next to each other in code.

src/components/chatbox/message.tsx
```diff
@@ -226,26 +224,94 @@ export const AssistantMessage = ({
...
+	text: string
+}
+
+export type MessageType =
+	| UserMessageData
+	| AssistantMessageData
+	| ApplicationMessageData
+
+// A pure, helper component that renders a list of messages, switching based on their type.
+export const MessageList = ({ messages }: { messages: MessageType[] }) => {
```
dstaley, line 276: love this!

dstaley (approved):

src/components/chatbox/message.module.css
zchsh: 💭 Would it be acceptable to use the "primary" state to indicate selection here, rather than custom coloration? Here's a quick mockup via Swingset for clarity:

CleanShot 2023-09-21 at 16 02 05@2x

<details>

<summary>Swingset code for Button</summary>

```jsx
const FeedbackExample = () => (
  <>
    <div data-theme="light">
      <div
        style={{
          display: "flex",
          gap: "24px",
          padding: "24px",
          background: "var(--token-color-surface-primary)",
        }}
      >
        <div
          style={{
            display: "grid",
            gap: "8px",
            gridTemplateColumns: "min-content min-content",
          }}
        >
          <span style={{ color: "var(--token-color-foreground-primary)" }}>
            Initial
          </span>
          <span></span>
          <Button
            onClick={() => console.log("clicked button")}
            aria-label="yes"
            color="secondary"
            icon={<SwingsetTestIcon name="thumbs-up" />}
          />
          <Button
            onClick={() => console.log("clicked button")}
            aria-label="No"
            color="secondary"
            icon={<SwingsetTestIcon name="thumbs-down" />}
          />
        </div>

        <div
          style={{
            display: "grid",
            gap: "8px",
            gridTemplateColumns: "min-content min-content",
          }}
        >
          <span style={{ color: "var(--token-color-foreground-primary)" }}>
            1
          </span>
          <span></span>
          <Button
            onClick={() => console.log("clicked button")}
            aria-label="Yes"
            icon={<SwingsetTestIcon name="thumbs-up" />}
          />
          <Button
            onClick={() => console.log("clicked button")}
            aria-label="No"
            color="secondary"
            icon={<SwingsetTestIcon name="thumbs-down" />}
          />
        </div>

        <div
          style={{
            display: "grid",
            gap: "8px",
            gridTemplateColumns: "min-content min-content",
          }}
        >
          <span style={{ color: "var(--token-color-foreground-primary)" }}>
            -1
          </span>
          <span></span>
          <Button
            onClick={() => console.log("clicked button")}
            aria-label="Yes"
            color="secondary"
            icon={<SwingsetTestIcon name="thumbs-up" />}
          />
          <Button
            onClick={() => console.log("clicked button")}
            aria-label="No"
            icon={<SwingsetTestIcon name="thumbs-down" />}
          />
        </div>
      </div>
    </div>

    <div data-theme="dark">
      <div
        style={{
          display: "flex",
          gap: "24px",
          padding: "24px",
          background: "var(--token-color-surface-primary)",
        }}
      >
        <div
          style={{
            display: "grid",
            gap: "8px",
            gridTemplateColumns: "min-content min-content",
          }}
        >
          <span style={{ color: "var(--token-color-foreground-primary)" }}>
            Initial
          </span>
          <span></span>
          <Button
            onClick={() => console.log("clicked button")}
            aria-label="yes"
            color="secondary"
            icon={<SwingsetTestIcon name="thumbs-up" />}
          />
          <Button
            onClick={() => console.log("clicked button")}
            aria-label="No"
            color="secondary"
            icon={<SwingsetTestIcon name="thumbs-down" />}
          />
        </div>

        <div
          style={{
            display: "grid",
            gap: "8px",
            gridTemplateColumns: "min-content min-content",
          }}
        >
          <span style={{ color: "var(--token-color-foreground-primary)" }}>
            1
          </span>
          <span></span>
          <Button
            onClick={() => console.log("clicked button")}
            aria-label="Yes"
            icon={<SwingsetTestIcon name="thumbs-up" />}
          />
          <Button
            onClick={() => console.log("clicked button")}
            aria-label="No"
            color="secondary"
            icon={<SwingsetTestIcon name="thumbs-down" />}
          />
        </div>

        <div
          style={{
            display: "grid",
            gap: "8px",
            gridTemplateColumns: "min-content min-content",
          }}
        >
          <span style={{ color: "var(--token-color-foreground-primary)" }}>
            -1
          </span>
          <span></span>
          <Button
            onClick={() => console.log("clicked button")}
            aria-label="Yes"
            color="secondary"
            icon={<SwingsetTestIcon name="thumbs-up" />}
          />
          <Button
            onClick={() => console.log("clicked button")}
            aria-label="No"
            icon={<SwingsetTestIcon name="thumbs-down" />}
          />
        </div>
      </div>
    </div>
  </>
);
```


</details>

src/components/chatbox/ai-feature-toast.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+const TOAST_AUTO_DISMISS_MS = 15000
+
+export function AIFeatureToast() {
+	const { session } = useAuthentication()
+
+	useEffect(() => {
+		// skip toast if we're past the expiration date
+		if (new Date() > COOKIE_EXPIRES_AT) {
+			// return
```
zchsh, line 24: ❓ Curious why the `return` statements here commented out - are we avoiding having these conditionals active until some future state? (Without the `returns` uncommented, seems like the toast would _always_ pop up if `session?.meta.isAIEnabled` is truthy, even if it's been previously dismissed!)

src/components/chatbox/message.module.css
```diff
@@ -0,0 +1,94 @@
...
+	margin-bottom: 24px;
+}
+
+.actionButtons {
+	display: flex;
+	flex-direction: row;
+	gap: 16px;
+}
+
+.ratingLike {
```
zchsh, line 75: Mentioned in the relevant thread, and this might be more a design question than anything else, but I think we could resolve the `!important` here by using `secondary` state for the initial rating button state (same as is now), and then `primary` state for the "selected" rating state (then no need for color overrides).

CleanShot 2023-09-21 at 16 02 05@2x

src/components/chatbox/message.tsx
```diff
@@ -0,0 +1,317 @@
...
+	/** copy rich text; not raw markdown */
+	const handleCopy = async () => {
+		// CSS Module classname is a reasonable "hook" to grab the element
+		// rendered by ReactMarkdown.
+		const el = document.getElementsByClassName(s.markdown)?.[0]
+
+		if (el) {
+			// https://www.nikouusitalo.com/blog/why-isnt-clipboard-write-copying-my-richtext-html/
+			const clipboardItem = new ClipboardItem({
+				// @ts-expect-error - innerText is available
```
zchsh, line 103: ⛏️ Nitpick: I think you can get rid of the `@ts-expect-error` here with a (somewhat wordy) type guard:

```suggestion
		if (el && 'innerText' in el && typeof el.innerText === 'string') {
			// https://www.nikouusitalo.com/blog/why-isnt-clipboard-write-copying-my-richtext-html/
			const clipboardItem = new ClipboardItem({
```

src/components/chatbox/message.tsx
```diff
@@ -0,0 +1,317 @@
...
+	return (
+		<div className={classNames(s.message, s.assistant)}>
+			<IconTile className={classNames(s.icon)}>
+				<IconWand24 style={{ width: 24, height: 24 }} />
+			</IconTile>
+			<div className={classNames(s.content)}>
+				<ReactMarkdown
+					className={s.markdown}
+					components={{
+						// @ts-expect-error - ignore this type incompatibility
```
zchsh, line 126: ⛏️ Nitpick: this is really outside the scope of this work, so totally optional, but if you hop over to `mdx-code-blocks/index.tsx` and drop in a placeholder type for `MdxPre` I think you could drop this `@ts-expect-error`:

```tsx
export function MdxPre({
	className,
	...restProps
}: {
	/**
	 * @TODO(zchsh): `@hashicorp/react-code-block` needs some type work,
	 * this is a temporary solution for compatibility with `PreProps`
	 * in that package.
	 */
	children: $TSFixMe
	className?: string
	hasBarAbove?: boolean
	theme?: 'light' | 'dark'
}) {
```

src/components/chatbox/message.tsx
```diff
@@ -0,0 +1,317 @@
...
+					</div>
+				</div>
+			</div>
+
+			<div className={classNames(s.gutter)}></div>
+		</div>
+	)
+}
+
+// TODO(kevinwang): error styling.
```
zchsh, line 232: 🎏 Flagging in case this is a `TODO` that needs doing before merge!

src/components/chatbox/utils.ts
```diff
@@ -0,0 +1,313 @@
+import { useEffect, useState, useRef } from 'react'
+
+// use a ref to check if an element's scrollbar is visible
+export function useScrollBarVisible() {
```
zchsh, line 4: 💭 ⛏️ Very optional - the first three helpers & hooks in this file (`useScrollBBarVisible`, `mergeRefs`, `useThrottle`) seem both generic and useful enough that they might be worth lifting out to `src/lib` or something!

src/components/chatbox/message.module.css
kellyschummer: @thiskevinwang @zchsh - I like the above approach with the primary state but would recommend we keep consistent with our active state -  Kevin and I agree and decided the simplify this flow even further and remove the input- Keeping the engagement to the 👍 👎 . 
Screenshot 2023-09-21 at 1 19 14 PM

src/components/command-bar/commands/search/unified-search/components/dialog-body/chat.module.css
```diff
@@ -0,0 +1,23 @@
...
+	/* override .linkOrButton tokens */
```
zchsh, line 2: 💭 Re: `!important` overrides being mentioned in that other thread - if we _do_ want to stick with the override coloration, maybe this "token replacement" approach could be a less "oh no we used `!important`" way to get the overrides in place! 🤔

src/components/command-bar/components/dialog/header.tsx
```diff
@@ -44,23 +44,30 @@ const CommandBarDialogHeader = () => {
...
-					onChange={(e: ChangeEvent<HTMLInputElement>) =>
-						setCurrentInputValue(e.target.value)
-					}
-					placeholder={inputPlaceholder}
-					ref={inputRef}
-					value={currentInputValue}
-				/>
+				{currentCommand.name == 'chat' ? (
+					// This is a dummy element to display text
+					// TODO(kevinwang): make this less hacky
```
zchsh, line 49: 🎏 Flagging this `TODO` in case it should be resolved before merging!

src/components/command-bar/index.tsx
```diff
@@ -175,7 +172,10 @@ const CommandBarProvider = ({ children }: CommandBarProviderProps) => {
 				<Command.Dialog isOpen={state.isOpen} onDismiss={toggleIsOpen}>
 					<Command.Header />
 					<Command.Body />
-					<Command.Footer />
+					{
+						// TODO(kevinwang): This conditional needs an abstraction
```
zchsh, line 176: 🎏 Flagging this unresolved `TODO` - hot take we could probably just leave this as-is and figure out a more robust solution later, if we want!

src/components/chatbox/message.module.css
zchsh: 📌 Oooh, related future thought, and likely outside the scope of this PR: should we consider consistency with the tutorials feedback form? cc @kellyschummer would love your take on this 🙇‍♂️

CleanShot 2023-09-21 at 15 40 50@2x

CleanShot 2023-09-21 at 15 41 30@2x

zchsh (approved):
✅ This is incredible work! 🤩

I can't claim to have a full understanding of all the chat mechanisms, nor the underlying API created for this work, but have done my best to review all the code where I felt capable, and opted-in and tested the experience - seems ready to me 👍

Happy to dive deeper if more thorough review coverage on the API routes and underlying API route feels essential 🙇‍♂️

src/components/chatbox/message.module.css
thiskevinwang: FYI @kellyschummer, I'm going to merge this PR, and follow up with the bulk of [VQA](https://app.asana.com/0/1100423001970639/1205523765153638/f) in a separate PR. 
- That'll include deleting the "Feedback Text Input"

src/components/command-bar/components/dialog/header.tsx
thiskevinwang: ACK - I'm going to take no action for now

src/components/command-bar/index.tsx
thiskevinwang: ACK - I'm going to take no action for now

src/components/chatbox/message.tsx
thiskevinwang: Will address this in an upcoming chunk of VQA work!

src/components/chatbox/message.tsx
thiskevinwang: f52efdc8a - this is unfortunately more lines of code, but it's a large improvement with clear and immediately actionable steps.

src/components/chatbox/message.tsx
thiskevinwang: That looks like it _should_ satisfy the compiler but it doesn't seem to work — going to skip this change

CleanShot 2023-09-22 at 11 29 03@2x

src/components/chatbox/ai-feature-toast.tsx
thiskevinwang: Thanks for catching this! (Accidentally committed testing code)
fixed in 240a273af

PR #2129, opened 2023-08-16
feat: add filter to openapi-view sidebar

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Adds sidebar filtering to OpenApiDocsView.

## 📸 Design Screenshots

| Before | After |
| - | - |
| before | after |

## 🧪 Testing

    - On large viewports, there should be a filter at the top of the sidebar
    - On small viewports, there should be a filter at the top of the mobile menu
    - On all viewports, the sidebar filtering should look & function as expected

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/1204678746647847/1205271542994749/f
[preview]: https://dev-portal-git-zsopenapi-sidebar-filter-hashicorp.vercel.app/
[/hcp/api-docs/vault-secrets]: https://dev-portal-git-zsopenapi-sidebar-filter-hashicorp.vercel.app/hcp/api-docs/vault-secrets

### DIFF
--- a/src/views/open-api-docs-view/components/open-api-sidebar-contents/index.tsx
+++ b/src/views/open-api-docs-view/components/open-api-sidebar-contents/index.tsx
@@ -6,6 +6,7 @@
 // Third-party
 import { useState } from 'react'
 // Components
+import FilterInput from 'components/filter-input'
 import { SidebarNavMenuItem } from 'components/sidebar/components'
 // Utils
 import { useNavItemsWithActive } from './utils/use-nav-items-with-active'
@@ -14,17 +15,14 @@ import { filterByTitle } from './utils/filter-by-title'
 import type { OpenApiNavItem } from 'views/open-api-docs-view/types'
 // Styles
 import s from './open-api-sidebar-contents.module.css'
-import FilterInput from 'components/filter-input'
 
 /**
  * Renders sidebar contents for OpenApiDocsView.
  */
 export function OpenApiSidebarContents({
 	navItems,
-	navResourceItems,
 }: {
 	navItems: OpenApiNavItem[]
-	navResourceItems: OpenApiNavItem[]
 }) {
 	const [filterValue, setFilterValue] = useState('')
 
@@ -36,39 +34,18 @@ export function OpenApiSidebarContents({
 
 	// Render a generic list of `SideBarNavMenuItem`
 	return (
-		<>
-			<div className={s.filterAndNav}>
-				<FilterInput
-					value={filterValue}
-					onChange={setFilterValue}
-					placeholder="Filter sidebar"
-				/>
-				<SidebarNavMenuItemsList items={filteredNavItems} />
-			</div>
-			{/* Render resources, if present */}
-			{navResourceItems?.length > 0 ? (
-				<SidebarNavMenuItemsList
-					items={[
-						{ divider: true },
-						{ heading: 'Resources' },
-						...navResourceItems,
-					]}
-				/>
-			) : null}
-		</>
-	)
-}
-
-/**
- * Renders an unordered list of nav items, with list styles reset.
- */
-function SidebarNavMenuItemsList({ items }: { items: OpenApiNavItem[] }) {
-	return (
-		<ul className={s.listResetStyles}>
-			{items.map((item: OpenApiNavItem, index: number) => (
-				// eslint-disable-next-line react/no-array-index-key
-				<SidebarNavMenuItem item={item} key={index} />
-			))}
-		</ul>
+		<div className={s.filterAndNav}>
+			<FilterInput
+				value={filterValue}
+				onChange={setFilterValue}
+				placeholder="Filter sidebar"
+			/>
+			<ul className={s.listResetStyles}>
+				{filteredNavItems.map((item: OpenApiNavItem, index: number) => (
+					// eslint-disable-next-line react/no-array-index-key
+					<SidebarNavMenuItem item={item} key={index} />
+				))}
+			</ul>
+		</div>
 	)
 }
--- a/src/views/open-api-docs-view/components/open-api-sidebar-contents/index.tsx
+++ b/src/views/open-api-docs-view/components/open-api-sidebar-contents/index.tsx
@@ -3,13 +3,10 @@
  * SPDX-License-Identifier: MPL-2.0
  */
 
-import { useMemo } from 'react'
-import { useRouter } from 'next/router'
-// Lib
-import { useActiveSection } from 'lib/hash-links/use-active-section'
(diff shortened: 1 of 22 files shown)


src/views/open-api-docs-view/components/open-api-sidebar-contents/index.tsx
```diff
@@ -23,64 +24,28 @@ export function OpenApiSidebarContents({
...
-	 */
-	const urlPathname = useMemo(() => newUrl(asPath).pathname, [asPath])
+	// Highlight active navItems, including `#hash` links via `useActiveSection`.
+	const navItemsWithActive = useNavItemsWithActive(navItems)
 
-	/**
-	 * Build an array of section slugs,
-	 * and pass these to `useActiveSection`.
-	 */
-	const sectionSlugs = useMemo(() => {
```
zchsh, line 43: I've split all the "active section" and highlighting logic out to a separate file for clarity.

"Split" view is probably more useful for this file 👍

src/views/open-api-docs-view/components/open-api-sidebar-contents/utils/use-nav-items-with-active.ts
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * Highlight any active matched sidenav items.
+ * These could match `#activeSection` or the full pathname.
+ *
+ * Note that if there are no `#hash` links in `navItems`, then the
+ * `useActiveSection` hook will not set up any IntersectionObservers,
+ * and the `activeSection` variable will be `undefined`. The highlighting
+ * of standard link matches will still function as expected.
+ */
+export function useNavItemsWithActive(
```
zchsh, line 81: Split this hook out, and tried to chunk out the component parts so that they're easier to understand.

src/views/open-api-docs-view/types.ts
```diff
@@ -83,10 +83,11 @@ export interface OpenApiDocsParams extends ParsedUrlQuery {
  */
 type DividerNavItem = { divider: true }
 type HeadingNavItem = { heading: string }
-type LinkNavItem = {
+export type LinkNavItem = {
 	title: string
 	fullPath: string
 	theme?: ProductSlug
+	isActive?: boolean
```
zchsh, line 90: Revised split-out hook has more well-constructed types, this property is optionally present on `LinkNavItem` and that's now properly reflected 👍

src/views/open-api-docs-view/components/open-api-sidebar-contents/utils/filter-by-title.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * Given a flat array of items, and a `filterValue`,*
+ * Return an array of filtered items, including only those items that
+ * are objects with a `title` attribute that matches the `filterValue`.
+ *
+ * If the provided `filterValue` is falsy, we return all items unchanged.
+ */
+export function filterByTitle<T = unknown>(
+	items: T[],
```
kendallstrautman, line 9: Should we just be explicit with the type here that we expect `title`? Or do you prefer the type guard? Curious if we need to give any feedback if there isn't a title on the items for DX.

src/views/open-api-docs-view/components/open-api-sidebar-contents/utils/use-nav-items-with-active.ts
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * Here, we remove the hash, so that we can compare to the `fullPath`
+ * of each nav item, which are not expected to have hashes,
+ * to the URL path that we're on.
+ *
+ * Note that domain does not matter, as we're just grabbing the full path.
+ */
+function usePathname() {
+	const { asPath } = useRouter()
+	return useMemo(() => newUrl(asPath).pathname, [asPath])
```
kendallstrautman, line 21: ❓ I'm surprised this wouldn't error without the base path as the second arg to the url constructor? Isn't `asPath` just a relative path?

kendallstrautman, line 21: Nice I'm glad that utility function is coming in handy. At first my brain was like 'where is the base path' 😆

src/views/open-api-docs-view/components/open-api-sidebar-contents/utils/use-nav-items-with-active.ts
```diff
@@ -0,0 +1,105 @@
...
+	navItems: OpenApiNavItem[]
+): OpenApiNavItem[] {
+	// Transform the incoming nav items into section slugs
+	const sectionSlugs = useMemo(
+		() => sectionSlugsFromNavItems(navItems),
+		[navItems]
+	)
+
+	// Determine which of the navItem sectionsSlugs is the active section slug
+	const activeSection = useActiveSection(sectionSlugs)
```
kendallstrautman, line 91: This is interesting behavior, when a filtered nav is also hitting on the section highlighting 🤔 I can see an argument for the consistency but while I was playing around with it the UX was a bit unintuitive when I had filter input data present and then the active item changed after scrolling.

kendallstrautman (approved):
Code is good! Just had one comment /opinion on the UX that might be worth thinking through with design.

src/views/open-api-docs-view/components/open-api-sidebar-contents/utils/filter-by-title.ts
zchsh: I'm using the vague type & type guard cause we do expect _some_ items to not have a title... this is a great point on DX though 🤔 I'll update to log a lil warning in dev if _none_ of the incoming items have a `title` property!

src/views/open-api-docs-view/components/open-api-sidebar-contents/utils/use-nav-items-with-active.ts
zchsh: ah great catch! for now i'll see if I can update to removing the active section highlighting when filtering is being applied. Will also check with design on this during final VQA!

zchsh: ✅ have updated. agree that the UX here can feel a lil funny sometimes though 🤔

🔥 super hot take, but I think our "sidebar filter" might be less critical to have in this type of use case, where we have all the searchable strings in the sidebar also in the main body area, so easy to `Cmd + F`. Not to say the filtering doesn't add something - just that the complexity it adds may not be worth the benefit. Something to iterate on I think!

PR #2138, opened 2023-08-21
feat(open-api-view): update hash URL on active section change

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Updates `OpenApiDocsView` to keep the URL in sync with the `#active-section` currently scrolled to on the page.

This page also updates the sidebar to use the top-level item as an `#overview` top-of-page link, rather than the previous `/hcp/api-docs/vault-secrets` full path link.

## 🤷 Why

For more convenient link sharing of URLs when visiting `OpenApiDocsView` pages and sharing their URLs.

## 🛠️ How

Note there is a little more complexity to this task than might initially meet the eye, as broadly speaking we have two types of `#anchor-link` targets on the page:
- `#section` anchor links, like [/hcp/api-docs/vault-secrets#open-app-secrets], which are monitored by `useActiveSection` and show up highlighted in the sidebar when their page area is scrolled to
- `#random-tiny-property` anchor links, like [/hcp/api-docs/vault-secrets#open-app-secrets_response_default_details]

When linking to `#section` areas, the "active section" highlighting and the URL bar are fully in sync, this is relatively straightforward.

When linking to `#random-tiny-property` areas, we do want the "active section" highlighting to update immediately... but we do _not_ want the URL bar to immediately reflect the new `#section` URL, as that would erase the `#random-tiny-property` URL that was just clicked.

This work is further complicated by `next/router` seemingly [failing to support `{ scroll: false }` option](https://github.com/vercel/next.js/discussions/35072).

With all this in mind, the way I've approached the solution is:

- 🔗 To implement basic URL `#section` syncing, implement a hook `useSyncedUrlHash`, which takes in the `activeSection` from our `useActiveSection` hook and ensures it's reflected in the URL
- 💔 To work around `next/router`'s `{ scroll: false }` option [seemingly being broken](https://github.com/vercel/next.js/discussions/35072), use `window.history.replaceState` in `useSyncedUrlHash` instead of `router.replace()` (as noted in code comments).
- 🤝 To avoid having `window.history.replaceState` conflict with `next/router`, implement a `useRouteChangeState` hook that indicates when it's safe to update the `#active-section` in the URL, so that `next/router` and `window.history.replaceState` can work nicely together
- 👯 To avoid overwriting `#active-section` in the URL right after a visitor has jumped to a `#random-tiny-property`, include a `wasHashJustChanged` property in the state returned by `useRouteChangeState`. If the hash was just changed, don't sync the active section to the URL.
- 👂 To avoid `wasHashJustChanged` becoming permanently `true` after a hash link jump, add a scroll event listener and revert `wasHashJustChanged` to `false` when the page is scrolled. This in turn means the URL `#section` updates resume.

## 📸 Design Screenshots

> **Note**: majority of work in this PR doesn't affect appearance directly, it's more about the URL behaviour. But one change that does have a visual impact is we now always have exactly one item highlighted in the sidebar.

| Before | After |
| - | - |
| before | after |

## 🧪 Testing

    - As you scroll the page, the first item in the sidebar should lose its "active" state, as it is now an `#overview` hash link
    - As you scroll the page, the URL should update with `#active-section` hash links to reflect the position on the page
    - When you scroll back to the top of the page, the URl should _not_ show `/hcp/api-docs/vault-secrets#overview`, it should instead revert to `/hcp/api-docs/vault-secrets`
- [ ] Try landing directly on the base URL, [/hcp/api-docs/vault-secrets]
    - Even after the page mounts, the URL should stay clean
    - Note: this is despite the "active section" technically being the `#overview` section (as shown by highlighting in the sidebar), testing this behaviour ensures our handling of the "first active section" is as expected (ie don't show it in the URL, it's not really useful to do so)
- [ ] Try landing directly on a "major element" hash link, like [/hcp/api-docs/vault-secrets#open-app-secrets]
    - The page should jump to the item, the active section should be reflected in the sidebar and should be retained in the URL
- [ ] From an existing scroll position, with a "major" active section, try clicking a hash link
    - For example, from [/hcp/api-docs/vault-secrets#open-app-secrets], click the anchor link for the `Default Response`
    - The browser URL should _retain_ the original hash link until you start scrolling.
    - Once you start scrolling, the URL should reflect the active section again, which in the example will be the next active section (`#open-app-secret`) since we've jumped closer to it
- [ ] Try landing directly on a "minor element" hash link, like [/hcp/api-docs/vault-secrets#open-app-secrets_response_default_details]
    - On initial load, the page should jump to the item
    - The active section ("major" heading elements) should be reflected in the sidebar once things mount
    - The URL should _retain_ the original hash link until you start scrolling
    - Once you start scrolling, the URL should reflect the active section

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/1204678746647847/1205211659139665/f
[preview]: https://dev-portal-git-zsopenapi-update-url-hash-hashicorp.vercel.app/
[/hcp/api-docs/vault-secrets#open-app-secrets]: https://dev-portal-git-zsopenapi-update-url-hash-hashicorp.vercel.app/hcp/api-docs/vault-secrets#open-app-secrets
[/hcp/api-docs/vault-secrets]: https://dev-portal-git-zsopenapi-update-url-hash-hashicorp.vercel.app/hcp/api-docs/vault-secrets
[/hcp/api-docs/vault-secrets#open-app-secrets_response_default_details]: https://dev-portal-git-zsopenapi-update-url-hash-hashicorp.vercel.app/hcp/api-docs/vault-secrets#open-app-secrets_response_default_details

### DIFF
--- a/src/components/badge/index.tsx
+++ b/src/components/badge/index.tsx
@@ -27,14 +27,6 @@ const Badge = ({
 	const hasText = !!text
 	const hasLabel = !!ariaLabel
 	const isIconOnly = hasIcon && !hasText
-	const isStatusBadge =
-		color == 'success' || color == 'warning' || color == 'critical'
-
-	if (isStatusBadge && !hasIcon) {
-		throw new Error(
-			'`Badge`s used for communicating status must have an icon to avoid relying on color alone.'
-		)
-	}
 
 	if (!hasIcon && !hasText) {
 		throw new Error(
--- a/src/lib/hash-links/use-active-section.ts
+++ b/src/lib/hash-links/use-active-section.ts
@@ -15,11 +15,10 @@ import getFullNavHeaderHeight from 'lib/get-full-nav-header-height'
  */
 export function useActiveSection(
 	slugs: string[],
-	isEnabled: boolean = true,
-	defaultSection?: string
+	isEnabled: boolean = true
 ): string {
 	const visibleHeadings = useRef<Set<string>>(new Set())
-	const [activeSection, setActiveSection] = useState<string>(defaultSection)
+	const [activeSection, setActiveSection] = useState<string>()
 	const previousY = useRef<number>()
 
 	/**
--- a/src/lib/hash-links/use-active-section.ts
+++ b/src/lib/hash-links/use-active-section.ts
@@ -15,10 +15,11 @@ import getFullNavHeaderHeight from 'lib/get-full-nav-header-height'
  */
 export function useActiveSection(
 	slugs: string[],
-	isEnabled: boolean = true
+	isEnabled: boolean = true,
+	defaultSection?: string
 ): string {
 	const visibleHeadings = useRef<Set<string>>(new Set())
-	const [activeSection, setActiveSection] = useState<string>()
+	const [activeSection, setActiveSection] = useState<string>(defaultSection)
 	const previousY = useRef<number>()
 
 	/**
--- a/src/views/open-api-docs-view/components/open-api-docs-mobile-menu-levels/index.tsx
+++ b/src/views/open-api-docs-view/components/open-api-docs-mobile-menu-levels/index.tsx
@@ -24,13 +24,11 @@ export function OpenApiDocsMobileMenuLevels({
 	productData,
 	navItems,
 	navResourceItems,
-	topOfPageSlug,
 }: {
 	// Product data, used to generate mobile menu levels.
 	productData: ProductData
 	navItems: OpenApiNavItem[]
 	navResourceItems: OpenApiNavItem[]
-	topOfPageSlug: string
 }) {
 	return (
 		<MobileMenuLevels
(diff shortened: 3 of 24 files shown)


braydenlove:
The `codeblocks` are disappearing at `1152px` and reappearing at `923px` when the sidebar goes away. They then disappear again at 792px. That feels a bit too early, can we make a custom breakpoint (as much as I don't want to) to have the `codeblocks` disappear at >~`420px` only? I want to make sure we aren't taking functionality away from folks who have multiple screens up or shrink their screen size. LMK what you think.

braydenlove:
Curious if this 8px padding is to allow for room when we have the zach corners?
image

If so I can live with it, if not I would like the text to ideally be left aligned with the code container like so:
image

braydenlove:
Can we update the `codeblocks` to have a corner radius of `5px` to be consistent? I know we are using the old `codeblock` component so ignore me if this causes chaos.

zchsh:
Note: closing this for now as it doesn't feel worth the complexity to ship, will likely revisit this at some point 👍 🎟️ [[future] Update #operation-anchor in URL as page is scrolled](https://app.asana.com/0/1204678746647847/1205211659139651/f).

zchsh:
Note: closing this for now as it doesn't feel worth the complexity to ship, will likely revisit this at some point 👍 🎟️ [[future] Update #operation-anchor in URL as page is scrolled](https://app.asana.com/0/1204678746647847/1205211659139651/f).

PR #2139, opened 2023-08-22
fix: build variant URLs as paths to fix local dev

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Fixes a tutorials-specific local development bug related of variants.

## 🤷 Why

For folks running the site locally from the `hashicorp/tutorials` repo, it seems the front-end has to be restarted for any new variants or variant options to function as expected. We want the front-end to work and display variants and their interactivity correctly, without requiring frequent restarts.

## 🛠️ How

The root cause of the local development bug seems to be related to middleware rewrites, which are generated only once when the dev server starts. In more detail:

- Rewrites allow links with search params like `?variants=variant:option` to be interpreted by `getStaticProps` as an additional `/variant:option` URL path segment built on the current tutorial URL
- Variants are switched on the front-end with `Link` elements, that link to variant URLs
    - When a variant URL is visited, new results are returned from `getStaticProps`, and this updates the variant data shown on the page
    - Note we do have a `VariantContext`, but this isn't exactly used as interactive client-side context, as it only updates when the variant data incoming from `getStaticProps` changes
- Upstream, variants URLs are built the format `?variants=variant:option`
- The upstream format of variant URLs _requires middleware rewrites_ to be up-to-date
- Middleware rewrites are only updated when the front-end process restarts, so in order for new variant URLs to function upstream, the front-end process must be restarted

This PR implements a short-term fix for the issue:

- Links are now constructed as path additions to the base tutorial URL
- In other words, we now use the `/<tutorialPath>/variant:option` URL format directly, rather than relying on middleware rewrites of `?variants=variant:option`
- This shift in URL structure means middleware rewrites do not need to be up to date (and in fact may not be needed at all? not sure), and the dev server does not need to be restarted for new variant and variants option URLs to function
- We use the `next/link` `as` prop to ensure the URL looks the same in the browser, appearing to be the search param style like `?variants=variant:option` 

## 🧪 Testing

- [ ] Work off a variants branch locally and test that new variant and option additions work as expected
    - Clone `hashicorp/tutorials`, and with Docker running, run `make`, this will start everything including a frontend we'll ignore at `http://localhost:3000`
    - Pull down this branch, modify `.env` to point this front-end to the Learn API running at `http://localhost:5001`
    - Start the front-end off this branch with `npm run start`, should start up at `http://localhost:3001`
    - Mess around with adding and removing variants, and interact with new variants on the front-end.
    - Variants and variant options added after starting the dev server should function as expected, even though they won't be accounted for in middleware rewrites.

## 💭 Anything else?

This PR does _not_ account for multiple combined variant options... not really if that's a thing, but the `?variants=key:value` format implies that `?variants=key1:value1,key2:value2` might be possible... It seems like our `getStaticProps` function supports only one additional path segment, so only one `key:value` variant pair... but figured worth mentioning in case I missed something!

[task]: https://app.asana.com/0/1202097197789424/1205317011687135/f
[preview]: https://dev-portal-git-zsdebug-variants-local-dev-hashicorp.vercel.app/

### DIFF
--- a/src/components/dropdown-disclosure/components/list-item/index.tsx
+++ b/src/components/dropdown-disclosure/components/list-item/index.tsx
@@ -112,6 +112,7 @@ const DropdownDisclosureButtonItem = ({
 const DropdownDisclosureLinkItem = ({
 	children,
 	href,
+	as,
 	icon,
 	rel,
 	target,
@@ -122,6 +123,7 @@ const DropdownDisclosureLinkItem = ({
 			<Link
 				className={s.link}
 				href={href}
+				as={as}
 				rel={rel}
 				target={target}
 				onClick={onClick}
--- a/src/components/dropdown-disclosure/components/list-item/types.ts
+++ b/src/components/dropdown-disclosure/components/list-item/types.ts
@@ -80,7 +80,9 @@ interface DropdownDisclosureLinkItemProps {
 	href: NativeAnchorProps['href']
 
 	/**
-	 * An alias to display when navigating to the href.
+	 * An optional URL to display instead of the provided `href`.
+	 * The `href` will still be navigated to, but the URL will appear to
+	 * be the value passed to `as`.
 	 */
 	as?: LinkProps['as']
 
--- a/src/components/dropdown-disclosure/components/list-item/types.ts
+++ b/src/components/dropdown-disclosure/components/list-item/types.ts
@@ -80,6 +80,11 @@ interface DropdownDisclosureLinkItemProps {
 	href: NativeAnchorProps['href']
 
 	/**
+	 * An alias to display when navigating to the href.
+	 */
+	as?: LinkProps['as']
+
+	/**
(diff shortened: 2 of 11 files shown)


kendallstrautman:
https://github.com/hashicorp/dev-portal/assets/36613477/29e4caaf-3e36-4176-9efa-8b5e2a7746fa

I'm seeing a state issue with the active button.

kendallstrautman:
I'm seeing a state issue with the variant button in the meta

https://github.com/hashicorp/dev-portal/assets/36613477/4ba036d3-13b9-442f-92ec-e6a9c1bc5cd5

kendallstrautman:
Also, to your note about the query param structure hinting at the ability to support multiple variants but the routing doesnt yet support, that is known and intentional. We only want to support one variant now but a reason behind the visual query param routes was to make it more intuitive down the line to add multiple variants.

PR #2143, opened 2023-08-28
chore(openapi): tweak code block visibility

zchsh:
## 🔗 Relevant links

- [Asana task](url) 🎟️

### DIFF
--- a/src/views/open-api-docs-view/components/operation-sections/operation-sections.module.css
+++ b/src/views/open-api-docs-view/components/operation-sections/operation-sections.module.css
@@ -14,6 +14,11 @@
 	grid-template-areas: 'header' 'details';
 	grid-template-columns: minmax(0, 1fr);
 
+	/* We show 'examples' in a one-column if the viewport size is large enough */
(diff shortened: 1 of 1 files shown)


braydenlove:
@zchsh this looks really good, curious if we can shore up the spacing here at all when it goes to one column? It looks like these is a bottom margin of `24px` which could be removed to make it feel less awk. I will let you decide tho as I know removing margins can get trick at specific breakpoints! 
image

braydenlove:
@zchsh when looking at the original designs it does look like there is a doubling up of spacing between the bottom margin and the various grid gaps. You might want to refine here so it gets a little closer to the original designs and ideally fixes the single column funk.

image
Figma [designs](https://www.figma.com/file/wjoC3ZVsA1kDuPsFLjEFuY/API-Template?type=design&node-id=526%3A34648&mode=design&t=Sl5EkBzHe2uZ1W0K-1)

image
PR screenshot

braydenlove (approved):
LGTM 🚢

PR #2145, opened 2023-08-28
chore(openapi): mod slugs to match source

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

In OpenAPI view, modifies all `#anchor-link` slug values so that they're truer to source, namely by retaining original casing.

## 🧪 Testing

    - All operation anchor links should now more closely match their operation IDs, like `#mySpecialOperation` rather than `#my-special-operation`
    - Similarly, all nested parameter slugs should more closely match their keys
    - With the above changes, all slugs should remain unique (not necessarily something easy to test in browser, probably better to assert this through code review, but clicking around a bunch would probably add some confidence)

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/1204678746647847/1205367854380355/f
[preview]: https://dev-portal-git-zsopenapi-anchors-match-source-hashicorp.vercel.app/
[/hcp/api-docs/vault-secrets]: https://dev-portal-git-zsopenapi-anchors-match-source-hashicorp.vercel.app/hcp/api-docs/vault-secrets

### DIFF
--- a/src/views/open-api-docs-view/utils/get-operation-props.ts
+++ b/src/views/open-api-docs-view/utils/get-operation-props.ts
@@ -4,7 +4,6 @@
  */
 
 // Third-party
-import { paramCase } from 'change-case'
 import slugify from 'slugify'
 // Local
 import { getRequestData, getResponseData, truncateHcpOperationPath } from './'
@@ -41,9 +40,7 @@ export async function getOperationProps(
 			}
 
 			// Create a slug for this operation
-			const operationSlug = slugify(paramCase(operation.operationId), {
-				lower: true,
-			})
+			const operationSlug = slugify(operation.operationId)
 
 			/**
 			 * Parse request and response details for this operation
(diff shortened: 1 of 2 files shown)


src/views/open-api-docs-view/utils/get-operation-props.ts
```diff
@@ -41,9 +40,7 @@ export async function getOperationProps(
 			}
 
 			// Create a slug for this operation
-			const operationSlug = slugify(paramCase(operation.operationId), {
-				lower: true,
-			})
+			const operationSlug = slugify(operation.operationId)
```
thiskevinwang, line 43: 👍

thiskevinwang (approved):
Looks good to me!

> [!NOTE] 
> For posterity: This PR is the result of a small nit that I raised after viewing the preview for #2138. I had questioned the need to transform the typically-camel-cased [operationId](https://spec.openapis.org/oas/v3.1.0#operation-object) that we get directly from an OpenAPI  spec. This PR reverts work to kebab-case the operation id

PR #2155, opened 2023-09-01
feat: /profile/account

thiskevinwang:
# Description 

This introduces the new `/profile/account` page.
- test: https://dev-portal-git-kwprofile-account-hashicorp.vercel.app/profile/account

CleanShot 2023-09-07 at 19 40 23@2x

CleanShot 2023-09-07 at 19 39 37@2x

- ~This hides breadcrumbs on `/profile/*` pages — cc @kellyschummer~
- This moves the external link to HCP under `/profile/account`

This is based on:
- #2161 

This work is in support of:
- #2128 

Figma:
- https://www.figma.com/file/0Op0f8SEDLB8HuOHzfXAwP/Chat_GPT?type=design&node-id=1045-9676&mode=dev

---
  - https://app.asana.com/0/0/1205398382936229

### DIFF
--- a/src/views/profile/account-view/index.tsx
+++ b/src/views/profile/account-view/index.tsx
@@ -25,7 +25,7 @@ export default function ProfileAccountView() {
 				breadcrumbLinks={[
 					{ title: 'Developer', url: '/' },
 					{
-						title: 'Account',
+						title: 'Basic info',
 						url: '/profile/account',
 						isCurrentPage: true,
 					},
@@ -43,20 +43,20 @@ const ProfileAccountViewContent = () => {
 	return (
 		<div>
 			<Heading level={1} weight="bold" size={500}>
-				Account
+				Basic info
 			</Heading>
 			<Text className={s.subheading}>
-				A read-only view of personal account info from&nbsp;
+				A read-only view of your profile details from&nbsp;
 				<a href="https://portal.cloud.hashicorp.com/account-settings">
 					HCP Account settings
 				</a>
 			</Text>
 
 			<Heading level={2} weight="bold" size={200} className={s.sectionTitle}>
-				Email
+				Primary email
 			</Heading>
 
-			<Card className={s.section} elevation="mid">
+			<Card className={s.section} elevation="base">
 				<Text className={s.description}>
 					Email address associated with your profile name
 				</Text>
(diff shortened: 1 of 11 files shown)


src/views/profile/account-view/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+				<Text className={s.description}>
+					Email address associated with your profile name
+				</Text>
+				<div className={s.copySnippetWrapper}>
+					<CopySnippet textToCopy={user?.email} />
+				</div>
+			</Card>
+
+			<Heading level={2} weight="bold" size={200} className={s.sectionTitle}>
+				HashiCorp ID
```
zchsh, line 61: 💭  Noticed most of the properties on `user`, eg `email`, `id`, etc, are optional. What happens here if the user doesn't have one of those properties - is that a scenario we kind of expect, or is this more a quirk with the types?

src/views/profile/account-view/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+			</Heading>
+
+			<Card className={s.section} elevation="mid">
+				<Text className={s.description}>Your ID within HashiCorp</Text>
+				<div className={s.copySnippetWrapper}>
+					<CopySnippet textToCopy={user?.id} />
```
zchsh, line 67: ⛏️  Nit: I don't _think_ the `?` check is necessary here, since we have `AuthenticatedView` wrapping. If it is, maybe after `const { user } = useAuthentication()` we do like `if (user === null) return null` or something to avoid the conditional checks throughout?

```suggestion
					<CopySnippet textToCopy={user.id} />
```

zchsh (approved):
✅ Looks great!

⛏️  Had some nitpicks around scenarios where the `user` property might be `null`, or where the `user.id` or `user.email` properties might not exist (since they're marked optional in types), but may not be the most helpful things to try to address at this stage 👍

src/views/profile/account-view/index.tsx
thiskevinwang: ID will never be missing if the user is authenticated. Similarly, I'm fairly confident email will never be missing either.

If either are the case, an empty CopySnippet would be shown which isn't particularly critical.

src/views/profile/account-view/index.tsx
thiskevinwang: removed that `?` check in https://github.com/hashicorp/dev-portal/pull/2155/commits/48c4aa04623c7a7707c06721eb61b7436e83c039

src/views/profile/account-view/index.tsx
thiskevinwang: udated the type in https://github.com/hashicorp/dev-portal/pull/2155/commits/48c4aa04623c7a7707c06721eb61b7436e83c039

kellyschummer (approved):
🥰 Updates look great! Thank you @thiskevinwang

PR #2156, opened 2023-09-01
zs.stub openapi preview tool

zchsh:
- feat: add api route to generate props
- feat: add necessary basic input components
- fix: make statusIndicatorConfig optional
- feat: function to call API route for static props
- feat: rough in input components, set up the view
- feat: add page file, don't render in prod
- chore: tidy up breadcrumb links

## 🔗 Relevant links

- [Asana task](url) 🎟️

### DIFF
--- a/src/pages/api/get-open-api-docs-view-props.ts
+++ b/src/pages/api/get-open-api-docs-view-props.ts
@@ -1,0 +1,113 @@
+import { getStaticProps } from 'views/open-api-docs-view/server'
+// Types
+import type { NextApiRequest, NextApiResponse } from 'next'
+import type { OpenAPIV3 } from 'openapi-types'
+import type { ProductSlug } from 'types/products'
+
+/**
+ * Boilerplate page configuration, we could in theory expose this so visitors
+ * to the preview tool could manipulate it, but we intentionally just
+ * hard-code here to keep the focus of the preview tool on OpenAPI spec
+ * contents.
+ */
+const GENERIC_PAGE_CONFIG = {
+	// basePath same no matter what, preview tool is on static route
+	basePath: '/openapi-docs-preview',
+	// No versioning in the preview tool, focus on one spec file at a time
+	context: { params: { page: [] } },
+	// Product slug, using HCP to just show a generic HashiCorp logo,
+	// so that the preview tool's focus can remain on the spec file contents
+	productSlug: 'hcp' as ProductSlug,
+	// Generic resource items, we can set more specific ones closer to launch
+	navResourceItems: [
+		{
+			title: 'Tutorial Library',
+			href: '/tutorials/library',
+		},
+		{
+			title: 'Certifications',
+			href: '/certifications',
+		},
+		{
+			title: 'Community',
+			href: 'https://discuss.hashicorp.com/',
+		},
+		{
+			title: 'Support',
+			href: 'https://www.hashicorp.com/customer-success',
+		},
+	],
+}
+
+/**
+ * We expected posted data to be an OpenAPI spec in JSON format.
+ * We also allow an optional schema `info.description`, which normally would be
+ * included in the spec content, so that authors can more easily develop
+ * and preview their `info.description` content using our preview tool.
+ */
+type ExpectedBody = {
+	openApiJsonString: string
+	openApiDescription: string
+}
+
+export default async function handler(
+	req: NextApiRequest,
+	res: NextApiResponse
+) {
+	// Reject non-POST requests, only POST is allowed
+	if (req.method !== 'POST') {
+		res.setHeader('Allow', ['POST'])
+		res.status(405).json({ error: 'Method not allowed' })
+	}
+
+	/**
+	 * Build the static props from the POST'ed page configuration data,
+	 * which includes the full OpenAPI spec as a string.
+	 */
+	const { openApiDescription, openApiJsonString } = JSON.parse(
+		req.body
+	) as ExpectedBody
+
+	/**
+	 * Construct some preview data just to match the expected `getStaticProps`
+	 * signature. The `versionId` and `releaseStage` don't really matter here.
+	 */
+	const versionData = [
+		{
+			versionId: 'preview',
+			releaseStage: 'preview',
+			sourceFile: openApiJsonString,
+		},
+	]
+
+	/**
+	 * Build static props for the page
+	 */
+	const staticProps = await getStaticProps({
+		// Pass the bulk of the page config
+		...GENERIC_PAGE_CONFIG,
+		// Pass the constructed version data
+		versionData,
+		/**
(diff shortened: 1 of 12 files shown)


src/pages/api/get-open-api-docs-view-props.ts
```diff
@@ -0,0 +1,113 @@
...
+ * We also allow an optional schema `info.description`, which normally would be
+ * included in the spec content, so that authors can more easily develop
+ * and preview their `info.description` content using our preview tool.
+ */
+type ExpectedBody = {
+	openApiJsonString: string
+	openApiDescription: string
+}
+
+export default async function handler(
```
zchsh, line 53: This is kind of the "driver" of the view - accepts OpenAPI spec `JSON` from the client, and sends back the "static props" needed to render an `OpenApiDocsView`.

Note that we've set up `GENERIC_PAGE_CONFIG` here to keep the preview input interface simpler. We _could_ in theory expose more inputs, but it didn't feel worth it at this time - we want this preview tool to focus in on the contents of the spec `.json` file.

src/pages/open-api-docs-preview/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+ * in production.
+ */
+export async function getStaticProps(): Promise<
+	GetStaticPropsResult<Record<string, never>>
+> {
+	/**
+	 * In production, return a 404 not found for this page.
+	 * In other environments (local, preview, and staging), show the page.
+	 */
+	return IS_PRODUCTION ? { notFound: true } : { props: {} }
```
zchsh, line 23: I haven't actually tried this before, but in theory this should mean that even after merging this PR to `main`, <https://developer.hashicorp.com/open-api-docs-preview> will `404`, and we'll only see this preview tool in local dev, preview deploys, and perhaps most importantly our staging deploy at <https://developer.hashicorp.services/open-api-docs-preview> 👀

src/views/open-api-docs-preview/components/file-string-input/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Render a very basic file input.
+ *
+ * This is a temporary solution until we have a React version of FileInput
+ * set up, which we'd likely do in the `web` monorepo.
+ *
+ * For now, this felt sufficient for an internal preview tool for OpenAPI specs.
+ */
+export function FileStringInput({
```
zchsh, line 12: I've added very basic input components. Intent is to rebuild the UI of this preview tool with the appropriate HDS components once they're available... but since this is an internal tool, felt pragmatic to try to deliver value more immediately, even with very basic-looking inputs.

src/views/open-api-docs-preview/components/open-api-preview-inputs/open-api-preview-inputs.module.css
```diff
@@ -0,0 +1,55 @@
...
+	border: none;
+	background: var(--highlight-color);
+	color: black;
+	cursor: pointer;
+}
+
+.generateButton {
+	flex-shrink: 0;
+
+	/* Swap blue and purple to match the --highlight-color */
```
zchsh, line 51: Did this as a bit of a hack, feels worth it to have the preview UI stand out a bit. Will definitely be removing this once we have the HDS components needed to implement the ✨ [actual designs](https://www.figma.com/file/wjoC3ZVsA1kDuPsFLjEFuY/API-Template?type=design&node-id=720-3306&mode=design&t=fmzKCVL2rUMWKEpb-4) ✨ .

kendallstrautman:
@zchsh do you have any quick links to open api spec files that I can use to test? I'm not sure exactly where those live.

zchsh:
@kendallstrautman Absolutely! Probably the easiest one to test with is HCP Vault Secrets, since there's the corresponding "actual view" to compare to - https://github.com/hashicorp/hcp-specs/blob/main/specs/cloud-vault-secrets/preview/2023-06-13/hcp.swagger.json. There are also other spec files in that repo that would be suitable for testing (though they won't have been authored for display in this way, so the content might look a lil off!)

src/views/open-api-docs-preview/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { OpenApiPreviewInputs } from 'views/open-api-docs-preview/components/open-api-preview-inputs'
+// Types
+import type { OpenApiDocsViewProps } from 'views/open-api-docs-view/types'
+
+/**
+ * Render an OpenApi docs view alongside preview inputs that allow
+ * dynamic, client-side updating of the props used to render the view.
+ */
+function OpenApiDocsPreviewView() {
+	const [staticProps, setStaticProps] = useState<OpenApiDocsViewProps>()
```
thiskevinwang, line 21: I would call this `[apiDocsViewProps, setApiDocsViewProps]` since these are not "static" — they're `fetch`ed from from an API route.

src/views/open-api-docs-preview/components/open-api-preview-inputs/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+	setStaticProps,
+}: {
+	setStaticProps: (v: OpenApiDocsViewProps) => void
+}) {
+	const [error, setError] = useState<{
+		title: string
+		description: string
+		error: string
+	}>()
+	const [isLoading, setIsLoading] = useState(false)
```
thiskevinwang, line 36: This is non-blocking suggestion to reduce the sheer # of  lines of code, and prop passing:

The `isLoading`,`error`, and data fetching code in this component can be conveniently abstracted away by a react-query `useQuery` hook... (and a fixed **querykey**; see more below)

Similarly, the lone `setStaticProps` state setter could be completely removed. The parent that owns `[staticProps, setStaticProps]` could replace `useState` and instead do a "read-only" usage of `useQuery` with the previously mentioned fixed **querykey**.

thiskevinwang (approved):
The feature works! Approved

There's opportunity to leverage react-query's `useQuery` to cut down on lines of code for handling `isLoading`, `error`, and data, and state-setter-prop-passing.

src/views/open-api-docs-preview/index.tsx
zchsh: Oh nice call! Much better name 💖

src/views/open-api-docs-preview/components/open-api-preview-inputs/index.tsx
zchsh: Ah neat - would love to explore this in a future PR! I tried for a bit, but ran into some complications around having the query only fetch initially (and only refetch) when the `updateStaticProps` function is called - wasn't immediately clear based on the docs the best way to do that. Very open to refactoring this later 🙇‍♂️

PR #2161, opened 2023-09-06
chore(auth): return `user.id`

thiskevinwang:
# Description

This is in support of https://github.com/hashicorp/dev-portal/pull/2128

### DIFF
--- a/next-auth.d.ts
+++ b/next-auth.d.ts
@@ -26,6 +26,8 @@ declare module 'next-auth' {
 		 * Object representing an authenticated user.
 		 */
 		user: {
+			/** The user's HashiCorp ID */
+			id?: string
 			/** The user's nickname. */
 			nickname?: string | null
 		} & DefaultSession['user']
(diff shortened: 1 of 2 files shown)


next-auth.d.ts
```diff
@@ -26,6 +26,8 @@ declare module 'next-auth' {
 		 * Object representing an authenticated user.
 		 */
 		user: {
+			/** The user's HashiCorp ID */
```
zchsh, line 29: 💭  Not actually sure this is correct, but taking a shot at fleshing out this comment to try to clarify my own understanding!

```suggestion
			/** The user's HashiCorp Cloud IDP ID. Optional, as some users may still have pre-Cloud-IDP accounts. */
```

zchsh (approved):
✅ Looks good! Had a clarifying question about the nature or origin of `user.id`, I'm not as familiar with our `next-auth` setup so apologies if it's off 😅

next-auth.d.ts
thiskevinwang: @zchsh (PR auto merged before I could respond here) 

HashiCorp "ID" and "Cloud IDP ID" are largely the same, IMO.

"ID" is a more generic label, whereas "Cloud IDP ID" includes the exact service name. I reached for the generic label and didn't give much deep thought after the fact.

PR #2162, opened 2023-09-07
feat: load redirects from ptfe-releases

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Tweak redirects loading to allow redirects to be loaded from closed repositories, and adds configuration to load redirects from the `hashicorp/ptfe-releases` repo.

### All redirects should continue to function

- [ ] The [preview] should look & function as expected
- [ ] Redirects loaded from the remote repositories should continue to function

### Redirects should be setup to load from `ptfe-releases`

Redirects should load from `ptfe-releases`, but _only_ once the `redirects.js` file exists in a `ref` known to our content API. To fake this in order to test that redirects will work once released, clone this PR's branch down, and in `getRedirectsFromContentRepo`, force the ref `main` for `ptfe-releases` rather than the latest ref known by our content API:

```js
// const latestContentRef = await getLatestContentRefForProduct(repoName)
const latestContentRef =
	repoName === 'ptfe-releases'
		? 'refs/heads/main'
		: await getLatestContentRefForProduct(repoName)
```

### Redirects for `ptfe-releases` should already work for local content dev

Redirects should load when running the site from the content repo `hashicorp/ptfe-releases`. To test this:

- Clone down `hashicorp/ptfe-releases`
- Modify the `website-start` script to use this branch of `dev-portal` for `website-preview`, as shown below
- Run the site locally with `npm run start`
- Redirects should function

| Build script modification |
| - |
| CleanShot 2023-09-08 at 14 53 51@2x |

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/1202097197789424/1205445365434398/f
[preview]: https://dev-portal-git-zsload-ptfe-redirects-hashicorp.vercel.app/

### DIFF
--- a/build-libs/redirects.js
+++ b/build-libs/redirects.js
@@ -194,7 +194,7 @@ async function getRedirectsFromContentRepo(
 	 * Evaluate the redirects file string, filter invalid redirects, and add
 	 * a host condition for proxied sites.
 	 *
-	 * TODO: remove `addHostCondition` once Sentinel is migrated (ie once
+	 * TODO(zachshilton): remove `addHostCondition` once Sentinel is migrated (ie once
 	 * `docs.hashicorp.com/sentinel` redirects to `developer.hashicorp.com`).
 	 */
 	/** @type {Redirect[]} */
@@ -251,7 +251,7 @@ async function buildProductRedirects() {
 			 * expect any attempt to fetch the redirects to 404. To account for this,
 			 * we've added a temporary try-catch block here.
 			 *
-			 * TODO: remove this try-catch block, once `hashicorp/ptfe-releases` has
+			 * TODO(zachshilton): remove this try-catch block, once `hashicorp/ptfe-releases` has
 			 * cut a release with a `redirect.js` file and that release has been
 			 * extracted by our content workflows. At that point, we'll expect the
 			 * redirects.js file to exist, and only then should 404s break the build.
--- a/build-libs/redirects.js
+++ b/build-libs/redirects.js
@@ -259,7 +259,7 @@ async function buildProductRedirects() {
 			 */
 			(async function getPtfeRedirects() {
 				try {
-					await getRedirectsFromContentRepo('ptfe-releases')
+					return await getRedirectsFromContentRepo('ptfe-releases')
 				} catch (e) {
 					if (e.toString() === 'HttpError: Not Found') {
 						console.warn(
--- a/build-libs/redirects.js
+++ b/build-libs/redirects.js
@@ -244,8 +244,7 @@ async function buildProductRedirects() {
 			getRedirectsFromContentRepo('terraform-docs-common'),
 			getRedirectsFromContentRepo('hcp-docs', '/redirects.js'),
 			/**
-			 * Note: HashiCorp Cloud Platform docs were recently in the process of
-			 * being cut over ` is in the process of adding a
+			 * Note: `hashicorp/ptfe-releases` is in the process of adding a
 			 * `redirects.js` file. Until a release is cut and our content API
 			 * has a `latestRef` corresponding to a commit with that file, we
 			 * expect any attempt to fetch the redirects to 404. To account for this,
--- a/build-libs/redirects.js
+++ b/build-libs/redirects.js
@@ -135,63 +135,72 @@ async function getLatestContentRefForProduct(product) {
 }
 
 /**
- * Fetches a redirects file for a given product from the given ref and evaluates the contents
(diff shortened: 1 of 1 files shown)


zchsh:
Note: At first I tried to completely isolate the change to load PTFE redirects, but the more I dove into what was actually happening already, the more it seemed to make sense to clean up some debt as part of the PR as well. This is why this PR _replaces_ `getRedirectsForProduct` rather than introducing a net new function.

Notes on specific changes:
- Previously we allowed passing a custom ref when loading redirects, but this was actually just ignored for all repos, we always replace it with the latestRef loaded from the content API
- Previously we had a HCP-docs related fallback (load from cloud.hashicorp.com if hcp-docs doesn't work), I removed this fallback as I believe that migration was a while back!

build-libs/redirects.js
```diff
@@ -135,63 +135,72 @@ async function getLatestContentRefForProduct(product) {
...
+	} else {
+		// Return early, in this build we can ignore repoName's redirects.
+		return []
 	}
-
+	/**
+	 * Evaluate the redirects file string, filter invalid redirects, and add
+	 * a host condition for proxied sites.
+	 *
+	 * TODO: remove `addHostCondition` once Sentinel is migrated (ie once
```
thiskevinwang, line 197: Request: Add your name to TODOs — the idea is not that you will necessarily be the one to address this TODO in the future, it's moreso an indicator of "who authored this TODO".

This is a nicer alternative to `git blame` / `git history`, which sometimes gets obfuscated by multi-line changes, and directs folks to incorrect TODO-authors.

```suggestion
	 * TODO(zachshilton): remove `addHostCondition` once Sentinel is migrated (ie once
```

build-libs/redirects.js
```diff
@@ -225,39 +234,46 @@ async function buildProductRedirects() {
...
+			getRedirectsFromContentRepo('hcp-docs', '/redirects.js'),
+			/**
+			 * Note: HashiCorp Cloud Platform docs were recently in the process of
+			 * being cut over ` is in the process of adding a
+			 * `redirects.js` file. Until a release is cut and our content API
+			 * has a `latestRef` corresponding to a commit with that file, we
+			 * expect any attempt to fetch the redirects to 404. To account for this,
+			 * we've added a temporary try-catch block here.
+			 *
+			 * TODO: remove this try-catch block, once `hashicorp/ptfe-releases` has
```
thiskevinwang, line 254:
```suggestion
			 * TODO(zachshilton): remove this try-catch block, once `hashicorp/ptfe-releases` has
```

build-libs/redirects.js
thiskevinwang: The red diff here is an example of me tagging myself in TODOs
https://github.com/hashicorp/dev-portal/pull/2128/commits/5499c850cef0ebd1de2320c5594b427b5fe9ad28

zchsh: ooh nice i love this pattern!

build-libs/redirects.js
```diff
@@ -225,39 +234,46 @@ async function buildProductRedirects() {
...
+			 * TODO(zachshilton): remove this try-catch block,
+			 * once `hashicorp/ptfe-releases` has cut a release with a `redirect.js`
+			 * file and that release has been extracted by our content workflows. At
+			 * that point, we'll expect the redirects.js file to exist, and only then
+			 * should 404s break the build.
+			 * Task: https://app.asana.com/0/1202097197789424/1205453036684673/f
+			 */
+			(async function getPtfeRedirects() {
+				try {
+					return await getRedirectsFromContentRepo('ptfe-releases')
```
thiskevinwang, line 262: Only thought is "do we want to just fetch ptfe-releases redirects from `main`?" This would decouple redirect edits from requiring a release, which in ptfe-releases is very cumbersome.

thiskevinwang (approved):

build-libs/redirects.js
zchsh: Great question! Rationale was that we should avoid this as a general rule in case content files are being moved around on main, but not yet released. Specifically, there was this recent PR that introduced the `redirects.js` file and also moved a bunch of files around:

- https://github.com/hashicorp/ptfe-releases/pull/740

If we load redirects from `main`, I think we'll end uploading redirects to files that aren't yet going to render because they haven't been extracted by our content API.

Let me know if that sounds right, very open to adjusting 🙇‍♂️

PR #2171, opened 2023-09-20
fix(openapi): support global HCP service paths

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Supports path truncation for global HCP services, which follow a different format than specific HCP product services.

## 🤷 Why

To better organize operations within global HCP service OpenAPI documentation.

## 🧪 Testing

    - The page should appear exactly as it does [upstream][upstream/hcp/api-docs/vault-secrets]
- [ ] Visit the [preview tool][/open-api-docs-preview]
    - When you upload a spec file for a global HCP service, you should no longer see the version number prefix in the operation path groupings.

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/1204678746647847/1205543368360320/f
[preview]: https://dev-portal-git-zsopenapi-support-global-hcp-grouping-hashicorp.vercel.app/
[/hcp/api-docs/vault-secrets]: https://dev-portal-git-zsopenapi-support-global-hcp-grouping-hashicorp.vercel.app/hcp/api-docs/vault-secrets
[/open-api-docs-preview]: https://dev-portal-git-zsopenapi-support-global-hcp-grouping-hashicorp.vercel.app/open-api-docs-preview


src/views/open-api-docs-view/utils/truncate-hcp-operation-path.ts
```diff
@@ -3,18 +3,47 @@
...
 
+/**
+ * Specific HCP service paths start with the service name,
+ * followed by a date-based version number,
+ * and always include "organization" and "project" parameters.
+ *
+ * We remove all of this prefix from the path for better grouping
+ * of operations within the specific spec.
+ */
+const SPECIFIC_SERVICE_PATTERN =
```
kendallstrautman, line 14: In the comments could you get a specific example or explain what this _should_ match. I find it helps when debugging regex in the future to have the human readable explanation. 

Also not sure if you wanted to use that GLOBAL_SERVICE_PATTERN in this via a template literal and `new RegExp(...)`

kendallstrautman (approved):

src/views/open-api-docs-view/utils/truncate-hcp-operation-path.ts
zchsh: Thanks for the feedback! 💖  🙌 

- **Re: examples**, I added a concrete + a generic example for each regex! Love this as a general rule when defining regexes, will try to do this proactively in the future 🙇‍♂️ 💯 
- **Re: sharing regex parts**, for now have left it a little less "DRY", and have left the version-date matching that's basically all `GLOBAL_SERVICE_PATTERN` does and the same version-date matching in `SPECIFIC_SERVICE_PATTERN` decoupled. I don't see a strong case for abstracting out the underlying regex at this point, but definitely happy to adjust if there are any strong opinions 🙇‍♂️

PR #2180, opened 2023-09-26
fix(openapi): revise default version logic

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Fixes an issue related to versioned used of `OpenApiDocsView`.

The issue is that In cases where a given OpenAPI spec has _multiple_ versions, but _no stable versions_, a so-called `latestStableVersion` ends up being _not found_, resulting in a 404 at the API docs URL (despite valid and useful preview versions existing).

This PR first clarifies intent by renaming to `defaultVersion`, which more accurately reflects that we don't _require_ a `stable` version to render OpenAPI docs. It implements and applies `defaultVersion` to ensure that if there are one or more versions of an OpenAPI document, we'll be able to find a default version (which may be `stable` or may be in a different release stage), and we'll render that version of the OpenAPI docs, avoiding any unexpected 404s.

Further details on the heuristic for finding the default version can be found in the code comments for the `findDefaultVersion` function.

## 🤷 Why

To allow OpenAPI docs that do not yet have a stable version to still render.

## 🛠️ How

- Implements a new `findDefaultVersion` function
- Uses `findDefaultVersion` to resolve the `latestStableVersion`-related 404 issue
- Various vaguely related chores
    - Cleans up unused imports
    - Avoids a potential `undefined` property in `getStaticProps`
    - Removes the now-unused implementation of `findLatestStableVersion`

## 🧪 Testing

> [!NOTE]
> The changes in this PR affect versioned `OpenApiDocsView` documentation, which is not yet in use. With this in mind, testing steps for this PR are focused on ensuring existing uses of `OpenApiDocsView` continue to look & behave exactly as they do upstream. For a demo of versioned documentation that is actually impacted by this work, see the PR [hashicorp/dev-portal-internal#119](https://github.com/hashicorp/dev-portal-internal/pull/119).

    - The page should look and function as it does [upstream][upstream/hcp/api-docs/vault-secrets]

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/1204678746647847/1205587495423433/f
[preview]: https://dev-portal-git-zsfix-openapi-default-version-hashicorp.vercel.app/
[/hcp/api-docs/vault-secrets]: https://dev-portal-git-zsfix-openapi-default-version-hashicorp.vercel.app/hcp/api-docs/vault-secrets


src/views/open-api-docs-view/server.ts
```diff
@@ -71,7 +68,7 @@ export async function getStaticProps({
 	serviceProductSlug = productSlug,
 	versionData,
 	basePath,
-	statusIndicatorConfig,
+	statusIndicatorConfig = null, // returned in getStaticProps, avoid undefined
```
zchsh, line 71: Bit of unrelated cleanup here, figured worth sneaking in 👍

src/views/open-api-docs-view/utils/find-default-version.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+import { OpenApiDocsVersionData } from '../types'
+import { sortDateVersionData } from './sort-date-version-data'
+
+/**
+ * Given an array of version data,
+ * Return the default version that should be shown.
+ *
+ * - If there is exactly one version, we treat it as the default version.
```
zchsh, line 13: Tried to be explicit about the heuristic here - hopefully it feels more intuitive now!

src/views/open-api-docs-view/server.ts
thiskevinwang: "must be JSON-serializable" would be a good distinction to note!

thiskevinwang (approved):
Seems reasonable to me!

src/views/open-api-docs-view/server.ts
zchsh, line 71:
```suggestion
	statusIndicatorConfig = null, // must be JSON-serializable
```

PR #2189, opened 2023-10-05
a11y: use inline link instead of plain anchor element

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-kssmall-a11y-fixes-hashicorp.vercel.app/profile/account) 🔎
- [Asana task](url) 🎟️

## 🗒️ What
Use `InlineLink` instead of plain `a` so that the link indication doesn't rely on color only. Also the aria label for the command bar header is aligned with the visible label. 

See issues outlined[ in this document.](https://docs.google.com/document/d/17VH69KP2TFNDkC7MEdP8-wUnec4UtcXfMUlCPX7eE6Y/edit#heading=h.38vj98htf15t)

## 🧪 Testing

- Sign in to the preview (with your HCP id added to accept list on edge config)
- Go to https://dev-portal-git-kssmall-a11y-fixes-hashicorp.vercel.app/profile/account
- You should see 'HCP Account Settings' link with an underline

- Use voice over or look in DOM to confirm that the aria-label for back icon in the command bar header says 'Return to search' instead of 'Back to search'
Screenshot 2023-10-05 at 11 07 39 AM

### DIFF
--- a/src/components/command-bar/commands/chat/index.tsx
+++ b/src/components/command-bar/commands/chat/index.tsx
@@ -17,7 +17,7 @@ const Icon = () => {
 		<IconArrowLeft24
 			tabIndex={0}
 			className={s.backArrow}
-			aria-label="Back to search"
+			aria-label="Return to search"
 			onKeyDown={(e) => {
 				if (e.key === 'Enter') {
 					setCurrentCommand('search')
--- a/src/views/profile/account-view/index.tsx
+++ b/src/views/profile/account-view/index.tsx
@@ -4,17 +4,18 @@
  */
 import { IconExternalLink16 } from '@hashicorp/flight-icons/svg-react/external-link-16'
 
+import useAuthentication from 'hooks/use-authentication'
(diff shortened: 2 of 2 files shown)


src/components/command-bar/commands/chat/index.tsx
```diff
@@ -17,7 +17,7 @@ const Icon = () => {
 		<IconArrowLeft24
```
kendallstrautman, line 17: Note this whole component should be a button and not an image with interactivity added. That would need to be a bigger refactor so pushing this update for now and separating that out. Issue is noted [in this doc ](https://docs.google.com/document/d/17VH69KP2TFNDkC7MEdP8-wUnec4UtcXfMUlCPX7eE6Y/edit#heading=h.d71g029sucyd)

kaitlynnefuery (approved):
@kendallstrautman LGTM

kellyschummer:
Thank you @kendallstrautman !

PR #2191, opened 2023-10-05
Some suggested a11y changes for AI integration feature

MelSumner:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-melsumner-a11y-ai-fix-hashicorp.vercel.app/) 🔎

## 🗒️ What

Return to search:
- removed the label (element) text for the "return to search" `button`
- removed the interactivity on the icon
- turned it all into a single button
- took the styles from the search button itself and tweaked it slightly (also this will need design review) 

Textarea:
- added visually hidden help text to the "Ask a question" `textarea` _<---this text will need to be made correct, there's just a placeholder there now. I added this because I asked a really long question and got an error message instead of an answer, but there were no instructions about how long my question could be._
- altered the `aria-label` text to be a little clearer for a non-sighted user

Message results:
- adds `aria-live` to user message so that the user with a screen reader will know that their question has been asked

## 🤷 Why

If merged, this PR would make some accessibility-related improvements to the new AI chat feature.

## 🛠️ How

I'm not familiar with React, so I wasn't able to make this code as clean as I would have preferred. There is still work to do here but I hope this draft PR will provide some additional support to help make this awesome new feature more inclusive. 

I tried to copy existing components/code where possible, and kept tweaking and testing with a screen reader until the outcomes were closer to the desired outcomes.

## 📸 Screenshots

Return to search button with screenreader text shown:
CleanShot 2023-10-05 at 18 12 15@2x

Textarea accessible name and description:
CleanShot 2023-10-05 at 18 04 45@2x

Demonstration of aria-live on user question:
CleanShot 2023-10-05 at 18 14 50@2x


kellyschummer:
@MelSumner Thank you!!! Looking at this styling- I'm leaning towards implementing our existing Button/Large/Tertiary - in this scenario. 
Screenshot 2023-10-05 at 5 08 21 PM copy

MelSumner:
@kellyschummer would it work the same on both light and dark mode?

kellyschummer:
Yes- Dark theme token values would display- that would be the only difference- Dark/foreground/action value

src/components/command-bar/commands/chat/index.tsx
```diff
@@ -7,33 +7,39 @@ import { IconArrowLeft24 } from '@hashicorp/flight-icons/svg-react/arrow-left-24
...
 import ChatBox from 'components/chatbox/chatbox'
 import { useCommandBar } from 'components/command-bar'
+import Text from 'components/text'
 
 import s from './chat.module.css'
 
-// An interactive Icon w/ "back" functionality
+// A button with an icon and text w/ "back" functionality
+// TODO this is hacky?
 const Icon = () => {
```
MelSumner, line 16: I know this should be able to be something else, but I don't know how to change it successfully.

thiskevinwang:
❤️ 🥹 @MelSumner  🎉🙏

MelSumner:
Ok so @kellyschummer I can't seem to get the icon color to be correct unless I inline style it, which feels bad. Could someone more familiar with React and this code base please check to see what might be going wrong? I've tried to screenshot/highlight the issue.  

CleanShot 2023-10-05 at 20 37 35@2x

src/components/chatbox/chatbox.module.css
```diff
@@ -105,6 +105,24 @@
...
+	width: 1px !important;
+	height: 1px !important;
+	margin: -1px !important;
+	padding: 0 !important;
+	overflow: hidden !important;
+	white-space: nowrap !important;
+	border: 0 !important;
+	clip: rect(1px, 1px, 1px, 1px) !important;
+	clip-path: inset(50%) !important;
+}
```
MelSumner, line 124: This is a common "visually hidden" or "screen-reader only" bit of CSS; do you have this as a helper class already available anywhere else?

src/components/chatbox/chatbox.tsx
```diff
@@ -380,6 +382,9 @@ const ChatBox = () => {
 							placeholder="Ask a question"
 							disabled={isLoading}
 						/>
+						<div id="user-instructions" className={s.userInstructions}>
+							The question can be 120 characters long.
```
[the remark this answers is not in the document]
kellyschummer: The question can be 200 characters long-

MelSumner, line 386:
```suggestion
							The question can be 200 characters long.
```

src/components/chatbox/chatbox.module.css
kendallstrautman: Yes we do! [The `g-screen-reader-only` class](https://github.com/hashicorp/dev-portal/blob/main/src/pages/style.css#L148-L166)

src/components/command-bar/commands/chat/index.tsx
```diff
@@ -31,9 +31,9 @@ const Icon = () => {
 }
 const chatCommand: CommandBarCommand = {
 	name: 'chat',
-	icon: <Icon />,
+	icon: <BackButton />,
 	inputProps: {
-		placeholder: () => 'Return to search',
+		placeholder: () => 'Hide me',
```
MelSumner, line 36: This seems to have to be there but also doesn't seem to do anything because it's not actually in input. If there is a way to do this better that would be good.

dstaley (approved):

PR #2193, opened 2023-10-07
Fix OAS 3.0 file parsing

schavis:
## 🔗 Relevant links

- Preview link: I don't have a branch preview link as I don't have permission to push commits to the repo (even though I was able to create a branch for some reason).
- Asana task: I'm not on the DevDot ENG team, so I don't have an Asana task for this work. But it's meant to unblock this work related to [JIRA epic #SPE-371](https://hashicorp.atlassian.net/browse/SPE-371).

## 🗒️ What

Adds code to handle parsing OAS files with:
- non-JSON request bodies by using the encoding as the parameter name so we can display it properly
- multiple response bodies for a given response code (e.g., a JSON response and an XML response). For now, the fix just filters out non-JSON responses, but we should eventually handle multi-responses more elegantly (maybe tabs?).

## 🤷 Why

The code currently does not reliably parse OAS 3.0 files. Using the default example file at https://petstore3.swagger.io/api/v3/openapi.json currently breaks the code. I am planning to convert Vault API docs to OAS 3.0 so I need this fixed to unblock that effort.

## 🛠️ How

Updated `getBodyParameterProps`:
 - explicitly checks for a `.properties` element and added code to handle the case where it is not present
 - removed assumption that the content type would be `application/json`

Updated `getResponseData`:
- converted the logic to a loop to account for codes with multiple response messages
- added a check on the content type to skip anything that wasn't `application/json`
- consolidated the various skip conditions so it's easier to track when we're bailing on an item

## 📸 Design Screenshots

N/A

## 🧪 Testing

- [ ] Confirm OAS pages using the OpenAPI docs view are unchanged
- [ ] Update an existing OAS page to use the PetStore API example file
- [ ] Confirm the page renders correctly when compared to the source schema

## 💭 Anything else?

I'm currently working on fixes for other blockers to converting the Vault API docs:

- We're not displaying information related to endpoint descriptions/summaries
- The current layout doesn't support links above the reference for hand-written content beyond additional resources. For example, our SDK and related tool listings, which should be at the top rather than hidden "below the fold".
- The current layout assumes the page is part of the HCP docset 
- The page breadcrumbs aren't rendering correctly (Vault is all-lowercase)
- The current layout doesn't allow for example request/response messages


schavis:
Replaced with https://github.com/hashicorp/dev-portal/pull/2201

PR #2194, opened 2023-10-09
cleanup: use global sr-only util class

dstaley:
## 🔗 Relevant links

- [Base PR](https://github.com/hashicorp/dev-portal/pull/2191)

## 🗒️ What

This PR makes some slight adjustments to use our global utility class for visually hidden elements in addition to tweaking how we style SVG elements within the Command Bar header to support `currentColor` without needing to use inline styles.

### DIFF
--- a/src/components/command-bar/commands/chat/chat.module.css
+++ b/src/components/command-bar/commands/chat/chat.module.css
@@ -4,9 +4,13 @@
  */
 
 .backButton {
-	.buttonIcon {
-		& svg {
-			color: var(--token-color-foreground-action) !important;
-		}
+	& svg {
+		/* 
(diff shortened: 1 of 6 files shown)


src/components/command-bar/commands/chat/chat.module.css
```diff
@@ -4,9 +4,13 @@
  */
 
 .backButton {
-	.buttonIcon {
-		& svg {
-			color: var(--token-color-foreground-action) !important;
-		}
+	& svg {
```
MelSumner, line 7: Thank you for figuring this out, it was driving me batty!

MelSumner (approved):
Thanks for these improvements!

PR #2211, opened 2023-10-24
chore: don't publish consul api docs from 2-15

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksmanually-filter-consul-api-version-hashicorp.vercel.app/) 🔎
- [Asana ticket](https://app.asana.com/0/1202097197789424/1205798063280001) 🎟️ 

## 🗒️ What

Manually filters out this version until we support configuration in the spec to filter out a spec file from publishing as per request in slack.

## 🧪 Testing

- [Visit the preview ](https://dev-portal-git-ksmanually-filter-consul-api-version-hashicorp.vercel.app/hcp/api-docs/consul)
- Validate that 2-15-2022 isn't an available version. There should be no version selector visible as we only have the single version [spec from 10-10-23](https://github.com/hashicorp/hcp-specs/tree/main/specs/cloud-global-network-manager-service/preview/2023-10-10)

## 💭 Anything else?

We should support config based option for authors to not publish a spec version.

### DIFF
--- a/src/pages/hcp/api-docs/consul/[[...page]].tsx
+++ b/src/pages/hcp/api-docs/consul/[[...page]].tsx
@@ -99,7 +99,7 @@ export const getStaticProps: GetStaticProps<
 	 * https://hashicorp.slack.com/archives/C5FSPUGDS/p1698178472462449
 	 */
 	const filteredVersionData = versionData.filter(
-		(version: ApiDocsVersionData) => version.versionId === '2022-02-15'
+		(version: ApiDocsVersionData) => version.versionId !== '2022-02-15'
 	)
 
 	// Generate static props based on page configuration, params, and versionData
--- a/src/pages/hcp/api-docs/consul/[[...page]].tsx
+++ b/src/pages/hcp/api-docs/consul/[[...page]].tsx
@@ -5,6 +5,7 @@
 
 // Lib
 import { fetchCloudApiVersionData } from 'lib/api-docs/fetch-cloud-api-version-data'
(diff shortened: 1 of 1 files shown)


src/pages/hcp/api-docs/consul/[[...page]].tsx
```diff
@@ -92,11 +93,20 @@ export const getStaticProps: GetStaticProps<
 	const versionData = await fetchCloudApiVersionData(
 		PAGE_CONFIG.githubSourceDirectory
 	)
+	/**
+	 * TODO: add in support for conditionally publishing specs in hcp-specs repo
+	 * For now just manually filtering out this spec as per request in slack.
+	 * https://hashicorp.slack.com/archives/C5FSPUGDS/p1698178472462449
+	 */
+	const filteredVersionData = versionData.filter(
+		(version: ApiDocsVersionData) => version.versionId !== '2022-02-15'
```
Comment (im2nguyen) [inline src/pages/hcp/api-docs/consul/[[...page]].tsx:102]: does this filter out this version or only include it?

it should be this instead right?

```suggestion
		(version: ApiDocsVersionData) => version.versionId !== '2022-02-15'
```

im2nguyen (approved):
small question, but otherwise this looks good!

src/pages/hcp/api-docs/consul/[[...page]].tsx
Comment (kendallstrautman) [reply src/pages/hcp/api-docs/consul/[[...page]].tsx:102]: I think I pushed up that change at the exact same time you suggested it! 😆 Thank you

vignesh-jayabalan (approved):

dstaley (approved):

PR #2213, opened 2023-10-26
fix: pin node & npm version in dockerfile

kendallstrautman:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

The docker build image workflow [has been failing ](https://github.com/hashicorp/dev-portal/actions/runs/6658627905/job/18095903036)at `npm install`. I was able to repro the same `errno ECONNRESET` error locally during `npm install` without changing anything in the file.

I think the docker image build was failing because it was installing the latest version of npm (10) yet still using node 18. When I ran this locally after pinning, `docker build -t dev-portal:local .` the build succeeded. 

It's a weird error so definitely open to input if there's other things I should try! 😅 

Screenshot 2023-10-26 at 4 41 55 PM

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # Copyright (c) HashiCorp, Inc.
 # SPDX-License-Identifier: MPL-2.0
 
-FROM docker.mirror.hashicorp.services/node:18-alpine AS deps
+FROM docker.mirror.hashicorp.services/node:18.18.2-alpine AS deps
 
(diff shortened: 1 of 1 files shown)


dstaley (approved):
Approving this to unblock folks, but this is definitely one of those "this shouldn't fix the error" situations. As a future PR, we should switch away from Alpine which has tons of weird, random, difficult to track down issues. https://martinheinz.dev/blog/92

PR #2215, opened 2023-10-31
enable custom next collection slugs in dev portal

heatlikeheatwave:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1204810520737800/1205154011279146/f) 🎟️
- [`learn-api-content-sync`](https://github.com/hashicorp/learn-api-content-sync/pull/88)
- [`learn-api`](https://github.com/hashicorp/learn-api/pull/187)

## 🗒️ What

If `next_collection` is defined on a `collection`, the frontend will use this to override the default behaviour for the Next / Previous button. If a next_collection is not passed provided, the next logical collection in the sidebar will be used (current behavior).

## 🤷 Why

Add support for authors to supply a custom href for the 'Next collection' button. This will be optional and will be defined in the [collection files](https://github.com/hashicorp/tutorials/blob/main/content/collections/boundary/credential-management.yml) in the tutorials repo as a plain string.

## 🧪 Testing

- [ ] Step 1: add `next_collection: terraform/aws-get-started` to any file in `/content/collections` in your local `tutorials` repo. [`example collections file`](https://github.com/hashicorp/tutorials/blob/main/content/collections/boundary/credential-management.yml)
- [ ] Step 2: checkout `heat/feat/next_collection_slug` in `learn-api` and run `make`
- [ ] Step 3: checkout `heat/feat/next-collection` in `learn-api-content-sync`, and update the LEARN API value to your local port
- [ ] Step 4: checkout `heat/feat/custom-next-button` in this repo. Visit the page for the file you added `next_collection` to.

### DIFF
--- a/src/lib/learn-client/api/collection/formatting.ts
+++ b/src/lib/learn-client/api/collection/formatting.ts
@@ -45,9 +45,9 @@ export function formatCollection(collection: ApiCollection): Collection {
 		level,
 		category: category || undefined,
 		tutorials: tutorials.map(formatToTutorialLite),
-		...(next_collection
-			? { nextCollection: formatToCollectionLite(next_collection) }
-			: {}),
+		nextCollection: next_collection
+			? formatToCollectionLite(next_collection)
+			: undefined,
 	}
 }
 
--- a/src/lib/learn-client/api/collection/formatting.ts
+++ b/src/lib/learn-client/api/collection/formatting.ts
@@ -30,6 +30,7 @@ export function formatCollection(collection: ApiCollection): Collection {
 		level,
 		category,
 		tutorials,
+		next_collection,
 	} = collection
 
 	return {
@@ -44,6 +45,9 @@ export function formatCollection(collection: ApiCollection): Collection {
 		level,
 		category: category || undefined,
 		tutorials: tutorials.map(formatToTutorialLite),
+		...(next_collection
+			? { nextCollection: formatToCollectionLite(next_collection) }
+			: {}),
 	}
 }
 
--- a/src/views/tutorial-view/index.tsx
+++ b/src/views/tutorial-view/index.tsx
@@ -145,8 +145,7 @@ function TutorialView({
 	const nextPreviousData = getNextPrevious({
 		currentCollection: collectionCtx.current,
 		currentTutorialSlug: slug,
-		nextCollectionInSidebar:
-			collectionCtx.current.nextCollection ?? tutorial.nextCollectionInSidebar,
+		nextCollectionInSidebar: tutorial.nextCollectionInSidebar,
 		formatting: {
 			getTutorialSlug,
 			getCollectionSlug,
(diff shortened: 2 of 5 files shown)


kendallstrautman:
Note: Once we release the api updates to staging, we can sync the staging database and see the changes in this preview!

src/lib/learn-client/api/collection/formatting.ts
```diff
@@ -44,6 +45,9 @@ export function formatCollection(collection: ApiCollection): Collection {
 		level,
 		category: category || undefined,
 		tutorials: tutorials.map(formatToTutorialLite),
+		...(next_collection
+			? { nextCollection: formatToCollectionLite(next_collection) }
+			: {}),
```
kendallstrautman, line 50: We should probably default to `undefined`?

src/views/tutorial-view/index.tsx
```diff
@@ -145,7 +145,8 @@ function TutorialView({
 	const nextPreviousData = getNextPrevious({
 		currentCollection: collectionCtx.current,
 		currentTutorialSlug: slug,
-		nextCollectionInSidebar: tutorial.nextCollectionInSidebar,
+		nextCollectionInSidebar:
+			collectionCtx.current.nextCollection ?? tutorial.nextCollectionInSidebar,
```
kendallstrautman, line 149: Let's actually handle this in the `server` file? So we could calculate `nextCollectionInSidebar` there.

src/lib/learn-client/api/collection/formatting.ts
heatlikeheatwave: Why should we default to undefined? This only adds a `next_collection` property if it exists

src/views/tutorial-view/index.tsx
heatlikeheatwave: I can do that!

kendallstrautman: Also I do think we'll need to adjust the href to be the dev portal path since the collection slug doesn't map to the actual path in developer. We have some helper functions, let me find them.

src/lib/learn-client/api/collection/formatting.ts
kendallstrautman: For sure, its up to you since its the same end result but the suggestion was more for consistency with the other learn client formatting functions.

src/views/tutorial-view/index.tsx
kendallstrautman: [getCollectionSlug!](https://github.com/hashicorp/dev-portal/blob/63ace1b5b948b09158b8f9f4c469f71637101c0c/src/views/collection-view/helpers/get-slug.ts#L38-L53)

heatlikeheatwave: The formatting is already handled in the helper [here](https://github.com/hashicorp/dev-portal/blob/main/src/views/tutorial-view/components/next-previous/helpers.ts#L57), so no need to change anything! (:

heatlikeheatwave: Moved it to server.ts in this commit: https://github.com/hashicorp/dev-portal/pull/2215/commits/40166997ad7272eb89e591b442a288b7847fb2ed

kendallstrautman (approved):

PR #2221, opened 2023-11-08
Redirect to /install from any paths with /downloads

heatlikeheatwave:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1204817897905865/1205712837415360/f) 🎟️

## 🗒️ What

- Change product install page path from `/[product]/downloads` to `/[product]/install`
- Redirects in `next.config.js` to handle the following:
  - `/[product]/downloads` to `/[product]/install`
  - `/[product]/downloads/enterprise` to `/[product]/install/enterprise`
  - `/vagrant/downloads/vmware` to `/vagrant/install/vmware`

## 🤷 Why
- From the Asana ticket:
> the page is called install, but the url is /downloads. we need this to be consistent, and from discovery, the preferred title / url is install. we need to change the url to be `https://developer.hashicorp.com/[product]/install`

## 🛠️ How

- Decided on adding the redirect to `next.config.js` rather than `middleware.ts` as the existing convention is to add dev portal redirects this way

## 📸 Design Screenshots

- There were no designs for this ticket

### Typical case:
- Navigate to [`/boundary/downloads`](https://dev-portal-git-heat-featinstall-redirects-hashicorp.vercel.app/boundary/downloads) and you should be redirected to [`/boundary/install`](https://dev-portal-git-heat-featinstall-redirects-hashicorp.vercel.app/boundary/install)

### Edge cases for /install path:
- Navigate to [`/nomad/downloads/enterprise`](https://dev-portal-git-heat-featinstall-redirects-hashicorp.vercel.app/nomad/downloads/enterprise) and you should be redirected to [`/nomad/install/enterprise`](https://dev-portal-git-heat-featinstall-redirects-hashicorp.vercel.app/nomad/install/enterprise)
- Navigate to [`/vagrant/downloads/vmware`](https://dev-portal-git-heat-featinstall-redirects-hashicorp.vercel.app/vagrant/downloads/vmware) and you should be redirected to [`/vagrant/install/vmware`](https://dev-portal-git-heat-featinstall-redirects-hashicorp.vercel.app/vagrant/install/vmware)

## 💭 Anything else?

I'm not sure if there are SEO considerations that need to be taken into account here. If someone with better SEO knowledge than me knows, please leave a comment!


dstaley (approved):

build-libs/redirects.js
```diff
@@ -309,6 +309,16 @@ async function buildDevPortalRedirects() {
 			destination: '/waypoint/integrations/hashicorp/:slug',
 			permanent: true,
 		},
+		{
+			source: '/:path/downloads',
+			destination: '/:path/install',
+			permanent: true,
```
kendallstrautman, line 315: If we wanted to be really explicit we could [use a regex ](https://nextjs.org/docs/pages/api-reference/next-config-js/redirects#regex-path-matching)to only match on product slugs we expect.

build-libs/redirects.js
```diff
@@ -309,6 +309,16 @@ async function buildDevPortalRedirects() {
...
+		},
+		{
+			source: '/:path/downloads/:slug',
+			destination: '/:path/install/:slug',
```
kendallstrautman, line 319: Is vagrant vmware the only install route with a subpath? Or are there others?

build-libs/redirects.js
```diff
@@ -309,6 +309,16 @@ async function buildDevPortalRedirects() {
...
+			permanent: true,
```
kendallstrautman, line 320: Same here I think we should match the two sub paths exactly with a regex?

kendallstrautman (changes requested):
I had one suggestion on tightening up the path matching. We'll also need to update any references to `/:product/download` in the rest of the application and we'll need to update the sitemaps. Also if we could get a plan for the source content updates that would be great since these are very high leverage pages for us and we don't want an SEO hit.

build-libs/redirects.js
heatlikeheatwave: Sure, I can do that. Just out of curiosity, is that to prevent any unintentional redirects or is there more to it than that?

heatlikeheatwave:
> Also if we could get a plan for the source content updates that would be great since these are very high leverage pages for us and we don't want an SEO hit.

Who should I reach out to get a plan for the source content updates?

build-libs/redirects.js
kendallstrautman: Yep exactly just to be really explicit and prevent unintentional redirects!

kendallstrautman (approved):
🚀

PR #2223, opened 2023-11-09
Move CTA graphics away from absolute positioning

RubenSandwich:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-rnhomepage-ctas-clickable-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1204770769465639/f) 🎟️

## 🗒️ What

Homepage CTAs indicate clickable, with the pointer hand, but aren't.
Specifically this was happening on the:
1. Certification CTA image
2. WAF CTA image

i.e. the image in this CTA card is not clickable:
Screenshot 2023-11-09 at 12 46 04 PM

## 🤷 Why

What was happening is because both the Card CTA and the Image were absolute positioning, the image was rendering above the Card which was causing the Card CTA anchor link to not be clickable over the CTA image.

## 🛠️ How

I fixed this problem by adding `pointer-events: none;` the CTA image. (Didn't I saw that didn't work before? Yes I did, but I must of done it incorrectly and not double checked. 🤦🏻 )

I tested a variety of ways to fix this issue before settling on removing the CTA image's absolute positioning:
1. Adding a lower z-index on the CTA image
    1. this cause it to be rendered before the CTA card
2. Adding another anchor tag that links the CTA image to the same location as the CTA card
    1. this caused another focusable element which degraded the accessibility experience of the webpage
3. [Modifying the CTA Card component, CardLink, to have the anchor tag not be absolutely positioned, and rather wrap it's children](https://github.com/hashicorp/dev-portal/blob/eeaaec0b9a89714a80c79e6ec3e514933abef51b/src/components/card-link/index.tsx#L48).
    1. this had too big of a blast/effect radius, as the CTA Card component is used in around 15 different places spread out across our docs site

## 🧪 Testing

1. Visit the developer homepage:
    1. [This PR](https://dev-portal-git-rnhomepage-ctas-clickable-hashicorp.vercel.app/)
    2. [Production](https://developer.hashicorp.com/)
2. Notice on [Production](https://developer.hashicorp.com/) that the anchor link does on the CTA image, on both Certification and WAF, is **not** clickable
3. Check that [This PR](https://dev-portal-git-rnhomepage-ctas-clickable-hashicorp.vercel.app/) makes the CTA image, on both Certification and WAF, is not clickable **is** clickable


mwickett (commented):
Click interactions seem right, but I'm seeing some text and image shifts from the current page in production. 

* The certification text shifts up vertically
* The WAF image size is slightly smaller

CleanShot 2023-11-09 at 15 23 37

RubenSandwich:
> Click interactions seem right, but I'm seeing some text and image shifts from the current page in production.
> 
> * The certification text shifts up vertically
> * The WAF image size is slightly smaller

Thank you for the eagle eyes and the video @mwickett! I will fix this today.

mwickett (approved):
Looks good!

PR #2217, opened 2023-11-03
Install page enhacement: version dropdown moved to PageHeader

heatlikeheatwave:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1204817897905865/1205712837415354/f) 🎟️
- [Figma](https://figmashort.link/GJTJF3)

## 🗒️ What

- Removed some dead code related to latest version as this is handled [here](https://github.com/hashicorp/dev-portal/blob/main/src/views/product-downloads-view/helpers.ts#L189-L204)

- Removed header subtitle:
Screenshot 2023-11-03 at 6 34 20 PM

- Moved `<VersionContextSwitcher />` to `PageHeader`

## 🤷 Why

For consistency, We'd like to update to the new component (version toggle/dropdown) used in our [documents](https://dev-portal-git-ksupdate-welcome-message-width-hashicorp.vercel.app/consul/docs) template and move it's placement to the page header.

### Mobile
Screenshot 2023-11-03 at 7 02 00 PM

### Desktop
Screenshot 2023-11-03 at 7 02 11 PM

## 🧪 Testing

- Navigate to [Vault install page](https://dev-portal-git-heat-featversion-select-update-hashicorp.vercel.app/vault/downloads)

PR #2227, opened 2023-11-10
patch: fix up terraform version selector sort

kendallstrautman:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ksfix-tf-order-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202801212949828/1205921409065634/f) 🎟️

## 🗒️ What

This logic was removed in #2217 ([see comment](https://github.com/hashicorp/dev-portal/pull/2217#discussion_r1382281828)) but caused the [terraform version on the install page](https://developer.hashicorp.com/terraform/install) to not be sorted properly. This is just a patch to fix the issue and we can readjust. 

**Preview**
Screenshot 2023-11-09 at 4 20 35 PM

**Upstream**
Screenshot 2023-11-09 at 4 20 50 PM

## 🧪 Testing

Validate that the version order on the [tf install page](https://dev-portal-git-ksfix-tf-order-hashicorp.vercel.app/terraform/install) is correct (latest first), [compared to production](https://developer.hashicorp.com/terraform/install).

### DIFF
--- a/src/components/version-context-switcher/index.tsx
+++ b/src/components/version-context-switcher/index.tsx
@@ -25,8 +25,25 @@ const VersionContextSwitcher = ({
 	options,
 }: VersionContextSwitcherProps): ReactElement => {
 	const currentProduct = useCurrentProduct()
-	const [selectedVersion, setSelectedVersion] =
-		useState<ContextSwitcherOption['value']>(initialValue)
+	let latestVersion = options[0].value
+	/** Update `latestVersion` to the latest version of the product
+	 * that has been released if it can be determined. This is needed
+	 * because not all products have `latest` listed as the first item
+	 * in the sorted array. E.g. `vault-enterprise` lists `+ent.hsm.fips1402`
+	 * variants first.
+	 */
+	for (let i = 0; i < options.length; i++) {
+		const versionLabel = options[i].label
+		const versionValue = options[i].value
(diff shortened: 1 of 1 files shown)


src/components/version-context-switcher/index.tsx
```diff
@@ -25,8 +25,25 @@ const VersionContextSwitcher = ({
 	options,
 }: VersionContextSwitcherProps): ReactElement => {
 	const currentProduct = useCurrentProduct()
-	const [selectedVersion, setSelectedVersion] =
-		useState<ContextSwitcherOption['value']>(initialValue)
+	let latestVersion = options[0].value
```
dstaley, line 28: I think this is the same algorithm? Could be nice to have it all part of one assignment, but I totally understand if this is just restoring code that was there previously.

```js
const latestVersion =
  options.find(({ label }) => label.includes("(latest)"))?.value ??
  options[0].value;
```

dstaley (approved):

anthonyvetter (approved):
LGTM!

src/components/version-context-switcher/index.tsx
kendallstrautman: Yes I do like this better. It was indeed a straight copy paste of what was there before. Hopefully we can remove this again while verifying the root cause of the TF sorting issue.

PR #2228, opened 2023-11-10
Bug fix: Terraform version dropdown on /install

heatlikeheatwave:
## 🗒️ What

- Terraform versioning was done in a `filterVersions` function in `/pages/terraform/install/index.tsx`.
- `filterVersions` was moved to `lib` so it could have its own test added to make sure this bug doesn't happen in the future
  - The only change in `filterVersions` was `latestPatchesOnly.push(`${x}.${y}.${z}`)` to `latestPatchesOnly.unshift(`${x}.${y}.${z}`)`. The `push` method here created the ASC order
- Added a test to make sure the returned versions are in DESC order (latest version first)
- Removed the code that was added back in as a patch fix in [this PR](https://github.com/hashicorp/dev-portal/pull/2227)

## 🤷 Why

- Terraform is the only product that had versions passed in a ASC order. This meant that the default version was `1.0.11` when users landed on `/terraform/install` because it was the first value.

## 🧪 Testing
- navigate to [/terraform/install](https://dev-portal-git-heat-fixterraform-versions-hashicorp.vercel.app/terraform/install)
- The Version dropdown shows `1.6.3 (latest)`

- navigate to [/vault/install](https://dev-portal-git-heat-fixterraform-versions-hashicorp.vercel.app/vault/install)
- The Version dropdown shows `1.15.2 (latest)`


src/lib/filter-versions.ts
```diff
@@ -0,0 +1,61 @@
...
+		} else {
+			tree[x][y] = Math.max(tree[x][y], z)
+		}
+	})
+
+	// Turn the reduced tree of latest patches only into an array
+	const latestPatchesOnly = []
+	Object.entries(tree).forEach(([x, xObj]) => {
+		Object.entries(xObj).forEach(([y, z]) => {
+			latestPatchesOnly.unshift(`${x}.${y}.${z}`)
```
heatlikeheatwave, line 49: Only change to this function is `unshift` here. Originally, it was `push`, which is why the versions were passed in the wrong order

src/lib/filter-versions.ts
```diff
@@ -0,0 +1,61 @@
...
+import { ReleasesAPIResponse } from 'lib/fetch-release-data'
+
+/**
+ * Pulled from terraform-website/pages/downloads/index.jsx on 2022-03-09:
+ * https://github.com/hashicorp/terraform-website/blob/master/pages/downloads/index.jsx#L55-L98
+ *
+ * Modified 2022-03-28 to replace semverGte with semverSatisfies,
+ * as the former seemed to return pre-releases, which is not what we want.
+ */
+function filterVersions(
```
heatlikeheatwave, line 14: Moved from this file: https://github.com/hashicorp/dev-portal/pull/2228/files#diff-82193e22f46ec6f8cf4dd14abf1b88cd8d9916bee43269ccb4ab0295839e87bb

src/components/version-context-switcher/index.tsx
```diff
@@ -25,25 +25,8 @@ const VersionContextSwitcher = ({
...
-			latestVersion = versionValue
-			break
-		}
-	}
-
-	const [selectedVersion, setSelectedVersion] = useState<
-		ContextSwitcherOption['value']
-	>(initialValue || latestVersion)
+	const [selectedVersion, setSelectedVersion] =
+		useState<ContextSwitcherOption['value']>(initialValue)
```
heatlikeheatwave, line 29: Removed this code as it was a patch fix: https://github.com/hashicorp/dev-portal/pull/2227

src/lib/filter-versions.ts
```diff
@@ -0,0 +1,60 @@
+import semverSatisfies from 'semver/functions/satisfies'
```
dstaley, line 1: I think this will need the license header otherwise a bot is gonna come in here and yell at us.

dstaley (approved):
Looks great! I love that you added a test.

heatlikeheatwave:
> Looks great! I love that you added a test.

Thank you! I always add at least one test whenever I fix a bug to ensure it doesn't happen again (:

PR #2242, opened 2023-11-21
HVD: Load preview of content in the Guide Detail View

RubenSandwich:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-rnhvd-guide-fetching-hashicorp.vercel.app/validated-designs) 🔎
- [Figma Designs](https://www.figma.com/file/s5mltxHQiTZ45PuqQ6b6JO/HVD-Phase-1?type=design&node-id=13-1210&mode=design&t=44oVeXpR7LdLnRer-0) 🎨 

Asana tickets
- [Scaffold guide template data fetching](https://app.asana.com/0/1204908934882872/1205942535702199/f) 🎟️
- [[guide template] set up sidebar sidecar layout](https://app.asana.com/0/1204908934882872/1205942535705364/f) 🎟️
- [[guide template] set up mdx content & page heading](https://app.asana.com/0/1204908934882872/1205942535705366/f) 🎟️

## 🗒️ What

Scaffold out the the Main and Detail view of HVD and have the Detail view load the actual content

### Main HVD page
Screenshot 2023-12-05 at 2 21 56 PM

### Detail HVD page
Screenshot 2023-12-05 at 2 22 05 PM

## 🧪 Testing

- [ ] Both the Main and the Detail view of HVD should be loading
- [ ] Main should be rendering
    - [ ] list of guides from [HVD content repo](https://github.com/hashicorp/hvd-docs)
- [ ] Detail should be rendering from the [HVD content repo](https://github.com/hashicorp/hvd-docs)
    - [ ] markdown of the selected guide
    - [ ] markdown headings of the selected guide
    - [ ] list of pages of the current guide

### DIFF
--- a/src/views/validated-designs/server.ts
+++ b/src/views/validated-designs/server.ts
@@ -114,7 +114,7 @@ export function getHvdCategoryGroups(): HvdCategoryGroup[] {
 			)
 
 			// find the existing HvdCategoryGroup, because we traverse the files in order the HvdCategoryGroup YAML will always come before the HvdGuide YAML
-			const categoryGroup = HvdCategoryGroups.find(
+			const categoryGroup = hvdCategoryGroups.find(
 				(categoryGroup: HvdCategoryGroup) => {
 					return categoryGroup.slug === slug
 				}
@@ -154,7 +154,7 @@ export function getHvdCategoryGroups(): HvdCategoryGroup[] {
 		}
 	})
 
-	return HvdCategoryGroups
+	return hvdCategoryGroups
 }
 
 export function getHvdCategoryGroupsPaths(): string[][] {
--- a/src/views/validated-designs/server.ts
+++ b/src/views/validated-designs/server.ts
@@ -13,6 +13,8 @@ import { HvdCategoryGroup } from './types'
 import { ValidatedDesignsGuideProps } from './guide'
 import { ValidatedDesignsLandingProps } from '.'
 
+import { serialize } from 'next-mdx-remote/serialize'
+
 const basePath = '/validated-designs'
 
 function loadMetadata(path: string): { title: string; description: string } {
@@ -172,16 +174,33 @@ export function getHvdCategoryGroupsPaths(): string[][] {
 	return paths
 }
 
-export function getHvdGuidePropsFromSlugs(
+// @TODO: calculate the actual url from the headers
+function getMarkdownHeaders(markdown: string) {
+	const headerRegex = /^(#{1,6})\s*(.+)/gm
+	let match
+	const headers = []
+
+	while ((match = headerRegex.exec(markdown)) !== null) {
+		headers.push({
+			title: match[2].trim(),
+			url: '#test',
+		})
+	}
+
+	return headers
+}
+
+export async function getHvdGuidePropsFromSlugs(
 	slugs: string[]
-): ValidatedDesignsGuideProps {
+): Promise<ValidatedDesignsGuideProps> {
 	const categoryGroups = getHvdCategoryGroups()
 	const [guideSlug, pageSlug] = slugs
 
 	const validatedDesignsGuideProps = {
 		title: '',
-		filePath: '',
+		mdxSource: null,
 		categorySlug: '',
+		headers: [],
 		currentPageIndex: 0,
 		basePath,
 		pages: [],
@@ -199,7 +218,14 @@ export function getHvdGuidePropsFromSlugs(
 					// If no pageSlug is provided, default to the first page
 					if (page.slug === pageSlug || (!pageSlug && index === 0)) {
 						validatedDesignsGuideProps.title = page.title
-						validatedDesignsGuideProps.filePath = page.filePath
+						const content = fs.readFileSync(page.filePath, 'utf8')
+
+						const headers = getMarkdownHeaders(content)
+						// @TODO: this does not add IDs to the headers, figure out why
+						const mdxSource = await serialize(content)
+
+						validatedDesignsGuideProps.headers = headers
+						validatedDesignsGuideProps.mdxSource = mdxSource
 						validatedDesignsGuideProps.currentPageIndex = index
 					}
 
--- a/src/views/validated-designs/server.ts
+++ b/src/views/validated-designs/server.ts
@@ -80,7 +80,7 @@ export function getHvdCategoryGroups(): HvdCategoryGroup[] {
 	 *
 	 * Note, is a spike and the code could be cleaned up / reformatted as per the needs of the view!
 	 */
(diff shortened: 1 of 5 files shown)


src/views/validated-designs/server.ts
```diff
@@ -75,56 +80,176 @@ export function getHvdLandingProps(): ValidatedDesignsLandingProps | null {
 	 *
 	 * Note, is a spike and the code could be cleaned up / reformatted as per the needs of the view!
 	 */
+	const HvdCategoryGroups: HvdCategoryGroup[] = []
```
heatlikeheatwave, line 83:
```suggestion
	const hvdCategoryGroups: HvdCategoryGroup[] = []
```

src/views/validated-designs/server.ts
```diff
@@ -75,56 +80,176 @@ export function getHvdLandingProps(): ValidatedDesignsLandingProps | null {
...
+
+		const slug = `${product}-${category}`
 		if (isCategoryMetadata) {
 			const { title, description } = loadMetadata(
 				path.join(HVD_CONTENT_DIR, item)
 			)
-			categoryGroups[categorySlug] = {
-				slug: categorySlug,
+
+			HvdCategoryGroups.push({
```
heatlikeheatwave, line 104:
```suggestion
			hvdCategoryGroups.push({
```

heatlikeheatwave (approved):

src/views/validated-designs/server.ts
```diff
@@ -75,56 +80,176 @@ export function getHvdLandingProps(): ValidatedDesignsLandingProps | null {
...
+	const categoryGroups = getHvdCategoryGroups()
+	// [[guide-slug], [guide-slug, page-slug]]
+	// e.g. [[terraform-operation-guide-adoption], [terraform-operation-guide-adoption, page-slug]]
+	const paths = []
+	for (const categoryGroup of categoryGroups) {
+		for (const guide of categoryGroup.guides) {
+			paths.push([guide.slug])
+			for (const page of guide.pages) {
+				paths.push([guide.slug, page.slug])
+			}
```
heatlikeheatwave, line 170: What do you think about refactoring this to use flatMap?

Not sure if this would be the exact syntax, but something like this:
```
return categoryGroups.flatMap(categoryGroup =>
  categoryGroup.guides.flatMap(guide => [
    [guide.slug],
     ...guide.pages?.map(page => [guide.slug, page.slug])
  ])
 ```

File: src/views/validated-designs/server.ts [hunk 4]
```diff
@@ -75,56 +80,176 @@ export function getHvdLandingProps(): ValidatedDesignsLandingProps | null {
...
+	const headerRegex = /^(#{1,6})\s*(.+)/gm
+	let match
+	const headers = []
+
+	while ((match = headerRegex.exec(markdown)) !== null) {
+		headers.push({
+			title: match[2].trim(),
+			url: '#test',
+		})
+	}
```
heatlikeheatwave, line 188: I think using a `while` loop in production could open up the possibility errors. For Regex.exec(), MDN suggests replacing it with [String.prototype.matchAll()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll) to make it less error-prone. What do you think?

src/views/validated-designs/server.ts
RubenSandwich: Ohh good point and much cleaner. Thanks @heatlikeheatwave! The only change is that the inner flatmap would need to be a map, other [guide.slug] would be removed from it's array and the structure I need for next.js static paths is:

```js
[
  [slug],
  [slug_part_1, slug_part_2]
]
```

src/views/validated-designs/server.ts
RubenSandwich: Honestly this is a hack till I can figure out how the rest of our code extracts and adds data to headers in markdown. (Which is how we get internal links working on markdown headers.) Late yesterday I think I found a thread I wanted to chase down here: https://github.com/hashicorp/dev-portal/blob/b59f2dbe6b0da9dc1fb98b1bf126cb9182ac2155/src/lib/serialize-integration-markdown.ts#L28.

src/views/validated-designs/server.ts
heatlikeheatwave: Ah yes, great point! Scratch the second flatMap then. Maybe something like this?
```
  return categoryGroups.flatMap(categoryGroup =>
    categoryGroup.map(({guides}, idx) => [
      [guides[idx].slug],
      ...guides[idx].pages?.map(page => [guides[idx].slug, page.slug])
    ])
  )
}
 ```

File: src/views/validated-designs/server.ts [hunk 5]
```diff
@@ -27,7 +33,7 @@ function loadMetadata(path: string): { title: string; description: string } {
 	}
 }
 
-export function getHvdLandingProps(): ValidatedDesignsLandingProps | null {
+export function getHvdCategoryGroups(): HvdCategoryGroup[] {
```
heatlikeheatwave, line 36: What do you think about breaking this function in two? One function is the code in the `if` block and the other is the code in the `else if` block.

heatlikeheatwave (commented):
Could you add tests or sample inputs/outputs in a code comment for these functions? Refactoring the code would be a lot easier that way.

src/views/validated-designs/server.ts
RubenSandwich: I am a bit indifferent about breaking it up, but I don't think breaking it up would clean up the code or how this function is used at all. Because ultimately separating out the "unpure" parts, the parts that could fail, from the "pure" parts wouldn't do much because both functions would always have to be called together anyway. Here is why.

I designed this function to both load the data and format it, because I need that data in 3 different places for Static Site Generation and to support Incremental Static Regeneration. This function recreates the data each time because when the HVD Docs repo is updated we will be firing a GitHub action that marks paths in our HVD docs as stale and will redownload the repo and rebuild those paths. (As opposed to requiring a new dev-portal build each time they are updated.)

The 3 places are below. These are wrappers around this function modify it's data slightly to fit the different required data needed:
* [guides list props build](https://github.com/hashicorp/dev-portal/blob/7d23744ff4b67d70d0817a5c4ada23f480a8d885/src/pages/validated-designs/index.tsx#L17)
* [guide detail pages static paths build](https://github.com/hashicorp/dev-portal/blob/7d23744ff4b67d70d0817a5c4ada23f480a8d885/src/pages/validated-designs/%5B...slug%5D/index.tsx#L16)
* [guide detail static props build](https://github.com/hashicorp/dev-portal/blob/7d23744ff4b67d70d0817a5c4ada23f480a8d885/src/pages/validated-designs/%5B...slug%5D/index.tsx#L39)

PR #2244, opened 2023-11-22
chore(deps): upgrade to next v14

dstaley:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-dsnext-13-5-6-hashicorp.vercel.app/) 🔎

## 🗒️ What

This PR updates Dev Portal to Next.js v14, updates TypeScript, updates `react-docs-page`, and removes Swingset.

## 🤷 Why

There have been a ton of changes to Next.js since the v13.0 release (which is what Dev Portal is currently using). One of the big changes was to middleware, which now support returning responses. This will enable us to build more effective rate-limiting. Also, dev server performance was significantly improved in the v13.5 series. Finally, App Router was made stable in v14, which is something we will want to explore adopting incrementally.

## 🛠️ How

1. Removes Swingset. Swingset has hard dependencies on Next v13, and instead of attempting to update it to support Next v14, this PR simply removes it. In the future, we'll be adding support for Storybook.
2. Updates TypeScript to pull in new types for the global `Response` used in Middleware. This is because recent versions of Next.js use the standard platform `Response` class instead of `NextResponse`, and the `.json()` method wasn't added until TypeScript 5.x.
3. Updates `react-docs-page` to fix an always false logical comparison caught by newer versions of TypeScript.
4. Updates Next.js to v14.

## 🧪 Testing

Ran visual regression testing on the top 100 pages, with 98 passing. The last 2 I believe are failing due to only content changes.

You can double check the results: [playwright-report.zip](https://github.com/hashicorp/dev-portal/files/13834857/playwright-report.zip).


next.config.js
```diff
@@ -61,90 +60,84 @@ const hideWaypointTipContent = {
...
+		'unist-util-visit',
+	],
+	webpack(config) {
+		config.plugins.push(HashiConfigPlugin())
+		return config
+	},
+	async headers() {
+		return [temporary_hideDocsPaths, hideWaypointTipContent]
+	},
+	async redirects() {
```
RubenSandwich, line 85: @dstaley when you have the time can you walk me through the redirection changes here? I'm a bit too new to completely understand these changes and would love to talk through them to make sure that we are not changing them in any meaningful way.

package.json
```diff
@@ -92,7 +92,7 @@
 		"rimraf": "^3.0.2",
 		"simple-git-hooks": "^2.7.0",
 		"stylelint": "^13.8.0",
-		"typescript": "^4.5.5",
+		"typescript": "^5.3.2",
```
RubenSandwich, line 95: snuck a typescript update in here too. nice 🎉

package.json
```diff
@@ -117,7 +117,7 @@
 		"@hashicorp/react-button": "^6.3.1",
 		"@hashicorp/react-code-block": "^6.4.0",
 		"@hashicorp/react-consent-manager": "^9.2.0",
-		"@hashicorp/react-docs-page": "^17.9.0",
+		"@hashicorp/react-docs-page": "^17.9.1-canary-20231122200025",
```
RubenSandwich, line 120: Are we comfortable putting this on a canary?

src/pages/[productSlug]/tutorials/[...tutorialSlug]/index.tsx
```diff
@@ -81,7 +81,16 @@ async function getStaticProps({
 	const { productSlug, tutorialSlug } = params
 
 	const productData = cachedGetProductData(productSlug) as LearnProductData
+	console.log(`calling getTutorialPageProps with`, {
+		productSlug,
+		tutorialSlug,
+	})
 	const props = await getTutorialPageProps(productData, tutorialSlug)
+	console.log(
```
Comment (RubenSandwich) [inline src/pages/[productSlug]/tutorials/[...tutorialSlug]/index.tsx:89]: Let's remove these console.logs

next.config.js
dstaley: Nothing about the redirects/rewrites has changed here. GitHub is just confused since the indentation has changed by removing the wrapping `withSwingset` call.

tsconfig.json
```diff
@@ -1,8 +1,20 @@
...
+		"lib": ["dom", "dom.iterable", "esnext"],
+		"jsx": "preserve",
+		"allowJs": true,
+		"esModuleInterop": true,
+		"forceConsistentCasingInFileNames": true,
+		"isolatedModules": true,
+		"moduleResolution": "node",
+		"resolveJsonModule": true,
+		"skipLibCheck": true,
+		"noEmit": true
```
RubenSandwich, line 17: nice to be explicit here on our setting versus just taking them from another project.

RubenSandwich: @dstaley mind going all the way and moving over the platform configs for Jest too from `@hashicorp/platform-configs`? That way we can fully remove the dependency `@hashicorp/platform-configs`. (No is an okay answer here, we can do that work in another PR.)

dstaley: Let's save that for a future PR as I'd like to evaluate the need for that package across all of our codebases and come up with an approach that works for both `dev-portal` and `web`!

RubenSandwich (approved):
Poked around and checked the code and everything looks good. 👍🏻

pbortnick (approved):

PR #2260, opened 2023-12-05
Harden SSG for when HVD docs do not exist

RubenSandwich:
Bug Fix for the issue of [breaking content preview/vercel builds for other teams.](https://hashicorp.slack.com/archives/C058H0MULTE/p1701812997016709)

## 🗒️ What

Due to HVD docs being built during Static Site Generation, i.e. build time, we expected the HVD docs directory to exist. If the provided env variable GITHUB_TOKEN does not have access to the repo then the directory cannot be built. This code makes it so that any HVD pages will return 404 if the HVD docs directory cannot be found.

## 🧪 Testing

Locally edit [HVD_CONTENT_DIR](https://github.com/hashicorp/dev-portal/blob/b59f2dbe6b0da9dc1fb98b1bf126cb9182ac2155/src/views/validated-designs/server.ts#L41) to be something nonsensical, like '/test' and then build. HVD pages shouldn't cause a build to fail, but rather return a 404. e.g.
* /validated-designs
* /validated-designs/terraform-operation-guides-adoption

### DIFF
--- a/src/pages/validated-designs/[...slug]/index.tsx
+++ b/src/pages/validated-designs/[...slug]/index.tsx
@@ -15,6 +15,13 @@ import {
 export async function getStaticPaths() {
 	const pagePaths = getHvdCategoryGroupsPaths()
 
+	if (!pagePaths) {
+		return {
+			paths: [],
+			fallback: false,
+		}
+	}
+
 	return {
 		paths: [
 			...pagePaths.map((path: string[]) => ({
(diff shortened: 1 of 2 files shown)


mwickett (approved):
I tested this locally with a nonsensical path and got a clean build, so approving based on that given it's meant to be a fix for failing builds.

nandereck (approved):
LGTM!

PR #2261, opened 2023-12-06
Making it easy for a user to copy the API URL in API docs

heatlikeheatwave:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-heat-choreselect-api-url-hashicorp.vercel.app/boundary/api-docs/auth-method-service) 🔎
- [Asana task](https://app.asana.com/0/1202097197789424/1204886638083759/f) 🎟️

### Demo Video
https://github.com/hashicorp/dev-portal/assets/55773810/63cb27f5-d31e-4347-9a96-b41be7b29602

## 🗒️ What

- A lil CSS and a splash of JS so users can copy API URL

## 🤷 Why

- Small request to improve UX

## 📸 Design Screenshots

N/A

## 🧪 Testing

- Visit the [Preview link](https://dev-portal-git-heat-choreselect-api-url-hashicorp.vercel.app/boundary/api-docs/auth-method-service) and copy the text in the screenshot below
Screenshot 2023-12-05 at 7 48 12 PM
- The accordion should not open. The cursor should be `text` and you should be able to easily copy the URL.

- Visit the [production page](https://developer.hashicorp.com/boundary/api-docs/auth-method-service) and try to copy the text shown in the screenshot above. It will not work, and the accordion will open.


EnMod (approved):
LGTM! 

I did notice that if you overshoot your text selection by even a pixel the accordion opens, but this makes sense because event propagation is only stopped on the `<span>` with class `.endpoint`.

heatlikeheatwave:
> LGTM!
> 
> I did notice that if you overshoot your text selection by even a pixel the accordion opens, but this makes sense because event propagation is only stopped on the `<span>` with class `.endpoint`.

Good catch! I made an adjustment for that

PR #2262, opened 2023-12-06
wip: convert css magic numbers to --hdsplus-spacing 

RubenSandwich:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-rncodemod-css-nums-to-vars-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1200392284811676/1206103199324366/f) 🎟️

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,7 +117,7 @@
 		"@hashicorp/react-button": "^6.3.1",
 		"@hashicorp/react-code-block": "^6.4.0",
 		"@hashicorp/react-consent-manager": "^9.2.0",
-		"@hashicorp/react-design-system-tokens": "^0.0.0-canary.99e8874",
+		"@hashicorp/react-design-system-tokens": "^0.0.0-nightly.d49b06f",
 		"@hashicorp/react-docs-page": "^17.9.0",
 		"@hashicorp/react-docs-sidenav": "^9.1.0",
 		"@hashicorp/react-enterprise-alert": "^7.0.1",
--- a/package.json
+++ b/package.json
@@ -117,6 +117,7 @@
 		"@hashicorp/react-button": "^6.3.1",
 		"@hashicorp/react-code-block": "^6.4.0",
 		"@hashicorp/react-consent-manager": "^9.2.0",
+		"@hashicorp/react-design-system-tokens": "^0.0.0-canary.99e8874",
 		"@hashicorp/react-docs-page": "^17.9.0",
 		"@hashicorp/react-docs-sidenav": "^9.1.0",
 		"@hashicorp/react-enterprise-alert": "^7.0.1",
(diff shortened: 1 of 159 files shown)


package.json
```diff
@@ -117,6 +117,7 @@
 		"@hashicorp/react-button": "^6.3.1",
 		"@hashicorp/react-code-block": "^6.4.0",
 		"@hashicorp/react-consent-manager": "^9.2.0",
+		"@hashicorp/react-design-system-tokens": "^0.0.0-canary.99e8874",
```
dstaley, line 120: Can yup update to [0.0.0-nightly.d49b06f](https://www.npmjs.com/package/@hashicorp/react-design-system-tokens/v/0.0.0-nightly.d49b06f)? The npm version is a bit of a fib since we've never published a stable version, so the "latest" is just the first published version.

RubenSandwich:
> What

dstaley:
> Thanks for your work on this! Truthfully, I am finding the code less readable because of the abstraction. Personally, I think it would be more useful to have spacing variables with the size in the name, e.g. `--token-spacing-4`.
> 
> What do you think about adding more clearly named spacing variables to `:root` in the global styles rather than importing this package? I'm not sure the use case requires adding a package. Thoughts?

@heatlikeheatwave These tokens are our official tokens, and are what we use for all our other sites, in addition to our designs in Figma. I definitely agree that it's confusing at first, seeing `--hdsplus-spacing-05` and having no concept of _what_ that means, but it does get easier as time goes on. That being said, we can absolutely discuss as a team whether to rename the tokens across all our codebases and figma designs!

RubenSandwich:
@dstaley and @heatlikeheatwave I agree with both of you and think I have a bit of a compromise that I just enabled.

While I am in favor of the added abstraction, we don't ultimately want our css variables tied to specific values because the reason why we are using variables is for the possibility of changing those values in the future IMO. But I agree with Heat that our current naming scheme makes it difficult to understand the amount of spacing we are applying. I have seen, and worked in, design systems that take a bit of a different naming approach to their spacing variables:
```
--spacing-xs
--spacing-s
--spacing-m
--spacing-l
--spacing-xl

// with some edge cases of
--spacing-card-s
```
etc.

I would be willing to see what we can internally rename the variables in our current spacing system to be more usable:
`--spacing-xs: var(--hdsplus-spacing-08);`

Till then I noticed that I forgot to add CSS autocomplete to these new tokens, and now enabled it in this PR so that hoving over these variables will display it's value:
Screenshot 2023-12-06 at 2 38 08 PM

RubenSandwich:
@heatlikeheatwave and @dstaley wrote a quick and hacky stylelint rule to catch our usage of magic nums for when a css variable exists for it:
Screenshot 2023-12-06 at 3 59 12 PM


I ran out of time today, but if I have more time this week I'll get autofix working as well.

kaitlynnefuery:
> @heatlikeheatwave and @dstaley wrote a quick and hacky stylelint rule to catch our usage of magic nums for when a css variable exists for it: Screenshot 2023-12-06 at 3 59 12 PM
> 
> I ran out of time today, but if I have more time this week I'll get autofix working as well.

@RubenSandwich I'm a fan of this!

PR #2271, opened 2023-12-08
chore: set up sentinel launch redirects

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Adds redirects that divert traffic from `docs.hashicorp.com` to `developer.hashicorp.com`, but only once `proxySettings.sentinel` have been removed.

## 🤷 Why

As soon as we launch Sentinel content on `developer.hashicorp.com`, we'll want to redirect traffic from the former `docs.hashicorp.com` domain. Since both domains are currently managed in the same Vercel project, it seems logical to put these redirects behind some sort of flag. We can then toggle the flag in order to complete the launch.

Our usual `config` flags don't seem suited to this use case, and we already have a kind of "flag" that captures proxied domain configuration in our `proxySettings`. So, this PR uses `proxySettings.sentinel` as the launch flag. To "toggle" it to launch, we'll remove `proxySettings.sentinel`.

## 🧪 Testing

For the purposes of testing this branch, we have some **temporary** measures in place:
- `proxySettings.sentinel` has been commented out, to simulate launch
- The `host` condition for the redirects we want to apply captures _both_ `docs.hashicorp.com` and an additional custom domain `sentinel-launch-redirects-test.hashicorp.vercel.app`. The former is what we actually need to care about for launch, but isn't feasible to mess around and test with. The latter has been set up for testing purposes.

To test that the work in this PR is functioning as expected, visit the [preview], where `proxySettings.sentinel` have been **temporarily** removed:

- [ ] The [default branch preview URL][preview] should behave as if it's `developer.hashicorp.com` in a post-launch scenario, with no redirects applied.
    - For example, the [root URL][preview] should _not_ redirect to [developer.hashicorp.com/sentinel]
    - Other `docs.hashicorp.com` URLs to redirect such as [/sentinel/downloads] should _not_ redirect to [developer.hashicorp.com/sentinel/install].
- [ ] The custom domain assigned to this branch, [sentinel-launch-redirects-test.hashicorp.vercel.app][custom], should behave as if it's the `docs.hashicorp.com` domain in a post-launch scenario.
    - For example, the [root URL][custom] should redirect to [developer.hashicorp.com/sentinel]
    - Other `docs.hashicorp.com` URLs to redirect such as [/sentinel/downloads][custom/sentinel/downloads] should redirect to [developer.hashicorp.com/sentinel/install].

Once this testing is done, we'll put the `proxySettings.sentinel` back in place before merging, as we're not yet ready to launch. Once the `proxySettings.sentinel` are back in place, the redirects added in this PR will no longer apply on either test domain, nor will they apply to `docs.hashicorp.com`, which we'll confirm with a quick visit to [sentinel-launch-redirects-test.hashicorp.vercel.app][custom].

## 💭 Anything else?

This change required a subtle change to how we split "simple" and "complex" redirects.

We now treat redirects with _any_ `has` condition as "complex". This means that now, redirects with `has` conditions will be handled by built-in redirects handling, rather than through our custom middleware.

Two notes on this change:
- This change seemed necessary to run the proof-of-concept for this PR using the one-off custom `sentinel-launch-redirects-test.hashicorp.vercel.app` domain
- This change has _minimal_ impact on the split of "simple" and "complex" redirects.
    - Upstream, we have `{ simpleRedirectsCount: 475, complexRedirectsCount: 47 }`
    - In this PR, we have `{ simpleRedirectsCount: 472, complexRedirectsCount: 50 }`, or `{ simpleRedirectsCount: 472, complexRedirectsCount: 49 }` when simulating a "launch" of the Sentinel migration by removing `proxySettings.sentinel`.

[task]: https://app.asana.com/0/1204759533834554/1206141921455040/f
[preview]: https://dev-portal-git-zsset-up-sentinel-launch-redirects-hashicorp.vercel.app/
[/sentinel/downloads]: https://dev-portal-git-zsset-up-sentinel-launch-redirects-hashicorp.vercel.app/sentinel/downloads
[/sentinel/install]: https://dev-portal-git-zsset-up-sentinel-launch-redirects-hashicorp.vercel.app/sentinel/install
[custom]: https://sentinel-launch-redirects-test.hashicorp.vercel.app
[custom/intro]: https://sentinel-launch-redirects-test.hashicorp.vercel.app/intro
[custom/sentinel/downloads]: https://sentinel-launch-redirects-test.hashicorp.vercel.app/sentinel/downloads
[developer.hashicorp.com/sentinel]: https://developer.hashicorp.com/sentinel
[developer.hashicorp.com/sentinel/install]: https://developer.hashicorp.com/sentinel/install

### DIFF
--- a/build-libs/__tests__/redirects.test.ts
+++ b/build-libs/__tests__/redirects.test.ts
@@ -22,7 +22,7 @@ function withHashiEnv(value, fn) {
 
 describe('splitRedirectsByType', () => {
 	test('splits simple and glob redirects', () => {
-		const { simpleRedirects, globRedirects } = splitRedirectsByType([
+		const { simpleRedirects, complexRedirects } = splitRedirectsByType([
 			{
 				source: '/:path',
 				destination: '/',
@@ -73,19 +73,8 @@ describe('splitRedirectsByType', () => {
 				destination: '',
 				permanent: true,
 			},
-			{
-				source: '/has-host',
-				destination: '/host',
-				permanent: true,
-				has: [
-					{
-						type: 'host',
-						value: 'www.host.com',
-					},
-				],
-			},
 		])
-		expect(globRedirects).toStrictEqual([
+		expect(complexRedirects).toStrictEqual([
 			{
 				source: '/:path',
 				destination: '/',
@@ -112,6 +101,17 @@ describe('splitRedirectsByType', () => {
 					{
 						type: 'cookie',
 						key: 'cookie',
+					},
+				],
+			},
+			{
+				source: '/has-host',
+				destination: '/host',
+				permanent: true,
+				has: [
+					{
+						type: 'host',
+						value: 'www.host.com',
 					},
 				],
 			},
--- a/build-libs/docs-dot-hashicorp-redirects.js
+++ b/build-libs/docs-dot-hashicorp-redirects.js
@@ -1,0 +1,103 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+//@ts-check
+
+/** @typedef { import("next/dist/lib/load-custom-routes").Redirect } Redirect  */
+/** @typedef { import("next/dist/lib/load-custom-routes").RouteHas } RouteHas */
+
+/**
+ * Note: Proxy settings are how we determine whether the Sentinel migration
+ * has been launched. Once the Sentinel migration to Dev Dot is complete,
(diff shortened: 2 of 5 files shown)


zchsh:
@dstaley Taggin' ya here since I think you'll have the most context on the `simple` redirects split! 🙇‍♂️

dstaley (approved):
LGTM! I'm not completely sure why redirects with host conditions ended up in simpleRedirects, but it definitely makes more sense for those to be in complexRedirects as you have here.

PR #2288, opened 2024-01-03
cleanup: remove .io cache warming

dstaley:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

This PR removes the .io cache warming portion of the `cache.yml` workflow.

## 🤷 Why

The `cache.yml` workflow previously warmed URLs from the .io sites as reported by Fathom. Since those products are no longer served from Dev Portal, it makes sense to remove that from the workflow, especially in light of Fathom's recent API limit imposition.

## 🛠️ How

Removed the relevant code.

### DIFF
--- a/scripts/warm-cache.ts
+++ b/scripts/warm-cache.ts
@@ -3,10 +3,9 @@
  * SPDX-License-Identifier: MPL-2.0
  */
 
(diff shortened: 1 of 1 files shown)


pbortnick (approved):
👍

PR #2309, opened 2024-01-25
feat: block curl requests to cdktf

dstaley:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

This PR uses middleware to return a 429 response for requests to `/terraform/cdktf` from `curl`.

```
➜  dev-portal git:(ds.middleware-curl-429) curl 'https://dev-portal-git-dsmiddleware-curl-429-hashicorp.vercel.app/terraform/cdktf'
{"error":"Please reach out to support@hashicorp.com so we can learn more about your use case."}
```

## 🤷 Why

This is because there is someone sending a request to this page every two seconds.

## 🛠️ How

We check the User-Agent header and pathname, and return a 429 response if it's for `/terraform/cdktf` from a UA containing `curl`.

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -48,11 +48,20 @@ function setHappyKitCookie(
 export async function middleware(req: NextRequest, ev: NextFetchEvent) {
 	const { geo } = req
 
-	// 07/07/2023 — Simple UA check
+	// UA checks to prevent misuse
 	const { ua } = userAgent(req)
-	const regexp = /(bytespider|bytedance)/i
-	if (regexp.test(ua)) {
+	if (/(bytespider|bytedance)/i.test(ua)) {
 		return Response.json(null, { status: 404 })
+	}
+
+	if (/curl/i.test(ua) && req.nextUrl.pathname === '/terraform/cdktf') {
+		return Response.json(
+			{
+				error:
+					'Please reach out to support@hashicorp.com so we can learn more about your use case.',
+			},
+			{ status: 429 }
+		)
 	}
 	// ----------------------
 


heatlikeheatwave (approved):

src/middleware.ts
```diff
@@ -48,12 +48,21 @@ function setHappyKitCookie(
...
-	// 07/07/2023 — Simple UA check
+	// UA checks to prevent misuse
 	const { ua } = userAgent(req)
-	const regexp = /(bytespider|bytedance)/i
-	if (regexp.test(ua)) {
+	if (/(bytespider|bytedance)/i.test(ua)) {
 		return Response.json(null, { status: 404 })
 	}
+
+	if (/curl/i.test(ua) && req.nextUrl.pathname === '/terraform/cdktf') {
```
RubenSandwich, line 57: @dstaley should this be a string contains check? I assume they are hitting subpaths under /terraform/cdktf as well.

heatlikeheatwave: I wanted to chime in here with a bit of context. There have been extraordinary curl requests to `/terraform/cdktf` (one every 250ms), and this change will prevent that. You can see it [here](https://vercel.com/hashicorp/~/monitoring/01HE36BTJ0CDCA9B231RFSPNAR) and slack thread [here](https://hashicorp.slack.com/archives/C058H0MULTE/p1698760465025109)

dstaley: @RubenSandwich Oddly enough, they _aren't_, which is the completely baffling thing. I wanted to minimize the blast radius of this change as much as possible, so that's why it's only for this one specific route. If we start to see traffic across other subpaths I'm definitely down to change it though!

PR #2310, opened 2024-01-26
Removing apt-key as this method of importing the key is deprecated.

naioja:
## 🗒️ What

Changing the way repository gpg sign keys are imported in Ubuntu.

## 🤷 Why

The currently used apt-key command is deprecated and it's not considered best practices to import all keys in the trusted.db gpg keystore.

## 🛠️ How

I used the same approach as here: https://github.com/hashicorp/dev-portal/blob/main/src/views/product-downloads-view/helpers.ts#L120

### DIFF
--- a/src/data/vault.json
+++ b/src/data/vault.json
@@ -103,8 +103,8 @@
 		{
 			"label": "Ubuntu/Debian",
 			"commands": [
-				"curl -fsSL https://apt.releases.hashicorp.com/gpg | sudo apt-key add -",
-				"sudo apt-add-repository \"deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main\"",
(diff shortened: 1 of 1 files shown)


jonathanfrappier (approved):
Tested against a clean jammy64 box. Steps work as provided.

heatlikeheatwave (approved):

RubenSandwich (approved):
Works and builds correctly. 👍🏻 

@naioja thanks for opening up this PR. Apologizes on the delays, you caught us just before an offsite.

RubenSandwich:
@heatlikeheatwave just the analyze build step worked just fine when deploying: https://github.com/hashicorp/dev-portal/actions/runs/7746504127/job/21124853757. 🤷🏻‍♂️

PR #2327, opened 2024-02-09
chore(sentinel migration): remove proxy settings & proxy config

pbortnick:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1204759533834554/1206183781878379/f) 🎟️

## 🗒️ What

This PR removes `proxySettings` and `proxyConfig` as well as any code dependent on those definitions.

## 🤷 Why

Sentinel was the final proxied item. Now that it is migrated, we no longer need these proxy settings.

## 🛠️ How

Thoroughly console logged within the affected areas to make sure the deleted items weren't actually controlling anything.


src/lib/env-checks.js
```diff
@@ -5,86 +5,10 @@
...
-	}, false)
-	return proxiedProductSlug
-}
-
-/**
- *
- * @param {string=} hostname
- * @returns {string | boolean | undefined}
- */
-function getMatchedDomain(hostname) {
```
pbortnick, line 44: This was only used in `isProxiedProduct()` which is now 🔪

src/lib/env-checks.js
```diff
@@ -5,86 +5,10 @@
...
-	return domainProductSlug
-}
-
-/**
- *
- * @param {string} productSlug
- * @param {string=} hostname
- * @returns {boolean}
- */
-function isProxiedProduct(productSlug, hostname) {
```
pbortnick, line 64: No longer need this check

build-libs/redirects.js
```diff
@@ -32,122 +23,12 @@ require('isomorphic-unfetch')
...
 	'docs.hashicorp.com': 'sentinel',
 	'test-st.hashi-mktg.com': 'sentinel',
 }
 
-// Redirect all proxied product pages
-// to the appropriate product domain
-//
-// Note: we do this for ALL domains, as we never want visitors to
-// see the original "proxied" routes, no matter what domain they're on.
-const productsToProxy = Object.keys(proxySettings)
```
pbortnick, line 48: Always an empty array since `proxySettings` = `{}`

dstaley (approved):

PR #2331, opened 2024-02-12
Clean up duplicated head tags

RubenSandwich:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-rn-chorefix-duplicated-meta-tags-hashicorp.vercel.app/vault/docs/auth/jwt/oidc-providers/adfs) 🔎
- [Asana task](https://app.asana.com/0/1205890531436279/1206594821752310/f) 🎟️

## 🗒️ What

This PR cleans up the usage of meta tags, mostly by removing the extra usage of the component <HashiHead> when it really isn't needed. It doesn't completely solve the duplicated meta tags problems, as the meta properties 'og:locale' and 'og:type' still seem to be rendering twice.

I went deep into the rabbit hole of what was happening in 'og:locale' and 'og:type', [and my best guess is because we are using property not name](https://github.com/hashicorp/react-components/blob/18198c8006c1b52e3d8ed2be1cf042e0b76b8eff/packages/head/index.tsx#L42), [though both are equal in the html spec from my understanding](https://stackoverflow.com/questions/20401085/is-it-possible-to-use-the-same-meta-tag-for-opengraph-and-schema-org/20429551#20429551). [The reason why this is a problem is because in next.js's head rendering code they check for collisions by using the name property.](https://github.com/zbialecki/next.js/blob/bf5256221f296d2196bd0615e82019318b062045/packages/next/next-server/lib/head.tsx#L87) Buuutttttt, even adding a name to those meta tags, and forcing a complete rebuild of the local next.js site, did not remove the duplicated rendering of those meta tags. 🤷🏻‍♂️ 

But because this PR cleans up quite a bit of duplicated or unneeded code, even if it does not completely fix the issue yet, I believe that it's still worth shipping.

## 🧪 Testing

This seems to happen on the transition from SSR to SPA. (So when you click on a link when a page is already opened.)

**How to test requires comparing and contrasting Prod to Preview:**
- [Prod Configure Vault with ADFS for OIDC](https://developer.hashicorp.com/vault/docs/auth/jwt/oidc-providers/adfs)
  - Click on the sidebar link "Use Cases"
  - Paste the duplicated checking meta tag code in the console from below
  - Notice the duplicated tags

- [Preview Configure Vault with ADFS for OIDC](https://dev-portal-git-rn-chorefix-duplicated-meta-tags-hashicorp.vercel.app/vault/docs/auth/jwt/oidc-providers/adfs)
  - Click on the sidebar link "Use Cases"
  - Paste the duplicated checking meta tag code in the console from below
  - Notice the duplicated tags
    - They should be less, but still usually: `og:locale`, `og:type`, and `article:publisher`.

The following function can be pasted into the DevTools console to check for duplicated meta tags:
```js
(function(){
    const metaTags = document.getElementsByTagName('meta');
    const metaContent = {};
    let duplicates = [];

    for (let i = 0; i < metaTags.length; i++) {
        const name = metaTags[i].name || metaTags[i].getAttribute('property');
        if (name && name !== "zd-site-verification") {
            if (metaContent[name]) {
                duplicates.push(name);
            } else {
                metaContent[name] = true;
            }
        }
    }

    console.log(duplicates);
})();
```


src/pages/_document.tsx
```diff
@@ -16,9 +15,7 @@ export default class MyDocument extends Document {
 	render() {
 		return (
 			<Html lang="en">
-				<Head>
-					<HashiHead />
-				</Head>
+				<Head></Head>
```
RubenSandwich, line 18: `next/document` head is not the same as `next/head`, [`next/document` is used for properties that does not change across all pages.](https://nextjs.org/docs/pages/building-your-application/routing/custom-document#caveats)

[We are already rendering HashiHead in our _app file.](https://github.com/hashicorp/dev-portal/blob/main/src/pages/_app.tsx#L103) So this one is unnecessary.

src/components/head-metadata/index.tsx
```diff
@@ -65,88 +65,86 @@ export default function HeadMetadata(props: HeadMetadataProps) {
 
 	return (
 		// TODO: OpenGraph image to be passed as the image prop here
-		<>
```
RubenSandwich, line 68: Removed extra nesting.

dstaley (approved):

PR #2348, opened 2024-02-22
feat: move next-mdx-remote inline

dstaley:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

This PR copies the code for `next-mdx-remote` v3.0.8 into `dev-portal`, and replaces references to `next-mdx-remote` with `lib/next-mdx-remote`.

## 🤷 Why

`dev-portal` currently uses `next-mdx-remote` v3.0.8, which supports MDX v1. `next-mdx-remote` v4 was updated to MDX v2, which is a significant breaking change incompatible with our existing collection of MDX documentation. We cannot currently, nor in the near future, upgrade to `next-mdx-remote` v4 due to the need to update our existing MDX documentation files to be compatible with MDX v2. This process will likely take several months. There is a possible future where a change to either React or Next.js breaks compatibility with `next-mdx-remote` v3, so instead of trying to maintain a public version of v3 via canary releases from the `next-mdx-remote` repository, this PR essentially "forks" the package. This will enable us to more easily make changes to the package to support our specific use case of MDX v1 with React v18+ (which is _technically_ not supported by MDX v1). The general goal will be to remain largely compatible with the API of `next-mdx-remote` v3 so that once our MDX v2 migration is complete we can upgrade from our in-tree version to the public version of `next-mdx-remote` v4+.

## 🛠️ How

The codebase for the v3.0.8 version of `next-mdx-remote` was copied into the `lib` folder (including tests). Then, references to `next-mdx-remote` were replaced with `lib/next-mdx-remote`.

## 🧪 Testing

This PR _should_ produce the exact same output as production. To verify, spot check that a few docs pages render the same in the preview as in production.


pbortnick (approved):
👍

PR #2352, opened 2024-02-23
Update HVD preview on main update

RubenSandwich:
## 🔗 Relevant links

- Preview link 🔎
- Asana task 🎟️

**This is a dev chore that only effects a currently existing github action, so no preview link or asana ticket.**

## 🗒️ What

Adds HVD to the content redeploy preview action. It seems to me that the purpose of this action is that if someone looks at the preview on main of one of the docs repos, it should reflect the latest version of dev-portal.

Pretty low use case IMO, but it's important to match them if they exist. 🤷🏻 

P.S. Having a document type that reads file differently depending on spacing is and always was a bad idea IMO. 🍵 

## 🧪 Testing

Because this is an addition to an already existing pipeline, I don't feel that there is the needs to go through the same testing procedure as the initial deployment. Rather just the addition can be tested by running the command locally:
```sh
curl -X POST ${{ secrets.VERCEL_DEPLOY_HOOK_HVD }}
```
You can get the value of secrets.VERCEL_DEPLOY_HOOK_HVD from [hvd-docs vercel project git settings](https://vercel.com/hashicorp/hvd-docs/settings/git), under the **Deploy Hooks** section with the name **Re-deploy when dev-portal/main updates on main**.

Once that request is made, [the hvd-docs Production deployment should rebuild.](https://vercel.com/hashicorp/hvd-docs/deployments)


heatlikeheatwave (commented):
👋 In [the PR](https://github.com/hashicorp/dev-portal/pull/1526) that added the other redeploy triggers they included how they tested this change. Could you add something to your PR comment with that info?

RubenSandwich:
> 👋 In [the PR](https://github.com/hashicorp/dev-portal/pull/1526) that added the other redeploy triggers they included how they tested this change. Could you add something to your PR comment with that info?

Sounds good. Added testing instructions for this addition to the github action.

heatlikeheatwave (approved):

PR #2371, opened 2024-03-11
build-libs: add redirect for ipsw packer plugin

lbajolet-hashicorp:
## 🔗 Relevant links

Related issue: https://github.com/hashicorp/packer/issues/12873

## 🗒️ What

The IPSW plugin was recently moved to integrations, and since we removed the old plugin system in Packer their docs were inaccessible.

This redirect should enable old-style plugin documentation links, and redirect them to the newly annointed documentation for IPSW.


nywilken (approved):
LGTM

PR #2386, opened 2024-03-19
feat: migrate from jest to vitest

dstaley:
## 🗒️ What

This PR switches our unit tests from Jest to Vitest.

## 🤷 Why

Jest has, historically, struggled with ES modules. Currently, we run Jest twice, once for CJS modules and again for ES modules. This makes it difficult to know which environment a given test needs to execute in, given that any test that imports a module relying on an ESM-only package will need to be run in the ESM mode.

Vitest, on the other hand, relies on Vite, a popular bundler, to handle module resolution, which has no such issues. It also uses a Jest-compatible API, meaning all our tests can remain as written (with some small adjustments).

## 🛠️ How

The vast majority of changes are replacing the `jest` global with `vi`. Snapshots also changed formatting slightly, so those have also been updated. Anything else has been specifically annotated with a comment in the PR to point out why a particular change was made.

## 🧪 Testing

You can test this locally by running `npm test`.


jest.config.ts
```diff
@@ -1,115 +0,0 @@
...
-	? `<rootDir>/(?!${ESM_TEST_DIRS.join('|')}).*/.*`
-	: `<rootDir>/(${ESM_TEST_DIRS.join('|')})/.*`
-
-/**
- * Build the final config.
- *
- * Note that when `isRunningInEsmMode`, options may be overridden
- * when spreading `...esmConfig`.
- */
-const config: Config.InitialOptions = {
```
RubenSandwich, line 94: good riddance

package.json
```diff
@@ -55,28 +53,26 @@
 		"@hashicorp/platform-types": "^0.4.0",
 		"@octokit/rest": "^19.0.5",
 		"@playwright/test": "^1.28.0",
-		"@testing-library/jest-dom": "^5.16.5",
-		"@testing-library/react": "^13.3.0",
+		"@testing-library/jest-dom": "^6.4.2",
+		"@testing-library/react": "^14.2.1",
```
RubenSandwich, line 57: thanks for updating these at the same time

vitest.config.mts
```diff
@@ -0,0 +1,30 @@
...
+
+export default defineConfig({
+	plugins: [react()],
+	test: {
+		globals: true,
+		environment: 'jsdom',
+		exclude: [...configDefaults.exclude, 'src/__tests__/e2e'],
+		setupFiles: ['dotenv/config', '.test/setup-vitest.js'],
+	},
+	resolve: {
```
RubenSandwich, line 13: Let's move to using the plugin [vite-tsconfig-paths](https://github.com/aleclarson/vite-tsconfig-paths), [as recommended by the docs](https://vitest.dev/guide/common-errors#cannot-find-module-relative-path), so we don't have to worry about updating these when we add alias's.

RubenSandwich: Hmmm, let me pull down this PR and play with it for 15 mins. Maybe the second try will work. Great job on this helper function btw.

RubenSandwich: Huh, I was able to pass the same number of test files, 70, with using vite-tsconfig-paths with no options:

`plugins: [react(), tsconfigPaths()],`

I made sure that the resolve code was commented out, and [ran with no cache.](https://vitest.dev/config/#cache)

dstaley: Oh weird! I was trying to use it at the beginning of the migration, so I wonder if I was confusing another error with it not working. Thanks for getting it up and running! Much better :)

RubenSandwich: Computers 🤷🏻‍♂️

config/__tests__/index.test.js
```diff
@@ -9,21 +9,18 @@ const { loadHashiConfigForEnvironment } = require('../index')
...
-		process.chdir(fixtureDir)
 	})
 
 	afterAll(() => {
-		process.chdir(cwd)
 		process.env.HASHI_ENV = undefined
 	})
 
 	test('loads configuration and handles extending', () => {
+		vi.spyOn(process, 'cwd').mockReturnValue(fixtureDir)
```
dstaley, line 21: Instead of _actually_ changing directories (which isn't supported in vitest), we simply mock the `process.cwd()` call to return the `fixtureDir`.

config/__tests__/index.test.js
```diff
@@ -9,21 +9,18 @@ const { loadHashiConfigForEnvironment } = require('../index')
...
 		expect(loadHashiConfigForEnvironment()).toMatchInlineSnapshot(`
-      Object {
+      {
```
dstaley, line 23: These changes are to align with Vitest's snapshot format.

package.json
```diff
@@ -55,28 +53,26 @@
...
 		"cross-env": "^7.0.3",
 		"flat": "^5.0.2",
 		"identity-obj-proxy": "^3.0.0",
-		"jest": "^27.5.1",
-		"jest-raw-loader": "^1.0.1",
+		"jsdom": "^24.0.0",
 		"next-remote-watch": "^2.0.0",
 		"nextjs-bundle-analysis": "^0.5.0-canary.1",
-		"nock": "^13.2.4",
+		"nock": "^14.0.0-beta.5",
```
dstaley, line 75: A beta version of `nock` is needed to intercept native `fetch` requests. Previously in Jest I believe it was using `node-fetch`, which the production version of nock is able to intercept.

src/components/accordion-disclosure/accordion-disclosure.test.tsx
```diff
@@ -60,9 +60,9 @@ describe('AccordionDisclosure', () => {
 		// best effort to query for plain divs rendered by Disclosure
 		const buttons = queryAllByRole('button')
 
-		expect(buttons[0].parentElement).not.toHaveClass('isNested')
-		expect(buttons[1].parentElement).toHaveClass('isNested')
-		expect(buttons[2].parentElement).toHaveClass('isNested')
+		expect(buttons[0].parentElement).not.toHaveClass(/isNested/)
```
dstaley, line 63: These `toHaveClass` changes were needed because Vitest correctly appends hashes to class names, whereas the Jest CSS module mock did not.

src/components/tutorial-card/tutorial-card.tsx
```diff
@@ -0,0 +1,76 @@
+/**
```
dstaley, line 1: This was moved to its own file to prevent a circular import.

src/data/product-data-map.ts
```diff
@@ -0,0 +1,29 @@
+/**
```
dstaley, line 1: This was created to eliminate dynamic `require` calls in test files.

src/lib/remark-plugins/rehype-sanitize/__tests__/rehype-sanitize.test.ts
```diff
@@ -136,13 +136,15 @@ describe('rehypeSanitize', () => {
...
 		// Assert that the AST we'll render does not have any JSX nodes
 		const paragraphNode = root.children[0]
-		if (!('children' in paragraphNode)) {
-			fail('Paragraph node is expected to have a link element as a child.')
-		}
-		const linkNode = paragraphNode.children[0]
-		if (!('properties' in linkNode)) {
-			fail('Link node is expected to have properties.')
-		}
+		expect(
```
dstaley, line 139: These were refactored to be actual `expect` calls.

nandereck (approved):
LGTM! 🎉

PR #2388, opened 2024-03-19
feat: replace esbuild with @swc/core

dstaley:
## 🗒️ What

This PR replaces our usage of esbuild to transpile compiled MDX to executable JavaScript with `@swc/core`.

## 🤷 Why

esbuild relies on executing the esbuild binary, which required special code to locate correctly. `@swc/core` requires no special code to load correctly in dev and production. Also, due to how esbuild was loaded, the Next.js dev server would print warnings about not being able to statically analyze the imports; this is not an issue with `@swc/core`. Finally, in certain cases, `@swc/core` is faster than esbuild.

## 🛠️ How

Replace the `esbuild.transform` call with `swc.transform`.

## 🧪 Testing

To test this change, click through a handful of documentation pages, comparing them to production. For the most part it should either work completely or be completely broken; I can't think of an instance where the differences would be subtle (famous last words 😅).


package-lock.json
```diff
@@ -10407,6 +10407,198 @@
...
+			"resolved": "https://registry.npmjs.org/@swc/core/-/core-1.4.8.tgz",
+			"integrity": "sha512-uY2RSJcFPgNOEg12RQZL197LZX+MunGiKxsbxmh22VfVxrOYGRvh4mPANFlrD1yb38CgmW1wI6YgIi8LkIwmWg==",
+			"hasInstallScript": true,
+			"dependencies": {
+				"@swc/counter": "^0.1.2",
+				"@swc/types": "^0.1.5"
+			},
+			"engines": {
+				"node": ">=10"
+			},
```
heatlikeheatwave, line 10421: I installed this locally on your branch and didn't have any issues, but I'm curious if the incongruity here will cause an issue with the app using `"npm": ">=8.0.0"`.

heatlikeheatwave (approved):

package-lock.json
dstaley: Great question! The `engines` defined in our `package.json` file is a subset of this `engines`, so it won't cause any issues. You can think of this one as being equivalent to:

```
"engines": {
  "node": ">=10",
  "npm": "*"
}
```

Meaning that _any_ version of npm is acceptable.

heatlikeheatwave: Thanks for explaining! Much appreciated

PR #2391, opened 2024-03-25
Harden and cleanup HVD content load/reload

RubenSandwich:
## 🔗 Relevant links

- [Asana task](url) 🎟️


package.json
```diff
@@ -8,9 +8,9 @@
 		"clean": "rimraf .next node_modules",
 		"lint": "next-hashicorp lint",
 		"lint:stylelint": "npx stylelint '**/*.css'",
-		"prebuild": "hc-tools ./scripts/generate-tutorial-variant-map.ts && hc-tools ./scripts/extract-hvd-content.ts",
```
RubenSandwich, line 11: This is not longer needed as because we wait on the hvd content promise during SSR that promise already fires during build in prod and on demand during dev.

scripts/extract-hvd-content.ts
```diff
@@ -96,7 +104,20 @@ export const HVD_FINAL_IMAGE_ROOT_DIR = '.extracted/hvd'
...
 		)
 		fs.cpSync(imageLocation, imageDestination, { recursive: true })
+
+		if (env === 'development') {
+			console.log(
+				'Loaded HVD content. Note this content will not be updated until "npm run start" is run again.'
+			)
+
+			// This is a hack to preserve some state between server restarts, as we cannot rely on JS variables persisting across restarts
+			process.env[alreadyLoadedDevEnvKey] = 'true'
```
RubenSandwich, line 114: Not a fan of this hack, but it only runs during dev so I think it's fine.

package.json
heatlikeheatwave: love this for us

scripts/extract-hvd-content.ts
```diff
@@ -46,16 +45,24 @@ export const HVD_CONTENT_DIR =
```
heatlikeheatwave, line 45: On L41:
"Otherwise, use the path generated from the `extractHvdContent` script"
->
"Otherwise, use the path generated from the `extractINGHvdContent` script

src/views/validated-designs/server.ts
```diff
@@ -221,20 +240,54 @@ export async function getHvdGuidePropsFromSlug(
...
-								remarkPlugins: [
-									[remarkPluginAnchorLinkData, { anchorLinks }],
-									rewriteStaticHVDAssetsPlugin,
-								],
-							},
-						})
+
+						let pageTitle
+						let pageDescription
+						let mdxSource
```
heatlikeheatwave, line 251: nit
```suggestion
						let pageTitle, pageDescription, mdxSource
```

heatlikeheatwave (approved):
LGTM

PR #2394, opened 2024-03-27
fix: account for releases without builds

dstaley:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

This PR updates our code that processes releases from the ofificial HashiCorp Releases API to account for versions that have no builds.

## 🤷 Why

Currently the "latest" (according to semver) release of Vault enterprise has a `builds` value set to `null`, which breaks code that expects `builds` to always be an array.

## 🛠️ How

By checking both the `version` string and `builds` property to calculate the latest version as the latest version with builds according to semver.

## 🧪 Testing

- [x] Confirm the [Vault Install Enterprise](https://dev-portal-git-dsfix-ent-builds-hashicorp.vercel.app/vault/install/enterprise) page reports 1.15.6 as the latest version

### DIFF
--- a/src/lib/fetch-release-data.ts
+++ b/src/lib/fetch-release-data.ts
@@ -27,14 +27,16 @@ export interface ReleaseVersion {
 	version: Version
 	shasums: string
 	shasums_signature: string
-	builds: {
-		name: HashiCorpProduct
-		version: Version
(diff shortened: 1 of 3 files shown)


RubenSandwich (approved):
Looks good and works.

PR #2399, opened 2024-03-28
fix: only fetch tutorials map once

dstaley:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

This PR refactors the `getStaticProps` implementation for `docs-view` to only generate/fetch the tutorials map once. In a future PR, I'm going to explore if we even need to fetch the tutorials map to rewrite URLs anymore.

## 🤷 Why

Prior to this PR, `getStaticProps` would generate/fetch the tutorials map multiple times, each taking ~2 seconds. This was because different steps of the implementation would fetch their own copy of the tutorials map independent of one another.

## 🛠️ How

In this PR, we fetch the tutorials map once at the beginning of `getStaticProps`, and pass it into the remote content loader (for link rewrites) and again into the `prepareNavData` utility.

## 🧪 Testing

- [ ] Click through a handful of docs pages and ensure that they render.
- [ ] Go to [preview](https://dev-portal-git-dstutorials-map-once-hashicorp.vercel.app/terraform/intro) and ensure the "Getting Started" link in the sidebar does not point to `learn.hashicorp.com`.

### DIFF
--- a/src/layouts/sidebar-sidecar/utils/prepare-nav-data-for-client.ts
+++ b/src/layouts/sidebar-sidecar/utils/prepare-nav-data-for-client.ts
@@ -10,7 +10,6 @@ import path from 'path'
 import {
 	getIsExternalLearnLink,
 	getIsRewriteableDocsLink,
-	getTutorialMap,
 	rewriteExternalDocsLink,
 	rewriteExternalLearnLink,
 } from 'lib/remark-plugins/rewrite-tutorial-links/utils'
@@ -59,8 +58,6 @@ function isNavDirectLink(value: NavNode): value is NavDirectLink {
 	return value.hasOwnProperty('href')
 }
 
-let TUTORIAL_MAP
-
 /**
  * Prepares all sidebar nav items for client-side rendering. Keeps track of the
  * index of each node using `startingIndex` and the `traversedNodes` property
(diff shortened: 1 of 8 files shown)


src/lib/learn-client/index.ts
```diff
@@ -2,19 +2,6 @@
  * Copyright (c) HashiCorp, Inc.
  * SPDX-License-Identifier: MPL-2.0
  */
-function getFetch() {
```
dstaley, line 5: This was removed in favor of native `fetch` to (hopefully) improve accuracy of Vercel's "Request Metrics" feature.

src/lib/remark-plugins/rewrite-tutorial-links/utils/get-tutorial-map.ts
```diff
@@ -3,7 +3,6 @@
  * SPDX-License-Identifier: MPL-2.0
  */
 
-import fetch from 'node-fetch'
```
dstaley, line 6: Removing to use the native global `fetch` for better integration with Vercel.

mwickett:
nice!

dstaley:
@heatlikeheatwave I think we can replace this async request to build the tutorials map with one statically generated prior to build. What do you think about that approach (in combination with this one where we pass the values instead of regenerating them).

src/layouts/sidebar-sidecar/utils/prepare-nav-data-for-client.ts
heatlikeheatwave, line 1: Looking at this file is giving me take home coding challenge flashbacks 😂

dstaley: fun fact: that's exactly where that component of the take home came from! (well the original component that was used for the same purpose on the .io sites)

heatlikeheatwave (approved):

src/layouts/sidebar-sidecar/utils/prepare-nav-data-for-client.ts
heatlikeheatwave: !!!

heatlikeheatwave:
> @heatlikeheatwave I think we can replace this async request to build the tutorials map with one statically generated prior to build. What do you think about that approach (in combination with this one where we pass the values instead of regenerating them).

I think this approach (generate once, pass to other functions) makes sense to me! Tested it locally (in addition to the preview) and it looks good 👍

PR #2401, opened 2024-03-29
feat: use pre-generated tutorials map

dstaley:
## 🗒️ What

This PR replaces our at-request-time generation of the Learn -> Dev Portal tutorials map with a pre-generated JSON file. This is to completely remove the Learn API from the SSR path.

## 🤷 Why

Each time we have to generate the tutorials map, we make 10 consecutive HTTP requests to fetch the entire list of tutorials. While this is heavily cached, it still means that _some_ pages will incur an additional 2-5s of latency. Furthermore, it's my understanding that the tutorial map is used to rewrite old links to tutorials; new links should always be pointing to the correct location, so we really don't need to update the tutorial map.

## 🛠️ How

This PR adds a `generate:tutorial-map` script that builds the map, and checks the map into source control. If we need to, we can re-run the script to generate the map, but given that the map is only for old links I don't think we'll need to rebuild it that often (if ever).

This PR also builds on the approach from #2399 where we updated locations making async calls to fetch the map with passed in values.

## 🧪 Testing

- [ ] Navigate to a handful of docs pages and ensure they render correctly.
- [ ] Go to [this page](https://dev-portal-git-dsgenerate-tutorials-map-hashicorp.vercel.app/terraform/intro) and ensure the "Getting Started" link in the sidebar does not point to `learn.hashicorp.com`.


src/lib/learn-client/index.ts
```diff
@@ -2,19 +2,6 @@
  * Copyright (c) HashiCorp, Inc.
  * SPDX-License-Identifier: MPL-2.0
  */
-function getFetch() {
```
dstaley, line 5: This was removed in favor of Node's native `fetch`.

src/lib/remark-plugins/rewrite-tutorial-links/utils/index.ts
```diff
@@ -8,7 +8,6 @@ export {
 	getIsExternalLearnPath,
 } from './get-is-external-learn-link'
 export { getIsRewriteableDocsLink } from './get-is-rewriteable-docs-link'
-export { getTutorialMap } from './get-tutorial-map'
```
dstaley, line 11: This was removed since nothing was using it anymore.

src/views/product-tutorials-view/helpers/__tests__/detect-and-reformat-learn-url.test.ts
```diff
@@ -3,34 +3,30 @@
...
  */
-const MOCK_TUTORIALS_MAP = {
-	'consul/gossip-encryption-secure':
-		'/consul/tutorials/gossip-encryption-secure',
-	'waypoint/aws-ecs': '/waypoint/tutorials/deploy-aws/aws-ecs',
-	'vault/getting-started-install':
-		'/vault/tutorials/getting-started/getting-started-install',
-	'cloud/amazon-peering-hcp': '/hcp/tutorials/networking/amazon-peering-hcp',
-}
+vi.mock('data/_tutorial-map.generated.json', async () => {
```
dstaley, line 15: Since we don't make a network call anymore, we mock the import instead of the network call.

nandereck (approved):
LGTM!

PR #2402, opened 2024-03-29
feat: enable opentelemetry tracing

dstaley:
## 🗒️ What

This PR enables [OpenTelemetry](https://nextjs.org/docs/pages/building-your-application/optimizing/open-telemetry) tracing, which allows us to have more insight into what our code is doing.

## 🤷 Why

Tracing allows us to have a deeper understanding of what our code is doing. This PR enables the default tracing provided by Next.js which tracks time spent in resolving components, `getStaticProps`, HTTP calls in `getStaticProps`, and rendering the route. By adding instrumentation spans around code we're looking to instrument, we can see the time spent in expensive functions. This PR wraps our MDX logic in spans to track both the time spent in the MDX compiler and the time spent in the SWC compiler.

## 🛠️ How

We first enable instrumentation by adding a `src/instrumentation.ts` file to the project, followed by wrapping sections of code we want to instrument with `setActiveSpan`.

Currently preview deployments are set to collect 100% of traces, with production set to collect 1% of traces. Local development is not traced.


EnMod (approved):
Looks like it will be useful!

PR #2403, opened 2024-04-02
cleanup: remove code for dark mode toast

dstaley:
## 🗒️ What

This PR removes code related to the Dark Mode toast notification.

## 🤷 Why

The dark mode toast notification hasn't been used since 6/15/2023, so it's safe to remove. This is also one less instance of `useRouter` that needs to be migrated to `next/navigation`.


EnMod (approved):
🧹

PR #2412, opened 2024-04-17
feat: use react-design-system-components

dstaley:
## 🗒️ What

This PR sets Dev Portal up to consume our `@hashicorp/react-design-system-components` package. To demonstrate that the components are working, it replaces the `CodeBlock` on the product install page with the new component.

## 🤷 Why

Dev Portal's components are visually similar to the components from the HashiCorp Design System, but their implementation predates the official implementation. As we adopted HDS components on hashicorp.com, we re-built the HDS components matching their Ember.js implementation. Those components are published as two packages:

- `@hashicorp/react-design-system-tokens`: This package contains our Dark Mode theme, since that's an extension of HDS.
- `@hashicorp/react-design-system-components`: This package contains React implementations of the official HDS Ember.js components published in `@hashicorp/design-system-components`.

This PR swaps our dark theme CSS file for the one from `@hashicorp/react-design-system-tokens` (which was based on Dev Portal's tokens).

## 🛠️ How

1. Import tokens in `src/pages/style.css`
2. Import HDS component CSS class names in `src/pages/_app.tsx` (this is because we can't import SCSS files inside CSS files; only JS files can import SCSS files).
3. Replace `react-code-block` with the new HDS version. 

## 📸 Design Screenshots

image

## 🧪 Testing

This PR was tested with a visual regression on the top 100 pages on Dev Portal and did not result in any visual changes aside from the expected changes to the `CodeBlock` on the install page.

## 💭 Anything else?

Additional comments regarding some changes to TypeScript are included in PR comments.


next-env.d.ts
```diff
@@ -1,4 +1,5 @@
 /// <reference types="next" />
+/// <reference types="next/image-types/global" />
```
dstaley, line 2: automatically added by Next.js

next.config.js
```diff
@@ -37,7 +37,6 @@ const hideWaypointTipContent = {
 }
 
 module.exports = withHashicorp({
-	nextOptimizedImages: true,
```
dstaley, line 40: `nextOptimizedImages` is an outdated approach to image handling that breaks Next's ability to import SVG files. Disabling this does not seem to have any impact, and restores the ability for Next to import SVG files, which is required for our `FlightIcon` component to work correctly.

package.json
```diff
@@ -80,9 +80,10 @@
 		"postcss-preset-env": "^6.7.0",
 		"react-test-renderer": "^18.2.0",
 		"rimraf": "^5.0.5",
+		"sass": "^1.75.0",
 		"simple-git-hooks": "^2.7.0",
 		"stylelint": "^13.8.0",
-		"typescript": "^5.3.2",
+		"typescript": "^5.4.5",
```
dstaley, line 86: Upgrading TypeScript was necessary to fix a bug with how TypeScript handles the `target` compiler option.

src/components/branded-card/index.tsx
```diff
@@ -4,7 +4,7 @@
  */
 
 import { CSSProperties } from 'react'
-import Image from 'next/legacy/image'
+import Image, { type StaticImageData } from 'next/legacy/image'
```
dstaley, line 7: Previously, `nextOptimizedImages` returned a string for these imports. Now that this is handled by Next.js natively, these imports return `StaticImageData`.

src/views/product-downloads-view/components/downloads-section/helpers.ts
```diff
@@ -11,7 +11,7 @@ import {
 export const generateCodeSnippetFromCommands = (
 	commands: PackageManager['commands']
 ): string => {
-	return commands.map((command: string) => `$ ${command}`).join('\n')
+	return commands.join('\n')
```
dstaley, line 14: The HDS code block does not expect shell session lines to be prepended with `$ `. If these are added, they are then copied by the Copy button.

src/views/product-downloads-view/server-helpers.ts
```diff
@@ -41,7 +41,7 @@ export const generatePackageManagers = async ({
 			const { commands } = packageManager
 			const installCodeHtml = await highlightString(
 				generateCodeSnippetFromCommands(commands),
-				'shell-session'
+				'bash'
```
dstaley, line 44: Without `$` prefixes, Prisma doesn't apply syntax highlighting. Switching to `bash` gets syntax highlighting working again.

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
 	"compilerOptions": {
 		"strict": false,
-		"target": "es5",
+		"target": "es2015",
```
dstaley, line 4: Bumping from `es5` to `es2015` was required since we use the spread operator on iterators; it does not impact the code generated by Next.js.

EnMod (approved):
Thanks for all the context in GH comments! Code LGTM, and so does preview.

PR #2418, opened 2024-04-22
Fix a bug when setting aria-describedby

RubenSandwich:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-rn-fixproduct-nav-open-in-a11y-hashicorp.vercel.app/terraform) 🔎
- [Asana task](https://app.asana.com/0/1205890709355230/1205890709355241/f) 🎟️

## 🗒️ What

Because of a bug, namely using += instead of = on a undefined variable, when setting `ariaDescribedBy` some information was not showing up in the accessibility tree. (In this case "opens in new window")

## 🧪 Testing

- [ ] [Make sure that "registry ▶️" link in the navbar](https://dev-portal-git-rn-fixproduct-nav-open-in-a11y-hashicorp.vercel.app/terraform) has the accessible description of: "(opens in new tab)"

You can view an element accessible tree properties by following this screenshot:
Screenshot 2024-04-24 at 4 05 09 PM

### DIFF
--- a/src/components/link/index.tsx
+++ b/src/components/link/index.tsx
@@ -37,12 +37,13 @@ const Link = ({ children, href, opensInNewTab, ...restProps }: LinkProps) => {
 	/**
 	 * Generate the final link's `aria-describedby` prop.
 	 */
-	let ariaDescribedBy: LinkProps['aria-describedby']
+	let ariaDescribedBy: LinkProps['aria-describedby'] = ''
 	if (shouldRenderScreenReaderOnlyMessage) {
-		ariaDescribedBy = opensInNewTabLabelId
+		ariaDescribedBy += opensInNewTabLabelId
 	}
+
 	if (restProps['aria-describedby']?.length > 0) {
-		ariaDescribedBy = ` ${restProps['aria-describedby']}`
+		ariaDescribedBy += ` ${restProps['aria-describedby']}`
 	}
 
 	return (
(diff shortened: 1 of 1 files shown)


src/components/link/index.tsx
```diff
@@ -39,10 +39,10 @@ const Link = ({ children, href, opensInNewTab, ...restProps }: LinkProps) => {
 	 */
 	let ariaDescribedBy: LinkProps['aria-describedby']
 	if (shouldRenderScreenReaderOnlyMessage) {
-		ariaDescribedBy += opensInNewTabLabelId
+		ariaDescribedBy = opensInNewTabLabelId
 	}
 	if (restProps['aria-describedby']?.length > 0) {
-		ariaDescribedBy += ` ${restProps['aria-describedby']}`
+		ariaDescribedBy = ` ${restProps['aria-describedby']}`
 	}
```
pbortnick, line 46: Is there ever a case where both `shouldRenderScreenReaderOnlyMessage` and `restProps['aria-describedby']?.length > 0` are true? If so, would we want to assign `ariaDescribedBy` as `opensInNewTabLabelId +  ${restProps['aria-describedby']}`. The way the code is currently leads me to believe that. Otherwise, why the addition assignment and the leading space character?

RubenSandwich: ohhh good point. I someone read that above line as an `else if` 🤦🏻 . I'll change the code to handle this case.

pbortnick (approved):

PR #2429, opened 2024-04-29
Default aria-describedby to null and only set when needed

RubenSandwich:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-rn-fixonly-set-aria-describedby-794e5c-hashicorp.vercel.app/hcp/docs/vault/what-is-hcp-vault/client) 🔎
- 🎟️ Asana task - None bug

## 🗒️ What

When changing aria-describedby to fix an undefined error last week, we introduce a new bug where we are adding an empty aria-describedby to a bunch of elements. This PR fixes that.

## 🧪 Testing

Screenshot 2024-04-29 at 2 10 52 PM

On the 

- [x] [In the current top bar](https://developer.hashicorp.com/hcp/docs/vault/what-is-hcp-vault/client) every link currently an empty aria-describedby. [Make sure that this PR removes that.](https://dev-portal-git-rn-fixonly-set-aria-describedby-794e5c-hashicorp.vercel.app/hcp/docs/vault/what-is-hcp-vault/client)

### DIFF
--- a/src/components/link/index.tsx
+++ b/src/components/link/index.tsx
@@ -37,13 +37,15 @@ const Link = ({ children, href, opensInNewTab, ...restProps }: LinkProps) => {
 	/**
 	 * Generate the final link's `aria-describedby` prop.
 	 */
-	let ariaDescribedBy: LinkProps['aria-describedby'] = ''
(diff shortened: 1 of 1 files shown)


EnMod (approved):
LGTM

PR #2440, opened 2024-05-10
fix: update repo-sync script for dev-portal-internal

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Fixes up the `repo-sync` workflow so that our `dev-portal-internal` repo can resume syncing.

## 🛠️ How

- Updates the workflow to align with more recent work in the workflow we originally copied
- Updates the PAT being used, as I think the previous one had expired

## 🧪 Testing

- Check out the results of a test of this modified action, at https://github.com/hashicorp/dev-portal-internal/pull/342/
    - Note that since `repo-sync` hadn't been run in a while, that PR did require some manual merge conflict resolution. All the changes seemed like trivial content changes, so my approach was to default to whatever's on `dev-portal` `main`, and discard whatever conflicting changes were present in `dev-portal-internal` `main`.

## 💭 Anything else?

Not at the moment!

[preview]: https://dev-portal-git-zsfix-internal-repo-sync-hashicorp.vercel.app/
[task]: https://app.asana.com/0/1206176289707208/1207289625383844/f

### DIFF
--- a/.github/workflows/repo-sync.yml
+++ b/.github/workflows/repo-sync.yml
@@ -137,7 +137,6 @@ jobs:
             console.log('Checking for merge conflicts')
             if (pull.mergeable_state === 'dirty') {
               console.log('Pull request has a conflict', pull.html_url)
-              await closePullRequest(pull_number)
               throw new Error('Pull request has a conflict, please resolve manually')
             }
             console.log('No detected merge conflicts')
--- a/.github/workflows/repo-sync.yml
+++ b/.github/workflows/repo-sync.yml
@@ -1,136 +1,154 @@
 # Note this is almost fully copied from
 # https://github.com/github/docs/blob/main/.github/workflows/repo-sync.yml
-# with the exception of a few modifications.
+# with the exception of:
+# - the `cron` schedule
+# - the `if` condition to allow this to run only in `dev-portal-internal`
+# - the `source_repo` has been updated to `dev-portal`
+# - the PAT credentials are from our `hashibot-web` account. They're managed in
+#   the `dev-portal-internal` repository secrets.
 
 on:
+  workflow_dispatch:
   schedule:
     - cron: '0 */2 * * *' # https://crontab.guru/every-2-hours
-  workflow_dispatch:
+
+permissions:
+  contents: write
+  pull-requests: write
 
 jobs:
   repo-sync:
-    # disallow scheduled runs for public repo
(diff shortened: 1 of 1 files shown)


zchsh:
@RubenSandwich I think the main issue with `repo-sync` was some expired credentials! I dived in a bit and updated the workflow to match the original workflow we'd copied - seems to be working, I tested it over in the `dev-portal-internal` repo and it yielded this PR: https://github.com/hashicorp/dev-portal-internal/pull/342/

Let me know what you think! 🙇‍♂️

.github/workflows/repo-sync.yml
```diff
@@ -1,136 +1,154 @@
...
-      - name: Close the pull request if no files remain
-        if: ${{ steps.find-pull-request.outputs.number && steps.pr-files.outputs.count == '0' }}
-        env:
-          GITHUB_TOKEN: ${{ secrets.PAT_WITH_REPO_AND_WORKFLOW_SCOPE }}
-        run: |
-          gh pr close ${{ steps.find-pull-request.outputs.number }} --repo $GITHUB_REPOSITORY
+            console.log('Checking for merge conflicts')
+            if (pull.mergeable_state === 'dirty') {
+              console.log('Pull request has a conflict', pull.html_url)
+              await closePullRequest(pull_number)
```
RubenSandwich, line 140: My only hesitation is here. Do we really want to close the pull request when there is a merge conflict? I think it would be better to:

1. leave the PR up, but comment that it has a merge conflict (Should we be notified when it does fail? I don't completely understand what we use dev-internal for yet.)
2. fail noisily

That way someone can use the existing PR as a starting point for fixing the merge conflict.

RubenSandwich (changes requested):
Just a question about how we want to handle merge conflicts.

.github/workflows/repo-sync.yml
zchsh: Sweet, I dig this! I'll remove the `closePullRequest()` so that we leave the PR up, so that it doesn't have to be re-opened 👍 I think the fail will be noisy in that the workflow will fail... If we want to go beyond that (eg notification of some type), I'd submit that we should treat that as separate from the current effort of getting `repo-sync` to run again.

RubenSandwich (approved):
Looks good. Thanks for updating this!

PR #2442, opened 2024-05-14
fix: announce page transitions

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Attempts to fix issue where some page transitions are not announced to screen readers.

> **Note**: WIP, in the midst of updating this PR, as it's now focused on ensuring page transitions are announced, rather than focus management during page transitions.

## 🤷 Why

The `RouteAnnouncer` built in to NextJS doesn't seem to function as expected in our use case, at least in some basic testing. It seems like this _might_ be related to `aria-live="assertive"` being used, rather than `aria-live="polite"`... at least in local testing, I seem to get more consistent page transition announcements when using `aria-live="polite"`... yet also occasionally get duplicate announcements (maybe this is just cause I've duplicated the `__next-route-announcer__` element though).

You can see in the first half of the video, with `aria-live="assertive"`, route announcements seem to be working at first, but then stop working. In the second half of the video, I switch to `aria-live="polite"`, and route announcements seem to be working more consistently.

https://github.com/hashicorp/dev-portal/assets/4624598/61bab98b-e887-4acc-84b7-53e3e7f4995c

This PR is an attempt to see if we can resolve the issue without diving into NextJS internals... but it seems like a sticky one, while there might be slight improvements, I seem to be getting inconsistent results and either missing or duplicate route announcements no matter what approach I try to take.

## 🛠️ How

- Pulls in the NextJS `RouteAnnouncer` code, includes it in `_app.tsx`
- Switches `aria-live="assertive"` to `aria-live="polite"

## 🧪 Testing

- [ ] Visit [upstream], confirm the issue exists
    - Quick example: [upstream/packer/docs/hcp]
    - For example, enable VoiceOver in macOS, navigate between sidebar items in docs (which are most often client-side page transitions),and confirm that during some client-side page transitions, the new page title is not announced
- [ ] 🚧 Visit the [preview], confirm page transitions are consistently announced exactly once (not actually working yet)
    - Quick example: [/packer/docs/hcp]
    - For example, enable VoiceOver in macOS, navigate between sidebar items in docs (which are most often client-side page transitions), and confirm that page transitions are consistently announced 

[task]: https://app.asana.com/0/1205890709355230/1205890709355271/f
[upstream]: https://developer.hashicorp.com
[upstream/packer/docs/hcp]: https://developer.hashicorp.com/packer/docs/hcp
[preview]: https://dev-portal-git-zsreset-focus-on-transition-hashicorp.vercel.app/
[/packer/docs/hcp]: https://dev-portal-git-zsreset-focus-on-transition-hashicorp.vercel.app/packer/docs/hcp

### DIFF
--- a/src/components/route-announcer/index.tsx
+++ b/src/components/route-announcer/index.tsx
@@ -1,0 +1,87 @@
+/**
+ * Note: this file is a near-duplicate of NextJS's included
+ * route announcer:
+ *
+ * https://github.com/vercel/next.js/blob/v14.2.1/packages/next/src/client/route-announcer.tsx
+ *
+ * It seems that `aria-live="assertive"`, as used there, results
+ * in the announcement not being read out at all (at least in some
+ * basic testing). This version uses `aria-live="polite"` instead,
+ * which seems to work as expected, with route changes being
+ * announced consistently.
+ */
+
+import React from 'react'
+import { useRouter } from 'next/router'
+
+const nextjsRouteAnnouncerStyles: React.CSSProperties = {
+	border: 0,
+	clip: 'rect(0 0 0 0)',
+	height: '1px',
+	margin: '-1px',
+	overflow: 'hidden',
+	padding: 0,
+	position: 'absolute',
+	top: 0,
+	width: '1px',
+
+	// https://medium.com/@jessebeach/beware-smushed-off-screen-accessible-text-5952a4c2cbfe
+	whiteSpace: 'nowrap',
+	wordWrap: 'normal',
+}
+
+export const RouteAnnouncer = () => {
+	const { asPath } = useRouter()
+	const [routeAnnouncement, setRouteAnnouncement] = React.useState('')
+
+	// Only announce the path change, but not for the first load because screen
+	// reader will do that automatically.
+	const previouslyLoadedPath = React.useRef(asPath)
+
+	// Every time the path changes, announce the new page’s title following this
+	// priority: first the document title (from head), otherwise the first h1, or
+	// if none of these exist, then the pathname from the URL. This methodology is
+	// inspired by Marcy Sutton’s accessible client routing user testing. More
+	// information can be found here:
+	// https://www.gatsbyjs.com/blog/2019-07-11-user-testing-accessible-client-routing/
+	React.useEffect(
+		() => {
+			console.log('asPath changed!')
+			console.log({
+				asPath,
+				previouslyLoadedPath: previouslyLoadedPath.current,
+			})
+			// If the path hasn't change, we do nothing.
+			if (previouslyLoadedPath.current === asPath) return
+			previouslyLoadedPath.current = asPath
+
+			if (document.title) {
+				console.log(
+					`setting routeAnnouncement to document.title: ${document.title}`
+				)
+				setRouteAnnouncement(document.title)
+			} else {
+				const pageHeader = document.querySelector('h1')
+				const content = pageHeader?.innerText ?? pageHeader?.textContent
+
+				console.log(`setting routeAnnouncement to content: ${content}`)
+				setRouteAnnouncement(content || asPath)
+			}
+		},
+		// TODO: switch to pathname + query object of dynamic route requirements
+		[asPath]
+	)
+
+	return (
+		<p
+			aria-live="assertive"
+			id="__next-route-announcer_debug__"
+			role="alert"
+			style={nextjsRouteAnnouncerStyles}
+		>
+			{routeAnnouncement}
+		</p>
+	)
+}
+
+export default RouteAnnouncer
--- a/src/components/route-announcer/index.tsx
+++ b/src/components/route-announcer/index.tsx
@@ -74,7 +74,7 @@ export const RouteAnnouncer = () => {
 
 	return (
 		<p
-			aria-live="assertive"
+			aria-live="polite"
 			id="__next-route-announcer_debug__"
 			role="alert"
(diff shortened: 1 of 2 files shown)


src/lib/hooks/use-body-focus-on-route-change.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Resets focus to the `body` element on route changes.
+ *
+ * Note that the `body` element must have `tabindex={-1}` to be focusable.
+ */
+export default function useBodyFocusOnRouteChange() {
+	const router = useRouter()
+
+	useEffect(() => {
+		const handleRouteChange = () => document.body.focus()
```
RubenSandwich, line 13: @zchsh In order to fix our issues. (See review video.)

We should:
1. Create an empty div in _document.tsx with the id `announce-route-change` with the aria-live="polite" set on it.
2. On route change put the text "Page changed to <new page title>." in the `announce-route-change` div
3. Check if focus already exists, if it doesn't focus on the body

I believe this will fix all of our current issues.

RubenSandwich (changes requested):
I recorded a review video that goes over some problems this PR introduces. (A video because I go over both the before and after of the page with voiceover.)

Basically us resetting to the body alone is not fixing our issues because:
1. It is not announcing the new page's title
2. It is sometimes announcing "chrome not responding" which is untrue and confusing

https://github.com/hashicorp/dev-portal/assets/1957315/66c481be-0a87-46c5-9064-cfa03ba1e1b7

zchsh:
@RubenSandwich This is great, thanks for the thorough review! 🙌 Was definitely kinda blindly following the task here, makes sense that re-focusing the body is not sufficient, and that really the issue here was "transitions not being announced" vs "missing transition focus management" 👍 Have renamed the task and will take another shot at this 🫡

zchsh:
@RubenSandwich I've taken a shot at making some updates here... testing locally, I can't seem to get the page title to reliably be announced _exactly once_... adding the `aria-live` region seems to work as expected, but I am seeing it announced twice on some page transitions:

https://github.com/hashicorp/dev-portal/assets/4624598/d0b9b999-57d8-40fc-9553-9aa4e248513f

Maybe related to unnecessary re-renders, or something? Feels like this could be hard to get perfect, I'm not sure... At first I'd tried this in `_document.tsx` as you suggested, but accessing the page `title` there doesn't seem as feasible, as it's [rendered server-side only](https://nextjs.org/docs/pages/building-your-application/routing/custom-document).

Re: focus management, it seems like we may be able to separate that a bit from the concern of "page titles being announced". If so, personally would love to split that to a separate PR to keep things focused - let me know what you think!

src/lib/hooks/use-page-title.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+export default function usePageTitle() {
+	const [pageTitle, setPageTitle] = useState('')
+	const router = useRouter()
+
+	useEffect(() => {
+		const updatePageTitle = () => {
+			setPageTitle(document.title)
+		}
+		updatePageTitle()
```
RubenSandwich, line 19: @zchsh it could be announcing twice because you are calling it here and then it also fires on the event `routeChangeComplete`.

RubenSandwich:
> Re: focus management, it seems like we may be able to separate that a bit from the concern of "page titles being announced". If so, personally would love to split that to a separate PR to keep things focused - let me know what you think!

Yeah agreed, let's split out the focus management stuff into another PR.

src/lib/hooks/use-page-title.ts
zchsh: Could be! I find that when I remove this, I only get intermittent announcements, and sometimes the page title announced is incorrect:

https://github.com/hashicorp/dev-portal/assets/4624598/48ab6183-e75e-49d5-86b8-703f4fe15bd4

Have tried a few other things but can't seem to get things to consistently announce only once on route change... 

Confusingly, the NextJS docs claim that [Next should announce client-side page transitions by default](https://nextjs.org/docs/architecture/accessibility#route-announcements). That doesn't seem to be what we're seeing here... so maybe another path forward could be to reach out to the NextJS team to get support on this... I'll reach out tomorrow!

RubenSandwich:
@zchsh it also looks like next.js did build their own "page transition announcer"; https://github.com/vercel/next.js/blob/91623ee4ecdbdce714e56b4445100975ce0b9674/packages/next/client/route-announcer.tsx. I wonder if it's working because we are not using router in dev portal? [The docs also seem to suggest that they prefer using next/link as well.](https://nextjs.org/docs/architecture/accessibility#route-announcements) 🤷🏻‍♂️

zchsh:
@RubenSandwich Ah yeah great question... maybe something to ask the next team! The linked commit seems to be from ~ 3 years ago, so hopefully their route announcer would work whether the `pages` or `app` dir is used for routing... but does seem possible that could be part of it 🕵️

RubenSandwich:
> The linked commit seems to be from ~ 3 years ago,

@zchsh [Sorry here RouteAnnouncer in their main branch (which they call canary).](https://github.com/hashicorp/dev-portal/pull/2442#issuecomment-2111252491)

zchsh:
@RubenSandwich I've taken another shot at this to try to demo what seems to be, maybe, an underlying issue in NextJS.

zchsh:
Closing this PR, see the 🎟 [Asana task](https://app.asana.com/0/1205890709355230/1205890709355271/f) for context!

PR #2455, opened 2024-05-17
fix: prevent nbsp characters from being manually copied

dstaley:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

This PR switches blank lines in our CodeBlocks to use a CSS pseudo element, which won't be copied when the user manually selects content to copy.

## 🤷 Why

Due to the way we split lines, empty lines aren't properly retained. We need to manually insert those blank lines, which requires that we give them some content so they take up space. Previously, we used the `nbsp;` character, but this creates issues when code is copied and pasted, since some languages don't allow this specific character.

## 🛠️ How

This PR switches the approach somewhat, by targeting empty lines via a specific CSS class. This class uses the `:after` pseudo element to render a space character, which isn't copied to the clipboard when a user selects code.

Unfortunately, since the blank lines have no content (not even a line terminating character), blank lines are not preserved when copied manually (they are when using the copy button). We can iterate and fix this later, but right now the important thing is preventing unsupported characters from ending up in our users' code.

## 🧪 Testing

- [ ] Go to [preview](https://dev-portal-git-dscodeblock-nbsp-hashicorp.vercel.app/terraform/language/meta-arguments/depends_on)
- [ ] Manually select the code in the code block and copy it to your clipboard.
- [ ] Open the console in your web browser.
- [ ] Type the first character of a template literal string <code>`</code>
- [ ] Paste the code
- [ ] Type the closing template literal symbol <code>`</code>
- [ ] Type `.includes('\xa0')`
- [ ] Press enter
- [ ] Ensure the output is `false` (meaning the string does not contain the non-breaking space character).

### DIFF
--- a/src/lib/rehype-code-plugins.ts
+++ b/src/lib/rehype-code-plugins.ts
@@ -27,9 +27,7 @@ export const rehypeCodePlugins: Pluggable[] = [
 						if (
 							lastChild.type === 'element' &&
 							lastChild.tagName === 'span' &&
-							lastChild.children.length === 1 &&
-							lastChild.children[0].type === 'text' &&
-							lastChild.children[0].value === '\xa0'
+							lastChild.children.length === 0
 						) {
 							node.children.pop()
 						}
@@ -37,9 +35,10 @@ export const rehypeCodePlugins: Pluggable[] = [
 						this.addClassToHast(node, `language-${this.options.lang}`)
 					},
 					line(node, line) {
-						// If the line is empty, add a non-breaking space
+						// If the line is empty, add a class so we can target
+						// it with CSS.
 						if (node.tagName === 'span' && node.children.length === 0) {
-							node.children.push({ type: 'text', value: '\xa0' })
+							this.addClassToHast(node, 'empty-line')
 						}
 					},
 					preprocess(code, options) {
--- a/src/pages/style.css
+++ b/src/pages/style.css
@@ -288,3 +288,8 @@ and negative margin.
 	font-size: 12px;
 	background: white;
 }
+
+/* Used to render blank lines in code blocks, but prevent them from being copied. */
+.hds-code-block__code .empty-line::after {
+	content: ' ';
+}


zchsh:
Thanks for tackling this - such a  ✨ FUN ✨ issue ahah! 💖 

Tested as described, and am indeed no longer seeing those NBSP characters, nor when using the `Copy` button... but there is a bit of wildness in my browser at least with "double-highlighting" (presumably of both the newlines, in pink, and the code itself, in green). And also seem to be missing some newlines in the pasted code:

| Upstream | This PR |
| - | - |
| before | after |

Maybe this is something we can live with... I did a bit more digging though, and found that _selecting_ and copying the text didn't seem to yield those weird whitespace characters in the first place (in my experience, at least). But, using the `Copy` button does. First showing this in browser console:

https://github.com/hashicorp/dev-portal/assets/4624598/5c95c7dc-df57-4098-9aa0-d96ea074e405

And then in VSCode:

https://github.com/hashicorp/dev-portal/assets/4624598/ea4f7f7b-7324-4cb1-8a1d-602bfc8280f1

As well it seems in some browsers (FireFox is where I found this), selecting and copying the text yields a pasted value with no newlines at all (upstream, and also in this PR, shown with preview link of this PR):

https://github.com/hashicorp/dev-portal/assets/4624598/69317339-bb61-4d97-998f-094d2a75e6e0

Looking back on the [original report in Slack](https://hashicorp.slack.com/archives/C01KCU4HDPY/p1715724298037879?thread_ts=1715724218.616259&cid=C01KCU4HDPY) I think this lines up... We kind of have two separate but somewhat related problems:

- `Select and copy yields no newlines` (in some browsers at least, probably relates to HTML structure / styles of Helios code block, or to our rehype processing, I have no idea in which domain a good solution might lie)
- `Copy button yields weird whitespace characters` (seems to affect all browsers)

Which is all to say... I think we may need a separate fix for the `Select and copy yields no newlines` issue, and I feel like the solution in this PR is probably headed more in that direction?

The "good news" (well kinda maybe?) is that we don't have to contend with the "select" and "copy button" issues being super-intertwined, so we don't have to try to solve both problems at once (i think this might've been one of the reasons we introduced `HiddenCopyContent` in the previous component... but heck if there is such a solution then that'd be awesome... i just feel like we tried going down that path and couldn't figure it out 😅 )

ANYWAYS... point being, for the `Copy button yields weird whitespace characters` we _may_ be able to fix the issue by adjusting the `processSnippet` function 🤔  And for the `Select and copy yields no newlines`, I feel like we somehow figured this out in the previous component... but i honestly can't remember how 🙈 i'll make a task for that latter issue, and dive in there!

dstaley:
> (presumably of both the newlines, in pink, and the code itself, in green)

This is actually from the Vercel toolbar!

zchsh:
@dstaley ahah oh man, the pink highlighting makes so much more sense now 🤦 Edited the comment to reflect that!

LeahMarieBush (approved):
LGTM!

zchsh (approved):
Looked at this again, definitely fixes the "weird whitespace in Copy button" issue so seems like a great improvement 👍 (Apologies for the wall of text earlier, went on quite the tangent there! 🙇‍♂️)

PR #2456, opened 2024-05-17
feat: use native fetch

dstaley:
## 🗒️ What

This PR removes unnecessary usage of `@vercel/fetch`.

## 🤷 Why

Something has happened on Vercel, resulting in HTTP requests made via `@vercel/fetch` to fail, with a `queryA ETIMEOUT`. `@vercel/fetch` was useful back in the Node v16 days, but now we have native global `fetch`, so we don't need to use it.

## 🛠️ How

Currently two instances of `@vercel/fetch` were breaking builds. Those have been removed, and the respective logic now relies on the global `fetch` object instead of `@vercel/fetch`.

### DIFF
--- a/scripts/generate-tutorial-variant-map.ts
+++ b/scripts/generate-tutorial-variant-map.ts
@@ -3,7 +3,6 @@
  * SPDX-License-Identifier: MPL-2.0
  */
 
-import createFetch from '@vercel/fetch'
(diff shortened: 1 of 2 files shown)


RubenSandwich (approved):
Looks good. 👍🏻

PR #2475, opened 2024-06-07
a11y: remove keyboard focus from InlineLinks with aria-hidden attribute

LeahMarieBush:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-leah-a11yaria-hidden-focus-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1205890709355230/1206539228551255/f) 🎟️

## 🗒️ What

This PR adds `tabIndex={-1}` to elements that have an aria-hidden attribute. These links are used as the targets for the associated permalinks so they should not be focusable. We cannot remove the targets because the permalinks do not work without them. More details in the [remark anchor links package](https://github.com/hashicorp/remark-plugins/blob/df606efc844319a2532ec54e4cf6ff2d575108ff/plugins/anchor-links/index.js#L122)

## 🤷 Why

During the accessibility review this came up as an a11y problem. It also shows as an error in the axe devtools plugin

## 🛠️ How

Since we don't want to remove the targets and don't want them focusable, a tabIndex of -1 was added. In the asana task it recommends removing the aria-hidden attribute but that should not happen since the permalink is what should be focused on.

## 🧪 Testing

1. Go to [dev portal link](https://developer.hashicorp.com/vault/docs/concepts/resource-quotas)
2. Turn on screen reader and tab through the page
3. Notice that tabbing through the exempt routes links includes the aria-hidden link targets
4. Go to [preview link](https://dev-portal-git-leah-a11yaria-hidden-focus-hashicorp.vercel.app/vault/docs/concepts/resource-quotas)
5. Turn on screen reader and tab through the page
6. Notice that tabbing through the exempt routes links does not include the aria-hidden link targets. Also note that clicking the links still works as intended.


mikegolus (approved):
lgtm

PR #2479, opened 2024-06-11
fix: remove unused func `isInternalLink`

prestonbourne:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1206176289707208/1207235163687853/f) 🎟️

## 🗒️ What

Small fix that removes an unused function from `.io` days

## 🤷 Why

1. Using exact string matching as opposed to something more robust like `new URL` raised an error for security
2. The function is unused across the repo


src/lib/utils.ts
heatlikeheatwave, line 1: I think you can delete this whole file! There doesn't appear to be any references to `chunk` in the code base either

heatlikeheatwave (approved):

PR #2482, opened 2024-06-13
fix sentinel page by removing overly strict error

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Fixes an issue where pages that use custom markup for code blocks may not render. The specific page which prompted this fix is [upstream/sentinel/docs/v0.17.x/language/parameters].

## 🛠️ How

For context, note the content we have for the broken page includes [a relatively rare use of HTML tags to construct a code block](https://github.com/hashicorp/sentinel/blob/4ad36966e46f00379352a39c05e2837e5499ce8c/website/content/sentinel/language/parameters.mdx?plain=1#L97):

custom-pre-block-in-mdx

When processing this MDX, our code-block component ends up erroring, since it doesn't see a strictly expected structure of children within the `<pre>` tag. Normally we expect `<pre>` to have a single component child, `<code>`, which in turn will be expected to have some props. But in this case, the first child of the `<pre>` is a text node.

This PR addresses the issue by removing strict error throwing. Instead, we return `null` when we can't retrieve the expected values, and fall back to rendering without the expected values in cases where they aren't available.

## 📸 Design Screenshots

| Before | After |
| - | - | 
| before | after |

## 🧪 Testing

- [ ] Confirm the issue exists by visiting [upstream/sentinel/docs/v0.17.x/language/parameters]
    - The page should 404, this is due to a server error with an existing document, but ultimately results in a 404 page, which implies the page does not exist
- [ ] Confirm the issue is fixed in the preview, by visiting [/sentinel/docs/v0.17.x/language/parameters]
    - Page should load and display content as expected

## 💭 Anything else?

It feels worth noting that this apparent `404` is in fact a `500`. I didn't realize this at first, and only recently realized I could've quickly opened the browser console to see this:

404-is-really-server-500

This feels worth noting as I wonder whether we'd want to try to adjust things so that the end user would see an accurate error page - eg "Something went wrong" - rather than a 404. This would likely relate to logging and monitoring as well (although we are getting a server 500 happening somewhere). Have made a backlog task to capture this thought: 🎟️ [consider different approach to 500s-rendering-as-404s](https://app.asana.com/0/1206176289707208/1207566071713496/f)

This also has me thinking about the risks of reaching for throwing errors when data or content isn't in the "expected shape", vs trying to find fallback approaches. So much of our content isn't rendered at build time, so while we _may_ sometimes throw these errors in a way that stops builds (generally a good thing, so that we can track down the source of the error, and respond accordingly), it seems like a _lot_ of the cases that could cause errors could easily slip through the cracks, and end up creating relatively silent errors. Not sure what to do about this, but wanted to make a note of it! I think it probably relates to the value of a content mono-repo - such a setup would facilitate 1) better content conformance checks, so we can help authors find these issues, and 2) the possibility of occasionally running a full static build of all docs content, which might take a while but would give us even more definitive feedback on whether all our pages are going to build as expected, or fail in some unexpected way.

[task]: https://app.asana.com/0/0/1206412320901525/f
[upstream/sentinel/docs/v0.17.x/language/parameters]: https://developer.hashicorp.com/sentinel/docs/v0.17.x/language/parameters
[preview]: https://dev-portal-git-zsfix-broken-sentinel-page-hashicorp.vercel.app/
[/sentinel/docs/v0.17.x/language/parameters]: https://dev-portal-git-zsfix-broken-sentinel-page-hashicorp.vercel.app/sentinel/docs/v0.17.x/language/parameters


src/components/dev-dot-content/mdx-components/mdx-code-blocks/index.tsx
```diff
@@ -70,12 +70,12 @@ function CodeBlockConfig({
 	hideClipboard,
 }: CodeBlockConfigProps) {
 	// This is only used for debugging purposes as we use shiki to assign the language tokens and styles
-	const propsOfFristChild = propsOfFirstChild(children)
```
zchsh, line 73: Renamed this function so I could fix the typo in the `const` name (otherwise it'd be a duplicate declaration).

Note with the modifications to the function, this may end up being `null`. This case was already handled - the `language` prop we pass may end up being `null`, which is not ideal, but doesn't seem worth stopping the whole page over.

src/components/dev-dot-content/mdx-components/mdx-code-blocks/index.tsx
```diff
@@ -205,7 +205,7 @@ export function MdxPre({ children, className, ...restProps }: MdxPreProps) {
 			className={classNames(className, {
 				[s.codeBlockMargin]: !restProps.hasBarAbove,
 			})}
-			value={normalizePlainCode(childrenOfFirstChild(children))}
```
zchsh, line 208: If we can't get `childrenOfFirstChild`, we fall back to just using `children`

src/components/dev-dot-content/mdx-components/mdx-code-blocks/index.tsx
```diff
@@ -220,20 +220,22 @@ function childrenOfFirstChild(children: ReactNode): ReactNode {
...
 		return firstElement.props.children
+	} else {
+		return null
 	}
-
-	throw new Error('First element of children is not an object with props')
 }
 
-function propsOfFirstChild(children: ReactNode): Record<string, unknown> {
+function getPropsOfFirstChild(
```
rmainwork, line 228: Small point but what do you think about calling this `getFirstChildProps`?

rmainwork (approved):
This looks great - thanks Zach!

I did make a note about the function name, but I'm not going to hold up the review for it - I'll leave it to you to decide if you want to make that change or not. 


🚀

src/components/dev-dot-content/mdx-components/mdx-code-blocks/index.tsx
zchsh: I dig it, thanks for the suggestion! Renamed and pushing the commit now 👍

PR #2492, opened 2024-06-24
feat: add signup to certs pages

prestonbourne:
## 🔗 Relevant links
- [Asana task](https://app.asana.com/0/1207484854701350/1207258153843315/f) 🎟️

## 🗒️ What

This PR adds a helper function, `isCertSlug`, and a wrapper component around [SignupFormArea](https://github.com/hashicorp/dev-portal/blob/dfc7d731612af4a3244e79a3b1069104102e7e58/src/views/certifications/components/signup-form-area/index.tsx#L17), to [`tutorials/[collectionSlug]`](https://github.com/hashicorp/dev-portal/blob/dfc7d731612af4a3244e79a3b1069104102e7e58/src/pages/%5BproductSlug%5D/tutorials/%5BcollectionSlug%5D.tsx) routes related to certification prep.

## 🤷 Why
On routes like [`terraform/tutorials/pro-cert`](https://developer.hashicorp.com/terraform/tutorials/pro-cert), users should be able to sign up for the certifications newsletter. 

### Notes ✏️
You'll likely notice the Asana ticket mentions the option of adding a link to https://developer.hashicorp.com/certifications#stay_informed instead of the entire component. 

I decided to go with adding the component since it would allow someone to signup without leaving their current page and context.

That being said, I see a valid argument in saying that this approach introduces some visual inconsistencies between the tutorials subroutes, so, just externally linking via a `p` tag at the bottom could be less obtrusive to the current ui

## 🗺️ Affected Routes

1. https://dev-portal-git-pres-featadd-signup-form-to-tut-fadd6c-hashicorp.vercel.app/terraform/tutorials/pro-cert
2. https://dev-portal-git-pres-featadd-signup-form-to-tut-fadd6c-hashicorp.vercel.app/terraform/tutorials/certification-associate-tutorials-003
3. https://dev-portal-git-pres-featadd-signup-form-to-tut-fadd6c-hashicorp.vercel.app/terraform/tutorials/certification-003
4. https://dev-portal-git-pres-featadd-signup-form-to-tut-fadd6c-hashicorp.vercel.app/consul/tutorials/certification-003
5. https://dev-portal-git-pres-featadd-signup-form-to-tut-fadd6c-hashicorp.vercel.app/vault/tutorials/associate-cert
6. https://dev-portal-git-pres-featadd-signup-form-to-tut-fadd6c-hashicorp.vercel.app/vault/tutorials/ops-pro-cert

## 🧪 Testing

Use the relevant links above for your convenience. I tried to find every affected page but there may be a few missing.

- [ ] Navigate to the affected Routes
- [ ] Ensure the newsletter sign up component is properly rendered
- [ ] Navigate to routes that should remain the same
- [ ] Ensure the component isn't being rendered on inappropriate route


src/views/collection-view/helpers/is-cert-slug.ts
prestonbourne, line 1: This will NOT work for plurals like "certs" or "certifications", figured keeping it readable and constrained would be best since the routes haven't been plural thus far but if something robust seems better I swap in something like:
`/cert(s?|ification(s?))/`
- cert: matches the base word "cert".
- s?: makes the 's' optional, so it matches both "cert" and "certs".
- ification: matches "certification".
- s?: makes the 's' optional, so it matches both "certification" and "certifications".

src/views/collection-view/components/cert-newsletter-signup/cert-newsletter-signup.module.css
prestonbourne, line 1: Seemed like this was an [agreed upon spacing value](https://github.com/hashicorp/dev-portal/blob/dfc7d731612af4a3244e79a3b1069104102e7e58/src/views/certifications/views/landing/landing.module.css#L64)

heatlikeheatwave (commented):
@prestonbourne Thanks for your work on this! What do you think about moving this logic from the view to the server.ts file? I was thinking it could be a new prop in layoutProps [here](https://github.com/hashicorp/dev-portal/blob/main/src/views/collection-view/server.ts#L117C8-L129)

src/views/collection-view/index.tsx
```diff
@@ -55,6 +56,7 @@ function CollectionView({
 					formatTutorialCard(t, collection)
 				)}
 			/>
+			{layoutProps.isCertificationPage && <CertNewsletterSignup />}
```
heatlikeheatwave, line 59:
```suggestion
			{layoutProps.isCertificationPage && (
				<>
					<div className={s.signUpForm}>
						<SignupFormArea />
					</div>
				</>
			)}
```

src/views/collection-view/index.tsx
```diff
@@ -15,6 +15,7 @@ import CollectionMeta from './components/collection-meta'
 import CollectionTutorialList from './components/collection-tutorial-list'
 import { formatTutorialCard } from 'components/tutorial-card/helpers'
 import { generateCollectionSidebarNavData } from './helpers/generate-collection-sidebar-nav-data'
+import CertNewsletterSignup from './components/cert-newsletter-signup'
```
heatlikeheatwave, line 18:
```suggestion
import { SignupFormArea } from 'views/certifications/components'
import s from './collection-view.module.css'
```

src/views/collection-view/index.tsx
heatlikeheatwave, line 1: Once the component is deleted, this code can be moved over as well as the css module

src/views/collection-view/index.tsx
heatlikeheatwave: could also add a `className` prop to `<SignupFormArea />` to cut down on the div soup

heatlikeheatwave (commented):
Thanks for making the previous changes.

Assuming the `CertNewsletterSignup` component and `isCertSlug` will only be used in `collection-view`, I think those files should be removed and the code moved to the `collection-view`. If those components and functions were used in several places, it would make sense to abstract them. For the current use case, I don't think it's needed. Let me know if I'm missing something here or if you have thoughts on this approach.

src/views/certifications/components/signup-form-area/index.tsx
```diff
@@ -14,9 +14,13 @@ import s from './signup-form-area.module.css'
...
 
-export function SignupFormArea() {
+interface SignupFormAreaProps {
+	className?: string
+}
+
+export function SignupFormArea({ className = '' }: SignupFormAreaProps) {
 	return (
-		<div className={s.root}>
+		<div className={`${s.root} ${className}`}>
```
prestonbourne, line 23: so that we don't have to wrap it in a `div` anytime we need to use it

src/views/collection-view/style.module.css
```diff
@@ -0,0 +1,7 @@
+.newsletterSignupArea {
```
prestonbourne, line 1: Had to make a new css file since the other 2 in `collection-view` were scoped to their relevant components already, figured a general one for the entire component would be best

prestonbourne:
> Thanks for making the previous changes.
> 
> Assuming the `CertNewsletterSignup` component and `isCertSlug` will only be used in `collection-view`, I think those files should be removed and the code moved to the `collection-view`. If those components and functions were used in several places, it would make sense to abstract them. For the current use case, I don't think it's needed. Let me know if I'm missing something here or if you have thoughts on this approach.

@heatlikeheatwave I guess since scope grew and it has to appear in 2 views now, a higher level of abstraction can stay? 
I figured keeping at as a computed field on the layoutProps would be better than adding a whole new key to [the type](https://github.com/hashicorp/dev-portal/blob/4210af506ec2be83b1493130d5e020e7af8d7ff9/src/lib/learn-client/types.ts#L95)

Not sure if you have any reservations about where the function was placed though ?

heatlikeheatwave (commented):
Just reposting my message here with the other option for visibility:

what do you think about passing a boolean prop to <SignupFormArea> and moving the css there?
It'd look like this:
```
// index.tsx

const APPEARANCE_SLUG = 'dark'

interface SignupFormAreaProps {
	isCertificationNewsletter?: boolean
}

export function SignupFormArea({
	isCertificationNewsletter = false,
}: SignupFormAreaProps) {
	return (
		<div
			className={classNames(s.root, {
				[s.certificationSignUp]: isCertificationNewsletter,
			})}
		>

// css module
.certificationSignUp {
	margin-top: 48px;
	@media (--dev-dot-tablet-up) {
		margin-top: 72px;
	}
}
```

then the implementation for your PR would look like
```
{layoutProps.isCertificationPrep && (
	<SignupFormArea isCertificationNewsletter={true} />
)}
```
All the other implementations of `<SignupFormArea />` would remain the same

src/views/certifications/components/signup-form-area/index.tsx
```diff
@@ -11,12 +11,16 @@ import {
...
 } from './components/newsletter-signup-form'
 import s from './signup-form-area.module.css'
+import { HTMLAttributes } from 'react'
+import classNames from 'classnames'
 
 const APPEARANCE_SLUG = 'dark'
 
-export function SignupFormArea() {
+export function SignupFormArea({
+	className = '',
```
heatlikeheatwave, line 20: No default value needed here
```suggestion
	className
```

src/views/collection-view/style.module.css
heatlikeheatwave, line 1: so bots don't yell 
```suggestion
/**
 * Copyright (c) HashiCorp, Inc.
 * SPDX-License-Identifier: MPL-2.0
 */

.newsletterSignupArea {
```

heatlikeheatwave (changes requested):
Just a couple small notes!

heatlikeheatwave (approved):

PR #2493, opened 2024-06-25
Set up preview tool for split page OpenAPI docs

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Implements a new OpenAPI preview tool, at  [/open-api-docs-preview-v2]. 

## 🤷 Why

The perceived value of this tool is:

- Provide a way for authors to quickly preview changes to OpenAPI specs. With authored content being a very key part of the value of these pages, and the interaction between the authored content and our rendering of that content being just as important, we want to create a tight feedback loop so that we can work closely with teams working on OpenAPI specs.

- Provide a way for `developer.hashicorp.com` web engineers to iterate on the new OpenAPI docs view, testing changes across many products, without having to set up new page routes and content source `.json` file fetching for every new product or new change.

Backing up a bit, we've chosen to split OpenAPI docs onto multiple pages, with one page for each operation, because single-page OpenAPI docs do not feel ideal.

From a platform perspective, the content can sometimes get long enough, or rather data-heavy enough, that incremental static regeneration fails due to response size limits on Vercel serverless functions. We've already seen some pretty bad failures due to this issue - see 🎟 [OpenAPI docs - Loading previous version doesn't seem to work anymore](https://app.asana.com/0/1207339219333499/1207348698328018/f) for example.

As well, from a UX perspective, the only clear advantage of a single long page is the ability to `⌘ + F` on the page to find content. Long pages already have issues with responsiveness, and this will only get worse as we introduce more useful, feature-rich OpenAPI docs. We're confident that we can implement a better search experience, by indexing OpenAPI docs content in our unified index (we're not currently doing this!). And we're confident that separate pages per operation will ultimately be a better experience for users, and a better foundation from which we can implement better OpenAPI docs.

As an example of what we consider to be a well-executed implementation of a separate page per operation, see https://test-hcp-vs.readme.io/reference/settier.

## 🛠️ How

- Creates a new `/api` route. This route accepts `POST` requests from the preview form, and saves the `POST`'d data with a unique ID, returning the ID to the client. It also accepts `GET` requests, with unique ID query parameter, to allow later retrieval of the stored data.
- Creates a new `views/open-api-docs-view-v2`. This is just a placeholder for now, the focus of this PR is on setting up a preview workflow, so the new view itself is only stubbed in.
- Copies & modifies the existing preview view to create `views/open-api-docs-preview-view-v2`. While this isn't too different from the existing work, and shares some components unchanged, the intent is to migrate to the new view and delete the old view & preview tool. With this in mind, shared code has been copied rather than abstracted, to avoid intertwining things and to make it easier to identify and delete the "old stuff" later.

## 📸 Design Screenshots

https://github.com/hashicorp/dev-portal/assets/4624598/27af9866-06c6-4da1-88ee-d4698c54a30f

## 🧪 Testing

- [ ] Visit [/open-api-docs-preview-v2], and test that the preview flow works
    - On initial load, you should see an empty state prompting you to fill out the preview form
    - Fill out the preview form, and submit it
    - After submitting the preview form, the page should reload
    - After the page reloads, you should see your submitted preview data transformed into server-generated props. Soon this will be an actually useful preview, but for now, to keep the scope of this PR reasonable, it's just placeholder.
    - Delete the cookie starting with `open-api-docs-preview-v2`. After deleting this cookie and reloading the page, you should be back to the started point. The preview data you previously uploaded still exists on the server (for now), but without the unique ID that was stored in the cookie, there's no way to retrieve it.

> **Note**: You can upload an actual OpenAPI spec if you'd like, but arbitrary JSON data also works for now, since we're not yet doing anything with the data, just showing the preview flow.

## 💭 Anything else?

Not at the moment!

[preview]: https://dev-portal-git-zssplit-page-api-docs-preview-hashicorp.vercel.app/
[task]: https://app.asana.com/0/1207339219333499/1207339701271604/f
[/open-api-docs-preview-v2]: https://dev-portal-git-zssplit-page-api-docs-preview-hashicorp.vercel.app/open-api-docs-preview-v2

### DIFF
--- a/src/pages/api/open-api-docs-preview-v2.ts
+++ b/src/pages/api/open-api-docs-preview-v2.ts
@@ -45,18 +45,19 @@ const MAX_FILE_AGE_MS = 1000 * 60 * 60 * MAX_FILE_AGE_HOURS
  *   the page or navigates to a new URL. This is particularly necessary as we're
  *   trying to render a multi-page preview, with operations on separate URLs.
  */
-export default async function handler(
-	req: NextApiRequest,
-	res: NextApiResponse
-) {
-	if (req.method === 'POST') {
-		return handlePost(req, res)
-	} else if (req.method === 'GET') {
-		return handleGet(req, res)
-	} else {
-		res.setHeader('Allow', ['POST', 'GET'])
-		res.status(405).json({ error: 'Method not allowed' })
-	}
+const reqHandlers = {
+  POST: handlePost,
+  GET: handleGet,
+};
+
+export default async function handler(req: NextApiRequest, res: NextApiResponse) {
+  const handle = reqHandlers[req.method];
+  if (handle) {
+    return handle(req, res);
+  } else {
+    res.setHeader('Allow', Object.keys(reqHandlers));
+    res.status(405).json({ error: 'Method not allowed' });
+  }
 }
 
 /**
--- a/src/pages/api/open-api-docs-preview-v2.ts
+++ b/src/pages/api/open-api-docs-preview-v2.ts
@@ -46,18 +46,21 @@ const MAX_FILE_AGE_MS = 1000 * 60 * 60 * MAX_FILE_AGE_HOURS
  *   trying to render a multi-page preview, with operations on separate URLs.
  */
 const reqHandlers = {
-  POST: handlePost,
-  GET: handleGet,
-};
+	POST: handlePost,
+	GET: handleGet,
+}
 
-export default async function handler(req: NextApiRequest, res: NextApiResponse) {
-  const handle = reqHandlers[req.method];
-  if (handle) {
-    return handle(req, res);
-  } else {
-    res.setHeader('Allow', Object.keys(reqHandlers));
-    res.status(405).json({ error: 'Method not allowed' });
-  }
+export default async function handler(
+	req: NextApiRequest,
+	res: NextApiResponse
+) {
+	const handle = reqHandlers[req.method]
+	if (handle) {
+		return handle(req, res)
+	} else {
+		res.setHeader('Allow', Object.keys(reqHandlers))
+		res.status(405).json({ error: 'Method not allowed' })
+	}
 }
 
 /**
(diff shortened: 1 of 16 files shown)


src/views/open-api-docs-preview-v2/components/file-string-input/index.tsx
heatlikeheatwave, line 1: This file and the css file above (`src/views/open-api-docs-preview-v2/components/file-string-input/file-string-input.module.css`) _appear_ to be the same as the existing ones for the api docs preview. Is there a reason why they were duplicated instead of moving the existing versions to the shared component dir?

src/views/open-api-docs-preview-v2/server.ts
```diff
@@ -0,0 +1,98 @@
...
+import type { GetServerSidePropsContext, GetServerSidePropsResult } from 'next'
+import type { OpenApiDocsViewV2Props } from 'views/open-api-docs-view-v2/types'
+import type { OpenApiPreviewV2InputValues } from './components/open-api-preview-inputs'
+import { OpenApiDocsPreviewV2Props } from '.'
+
+/**
+ * We expect the `HASHI_ENV` environment variable to be set to `production`
+ * only if we're actually in production.
+ */
+const IS_PRODUCTION = process.env.HASHI_ENV === 'production'
```
heatlikeheatwave, line 20: Since the page is available in production (the current open api docs preview), can this be changed to `process.env.VERCEL_ENV === 'production'`?
```suggestion
const IS_PRODUCTION = process.env.VERCEL_ENV === 'production'
```

heatlikeheatwave (approved):

src/views/open-api-docs-preview-v2/components/file-string-input/index.tsx
zchsh: Yep, this was intentional! The idea is we'll transition from the existing preview tool to this new preview tool, and then delete the old preview tool directories. Feels like it may be easier to make this transition if we avoid sharing code between them.

That being said, could definitely see the case for abstracting some of the more generic input components, as you've mentioned... I guess I just didn't feel a strong urge to create an abstraction at this point - I think we'd want to make sure the abstracted components are actually ready to be used in a generic way, no edge cases etc, and I'm not fully confident that's the case.

src/views/open-api-docs-preview-v2/server.ts
zchsh: Yeah great point! TBH I'm not really sure why we have `HASHI_ENV`... maybe related to testing production-ish things locally or something? 🤔 Does seem worth switching to use `VERCEL_ENV`, feels more standard.

src/pages/api/open-api-docs-preview-v2.ts
```diff
@@ -0,0 +1,178 @@
...
+	try {
+		const timestamp = Date.now()
+		const uniqueFileId = `ts${timestamp}_${randomUUID()}`
+		const newTempFile = getTempFilePath(TMP_DIR, uniqueFileId)
+		fs.writeFileSync(newTempFile, JSON.stringify(previewData, null, 2))
+		res.status(200).json({ uniqueFileId })
+	} catch (error) {
+		res.status(500).json({ error: error.toString() })
+	}
+}
```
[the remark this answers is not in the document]
zchsh: Sweet, I dig this! I've updated to split out `writeDataToTmpFile` ✅

src/pages/api/open-api-docs-preview-v2.ts
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * Handle GET requests, attempting to read stored props from `/tmp`.
+ *
+ * We also delete old files in the temporary directory on every request
+ * received to this endpoint, to ensure files don't hang around any longer
+ * than they need to for author previewing purposes.
+ */
+async function handleGet(req, res) {
+	// Delete old files in the temporary directory, to keep it clean.
+	deleteOldFiles(TMP_DIR, MAX_FILE_AGE_MS)
```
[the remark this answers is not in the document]
zchsh: I think we might need it to be blocking to avoid having the function execution cut off ([ref](https://github.com/vercel/next.js/discussions/14077#discussioncomment-25105)) (maybe possible to do "background tasks" in middleware now, but I'm not sure about API functions so would rather not risk it).

I did switch over to batching the deletion of old files with `Promise.all`, that might speed things up just slightly so we're blocking for less time 🫡

PR #2512, opened 2024-07-15
fix: cut off card text

LeahMarieBush:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-leah-fixcut-off-text-hashicorp.vercel.app/terraform) 🔎
- [Asana task](https://app.asana.com/0/1206176289707208/1205890709355288/f) 🎟️

## 🗒️ What

This PR fixes the issue where text is cut off for cards when the font size is increased. It removes `max-height` and relies instead on the `webkit-line-clamp` css property to properly truncate the lines. As a fallback for browsers that do not support line clamp, the text will just not be truncated for those cards. However, most browsers support line-clamp so that will not be very common (only ie does not support it)

## 📸 Design Screenshots

Before
Screenshot 2024-07-15 at 4 21 56 PM

After
Screenshot 2024-07-15 at 4 23 28 PM

## 🧪 Testing

1. Go to [prod](https://developer.hashicorp.com/terraform)
2. Go to your browser settings to increase the font size (for chrome you can increase the minimum font size to see this)
3. Scroll down to 'Popular use cases'
4. Note that the cards on that page get cut off in the middle of the description text
5. Go to [preview](https://dev-portal-git-leah-fixcut-off-text-hashicorp.vercel.app/terraform)
6. Scroll down to 'Popular use cases'
7. Verify that the cards are no longer cutting off the description text


RubenSandwich (approved):
Looks and works great!

PR #2515, opened 2024-07-16
fix: remove breadcrumbs without links

LeahMarieBush:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-leah-featbreadcrumbs-hashicorp.vercel.app/terraform/enterprise/v202404-1/releases/2024/v202402-1) 🔎
- [Asana task](https://app.asana.com/0/1206176289707208/1207746622331000/f) 🎟️

## 🗒️ What

This PR fixes the breadcrumbs bar so it does not show items without urls. The reason it is showing them is because a `forEach` was chained after the `filter` and that does not work as intended in javascript. I changed it so the `forEach` is on the new array that is formed from the `filter` method.

## 🧪 Testing

1. Go to [prod](https://developer.hashicorp.com/terraform/enterprise/v202404-1/releases/2024/v202402-1)
2. Notice that the 2018 and 2024 breadcrumbs are not clickable
3. Go to [preview](https://dev-portal-git-leah-featbreadcrumbs-hashicorp.vercel.app/terraform/enterprise/v202404-1/releases/2024/v202402-1)
4. The 2018 and 2024 links are no longer in the breadcrumbs

### DIFF
--- a/src/components/breadcrumb-bar/index.tsx
+++ b/src/components/breadcrumb-bar/index.tsx
@@ -27,17 +27,16 @@ function BreadcrumbBar({
 	// For now, we want to strictly require that all
 	// breadcrumb link URLs, if present, are relative rather
 	// than absolute links
-	links
-		.filter((l) => Boolean(l.url))
-		.forEach((l) => {
-			if (isAbsoluteUrl(l.url)) {
-				throw new Error(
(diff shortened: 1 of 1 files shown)


prestonbourne (approved):
🌟

PR #2519, opened 2024-07-22
Add site verificaiton meta tag for AHREF service.

aaronvanderlip:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Adding a verification meta 
## 🤷 Why

This change will enable additional reporting options in the aref tool.


src/components/head-metadata/index.tsx
```diff
@@ -108,6 +108,10 @@ export default function HeadMetadata(props: HeadMetadataProps) {
 				name="google-site-verification"
 				content="zRQZqfAsOX-ypXfU0mzAIzb5rUvj5fA4Zw2jWJRN-JI"
 			/>
+			<meta
+				name="ahrefs-site-verification"
+				content="9d739b21c466a84dae91df1c79396081a9072152e64f765c9390fed6f249fb84"
```
RubenSandwich, line 113: This is alright for now, but long term we should have all of these in some sort of config file.

RubenSandwich (approved):
👍🏻

src/components/head-metadata/index.tsx
kaitlynnefuery: @aaronvanderlip @RubenSandwich Should this live in https://github.com/hashicorp/hc-centralized-dns? I see other `google-site-verification`s there. 

I'd ask someone on CloudSec for confirmation, but I wonder if putting it here would then verify Developer and .com in one go: https://github.com/hashicorp/hc-centralized-dns/blob/main/terraform/active_domains/hashicorp_com/route53_records.tf

aaronvanderlip: @kaitlynnefuery ahref wants proof that you own the domain and provides a number of ways to go about it. For this task I chose a verification system that I could control more directly, adding meta tag to a front end file vs a system that I don't control directly, DNS. 

There is also `<meta name="google-site-verification" content="zRQZqfAsOX-ypXfU0mzAIzb5rUvj5fA4Zw2jWJRN-JI">` currently in the HEAD element so we historically don't seem to have settled on a method.

RubenSandwich: I would actually prefer for these values, google  in our [base config file.](https://github.com/hashicorp/dev-portal/blob/bec1da705ba80add46c3d2dac33f5c9bedb6fafa/config/base.json#L25). Namely I mean moving all of our site verification keys to it; zd-site-verification, google-site-verification, ahrefs-site-verification.

Though it's a low priority to me because so far we are keeping these all in one file.

PR #2523, opened 2024-07-24
Upload source maps during prod build

RubenSandwich:
## 🔗 Relevant links

- [Asana task](url) 🎟️


scripts/upload-source-maps.ts
```diff
@@ -0,0 +1,60 @@
...
+import { execSync } from 'child_process'
+
+const main = () => {
+	if (process.env.VERCEL_ENV === 'development') {
+		return
+	}
+
+	const LATEST_SHA = process.env.VERCEL_GIT_COMMIT_SHA
+	const PATH_PREFIX = 'https://developer.hashicorp.com/_next/static/'
+	const SERVICE = 'developer.hashicorp.com'
```
zchsh, line 14: I think we have two other possibilities when `process.env.VERCEL_ENV !== 'development'`:

- `preview`, for example for PR previews, branch deploys such as off our `staging` branch, etc
- `production`, the deployment that's served from `developer.hashicorp.com`

I noticed in DataDog we have two RUM applications set up, each with their own `SERVICE` identifier:

- `non-prod.developer.hashicorp.com`
- `developer.hashicorp.com`

Would it make sense to specifically target `developer.hashicorp.com` with `process.env.VERCEL_ENV === 'production'`, in order to reduce non-production noise in our production application? We could upload the other source maps to the `non-prod.developer.hashicorp.com` service.

Notably, I think we already have the necessary config to distinguish between the non-prod & prod services:

- [base.json config for non-prod](https://github.com/hashicorp/dev-portal/blob/70fb5226724a8cca166384645f88e26d950e3f26/config/base.json#L29)
- [production.json config](https://github.com/hashicorp/dev-portal/blob/70fb5226724a8cca166384645f88e26d950e3f26/config/production.json#L11)

scripts/upload-source-maps.ts
```diff
@@ -0,0 +1,60 @@
...
+
+	const DATADOG_API_KEY = process.env.DD_API_KEY
+
+	try {
+		const status = execSync(
+			`DATADOG_API_KEY=${DATADOG_API_KEY} npx @datadog/datadog-ci sourcemaps upload .next/static/ --service=${SERVICE} --release-version=${LATEST_SHA} --minified-path-prefix=${PATH_PREFIX}`
```
zchsh, line 20: Noticed this error in the Vercel build logs:

CleanShot 2024-07-25 at 11 11 34@2x

It seems like "running `datadog-ci` within the git repo" should in theory be happening, but isn't working as expected 🤔 It seems like we can [manually set the relevant info ourselves](https://github.com/DataDog/datadog-ci/tree/master/src/commands/sourcemaps#link-errors-with-your-source-code)... but I am curious why the DataDog CLI doesn't seem to be finding any git remotes...  maybe a Vercel thing? 🤷 I'm lost 😬

scripts/upload-source-maps.ts
```diff
@@ -0,0 +1,60 @@
...
+
+	// 	deployStatus
+	// 		.toString()
+	// 		.split('\n')
+	// 		.forEach((line) => {
+	// 			console.log(line)
+	// 		})
+
+	// 	console.log('Deployment created successfully')
+	// }
```
zchsh, line 57: 🧹 Seems like we can remove this commented code, unless it's here for testing!

scripts/upload-source-maps.ts
```diff
@@ -0,0 +1,60 @@
...
+		const status = execSync(
+			`DATADOG_API_KEY=${DATADOG_API_KEY} npx @datadog/datadog-ci sourcemaps upload .next/static/ --service=${SERVICE} --release-version=${LATEST_SHA} --minified-path-prefix=${PATH_PREFIX}`
+		)
+
+		status
+			.toString()
+			.split('\n')
+			.forEach((line) => {
+				console.log(line)
+			})
```
zchsh, line 28: As an alternative to storing `status` in a `const` and then logging it out _after_ the command is done, I think we could use the [`stdio` option](https://nodejs.org/api/child_process.html#child_processexecsynccommand-options) here to log output from the `execSync` command as it happens:

```suggestion
        { stdio: 'inherit' }
		)
```

scripts/upload-source-maps.ts
```diff
@@ -0,0 +1,60 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+import { execSync } from 'child_process'
+
+const main = () => {
+	if (process.env.VERCEL_ENV === 'development') {
+		return
+	}
```
zchsh, line 10: 🧪 I feel like it could be good to have a way to confirm that the basics of source maps are working locally... maybe when `VERCEL_ENV` is `development`, we could take on the optional [`--dry-run` parameter](https://github.com/DataDog/datadog-ci/tree/master/src/commands/sourcemaps#upload) and leave everything else the same?

zchsh (commented):
🙌 This looks great so far!

🤔 Main feedback is whether we should use the `non-prod.developer.hashicorp.com` service in DataDog for non-prod source maps. I think this might be necessary since it's already set up in our `config` files!

---

🧪 I'm not as familiar with DataDog, so with the empty PR description I had to poke around a bit to try to test things! Given that in the [Vercel logs, line 748](https://vercel.com/hashicorp/dev-portal/Fxz1UAjbSfgkX4H6m5tgVBcNwpd7#L748) we see this version identifier:

CleanShot 2024-07-25 at 11 37 24@2x

I think we would expect to see source maps uploaded on the `non-prod.developer.hashicorp.com` service with version `70fb5226724a8cca166384645f88e26d950e3f26`:

https://app.datadoghq.com/rum/performance-monitoring?query=%40session.type%3Auser%20%40application.id%3A40a2019f-88ad-4076-ab52-ffd074064eed%20version%3A70fb5226724a8cca166384645f88e26d950e3f26

I imagine part of the reason this isn't currently happening is since we're always uploading to the `developer.hashicorp.com` service, but I couldn't find a version `70fb5226724a8cca166384645f88e26d950e3f26` in that service... that being said, I might be going about testing in a weird way here, very open to other approaches!

Thinking out loud, but could we (temporarily) intentionally throw an error on a specific page in this PR? I think then we could use that page to test both that we're uploading source maps (which seems to be confirmed by the Vercel logs in this PR), and that we're uploading them in a way where we can trace them back when an error happens.

scripts/upload-source-maps.ts
heatlikeheatwave: I think this is coming up it's trying to get metadata. From the DD docs.

> If you run datadog-ci sourcemaps upload within a Git working directory, Datadog collects repository metadata. 
More info here: https://github.com/DataDog/datadog-ci/tree/master/src/commands/sourcemaps#link-errors-with-your-source-code

RubenSandwich:
Handing off this PR to @heatlikeheatwave.

scripts/upload-source-maps.ts
heatlikeheatwave: Update: the error went away when I added the flag `--disable-git` on my test branch

heatlikeheatwave:
> 🙌 This looks great so far!
> 
> 🤔 Main feedback is whether we should use the `non-prod.developer.hashicorp.com` service in DataDog for non-prod source maps. I think this might be necessary since it's already set up in our `config` files!
> 
> 🧪 I'm not as familiar with DataDog, so with the empty PR description I had to poke around a bit to try to test things! Given that in the [Vercel logs, line 748](https://vercel.com/hashicorp/dev-portal/Fxz1UAjbSfgkX4H6m5tgVBcNwpd7#L748) we see this version identifier:
> 
> CleanShot 2024-07-25 at 11 37 24@2x
> 
> I think we would expect to see source maps uploaded on the `developer.hashicorp.com` service with version `70fb5226724a8cca166384645f88e26d950e3f26`... I couldn't find a version `70fb5226724a8cca166384645f88e26d950e3f26` in that service, but I could find it for`non-prod.developer.hashicorp.com`, although without source maps (as far as I can tell):
> 
> https://app.datadoghq.com/rum/performance-monitoring?query=%40session.type%3Auser%20%40application.id%3A40a2019f-88ad-4076-ab52-ffd074064eed%20version%3A70fb5226724a8cca166384645f88e26d950e3f26
> 
> That being said, I might be going about testing in a weird way here, very open to other approaches!
> 
> Thinking out loud, but could we (temporarily) intentionally throw an error on a specific page in this PR? I think then we could use that page to test both that we're uploading source maps (which seems to be confirmed by the Vercel logs in this PR), and that we're uploading them in a way where we can trace them back when an error happens.


@zchsh Tested out your idea (to throw intentional error in preview and track in `non-prod` service). You can see the unminified error [here](https://app.datadoghq.com/rum/sessions?query=%40type%3Aerror%20%40application.id%3A40a2019f-88ad-4076-ab52-ffd074064eed%20-%40error.source%3Anetwork%20%40session.type%3Auser%20version%3A3e2af5b85963f3a28d7570ae060990aa512f0695&agg_m=count&agg_m_source=base&agg_t=count&cols=&event=AgAAAZDrLNNJeFyzhAAAAAAAAAAYAAAAAEFaRHJMTmM2QUFBMTd0aDVtUTZmM2dBTQAAACQAAAAAMDE5MGViMmMtZDUwYy00OTc1LThiMzYtYTJlYzkzNjA1MTI2&fromUser=false&refresh_mode=sliding&track=rum&viz=stream&from_ts=1721327766712&to_ts=1721932566712&live=true)! 🎉

scripts/upload-source-maps.ts
zchsh: Sweet, glad the error's gone! 🎉

In a perfect world, I think if we could resolve the issue without the `--disable-git` flag, then we might be able to link back to the source repo (screenshot from [here](https://app.datadoghq.com/rum/sessions?query=%40type%3Aerror%20%40application.id%3A40a2019f-88ad-4076-ab52-ffd074064eed%20-%40error.source%3Anetwork%20%40session.type%3Auser%20version%3A3e2af5b85963f3a28d7570ae060990aa512f0695&agg_m=count&agg_m_source=base&agg_t=count&cols=&event=AgAAAZDrLNNJeFyzhAAAAAAAAAAYAAAAAEFaRHJMTmM2QUFBMTd0aDVtUTZmM2dBTQAAACQAAAAAMDE5MGViMmMtZDUwYy00OTc1LThiMzYtYTJlYzkzNjA1MTI2&fromUser=false&refresh_mode=sliding&track=rum&viz=stream&from_ts=1721327766712&to_ts=1721932566712&live=true)):

CleanShot 2024-07-26 at 12 02 52@2x

Feels kinda unnecessary though... the setup you've done to get the above link working feels like it's the really valuable thing here 👍 Finding out how to get DataDog's `git` stuff to work feels like something we can tack on later as a very optional enhancement if we really want to 👍

scripts/upload-source-maps.ts
```diff
@@ -0,0 +1,69 @@
...
+		files.forEach((file) => {
+			if (file.endsWith('.js.map')) {
+				unlinkSync(join(dirPath, file))
+			}
+		})
+	} catch (error) {
+		console.error(error)
+
+		console.log('Failed to delete source maps from pages dir')
+	}
```
zchsh, line 66: ⛏️ Nitpick: I think this might be slightly faster and more thorough:

```js
// Remove source maps after uploading
execSync(`rm -f .next/static/**/*.map`)
```

scripts/upload-source-maps.ts
```diff
@@ -0,0 +1,69 @@
...
+const main = () => {
+	if (process.env.VERCEL_ENV === 'development') {
+		return
+	}
+
+	const LATEST_SHA = process.env.VERCEL_GIT_COMMIT_SHA
+	// const PATH_PREFIX = 'https://developer.hashicorp.com/_next/static/'
+	// const SERVICE = 'developer.hashicorp.com'
+	const PATH_PREFIX =
+		'https://dev-portal-git-rn-featbuild-source-maps-during-deploy-hashicorp.vercel.app/_next/static/'
```
zchsh, line 18: I think this'll need to somehow account for variable deployment URLs!

TLDR I think we could do like:

```suggestion
	const PATH_PREFIX = '/_next/static/'
```

In more detail:

- One option is to use the full path, like `https://${process.env.VERCEL_URL}/_next/static/`... though this won't work for testing dry runs in local dev (maybe a non-issue)
- Another option is to use an absolute path ([docs here](https://github.com/DataDog/datadog-ci/tree/master/src/commands/sourcemaps#commands), which feels maybe more convenient?

scripts/upload-source-maps.ts
```diff
@@ -0,0 +1,69 @@
...
+	const SERVICE = 'non-prod.developer.hashicorp.com'
```
zchsh, line 19: I think this'll need to be variable based on deployment environment. Something like:

```suggestion
	const SERVICE = process.env.VERCEL_ENV === 'production' ? 'developer.hashicorp.com' : 'non-prod.developer.hashicorp.com'
```

heatlikeheatwave: Good call!

scripts/upload-source-maps.ts
heatlikeheatwave: what do you mean by working locally? that we can test uploading them to datadog locally?

scripts/upload-source-maps.ts
heatlikeheatwave: I ended up going with process.env.VERCEL_ENV === 'production' ? 'https://developer.hashicorp.com/_next/static/' : `https://${process.env.VERCEL_BRANCH_URL}/_next/static/` here

scripts/upload-source-maps.ts
zchsh: More just to see that `datadog-ci` picks up on the `*.map` files, and that they're removed after the script runs! I believe the `--dry-run` flag will mean source maps aren't uploaded. Definitely feels optional 👍

PR #2528, opened 2024-07-29
bug fix: Remove error for testing sourcemaps

heatlikeheatwave:
## 🗒️ What

- remove error that was used to test sourcemaps RUM


RubenSandwich (approved):
🚢

PR #2543, opened 2024-08-15
feat(Analytics): implement PostHog

nandereck:
## 🔗 Relevant links
- [Preview link](https://dev-portal-git-nels-analyticsadd-posthog-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202787000646550/1208027559696437/f) 🎟️

## 🗒️ What
This PR adds `PostHog` to https://developer.hashicorp.com by loading the PostHog script using our `ConsentManager` component.

### PostHog Installation Method

This implementation follows the [`HTML Snippet` PostHog installation](https://posthog.com/docs/libraries/js#option-1-add-the-javascript-snippet-to-your-html-span-classbg-gray-accent-light-darkbg-gray-accent-dark-text-gray-font-semibold-align-middle-text-sm-p-1-roundedrecommendedspan) instead of the Next.js installation. This was done for several reasons:
- We follow this method for loading several other scripts via `ConsentManager`
- There is talk of migrating to a Consent Management Platform which would likely load scripts via GTM. If that's the case, transitioning to GTM would largely just be a matter of copying the HTML snippet into GTM. That said, I could be missing something so very open to a Next.js-specific implementation!

### PostHog Setup

In PostHog, separate projects have been created for [`Development`](https://eu.posthog.com/project/29988) and [`Production`](https://eu.posthog.com/project/29987). 
This follows [PostHog's recommendation](https://posthog.com/tutorials/multiple-environments#using-multiple-projects) and [matches our Segment implementation for hashicorp.com](https://github.com/hashicorp/web/blob/0f7d2ffc8b2381370e98559d39d430c3a0a826cc/apps/www/next.config.js#L8-L10). Having separate projects enables two things: 

1. keeps data from our production instance separate from data in non-production environments
2. allows for testing PostHog in development / preview environments

## 🧪 Testing
If you need access to PostHog, reach out to me and I can send an invite.

### Validate non-production analytics are sent to the PostHog `Development` project
- [ ] Navigate to an [the v0.1.0 release notes for Boundary in the preview](https://dev-portal-git-nels-analyticsadd-posthog-hashicorp.vercel.app/boundary/docs/release-notes/v0_1_0)
- [ ] Navigate to the activity tab in the [`Development` project in PostHog](https://eu.posthog.com/project/29988/activity/explore)
- [ ] Validate that a `Pageview` event is created for the visit to the above page. It should look like this:
<details>
<summary>Screenshot</summary>

</details>

- [ ] Navigate to the [`Production` project in PostHog](https://eu.posthog.com/project/29987/products)
- [ ] Validate that there is no activity tab and therefore a `Pageview` event is not created


nandereck:
@heatlikeheatwave - if you have a free minute, would you mind giving this PR a looksy? I'd love some dev-portal eyes to sanity check the approach / implementation. Very open to alternatives here!

heatlikeheatwave (commented):
Thanks for doing this! I understand the rationale of adding the HTML snippet (existing practice, may move it to GTM in the future). AFAIK, for Segment & GTM the only way to implement them is with a js snippet (I could be wrong about this though!). It makes sense we'd load a script tag for those services. However, PH has a next.js guide. 

I think for this use case we should follow the [next.js guide](https://posthog.com/docs/libraries/next-js). What do you think about this approach?

nandereck:
> I think for this use case we should follow the [next.js guide](https://posthog.com/docs/libraries/next-js). What do you think about this approach?

The only thing that gave me pause here was that there didn't seem to be a lightweight way of gating the script behind `ConsentManager` without having to change how `ConsentManager` works. Let me double-check this assumption and explore what the Next.js installation might look like.

heatlikeheatwave:
> > I think for this use case we should follow the [next.js guide](https://posthog.com/docs/libraries/next-js). What do you think about this approach?
> 
> The only thing that gave me pause here was that there didn't seem to be a lightweight way of gating the script behind `ConsentManager` without having to change how `ConsentManager` works. Let me double-check this assumption and explore what the Next.js installation might look like.
> 
> (Also, while this is by no means finalized, the idea has been floated that we migrate to a consent management platform. Given that, I'd be a little hesitant to extend `ConsentManager` if it's a heavy lift)

By gating the script behind the consent manager, do you mean not running the script until the user approves to being tracked?

nandereck:
Looking at this a bit more, I think we would need `ConsentManager` to expose some state (prob with context) that can be accessed by PostHog.  While this isn't a huge deal, I'm a bit hesitant given that 1) the above implementation does the trick and 2) there's potential to replace `ConsentManager`.

@heatlikeheatwave - what do you think?

src/hooks/use-posthog-analytics.ts
```diff
@@ -0,0 +1,35 @@
...
+
+declare global {
+	interface Window {
+		posthog?: $TSFixMe
+	}
+}
+
+function onRouteChangeComplete() {
+	// PostHog documentation for capturing pageviews in SPA with the JS web installation:
+	// https://posthog.com/docs/libraries/js#single-page-apps-and-pageviews
```
zchsh, line 12: ⛏️ Nitpick: I was at first confused how we were capturing `pageview` events on load. The docs link definitely helped! Felt like it may be useful to surface the "PostHog captures stuff automatically" directly as well:

```suggestion
	/**
	 * PostHog automatically captures a `pageview` for initial page loads.
	 * Subsequent client-side navigation events have to be captured manually,
	 * which is why we have to set up this `onRouteChangeComplete` event.
	 * 
	 * PostHog documentation for capturing pageviews in SPA with the JS web installation:
	 * https://posthog.com/docs/libraries/js#single-page-apps-and-pageviews
	 */
```

src/lib/consent-manager-services/developer.ts
```diff
@@ -0,0 +1,19 @@
...
...&(t=t[o[0]],e=o[1]),t[e]=function(){t.push([e].concat(Array.prototype.slice.call(arguments,0)))}}(p=t.createElement("script")).type="text/javascript",p.async=!0,p.src=s.api_host.replace(".i.posthog.com","-assets.i.posthog.com")+"/static/array.js",(r=t.getElementsByTagName("script")[0]).parentNode.insertBefore(p,r);var u=e;for(void 0!==a?u=e[a]=[]:a="posthog",u.people=u.people||[],u.toString=function(t){var e="posthog";return"posthog"!==a&&(e+="."+a),t||(e+=" (stub)"),e},u.people.toString=function(){return u.toString(1)+".people (stub)"},o="capture identify alias people.set people.set_once set_config register register_once unregister opt_out_capturing has_opted_out_capturing opt_in_capturing reset isFeatureEnabled onFeatureFlags getFeatureFlag getFeatureFlagPayload reloadFeatureFlags group updateEarlyAccessFeatureEnrollment getEarlyAccessFeatures getActiveMatchingSurveys getSurveys getNextSurveyStep onSessionId setPersonProperties".split(" "),n=0;n<o.length;n++)g(u,o[n]);e._i.push([i,s,a])},e.__SV=1)}(document,window.posthog||[]);posthog.init("${process.env.POSTHOG_PROJECT_API_KEY}",{api_host:'https://eu.i.posthog.com', person_profiles: 'identified_only', capture_pageview: false})`,
```
zchsh, line 15: 🚧 Blocking change (not actually sure of this, but want to make sure before approving):

I think `onRouteChangeComplete` and by extension the `use-posthog-analytics` hook will only fire on client-side navigation events, so will _not_ log a `$pageview` event on initial page loads. Tested locally and this seems to be the case.

In order to ensure initial page loads are captured, I _think_ we may want to omit `capture_pageview: false` here:

```suggestion
		body: `!function(t,e){var o,n,p,r;e.__SV||(window.posthog=e,e._i=[],e.init=function(i,s,a){function g(t,e){var o=e.split(".");2==o.length&&(t=t[o[0]],e=o[1]),t[e]=function(){t.push([e].concat(Array.prototype.slice.call(arguments,0)))}}(p=t.createElement("script")).type="text/javascript",p.async=!0,p.src=s.api_host.replace(".i.posthog.com","-assets.i.posthog.com")+"/static/array.js",(r=t.getElementsByTagName("script")[0]).parentNode.insertBefore(p,r);var u=e;for(void 0!==a?u=e[a]=[]:a="posthog",u.people=u.people||[],u.toString=function(t){var e="posthog";return"posthog"!==a&&(e+="."+a),t||(e+=" (stub)"),e},u.people.toString=function(){return u.toString(1)+".people (stub)"},o="capture identify alias people.set people.set_once set_config register register_once unregister opt_out_capturing has_opted_out_capturing opt_in_capturing reset isFeatureEnabled onFeatureFlags getFeatureFlag getFeatureFlagPayload reloadFeatureFlags group updateEarlyAccessFeatureEnrollment getEarlyAccessFeatures getActiveMatchingSurveys getSurveys getNextSurveyStep onSessionId setPersonProperties".split(" "),n=0;n<o.length;n++)g(u,o[n]);e._i.push([i,s,a])},e.__SV=1)}(document,window.posthog||[]);posthog.init("${process.env.POSTHOG_PROJECT_API_KEY}",{api_host:'https://eu.i.posthog.com', person_profiles: 'identified_only'})`,
```

zchsh (changes requested):
🙌 This seems great so far!

🚧 Only blocking comment is around `capture_pageview: false`. I think we may want to leave that enabled, otherwise I think we'll only log client-side navigation events, and will miss initial page loads.

💭 Vaguely related to the above, maybe we could split out the `posthog.init()` config options, and then stringify them into `body`, so they're easier to modify:

```js
/**
 */

// `options` are passed to `posthog.init()` in the `body` below.
const options = {
	api_host: 'https://eu.i.posthog.com',
	person_profiles: 'identified_only',
}

const developerConsentManagerServices = [
	{
		name: 'PostHog',
		description:
			'PostHog enables analytics, session replay, feature flags, etc. on our web properties',
		category: 'Analytics',
		async: true,
		// `capture_pageview` is set to `false` so that we can manually track page views.
		// This occurs in the `use-posthog-analytics.ts` hook.
		body: `!function(t,e){var o,n,p,r;e.__SV||(window.posthog=e,e._i=[],e.init=function(i,s,a){function g(t,e){var o=e.split(".");2==o.length&&(t=t[o[0]],e=o[1]),t[e]=function(){t.push([e].concat(Array.prototype.slice.call(arguments,0)))}}(p=t.createElement("script")).type="text/javascript",p.async=!0,p.src=s.api_host.replace(".i.posthog.com","-assets.i.posthog.com")+"/static/array.js",(r=t.getElementsByTagName("script")[0]).parentNode.insertBefore(p,r);var u=e;for(void 0!==a?u=e[a]=[]:a="posthog",u.people=u.people||[],u.toString=function(t){var e="posthog";return"posthog"!==a&&(e+="."+a),t||(e+=" (stub)"),e},u.people.toString=function(){return u.toString(1)+".people (stub)"},o="capture identify alias people.set people.set_once set_config register register_once unregister opt_out_capturing has_opted_out_capturing opt_in_capturing reset isFeatureEnabled onFeatureFlags getFeatureFlag getFeatureFlagPayload reloadFeatureFlags group updateEarlyAccessFeatureEnrollment getEarlyAccessFeatures getActiveMatchingSurveys getSurveys getNextSurveyStep onSessionId setPersonProperties".split(" "),n=0;n<o.length;n++)g(u,o[n]);e._i.push([i,s,a])},e.__SV=1)}(document,window.posthog||[]);posthog.init("${process.env.POSTHOG_PROJECT_API_KEY}", ${JSON.stringify(options)});`,
	},
]

export default developerConsentManagerServices
```

nandereck:
> 🙌 This seems great so far!
> 
> 🚧 Only blocking comment is around `capture_pageview: false`. I think we may want to leave that enabled, otherwise I think we'll only log client-side navigation events, and will miss initial page loads.

You're 100% correct. Updated [here](https://github.com/hashicorp/dev-portal/pull/2543/commits/0bb3760158666c8d62cc19d8fda7d4ddd913f91c)!


> 💭 Vaguely related to the above, and very much _not_ blocking, maybe we could split out the `posthog.init()` config options, and then stringify them into `body`, so they're easier to modify:

I dig this a lot. Also, updated [here](https://github.com/hashicorp/dev-portal/pull/2543/commits/0bb3760158666c8d62cc19d8fda7d4ddd913f91c).

zchsh (approved):

PR #2550, opened 2024-08-19
implement adr-001 env changes for unified docs

zchsh:
This PR implements #2548.

In more detail, this PR introduces three new environment variables:

- `DOCS_STATIC_PATHS_API` - will be set to <https://content.hashicorp.com> initially. Intent is to scope work around this functionality separately from the unified docs API migration.
- `MKTG_CONTENT_DOCS_API` - will be set to <https://content.hashicorp.com> initially. Docs functions that are not yet migrated will use this env variable.
- `UNIFIED_DOCS_API` - will be set to point to our new unified docs API, currently <https://web-presence-experimental-docs.vercel.app/>. The migration process will be to switch individual docs sections to use this new variable.

This PR also removes the existing `MKTG_CONTENT_API` env variable, as with the introduction of the above environment variables, it is now unused.

### DIFF
--- a/.env
+++ b/.env
@@ -9,7 +9,58 @@ NEXT_PUBLIC_ALGOLIA_SEARCH_ONLY_API_KEY=2cd9898a6c253bfa3965d2b62a4f7f3d
 # which takes precedence over the NEXT_PUBLIC_ALGOLIA_INDEX env var.
 # NEXT_PUBLIC_ALGOLIA_INDEX=product_WAYPOINT
 
-MKTG_CONTENT_API="https://content.hashicorp.com"
+#
+# EXISTING DOCS API
+#
+# `content.hashicorp.com`, deployed from `mktg-content-workflows`, is currently
+# used to provide docs content to `dev-portal`.
+#
+# We intend to migrate away from this use, and adopt our new unified docs
+# service, which is backed by a single content repository. We want to approach
+# this migration incrementally. This `MKTG_CONTENT_DOCS_API` environment
+# variable is intended for use during migration, allowing not-yet-migrated
+# content sources to continue using the existing `content.hashicorp.com` API.
+#
+# Once we've fully migrated to use `UNIFIED_DOCS_API`, we'll be able to remove
+# this environment variable.
+MKTG_CONTENT_DOCS_API="https://content.hashicorp.com"
+
+#
+# NEW UNIFIED DOCS API
+#
+# The new unified docs service is a single content repository that serves
+# documentation content for all HashiCorp products. This service is intended to
+# replace the docs endpoints of the existing `content.hashicorp.com` API.
+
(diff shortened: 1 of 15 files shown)


.env
```diff
@@ -9,7 +9,58 @@ NEXT_PUBLIC_ALGOLIA_SEARCH_ONLY_API_KEY=2cd9898a6c253bfa3965d2b62a4f7f3d
...
+# page view analytics.
+#
+# For this "static paths from analytics" purpose, we do not have a replacement
+# in mind in the short-term.
+#
+# We could potentially migrate the existing service to some other location,
+# perhaps refactoring it (as it probably needs updates). However, we also have
+# the option to leave this service alone for now, which may help us focus on the
+# more critical parts of the migration to our new unified docs service.
+DOCS_STATIC_PATHS_API="https://content.hashicorp.com"
```
prestonbourne, line 63: huge fan of how you broke them into different vars by functionality 

down the line this will make feature flagging or even just a simple function that returns differently based on where its being used alot easier to implement.

RE: [the pr that started this conversation](https://github.com/hashicorp/web-presence-experimental-docs/pull/23)
this makes me wonder how you're looking at previews now. As someone is working on unified docs is it more useful to see 
1. A preview of how it would look if all routes fetched from `UNIFIED_DOCS_API`
2. A preview of what it'll look like during migration, ie the preview should respect the existing `MKTG_CONTENT_DOCS_API` whilst showing some routes that pull from `UNIFIED_DOCS_API`

Intuitively i lean towards the second. it stands to reason that if we hypothetically plan to do this migration route by route, there should be a way for one to configure previews to fetch from the `UNIFIED_DOCS_API` route by route *before* or in sync with when `dev-portal` expects the change to occur

zchsh: Yeah this is a great point! The way I'm looking at it, I think both modes of preview will be valuable.

As you mentioned, the latter "respect where the migration is so far" mode probably makes sense as a default, as it will probably be more useful to content authors. Folks who are doing prototype work will still want a way to force _all_ docs routes to use a custom docs content API address, regardless of what's actually been migrated.

To go into a bit more detail in case it's helpful, both preview modes should both be possible with this revised `.env` setup:

1. `All routes preview` - intent being to force global use of the new unified docs API across all routes, ignoring any in-progress migration work such as feature flags. To do this, we'd set both the `UNIFIED_DOCS_API` and `MKTG_CONTENT_DOCS_API` env variables to the `localhost` address (for local previews) or to the API's deploy preview URL (for deploy previews).
2. `Migrated routes preview` - intent being to respect the existing migration status, which as of right now is that nothing is migrated. To do this, we'd set only the `UNIFIED_DOCS_API` env variable to the `localhost` address (for local previews) or to the API's deploy preview URL (for deploy previews).

prestonbourne (approved):

PR #2537, opened 2024-08-06
feat: create new product nav data structure

LeahMarieBush:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1206815131022336/1207832908083300/f) 🎟️

## 🗒️ What

This PR creates the new product nav data structure and fills in the correct information. When the new nav-panel package is ready to be imported, this data will be used to pass into the component.

## 🤷 Why

Currently our product dropdown only supports name and icon. Since the new designs feature more information, the data structure had to be updated to support it.

## 🧪 Testing

Since this is not being used anywhere yet, it can not be tested. The nav data will be used once [this Asana task](https://app.asana.com/0/1206815131022336/1207832908083297/f) is ready to start.

PR #2553, opened 2024-08-21
refactor: Use shared dropdown panel in nav

EnMod:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-feat-use-shared-dropdown-panel-hashicorp.vercel.app) 🔎
- [Asana task](https://app.asana.com/0/1207041417996032/1207832908083297/f) 🎟️
- [Figma spec](https://www.figma.com/design/gknlCxqoNsfrW5FAt11z9u/Developer-Menu?node-id=0-1&t=YCZpacxsob8OVNHm-0)

## 🗒️ What

This PR imports the `ProductPanel` component from `@hashicorp/react-components` for use in the global nav, replacing the Products dropdown on generic pages on both desktop and mobile. This is a follow-up to #2537, utilizing the new data structure prepared in that PR.

## 🤷 Why

By using this shared panel, we can maintain a single dropdown component across both `dev-portal` and `web`, with minimal adjustments to align to the styles in `dev-portal`.

## 🛠️ How

Importing a canary of the `@hashicorp/react-components` package was the first step, as this canary exports a `ProductPanel` (fixes from https://github.com/hashicorp/web/pull/1865, primarily) ready for external use. After that, I moved forward with implementation in `dev-portal`, replacing the "Products" dropdown with the new panel. This required some CSS hacking that will be documented in this PR; open to discussions on working around the need for these hacks!

Finally, side panel content support was added via https://github.com/hashicorp/web/pull/1874, and other adjustments made to align with our spec.

## 📸 Design Screenshots

<details><summary>Spec (includes side content)</summary>

Screenshot 2024-08-21 at 16 11 19

</details> 

<details><summary>Before</summary>

image

</details> 

<details><summary>After</summary>

Screenshot 2024-09-04 at 15 24 50

</details> 

## 🧪 Testing

- [ ] Open the preview link.
- [ ] Hover over the Products dropdown.
- [ ] Validate that the newly imported `ProductPanel` is shown instead of the product list.
- [ ] Shrink the browser width down to mobile.
- [ ] Open the mobile menu.
- [ ] Validate that the newly imported `ProductPanel` is shown instead of the "Main menu" and "Products" sections shown in production.

## 💭 Anything else?

Looking for feedback on implementation (especially from engineers more familiar with `dev-portal`) to see if there's a more elegant way to retrofit the new component into our current `dev-portal` code 🙏


package.json
```diff
@@ -209,7 +211,7 @@
 		"zod": "^3.20.2"
 	},
 	"overrides": {
-		"next": "$next",
+		"next": "^14.2.1",
```
heatlikeheatwave, line 214: Can you share what this change is for?

src/components/navigation-header/components/dropdown-menu/index.tsx
```diff
@@ -204,105 +205,120 @@ const NavigationHeaderDropdownMenu = ({
 				id={menuId}
 				style={{ display: isOpen ? 'block' : 'none' }}
 			>
-				{itemGroups.map(
```
mikegolus, line 207: Nit: Reducing the amount of inline return statements in this JSX could improve readability. In this case, `ItemGroup` or even `ItemGroups` could be defined as a separate component.

src/components/navigation-header/components/dropdown-menu/index.tsx
```diff
@@ -204,105 +205,120 @@ const NavigationHeaderDropdownMenu = ({
...
-										{label}
-									</Text>
-								)}
-								<ul
-									aria-labelledby={itemGroupLabelId}
-									className={classNames(s.itemGroup, {
-										[s.twoColumns]: items.length >= 10,
-									})}
-								>
-									{items.map(
```
mikegolus, line 236: Nit: Similar to my comment above, perhaps readability could be improved by making `Item` a separate component.

package.json
EnMod: Ah, good catch! This was meant to allow `dev-portal` to run locally without issue. I have been getting an error stating that Next didn't understand what `$next` was referencing, I believe 🤔

The issue seems to be resolved after a downgraded my Node version, so I'll revert this :+1:

EnMod: Addressed here:
3bd0896

src/components/command-bar/components/dialog/command-bar-dialog.module.css
heatlikeheatwave, line 1: Can you share context on why these changes were necessary when this change was scoped to the nav?

src/components/mobile-menu-levels-generic/index.tsx
```diff
@@ -27,14 +28,17 @@ import s from './mobile-menu-levels-generic.module.css'
...
-			<ul className={s.mobileMenuNavList}>
-				<SidebarNavMenuItem item={{ heading: 'Main Menu' }} />
-				{generateTopLevelSubNavItems().map((item: $TSFixMe, index: number) => (
-					// eslint-disable-next-line react/no-array-index-key
-					<SidebarNavMenuItem item={item} key={index} />
-				))}
-			</ul>
+			<div className={s.mobileMenuAuthContainer}>
+				<MobileAuthenticationControls />
+			</div>
```
heatlikeheatwave, line 33:
```suggestion
			<div className={s.mobileMenuAuthContainer}>
				<MobileAuthenticationControls />
			</div>
```

src/components/navigation-header/components/dropdown-menu/index.tsx
EnMod: @mikegolus This is a good point, and it would definitely help readability. 

Unfortunately, the JSX for `Item` and `itemGroup` rely on many functions defined in `DropdownMenu` and invoked in the JSX. To migrate `Item` and `ItemGroup` to their own components would require significant refactoring to ensure all the functions called by the original code all run as expected, passed in as props or otherwise reimplemented to work 😭 

I will make a note of this and your other comment though, to follow-up on!

src/components/command-bar/components/dialog/command-bar-dialog.module.css
EnMod: @heatlikeheatwave Apologies for not highlighting this change 🙇 This addresses some QA feedback from Kaitlynne re: accessibly clickable area around the "Give feedback" link. Given the choice between "address it now" or "wait till later", I chose the former, though I can revert this change to tighten up the scope of this PR.

src/components/mobile-menu-levels-generic/index.tsx
EnMod: Unfortunately that component doesn't support `className` in its types or props. I can add support, however! It would be a more elegant solution than the extra div for sure.

EnMod: Support added here:
9ceef62

With a follow-up commit to tweak the `margin-bottom` here, though it might even be best to remove `margin-bottom` on the component proper 🤔 
8300524

LeahMarieBush:
I found a few design differences between the preview and figma:

Mobile learn more panel items look too far apart
Screenshot 2024-09-04 at 4 21 09 PM

Mobile space between the last item and the promo looks too big
Screenshot 2024-09-04 at 4 21 27 PM

Mobile space between first item and the buttons above looks too big
Screenshot 2024-09-04 at 4 21 51 PM

src/components/mobile-menu-container/index.tsx
```diff
@@ -41,7 +44,10 @@ const MOBILE_MENU_MOTION = {
  * Handles rendering the Sign In and Sign Up UI elements in mobile viewports.
  * Intended to be used alongside `MobileMenuContainer`.
  */
-const MobileAuthenticationControls = () => {
+
+const MobileAuthenticationControls = ({
+	className,
+}: MobileAuthenticationControlsProps) => {
```
heatlikeheatwave, line 50: If you change it to this, then you can revert the changes to `src/components/mobile-menu-container/types.ts`. You only need to add `HTMLAttributes` to the line importing from react, e.g. `import { ForwardedRef, forwardRef, HTMLAttributes } from 'react'`

```suggestion
}: HTMLAttributes<HTMLDivElement>) => {
```

LeahMarieBush: Awesome suggestion! 🎉 I updated the types in [this commit](https://github.com/hashicorp/dev-portal/pull/2553/commits/05a1eb4741dc35e7602004a8aca13d2deb2202cf)

src/components/navigation-header/components/nav-bar-list-container/index.tsx
```diff
@@ -22,8 +23,8 @@ export default function NavBarListContainer({
 	children: ReactNode
 }) {
 	return (
-		<nav className={s.nav}>
+		<NavigationMenu.Root className={s.nav}>
 			<ul className={s.ul}>{children}</ul>
```
heatlikeheatwave, line 27: If the nav is replaced by `<NavigationMenu.Root>`, should the `<ul>` also be replaced by `<NavigationMenu.List>`?

src/contexts/mobile-menu.tsx
```diff
@@ -19,7 +19,7 @@ import { useNoScrollBody } from 'hooks/use-no-scroll-body'
 /**
  * Should correspond to --dev-dot-hide-mobile-menu
  */
-const DEFAULT_NAV_HEADER_DESKTOP_WIDTH = 924
+const DEFAULT_NAV_HEADER_DESKTOP_WIDTH = 1230
```
heatlikeheatwave, line 22: just curious about how you landed on `1230` here

EnMod: Deferring to @LeahMarieBush as I think she had made this adjustment :+1:

heatlikeheatwave (changes requested):
Thanks for all your work on this! I tagged Ruben as a reviewer to make sure this follows WCAG guidelines.

The only concern I have right now is that the tab order doesn't follow the visual order of elements in the product panel for desktop. IIRC, keyboard navigation should follow visual order of elements (left to right, top to bottom). Tabbing through the product panel doesn't follow this order (tabbing goes top to bottom, left to right). 

More info [here ](https://webaim.org/techniques/keyboard/) in the `Navigation order` section.

src/contexts/mobile-menu.tsx
LeahMarieBush: @heatlikeheatwave This got changed because the new product panel is wider than the previous. 1230 is around when the panel gets to the edge of the screen. Without this, the product panel would've pushed off the side of the page and caused white space on the right side

heatlikeheatwave: That makes sense. Did you try changing the position of the panel, so it is more centered in the screen? Like so 👇 

Screenshot 2024-09-05 at 14 54 01

heatlikeheatwave:
> Thanks for all your work on this! I tagged Ruben as a reviewer to make sure this follows WCAG guidelines.
> 
> The only concern I have right now is that the tab order doesn't follow the visual order of elements in the product panel for desktop. IIRC, keyboard navigation should follow visual order of elements (left to right, top to bottom). Tabbing through the product panel doesn't follow this order (tabbing goes top to bottom, left to right).
> 
> More info [here ](https://webaim.org/techniques/keyboard/) in the `Navigation order` section.

@ruben took a look and gave it the okay -- as long as it's consistent across the dropdown menus!

EnMod:
@heatlikeheatwave By "it" do you mean the left-to-right, top-to-bottom tab ordering? @LeahMarieBush just got some changes into the package that restore the original product/tab ordering.

We wanted to follow that up with support for a separate product list in `lib/products.ts`, to ensure the correct ordering of products on mobile. Does this make sense?

RubenSandwich:
@EnMod and @heatlikeheatwave For focus order here is what the [WCAG 2.1 SC 2.4.3 Focus Order (Level A) says:](https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html#:~:text=The%20focus%20order,their%20Web%20pages.)

> The focus order may not be identical to the programmatically determined reading order (see Success Criterion 1.3.2) as long as the user can still understand and operate the Web page. Since there may be several possible logical reading orders for the content, the focus order may match any of them. However, when the order of a particular presentation differs from the programmatically determined reading order, users of one of these presentations may find it difficult to understand or operate the Web page. Authors should carefully consider all these users as they design their Web pages.

So it's not always above up and down navigation with tabbing, it is about "logical navigation". To get even deeper, using tabs is navigation menus is actually frown upon. [(See this ARIA practices example.)](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-navigation/) As you are supposed to be able to tab from widget to widget and menus **should** be treated as a single widget. (With the arrows keys used inside widgets for navigation.) Which when you think about it makes sense because otherwise someone will have to press tab a bunch of times to travel through your website.

Buuuttt this is a place where everyone breaks the rules, cause folks new to a11y demanded that everything be "tab-able" because they didn't understand that there are deeper navigation patterns. There heart is in the right place, but they didn't go deeper. 🤦🏻 (don't get me started even more on it.)

So as long as we have "logical navigation" we are fine.

RubenSandwich (changes requested):
Few accessibility issues I found:

1. The learn List items seem to be missing focus indicators.
2. The main navigation links do not have accessible names:

Screenshot 2024-09-05 at 3 46 02 PM

I suggest putting both the name and the description in an aria-label on the <li>.

src/components/navigation-header/components/nav-bar-list-container/index.tsx
EnMod: That's a good idea @heatlikeheatwave, implemented in this commit:
0a2b19f

src/contexts/mobile-menu.tsx
EnMod: @heatlikeheatwave Ah, so the panel would drop down as if it was aligned to the logo, instead of to the "Products" trigger in the nav? I think that works, would have to test it out of course!

However, is there a plan to have more dropdowns in the nav in the future? If there is, I would assume they would drop down aligned to their triggers, which might make this product dropdown look out-of-place by comparison 🤔

heatlikeheatwave: > Ah, so the panel would drop down as if it was aligned to the logo, instead of to the "Products" trigger in the nav?

Yeah, exactly! That way we don't merge in a change that impacts the layout of other pages where this menu does not exist.

> However, is there a plan to have more dropdowns in the nav in the future? 

Not that I currently know of. The nav is different for product documentation & tutorials, and seems like it will be for the foreseeable future (unless there's another dev portal nav project in the works?).

I think it makes sense to adjust the position of the nav to fit within the existing breakpoints as that's our current use case, rather than making a change that impacts the layout of _all_ pages, but could accommodate a future and unknown nav change. Does that make sense?

EnMod: This makes sense @heatlikeheatwave ! This approach will still leave the nav looking kinda cut off between widths `924px` and `1013px` though by my calculations (`1013px` = `989px` dropdown width on desktop + `24px` to align the panel as it is in your screensheet, and `924px` is when the panel would change layout to mobile on developer). 

Given the small window of viewports this cutoff will appear in, that seems like a fair trade to me to avoid affecting the layout of other pages while minimizing visual issues. Is this reasonable?

RubenSandwich (approved):
a11y looks good. Thanks @EnMod!

src/contexts/mobile-menu.tsx
EnMod: @heatlikeheatwave It does make sense :+1: I've implemented this change locally and it does the trick, but I'm getting feedback from design on how this looks before I commit anything.

EnMod: @heatlikeheatwave Update: design is onboard with moving the panel over! Ben suggested also reducing nav item padding, but I will see how the panel behaves with just moving it over first, as that padding change will affect both usages of the panel.

heatlikeheatwave: Awesome! Thanks for following up with design about it. I appreciate all the extra work you've put into this feature (:

EnMod: Thanks @heatlikeheatwave 🎉 

I've implemented the move-over in this commit:
e8809fb9e7369a05635c23cace9d53795fccd2de

Changing the padding of items did nothing to the width of the dropdown, I'm assuming because of the use of CSS grid for layout. I don't think the `16px` of saved space is a blocker for launch, personally, but it is something to see if we can address in a follow-up if we don't get to it 👍

src/components/navigation-header/components/dropdown-menu/dropdown-menu.module.css
```diff
@@ -63,9 +63,13 @@
 	border-bottom-left-radius: 6px;
 	border-bottom-right-radius: 6px;
 	box-shadow: var(--token-elevation-higher-box-shadow);
-	min-width: 252px;
+	min-width: max-content;
+	position: fixed;
```
heatlikeheatwave, line 67: Is it possible to change this back to `absolute` from`fixed`? Asking because having it as `fixed` has caused visual regressions. If you scroll down the page and then open the new menu on desktop on the homepage, it is not at the top of the page as expected. Same for other pages

Screenshot 2024-09-09 at 16 16 04


Screenshot 2024-09-09 at 16 14 42

EnMod: Yikes! Good call. Addressed here:
ec41196

I had gone for `fixed` initially for a less-bespoke `left` value to get the panel aligned. Is there anything you would suggest other than the hardcoded `left: -216px` I now have in place?

src/components/mobile-menu-levels-generic/mobile-menu-levels-generic.module.css
```diff
@@ -4,10 +4,15 @@
  */
 
 .mobileMenuContainer {
-	padding: 24px;
 	z-index: 4;
 }
 
+.mobileMenuAuthContainer {
+	padding: 16px;
+	padding-bottom: 0;
```
heatlikeheatwave, line 12:
```suggestion
	padding: 16px 16px 0 16px;
```

heatlikeheatwave (approved):

PR #2562, opened 2024-09-05
Revert "Remove "Edit on GitHub" link from terraform docs"

rmainwork:
This PR reverts hashicorp/dev-portal#2540, which was overly broad in it's scope when hiding the "Edit on GitHub" link at the bottom of dev docs. This caused the link to disappear for all Terraform products rather than just Terraform Enterprise(the intended target). 

[A ticket to re-do this logic with narrower scope has been opened](https://app.asana.com/0/1207947026228781/1208226840548303/f)

### DIFF
--- a/src/views/docs-view/server.ts
+++ b/src/views/docs-view/server.ts
@@ -418,9 +418,9 @@ export function getStaticGenerationFunctions<
 			 * Note: If we need more granularity here, we could change this to be
 			 * part of `rootDocsPath` configuration in `src/data/<product>.json`.
 			 */
-			const isPublicContentRepo = !['hcp', 'sentinel', 'terraform'].includes(
-				product.slug
-			)
+			const isHcp = product.slug == 'hcp'
+			const isSentinel = product.slug == 'sentinel'
+			const isPublicContentRepo = !isHcp && !isSentinel
 			if (isPublicContentRepo) {
 				layoutProps.githubFileUrl = githubFileUrl
 			}


src/views/docs-view/server.ts
```diff
@@ -418,9 +418,9 @@ export function getStaticGenerationFunctions<
 			 * Note: If we need more granularity here, we could change this to be
 			 * part of `rootDocsPath` configuration in `src/data/<product>.json`.
 			 */
-			const isPublicContentRepo = !['hcp', 'sentinel', 'terraform'].includes(
-				product.slug
-			)
+			const isHcp = product.slug == 'hcp'
+			const isSentinel = product.slug == 'sentinel'
+			const isPublicContentRepo = !isHcp && !isSentinel
```
heatlikeheatwave, line 423: Since `isHcp` and `isSentinel` aren't used elsewhere, I think we can simplify it to this:
```suggestion
			const isPublicContentRepo = product.slug === 'hcp' && product.slug === 'sentinel'
```
WDYT?

rmainwork: I like this! The only thing that gives me pause for thought about this is that this PR is just to revert the code to it's previous state. What if we kept this as-is in this PR and then I borrowed this snippet for my PR to correct the behavior? Would you be good with that?

rmainwork: [Thusly](https://github.com/hashicorp/dev-portal/commit/ee172e16c9d8c8cda876731990d1b4f22adf70ec)

heatlikeheatwave (approved):

rmainwork:
Thanks Heat!

PR #2563, opened 2024-09-09
get latest redirects based on nav-data not version metadata

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

This PR avoids using the `ref` from `version-metadata` in order to fetch the "latest" redirects from content source repositories.

## 🤷 Why

This work is in service of [[DEVDOT-???] Release branches as docs single source of truth](https://hermes.hashicorp.services/document/1VEbwSwoO2dB6rb-pBUY9CKOVoislVo3YcbuRTEbnMew?draft=true). At present, our reliance on `version-metadata` is out of convenience, and assumes that `version-metadata` will have its `ref` and `sha` updated on _every_ content upload, even when there have been zero changes to version metadata. This method of constant updates seems to be causing issues that outweigh the convenience of this approach.

## 🛠️ How

This PR refactors `getLatestContentRefForProduct`, which upstream fetches the latest `ref` from `version-metadata`, into `getLatestContentShaForProduct`, which now fetches the latest `sha` from known `nav-data` endpoints.

## 🧪 Testing

Tests have been added to assert that `redirects` are successfully fetched with this approach. In addition, we have a more ephemeral demo, formatted as a test file for convenience, that asserts that the redirects fetched using the `sha` from `nav-data` are identical to redirects fetched using the `ref` from `version-metadata`:

```typescript
/**
 * OVERVIEW
 *
 * Intent here is to demo that redirects fetched using the latest SHA from
 * content match those fetched with refs from the latest version metadata.
 *
 * Note that as we make changes to how version metadata is updated, we expect
 * this demo will no longer work. This is intentional, as version metadata will
 * start reflecting the `ref` and `sha` at the last point the version metadata
 * was updated, which will often fall behind the `sha` of the latest content
 * upload.
 *
 * HOW TO USE THIS FILE
 *
 * Drop this file in `build-libs/__tests__/latest-content-sha-demo.test.ts`,
 * in the `dev-portal` repo, then run `npm run test -- latest-content-sha-demo`.
 */

import getLatestContentShaForProduct from '../get-latest-content-sha-for-product'
import fetchGithubFile from '@build-libs/fetch-github-file'
import { PRODUCT_REDIRECT_ENTRIES } from '@build-libs/redirects'

/**
 * Mirror the previous implementation of the function, which fetched
 * version metadata. We'll use this to compare the new `sha` approach with
 * the existing `sha` values that correspond to each `ref`.
 */
async function getLatestVersionMetadata(product) {
	const contentUrl = new URL('https://content.hashicorp.com')
	contentUrl.pathname = `/api/content/${product}/version-metadata/latest`
	return await fetch(contentUrl.toString()).then((res) => res.json())
}

/**
 * Run demo "tests"
 */
describe('latestContentShaDemo', () => {
	PRODUCT_REDIRECT_ENTRIES.forEach(({ repo, path }) => {
		/**
		 * Demo that the SHA matches between nav-data and version metadata.
		 * This is mostly what we care about, we're asserting that the `sha`
		 * in `nav-data` references an equivalent point in git history as the
		 * `ref` in version metadata.
		 */
		it(`demos that the latest SHA for nav-data in "${repo}" matches the SHA in the latest version metadata`, async () => {
			const latestSha = await getLatestContentShaForProduct(repo)
			const latestVersionMetadata = await getLatestVersionMetadata(repo)
			expect(latestSha).toBe(latestVersionMetadata.result.sha)
		})

		/**
		 * Demo that the redirects file is identical, which is mostly asserting
		 * that `fetchGithubFile` can work with a `ref` or `sha`. We skip
		 * private repos, as fetching that data would require authentication.
		 */
		if (['hcp-docs', 'sentinel', 'ptfe-releases'].includes(repo)) {
			console.log(`Skipping test for "${repo}" repo, as it's a private repo.`)
		} else {
			it(`demos that the redirects fetched from "${repo}" using the latest version metadata ref match redirects fetched from "${repo}" using the latest nav-data SHA`, async () => {
				// Previous approach, using version metadata
				const latestVersionMetadata = await getLatestVersionMetadata(repo)
				const redirectsFromRef = await fetchGithubFile({
					owner: 'hashicorp',
					repo: repo,
					path: path,
					ref: latestVersionMetadata.result.ref,
				})
				// New approach, using SHA from nav-data
				const latestSha = await getLatestContentShaForProduct(repo)
				const redirectsFromSha = await fetchGithubFile({
					owner: 'hashicorp',
					repo: repo,
					path: path,
					ref: latestSha,
				})
				expect(redirectsFromRef).toBe(redirectsFromSha)
			})
		}
	})
})
```

## 💭 Anything else?

Not at the moment!

[preview]:  https://dev-portal-git-zsref-for-redirects-hashicorp.vercel.app
[task]: https://app.asana.com/0/1207899860738460/1208264613311743/f


build-libs/__tests__/get-latest-content-sha-for-product.test.ts
rmainwork, line 1: I may be overlooking something here, but why is this a `.ts` file while the other two are `.js`?

rmainwork (approved):
LGTM 🚀 

Not going to hold the review up over that comment - was just curious.

build-libs/__tests__/get-latest-content-sha-for-product.test.ts
zchsh: I think we'd prefer to use TS in both places... but we import the `redirects` stuff into `next.config.js`, so it needs to be JS rather than TS. Tests can be TS since they don't need to run directly in that context 👍

rmainwork: Ah! Yeah that makes sense.

build-libs/get-latest-content-sha-for-product.js
```diff
@@ -0,0 +1,53 @@
...
+	nomad: '/api/content/nomad/nav-data/latest/docs',
+	vault: '/api/content/vault/nav-data/latest/docs',
+	vagrant: '/api/content/vagrant/nav-data/latest/docs',
+	packer: '/api/content/packer/nav-data/latest/docs',
+	consul: '/api/content/consul/nav-data/latest/docs',
+	'terraform-docs-common':
+		'/api/content/terraform-docs-common/nav-data/latest/docs',
+	'hcp-docs': '/api/content/hcp-docs/nav-data/latest/docs',
+	'ptfe-releases': '/api/content/ptfe-releases/nav-data/latest/enterprise',
+	sentinel: '/api/content/sentinel/nav-data/latest/sentinel',
```
heatlikeheatwave, line 28: Are the other Terraform repos with nav data out of scope for this PR?

zchsh: Great question! I don't think we need to worry about other content source repos in this context, as this is the same set of content sources from which we fetch redirects [upstream](https://github.com/hashicorp/dev-portal/blob/24199bc75838857d31f7a87cafc55df8c9b50af1/build-libs/redirects.js#L117).

Or to put it another way, it seems that all Terraform redirects are authored in either `terraform-docs-common` or `ptfe-releases` (maybe to reduce the risk of collisions?).

build-libs/__tests__/get-latest-content-sha-for-product.test.ts
```diff
@@ -0,0 +1,33 @@
...
+
+describe('getLatestContentShaForProduct', () => {
+	PRODUCT_REDIRECT_ENTRIES.forEach(({ repo, path }) => {
+		it(`fetches the latest SHA for the "${repo}" repo`, async () => {
+			const latestSha = await getLatestContentShaForProduct(repo)
+			expect(typeof latestSha).toBe('string')
+		})
+
+		if (['hcp-docs', 'sentinel', 'ptfe-releases'].includes(repo)) {
+			console.log(`Skipping test for "${repo}" repo, as it's a private repo.`)
```
heatlikeheatwave, line 18: nit
```suggestion
			console.log(`Skipping test for private repo "${repo}"`)
```

build-libs/get-latest-content-sha-for-product.js
heatlikeheatwave: Got it -- thanks for explaining!

heatlikeheatwave (approved):
LGTM

RubenSandwich (approved):
I agree with this change, we might as well go to the source of truth rather than having a redirects in our logic.

prestonbourne (approved):

PR #2576, opened 2024-09-23
chore: Refactor logic for docs versions dropdown and added tests

heatlikeheatwave:
## 🗒️ What

- Refactor logic for version dropdown to also work for release notes pages and moved function out of `server.ts` to make it testable (now in `fetchValidVersions()`)
- Added function `isReleaseNotesPage`
- Added tests for:
  - `isReleaseNotesPage()`
  - `fetchValidVersions()`
  - `getValidVersions()`

## 🤷 Why

- Previous logic did not have to account for release notes pages where a version may be in the path 1-2 times, which created a bug where incorrect document versions were fetched.

For example, the previous logic would call fetch versions that contained the directory index page, rather than the specific release notes page within that directory. The previous version called the API endpoint with `/releases/2024` on the path `/releases/2024/v202409-1`, rather than the full path (`/releases/2024/v202409-1`).

### TFE
- Visit TFE docs landing [page](https://dev-portal-git-heat-chore-version-dropdown-rel-afa33f-hashicorp.vercel.app/terraform/enterprise)
- Version dropdown should be visible

### PTFE
- Visit PTFE release notes [page](https://dev-portal-git-heat-chore-version-dropdown-rel-afa33f-hashicorp.vercel.app/terraform/enterprise/releases/2024/v202409-1)
- Open the dropdown
- It should match this screenshot:
Screenshot 2024-09-23 at 17 33 11

### Vault
- Navigate to versioned release notes [page](https://dev-portal-git-heat-chore-version-dropdown-rel-afa33f-hashicorp.vercel.app/vault/docs/release-notes/1.15.0)
- Version dropdown should match this screenshot:
Screenshot 2024-09-23 at 17 36 36
- Navigate to any other page (that is not a versioned release notes page), such as the release notes landing [page](https://dev-portal-git-heat-chore-version-dropdown-rel-afa33f-hashicorp.vercel.app/vault/docs/release-notes)
- Version dropdown should be visible

### Versioned TFE 404 page
- Navigate to this [page](https://dev-portal-git-heat-chore-version-dropdown-rel-afa33f-hashicorp.vercel.app/terraform/enterprise/v202210-1/releases/2023/v202310-1)
- Page should contain the text "This page does not exist for version v202210-1."


src/lib/docs/is-release-notes-page.ts
```diff
@@ -0,0 +1,22 @@
...
+ * - `releases/YYYY/vYYYYMM-NN`
+ * - `/releases/YYYY/vYYYYMM-NN`
+ * - `release-notes/vX.X.X` or `/release-notes/X.X.X`
+ *
+ * @param path - The path to be checked.
+ * @returns `true` if the path matches the release notes pattern, otherwise `false`.
+ */
+export const isReleaseNotesPage = (path: string) =>
+	/(\/?releases\/\d{4}\/(v\d{6}-\d{1}))$|\/?release-notes\/(v\d+[.|_]|(\d+[.|_]))\d+[.|_]([0-9]|x)$/i.test(
+		path
```
aaronvanderlip, line 21: Regex can be a challenge for humans to parse. There could be an advantage here to creating an array of Regex conditions for each of the 4 patterns, and then using array.some() to test for a match.


```
const  patterns = [ /regex-pattern1/, /regex-pattern-2/]

return patterns.some(pattern => pattern.test(path))

```

src/views/docs-view/utils/__tests__/get-valid-versions.test.ts
```diff
@@ -0,0 +1,98 @@
...
+ */
+
+import { describe, it, expect, vi } from 'vitest'
+import { getValidVersions } from '../get-valid-versions'
+import type { VersionSelectItem } from '../../loaders/remote-content'
+
+// Mock fetch
+global.fetch = vi.fn() as typeof fetch
+
+describe('getValidVersions', () => {
```
aaronvanderlip, line 13: 🙇

src/views/docs-view/utils/set-document-path.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import { isReleaseNotesPage } from 'lib/docs/is-release-notes-page'
+import path from 'node:path'
+import { getValidVersions } from './get-valid-versions'
+import { VersionSelectItem } from '../loaders/remote-content'
+
+export async function fetchValidVersions(
```
aaronvanderlip, line 11: An observation and opinion about utils. This function is used once, and in general I become wary of structuring code around a utils pattern as it can become kind of a junk drawer whereas a function used once could also be co-located with the code that calls it. Nothing to change here, but could be a topic for 'how we work' for some later date.

src/views/docs-view/server.ts
```diff
@@ -387,15 +387,11 @@ export function getStaticGenerationFunctions<
...
 			 * Filter versions to include only those where this document exists
 			 */
-			// Construct a document path that the content API will recognize
-			const pathWithoutVersion = pathParts
-				.filter((part) => part !== versionPathPart)
-				.join('/')
-			const fullPath = `doc#${path.join(basePathForLoader, pathWithoutVersion)}`
-			// Filter for valid versions, fetching from the content API under the hood
-			const validVersions = await getValidVersions(
+			const validVersions = await fetchValidVersions(
```
aaronvanderlip, line 390: Nice improvement. I think the risk with these types of sprawling function is to add yet another line of transforms so moving those transforms into `getValidVersions` is move towards better encapsulation.

src/views/docs-view/utils/set-document-path.ts
```diff
@@ -0,0 +1,59 @@
...
+		/**
+		 * Check specific to PTFE releases notes page, which may have a version in the path twice
+		 * e.g. v202409-2/releases/2024/v202407-1
+		 * Remove the first version instance, which is the docs version
+		 * e.g. releases/2024/v202407-1
+		 * the mdx file this page pulls from has a version in the title (e.g. 202407-1mdx)
+		 * the second version is the path should not be removed for this reason.
+		 * This block is here because the default (else statement below)
+		 * removes all versions from the path, which is not desired for release notes.
+		 */
```
aaronvanderlip, line 33: Setting up your future self for success with this comment.

aaronvanderlip (approved):
Very solid work. I appreciate the commitment to testing.

src/views/docs-view/utils/set-document-path.ts
heatlikeheatwave: That's a great point. My thinking here was that it would be easier to test if this functionality was moved out of server.ts, but I suppose I could define the function at the top of the file and export it so it can still be tested. WDYT?

aaronvanderlip: I would move the function to the top of server.ts and export it. One less file, one less import, one less context jump for the future developer.

src/lib/docs/is-release-notes-page.ts
heatlikeheatwave: Do you think having the examples in the function comment would help with this?
```
 * - `vYYYYMM-NN/releases/YYYY/vYYYYMM-NN`
 * - `releases/YYYY/vYYYYMM-NN`
 * - `/releases/YYYY/vYYYYMM-NN`
 * - `release-notes/vX.X.X` or `/release-notes/X.X.X`
 ```

heatlikeheatwave: done in https://github.com/hashicorp/dev-portal/pull/2576/commits/2a22362c9e179b1e8d50ac53dbe05185bca28e9e

PR #2581, opened 2024-09-30
Add alias for docs sidebar filtering

im2nguyen:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-docs-sidebar-alias-hashicorp.vercel.app/?vercelToolbarCode=EBBgVNHMkCUIng-) 🔎
- [Asana task](https://app.asana.com/0/1202520168081556/1202841933621138/f) 🎟️

## 🗒️ What

Right now we only filter on the word in the sidebar explicitly. This feature would be to add hidden aliases to these sidebar items. As an example, adding alias "bar" to sidebar item "foo", so that when a user filtered on the text "bar", "foo" would be shown in the results. 

## 🤷 Why

The issue for the site and the reason this is being considered is because we have "use case" items in the side bar and we have products that focus on these use cases. So a user filtering on a product (or subproduct/feature) could be shown use cases as results. 

## 🛠️ How

Add an optional `alias` attribute to the docs nav JSON file.

```json
...
      {
        "title": "Compute values",
        "path": "write-configuration/expressions",
        "alias": "expressions"
      },
      {
        "title": "Manipulate data",
        "path": "write-configuration/functions",
        "alias": "functions"
      },
...
```

## 📸 Design Screenshots

This adds sidebar filtering by alias to DevDot.

image

### DIFF
--- a/src/components/sidebar/helpers/get-filtered-nav-items.ts
+++ b/src/components/sidebar/helpers/get-filtered-nav-items.ts
@@ -23,6 +23,8 @@ export const getFilteredNavItems = (
 		return items as NavItemWithMetaData[]
 	}
 
+	console.log(items)
+
 	const filteredItems = []
 
 	items.forEach((item: EnrichedNavItem) => {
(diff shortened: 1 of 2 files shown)


src/components/sidebar/helpers/get-filtered-nav-items.ts
```diff
@@ -23,6 +23,8 @@ export const getFilteredNavItems = (
 		return items as NavItemWithMetaData[]
 	}
 
+	console.log(items)
```
pbortnick, line 26: @im2nguyen can we delete this console log?

im2nguyen:
hello! there’s not a super easy way to test out the docs sidebar filtering. 

i pushed a commit here where i added aliases to the terraform docs. https://github.com/hashicorp/terraform-docs-common-internal/pull/366. if you search expressions, you should also see computed values. you’d need to build the dev-portal pr locally and spin up the terraform-docs-common-internal repo using that local dev-portal image

pbortnick (approved):
✅ Nice feature!

PR #2582, opened 2024-10-02
fix: handle 500 errors that display as 404 errors

LeahMarieBush:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-leah-fix500-error-issues-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1207947026228781/1207566071713496/f) 🎟️

## 🗒️ What

This PR fixes scenarios where 500 errors render as 404 errors.  

## 🤷 Why

When there is a 500 error in the browser the `_error.js` file will received that as a 404 error from the response for an unknown reason [(maybe due to the second caveat on this page?)](https://nextjs.org/docs/pages/building-your-application/routing/custom-error#caveats).

## 🛠️ How

To fix this, I added the `500.tsx` file that will be rendered when a 500 error is detected in the browser. This fixes the rendered of the error page and also makes sure the correct error page analytics are recorded. 

## 📸 Design Screenshots

Before - 404 versioned docs page is rendered on 500 error
Screenshot 2024-10-03 at 3 40 11 PM

After - Generic 500 page is rendered on 500 error
Screenshot 2024-10-03 at 3 26 01 PM

## 🧪 Testing

There are a few different scenarios that need to be tested:

500 error- The before link should be showing the versioned docs 404 page (This page does not exist for version v0.17.x.). The after link should be showing the generic 500 error page ()
These links do not always give a 500 error. If the page renders with the correct content, try changing the version in the dropdown a few times until you see a 500 error. Also the pages may take a while to load.
[Before](https://developer.hashicorp.com/terraform/cdktf/v0.17.x/api-reference/go)
[After](https://dev-portal-git-leah-fix500-error-issues-hashicorp.vercel.app/terraform/cdktf/v0.17.x/api-reference/go)

404 versioned docs error- Verify that both before and after render the same for a 404 error on a versioned docs page
[Before](https://developer.hashicorp.com/terraform/cdktf/v0.17.x/api-reference/leah)
[After](https://dev-portal-git-leah-fix500-error-issues-hashicorp.vercel.app/terraform/cdktf/v0.17.x/api-reference/leah)

404 error- Verify that both before and after render the same for a 404 error
[Before](https://developer.hashicorp.com/leah)
[After](https://dev-portal-git-leah-fix500-error-issues-hashicorp.vercel.app/leah)

## 💭 Anything else?

More details on nextjs 500 error handling can be found [here](https://nextjs.org/docs/pages/building-your-application/routing/custom-error#500-page)


heatlikeheatwave (commented):
Thanks for picking up this task! What do you think about pivoting here and refactoring the error handling/data flow? I think the issue here is upstream. My suspicion is that either the issue is in `_error` (possibly because of getServerSideProps, which I think is unnecessary now that sentinel has been migrated) or somewhere in the remote loader (perhaps [here](https://github.com/hashicorp/dev-portal/blob/main/src/views/docs-view/loaders/remote-content.ts#L246-L249)). While adding a 500 error page does work to show a different error message to users when the max page size has been exceeded, it may not address the underlying problem. What are your thoughts on this?

heatlikeheatwave (approved):

LeahMarieBush:
I'm waiting until tomorrow to merge this bc of HashiConf today

PR #2584, opened 2024-10-11
chore: alert banner for HashiConf

heatlikeheatwave:
## 🔗 Relevant links

- [Preview](https://dev-portal-git-heat-hashiconfalert-banner-day-one-hashicorp.vercel.app/)
- [Asana task](https://app.asana.com/0/1207947026228781/1208503549555255/f) 🎟️

## 🗒️ What

- alert banner to watch live stream of HashiConf on October 15th, 2024 8:30 am - 12:30 pm EDT

## 🧪 Testing

- Visit the [preview link ](https://dev-portal-git-heat-hashiconfalert-banner-day-one-hashicorp.vercel.app/)and see the alert banner is not there
- Change the date & time on your machine to October 15th 8:30 AM and refresh the page
- The alert banner should be visible
- Change the date & time on your machine to October 15th 12:31 PM and refresh the page
- The alert banner should **not** be visible

### DIFF
--- a/src/layouts/base-layout/index.tsx
+++ b/src/layouts/base-layout/index.tsx
@@ -6,7 +6,7 @@
 // Third-party imports
 import classNames from 'classnames'
 import AlertBanner from '@hashicorp/react-alert-banner'
-import { HTMLAttributes, useState } from 'react'
+import { HTMLAttributes, useEffect, useState } from 'react'
 
 // HashiCorp imports
 import usePageviewAnalytics from '@hashicorp/platform-analytics'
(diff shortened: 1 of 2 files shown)


src/layouts/base-layout/index.tsx
```diff
@@ -60,11 +60,42 @@ const BaseLayout = ({
 	useScrollPercentageAnalytics()
 	const [showSkipLink, setShowSkipLink] = useState(false)
 
+	// TODO: remove after HashiConf 2024
+	// start of code to remove
+	const [isDuringHashiConf, setIsDuringHashiConf] = useState(false)
+
+	useEffect(() => {
+		const now = new Date().getTime() // current date & time
```
RubenSandwich, line 68: @heatlikeheatwave we have a timezone bug here. Namely we are getting a random timezone but then comparing it to PST, which could mean they are already in Hashiconf time when Hashiconf hasn't even started yet. We need to get the users timezone in PST to accurately compare to PST.

So now should be:
`const now = new Date().toLocaleString('en-US', { timeZone: 'America/Los_Angeles' })`

RubenSandwich (changes requested):
We have a timezone bug as we are compare across timezones.

src/layouts/base-layout/index.tsx
RubenSandwich: Actually that is just going to just return the string. I think we should import something like: https://day.js.org/en/ and use it's date manipulation functions as that is going to make comparing across timezones easier.

heatlikeheatwave: Shouldn't it be EDT since the conference is in Boston?

heatlikeheatwave: Also, great catch with this!

RubenSandwich: Dates. Why you so hard. 😢

aaronvanderlip: Y'all are making it happen with more 👁️ on the code.

heatlikeheatwave: @RubenSandwich I tested this again, and I don't think this has the bug you described. I changed my location and time on my machine to west coast and europe; the visibility of the banner worked as expected. What am I missing that you're seeing? 

I do agree that the pacific time zone (.e.g -8:00 in the time stamp) can be changed. That was a mistake on my part.

src/layouts/base-layout/index.tsx
```diff
@@ -6,7 +6,7 @@
 // Third-party imports
 import classNames from 'classnames'
 import AlertBanner from '@hashicorp/react-alert-banner'
-import { HTMLAttributes, useState } from 'react'
+import { HTMLAttributes, useEffect, useState } from 'react'
```
aaronvanderlip, line 9: I don't think the `useEffect` is used? But as long as the linter doesn't complain and it builds I have no issue with it.

aaronvanderlip (approved):

PR #2588, opened 2024-10-17
add API docs for HCP RBAC

zchsh:
## 🔗 Relevant links

- [Asana task](url) 🎟️

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33101,7 +33101,6 @@
 			"version": "2.3.2",
 			"resolved": "https://registry.npmjs.org/fsevents/-/fsevents-2.3.2.tgz",
 			"integrity": "sha512-xiqMQR4xAeHTuB9uWm+fFRcIOgKBMiOBP+eXiyT7jsgVCq1bkVygt00oASowB7EdtpOHaaPgKt812P9ab+DDKA==",
-			"dev": true,
 			"hasInstallScript": true,
 			"optional": true,
 			"os": [
--- a/src/pages/hcp/api-docs/rbac/[[...page]].tsx
+++ b/src/pages/hcp/api-docs/rbac/[[...page]].tsx
@@ -1,0 +1,114 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+// Lib
+import { fetchCloudApiVersionData } from 'lib/api-docs/fetch-cloud-api-version-data'
+// View
+import OpenApiDocsView from 'views/open-api-docs-view'
+import {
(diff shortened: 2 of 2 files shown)


chris-hut (approved):
After [confirming with the team](https://hashicorp.slack.com/archives/C04RWBAC30W/p1729198911780659?thread_ts=1728665500.083449&cid=C04RWBAC30W), this looks perfect to us!
Thanks so much!

emilypersson1 (approved):
thank you!

xargs-P (changes requested):
Hello 👋  , I have open questions in Slack about how the data is displayed and which apis are revealed. Please hold on merging 🙏  while we sort this out.

chris-hut (changes requested):
[After talking with Trevor:](https://hashicorp.slack.com/archives/C04RWBAC30W/p1729201420271879?thread_ts=1728665500.083449&cid=C04RWBAC30W), could we:
- Hide the org-create API endpoint (`​Create - post:  ​/​resource-manager​/​projects`)
- Combine the duplicate `resource-manager/resources` and `resource-manager/organizations` sections?
image

pbortnick (approved):
👍

zchsh:
Hiya @chris-hut and @xargs-P ! Thanks for catching the issue with the path-based grouping. That should be sorted now!

I've also implemented some temporary shim code to delete the operation at `schema.paths['/resource-manager/2019-12-10/projects'].post` before rendering the document. This should hide the org-create endpoint.

For the "hidden endpoints", I think at some point very soon we'll want to find a more sustainable way to achieve this. For example, we could omit the operation entirely from the source spec file in hashicorp/hcp-specs. Or if that's not an option, and it needs to be included in that repo but hidden on our docs site, we could explore OpenAPI extensions, eg as discussed in [this thread](https://github.com/OAI/OpenAPI-Specification/issues/433). (cc @emilypersson1 for this last bit, in case this merits a feature ticket or whatnot! 🙇 )

emilypersson1:
I added an [asana task](https://app.asana.com/0/1205712940742344/1206060189410073) for hiding private endpoints as a future idea. ideally we can remove the endpoint from the spec for now, but if the RBAC team needs this functionality, we can take a look at prioritizing this work!

zchsh:
I've updated this PR to use our new OpenAPI docs template. It should be an equivalent user experience to the previous template, the big change is each operation is now on a separate page (we were having issues with some specs creating super large single pages otherwise).

As mentioned earlier, we have a temporary shim code to delete the operation at schema.paths['/resource-manager/2019-12-10/projects'].post before rendering the document. I've retained this shim when moving to the new template.

PR #2591, opened 2024-10-17
unified docs: add per-product docs loader flag

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

This PR introduces a feature flag for migrating to Unified Docs.

## 🤷 Why

To allow incremental migration to our Unified Docs Repository, specifically by allowing content source repositories to be migrated individually. This continues the work started in #2550, adding feature flags to allow use of the `UNIFIED_DOCS_API` `.env` variable introduced in that PR.

## 🛠️ How

- Introduces a `unified_docs_migrated_repos` flag, which is expected to be an array of strings representing content source repositories
    - Note that "content source repository" is distinct from "product". For example, the `terraform` product has content sourced from multiple repositories, such as `terraform-docs-common`, `ptfe-releases`, `terraform-plugin-mux`, and so on.
- Builds a utility function `getContentApiBaseUrl(repoName)`, which makes a relatively simple check against the `unified_docs_migrated_repos` flag
    - Splitting out this utility felt intuitive, as in the first few use cases for the flag, it became clear there'd be identical logic since the situation is usually "we're about to make a content API call in the context of some repo, which URL should we make the request to?"
    - Splitting this out also provided a space to add some detailed comments on caveats and limitations related to the unified docs migration so far. In brief, this flag currently represents the migration of "fetching documents", but there are other yet-to-be-implemented content API functions that may make sense to migrate at the same time as "fetching documents". I've put as much detail as I thought might be helpful in comments within the utility file that expand on this.
- Used `getContentApiBaseUrl` utility function in docs-view remote loader
    - If a content source repo has been migrated, we should fetch `version-metadata`, `nav-data`, and `doc` documents from the unified docs API.
- Used `getContentApiBaseUrl` utility function in get-valid-versions
    - If a content source repo has been migrated, we should fetch version metadata on each document from the unified docs API via `/api/content-versions`. Note that this route has [yet to be fully implemented](https://app.asana.com/0/1207899860738460/1208471953509347/f).

Related but less consequential changes:

- Updates `.env` UNIFIED_DOCS_API to remove trailing slash, for consistency

## 📸 Design Screenshots

CleanShot 2024-10-18 at 11 37 23@2x

## 🧪 Testing

- Check out [hashicorp/web-unified-docs](https://github.com/hashicorp/web-unified-docs), and run `npm i && npm run prebuild && npm run dev` to start the API locally at `http://localhost:8080`
- Check out [hashicorp/dev-portal](https://github.com/hashicorp/dev-portal) to this branch, `zs.add-unified-docs-document-flag`. Modify the `.env` file so that `UNIFIED_DOCS_API= http://localhost:8080`, then modify `config/development.json` to add a flagged content source repo, for example `flags: { "unified_docs_migrated_repos": ["terraform"] }`. Then run `npm i && npm run start`, and visit the docs associated with the flagged content source repo. These docs should be served from the locally running `web-unified-docs` instance. All other docs should be served from our existing content API (`content.hashicorp.com`).

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/799513369421962/1208471953509340/f
[preview]: https://dev-portal-git-zsadd-unified-docs-document-flag-hashicorp.vercel.app/


RubenSandwich (approved):
Looks and works. 👍🏻

src/lib/unified-docs-migration-utils.ts
```diff
@@ -0,0 +1,93 @@
...
+		? UNIFIED_DOCS_API
+		: MKTG_CONTENT_DOCS_API
+	return contentApiBaseUrl
+}
+
+/**
+ * Courtesy helper, warn about any missing content API related
+ * environment vars during development
+ */
+function checkEnvVarsInDev() {
```
heatlikeheatwave, line 77: Given that the `MKTG_CONTENT_DOCS_API` and `UNIFIED_DOCS_API`, is there a situation where these env vars would be missing?

src/lib/unified-docs-migration-utils.ts
```diff
@@ -0,0 +1,93 @@
...
+	if (process.env.NODE_ENV === 'development') {
+		const missingEnvVars = []
+		if (!MKTG_CONTENT_DOCS_API) missingEnvVars.push('MKTG_CONTENT_DOCS_API')
+		if (!UNIFIED_DOCS_API) missingEnvVars.push('UNIFIED_DOCS_API')
+		if (missingEnvVars.length > 0) {
+			const message = [
+				'Missing environment variable required to fetch remote content:',
+				missingEnvVars.map((v) => `  - \`${v}\``).join('\n'),
+				'Reach out to #team-web-platform to get the proper value(s).',
```
heatlikeheatwave, line 86:
```suggestion
				'Reach out to #team-web-support to get the proper value(s).',
```

src/lib/unified-docs-migration-utils.ts
```diff
@@ -0,0 +1,93 @@
...
+ * the content for the repository in question has been migrated to the
+ * unified docs repository.
+ */
+function getContentApiBaseUrl(repoName: string): string {
+	checkEnvVarsInDev()
+	const contentApiBaseUrl = checkIsMigratedToUnifiedDocs(repoName)
+		? UNIFIED_DOCS_API
+		: MKTG_CONTENT_DOCS_API
+	return contentApiBaseUrl
+}
```
heatlikeheatwave, line 71: Can we add tests for this?

heatlikeheatwave (commented):
Thanks for your work and adding this logic to the app. I would love to see tests for the functions added in `src/lib/unified-docs-migration-utils.ts`, if possible!

src/lib/unified-docs-migration-utils.ts
zchsh: Good point, it shouldn't happen since they are hard-coded in `.env`! Added a comment to this effect. Happy to delete this immediately if we feel strongly it should go, otherwise seems fine to leave it in for now in case it does come up somewhere unexpected.

zchsh:
@heatlikeheatwave Great call on tests, I added some here, mocking `.env` and global `__config`. Let me know if those look alright to you! 🙇

LeahMarieBush (approved):
👍 LGTM!

RubenSandwich (changes requested):
@zchsh QQ before we merge, how will we set his from the unified docs side? Like how can we customize unified docs PR builds to only have it run on certain docs?

zchsh:
@RubenSandwich This PR uses our existing `__config` flags mechanism, which in brief looks at `HASHI_ENV` and loads config (which includes flags) from the associated `config/<HASHI_ENV>.json` file. We fall back to `config/development.json` if `HASHI_ENV` isn't set, and maybe also worth noting is that all configs extend `config/base.json`.

As an example of how we'd test unified docs product-by-product, once we're ready to start testing Terraform content in Unified Docs, we'll modify `config` files to flag the products we want to test in the environments we want to test. Lots of options depending how granular we want to get... the two ones I think are most likely:
- Modify `unified_docs_migrated_repos` in `config/development.json`, which will allow us to test wherever `HASHI_ENV` is not set or is set to `development`. Currently this includes running `dev-portal` locally, and _all_ deploys from `web-unified-docs`, since we don't set `HASHI_ENV` there as far as I know.
- Introduce new `config/<environment>.json` files, perhaps similar to what we did for HVD previews. How many of these we might wanna introduce is probably mostly a factor of how isolated we want the Terraform unified docs testing to be.

Hope this helps, let me know if there's more detail I can add! 🙇

RubenSandwich (approved):
👍🏻

PR #2593, opened 2024-10-19
feat: datagrail consent manager

prestonbourne:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-presfeatconsent-manager-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1206122994980385/1208560619615741/f) 🎟️

## 🗒️ What

- Adds `lib/consent-manager`
- Adds types for `js-cookie` (the package was already there, without the type declarations)
- uninstalls `@hashicorp/react-consent-manager`

## 🧪 Testing

- [x] Open the preview
- [x] Verify that the consent manager is present
- [x] You can toggle preferences
- [x] You can close the consent manager
- [x] Clicking on the button in the footer reopens the consent manager

## 💭 Anything else?

> [!IMPORTANT] 
> Before merging, verify that the scripts that would usually be loaded by the hashicorp consent manager can now be loaded by the datagrail consent manager, this can be checked via Google Tag Manager and (Okta -> Datagrail)

For the case of Dev Portal that would be Posthog. ([See here](https://github.com/hashicorp/dev-portal/blob/28981338ac976580d87f27936ace43a2cd161cbd/src/lib/consent-manager-services/developer.ts))
and the `oss` preset services: https://github.com/hashicorp/react-components/blob/main/packages/consent-manager/loader/services.ts

As of Now Dev Portal manages this through segment and they need to be migrated to GTM


RubenSandwich (approved):
Looks good. 👍🏻

RubenSandwich (changes requested):
Three strange things I found: (But only 2 need to be addressed in this PR.)

1. Some a11y issues. (This doesn't need to be addressed in this PR/) I will post more in our team channel as I would prefer to have datagrail fix these issues.

2. When clicking "Update Preferences" the first click seems to not work, it only works on the second click. (Might also be a datagrail issue.) 

3. For the `.dg-main-content` CSS class, the `overflow-y` value should be `auto`, not `scroll`. Because scroll will always show the scrollbar even when it's not needed. (See pictures.)

Screenshot 2024-10-21 at 12 49 32 PM

Screenshot 2024-10-21 at 12 47 25 PM

zchsh (commented):
I followed the testing steps in both my usual browser (Arc aka Chrome with bells and whistles) and Firefox. In both cases I couldn't seem to get the preferences to persist...

Arc:

https://github.com/user-attachments/assets/d97c1254-ac99-4ee5-a6e9-f66687c0b13c

Firefox:

https://github.com/user-attachments/assets/223e5272-92e4-47f1-93ca-d0370e675c19

I'm still waiting for access to DataGrail... happy to dive deeper here if it'd be helpful 🙇

prestonbourne:
@zchsh hey! the cookies not persisting is expected in the preview, we can't write to `vercel.app` domains 

See here: https://vercel.com/guides/can-i-set-a-cookie-from-my-vercel-project-subdomain-to-vercel-app

zchsh (approved):
✅ Makes sense about the cookies! I followed testing steps and everything looks good.

zchsh:
`overflow-y` value is fixed, and we're co-ordinating with Google Tag Manager changes, so going to merge this shortly.

RubenSandwich (approved):
Awesome thanks for doing the y offset!

PR #2594, opened 2024-10-21
OpenAPI V2 - implement sidebar

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Builds out the sidebar for our new version of OpenAPI docs view. This new version of the view splits individual operations onto separate pages.

## 🤷 Why

Our current OpenAPI docs view renders all operations on a single page. For some spec files, the page data can get heavy enough that it becomes very slow and even error-causing to try to handle it.

## 🛠️ How

- Implements composable versions of existing sidebar link components, so we're less entangled with the rest of the site, and so we can more easily iterate on our OpenAPI docs view sidebar in the near future
- Implements approaches to operation grouping and to schema transforms, both of which are relevant to the sidebar (for operation groups and for the overview page label respectively)
- Updates the V2 preview tool's `get-props-from-preview-data.ts` so that we're showing all the same links and features we do on upstream production OpenAPI docs pages

## 📸 Design Screenshots

Group by `tag` example, with [HCP Consul Central][hcp-consul-central-spec]:

| Before | After |
| - | - |
| group-by-tag-hcp-consul-before | group-by-tag-hcp-consul-after |

Group by `path` example, with [HCP Vault Secrets][hcp-vault-secrets-spec]:

| Before | After |
| - | - |
| group-by-path-hcp-vault-secrets-before | group-by-path-hcp-vault-secrets-after |

## 🧪 Testing

- [ ] Test out a spec with "group by path" using the preview tool at [/open-api-docs-preview-v2]
    - For example, compare an upload of the [HCP Vault Secrets][hcp-vault-secrets-spec] with the sidebar in the upstream [/hcp/api-docs/vault-secrets](https://developer.hashicorp.com/hcp/api-docs/vault-secrets)
- [ ] Test out a spec with "group by tag" using the preview tool at [/open-api-docs-preview-v2]
    - For example, compare an upload of the [HCP Consul Central][hcp-consul-central-spec] with the sidebar in the upstream [/hcp/api-docs/consul](https://developer.hashicorp.com/hcp/api-docs/consul)

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/1207339219333499/1208591364823438/f
[preview]: https://dev-portal-git-zsopen-api-v2-sidebar-hashicorp.vercel.app
[/open-api-docs-preview-v2]: https://dev-portal-git-zsopen-api-v2-sidebar-hashicorp.vercel.app/open-api-docs-preview-v2
[hcp-consul-central-spec]: https://github.com/hashicorp/hcp-specs/blob/main/specs/cloud-global-network-manager-service/preview/2023-10-10/hcp.swagger.json
[hcp-vault-secrets-spec]: https://github.com/hashicorp/hcp-specs/blob/main/specs/cloud-vault-secrets/stable/2023-06-13/hcp.swagger.json


src/pages/api/open-api-docs-preview-v2.ts
```diff
@@ -8,13 +8,12 @@ import path from 'path'
...
 // Types
 import type { NextApiRequest, NextApiResponse } from 'next'
-import { OpenApiPreviewV2InputValues } from 'views/open-api-docs-preview-v2/components/open-api-preview-inputs'
 
 /**
  * Setup for temporary file storage
  */
 // Determine if we're deploying to Vercel, this affects the temporary directory
-const IS_VERCEL_DEPLOY = process.env.VERCEL_ENV !== 'development'
+const IS_VERCEL_DEPLOY = typeof process.env.VERCEL_ENV === 'string'
```
zchsh, line 16: Here we fix some `IS_VERCEL_DEPLOY` check logic.

Previously if `VERCEL_ENV` was equal to anything other than `development`, then we'd have `IS_VERCEL_DEPLOY` as `true`. This isn't accurate in cases where `VERCEL_ENV` is undefined, which is sometimes the case when running the site locally.

src/views/open-api-docs-preview-v2/index.tsx
```diff
@@ -54,7 +54,10 @@ function OpenApiDocsPreviewViewV2({
 							<div style={{ padding: '24px', maxWidth: '35em' }}>
 								<h1>OpenAPI Preview Tool</h1>
 								<p>
-									{`Please use the input form on this page to upload your spec. After submitting the form, the page should reload and display a preview.`}
+									{`Please use the input form on this page to upload your spec. After submitting the form, the page should reload and display a preview. For security reasons, after around an hour, your submitted file will be deleted the next time this page is loaded. You will need to re-submit your spec if you're working with this tool for longer than an hour.`}
+								</p>
+								<p>
+									{`This page may unexpectedly after navigation due to differences in how this preview tool works relative to production contexts. Reloading the page may resolve the issue. If reloading the page doesn't do anything, you may need to re-submit your spec file. If you're experiencing frequent issues requiring page reloads, reach out in #team-web-support in Slack.`}
```
zchsh, line 60: Adds a warning about reloads.

I noticed this behaviour when running things locally. I think it has to do with `next/link` client-side navigation, sometimes it seems like navigating with the sidebar skips over `getServerSideProps`. It seems to happen less in the deploy preview, and can be resolved by reloading the page after navigating, but I wanted to mention it in case it does start happening.

If needed, one way we could work around this is to add some switch that allows us to render sidebar links as plain `<a />` rather than `next/link`. But I'm hesitant to add complexity for the sake of the preview tool, especially if this comes up more locally than anything else, since the preview tool is meant to be used off of our staging URL, `developer.hashicorp.services`.

src/views/open-api-docs-preview-v2/utils/get-props-from-preview-data.ts
```diff
@@ -28,10 +34,44 @@ export default async function getPropsFromPreviewData(
 	if (!previewData) {
 		return null
 	}
-	// Use the incoming preview data to generate static props for the view
-	return await getStaticProps({
```
zchsh, line 32: I've updated `get-props-from-preview-data` so that we can show previewed specs more or less exactly how the production pages will appear. Specifically, I've added a schema transform here to replace `HashiCorp Cloud Platform` with `HCP` in the title (nearly all specs need this), as well as some generic resource links.

I've also hooked up the `Group operations by path` input from the form into the newly introduced `getOperationGroupKey` grouping control function.

src/views/open-api-docs-view-v2/components/sidebar-link/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+ * creating a re-usable component that captures the composition pattern.
+ *
+ * A component that captures a common pattern of composition should _not_ need
+ * any conditional statements. The props interface should be very simple - the
+ * point being to _reduce_ the complexity of repeated the identical composed
+ * pattern. If "edge cases" need to be handled, then the consumer can "eject"
+ * from the composed pattern by copying and pasting the body of the composed
+ * component and making changes from there.
+ */
+export function SidebarLink({
```
zchsh, line 44: I've added a number of `sidebar` link components. I've gone into detail in comments why this felt right (as opposed to using the existing `SidebarNavMenuItem` components). The TLDR is that we have design refinements we want to make to OpenAPI specs, and it'll be hard to move on those if we're entangled with the sidebars used on almost every other page on the site. Happy to expand on this further if it'd be helpful.

I think the important thing here is 1) this seems like a reasonable approach, and 2) the look and feel matches the previous sidebar components (for now).

src/views/open-api-docs-view-v2/index.tsx
```diff
@@ -19,40 +23,32 @@ import type { OpenApiDocsViewV2Props } from './types'
  */
 export default function OpenApiDocsViewV2({
```
zchsh, line 24: In the main view here, we're now consuming the new props passed from `getStaticProps`, and rendering those with the newly established components.

src/views/open-api-docs-view-v2/schema-transforms/schema-transform-shorten-hcp.ts
```diff
@@ -0,0 +1,25 @@
...
+function shortenHcp(s: string): string {
+	return s.replace('HashiCorp Cloud Platform', 'HCP')
+}
+
+/**
+ * Given an OpenAPI schema document,
+ * Return the document with the title modified, with any instances of
+ * "HashiCorp Cloud Platform" replaced with "HCP".
+ */
+export function schemaTransformShortenHcp(
```
zchsh, line 15: The broad intent with this `V2` version of our OpenAPI docs view is to first migrate all existing OpenAPI pages to it, then to delete the old `open-api-docs-view`, then rename this new one to remove the `V2` prefixes.

With this in mind, it felt like a good idea to copy over any utilities from within the old view folder. This way, when migration is complete, the old and new views shouldn't be entangled at all, and we should have an easier time deleting the old view folder.

src/views/open-api-docs-view-v2/server.ts
```diff
@@ -3,41 +3,63 @@
...
 } from 'views/open-api-docs-view-v2/types'
 
+/**
+ * Build static props for an OpenAPI docs view.
+ *
+ * There are two main views:
+ * - Landing view, for the basePath, when no operationSlug is provided
+ * - Operation view, for the specific operationSlug that's been provided
+ */
 export async function getStaticProps({
```
zchsh, line 32: Main change here is building the sidebar link props to pass to the view. We use the newly established `getOperationObjects` and `groupItemsByKey` functions to achieve this.

src/views/open-api-docs-view-v2/types.ts
```diff
@@ -3,28 +3,41 @@
...
+import type { LandingContentProps } from './components/landing-content'
+import type { OpenAPIV3 } from 'openapi-types'
+import type { OperationContentProps } from './components/operation-content'
+import type { OperationObject } from './utils/get-operation-objects'
+import type { ProductSlug } from 'types/products'
 
 /**
  * Shared props are common to both the "landing" and "operation" views.
  */
 export interface SharedProps {
```
zchsh, line 16: In general I've opted to avoid reaching for type abstractions, as in the past I think we've been a bit overzealous in trying to create shared types, resulting in highly entangled components that can be difficult to iterate on. I find it a lot easier to declare things explicitly.

The downside is it can feel a bit repetitive, but I find it makes it much easier to change things without worrying about side-effects elsewhere in the app. And of course, TypeScript will still do the job of warning us about places that'll be affected by any type changes, regardless of how fancy our type abstractions are.

src/views/open-api-docs-view-v2/utils/get-operation-group-key-from-path.ts
```diff
@@ -0,0 +1,18 @@
...
+import { truncateHcpOperationPath } from './truncate-hcp-operation-path'
+import type { OperationObject } from './get-operation-objects'
+
+/**
+ * Given an operation object, derive a string representing the
+ * first two segments of the operation's path, and
+ * Return the string for use in grouping the operation with other
+ * operations that share the first two path segments.
+ */
+export function getOperationGroupKeyFromPath(o: OperationObject) {
```
zchsh, line 15: In the context of the new setup, which accepts a `getOperationGroupKey` function, this function allows us to group operation objects by path (for situations where path-based grouping is preferred over tag-based grouping).

src/views/open-api-docs-view-v2/utils/get-operation-objects.ts
```diff
@@ -0,0 +1,45 @@
...
+	type: string
+}
+
+/**
+ * Given an OpenAPI document, build an array of operation objects,
+ * each including the properties defined for operations in the OpenAPI spec,
+ * as well as the `path` and `type` (ie GET, POST, etc) of the operation, and
+ * Return the array of operation objects.
+ */
+export function getOperationObjects(
```
zchsh, line 24: Gathers operation objects from the provide `openApiDocument`.

We'll probably be able to re-use the result of this function in `getStaticProps` to filter for and then transform the props for a specific operation page.

src/views/open-api-docs-view-v2/utils/group-items-by-key.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+/**
+ * Given a flat array of items, and a getItemGroupKey function,
+ * Return an array of item groups, with item groups being constructed
+ * based on keys derived from the provided getItemGroupKey function.
+ */
+export function groupItemsByKey<Item>(
```
zchsh, line 11: I had a think through approaches to grouping operation objects, and found my way to this pretty generic function. This gets used with `getOperationGroupKey` passed as the `getItemGroupKey` function in order to group operations using arbitrary logic.

src/views/open-api-docs-view-v2/utils/truncate-hcp-operation-path.ts
```diff
@@ -0,0 +1,42 @@
...
+ * `/<version>/...(anything)`
+ */
+const GLOBAL_SERVICE_PATTERN = /\/\d\d\d\d-\d\d-\d\d/
+
+/**
+ * Truncates HCP operation paths for brevity, removing the service, version,
+ * organization, and project prefixes for specific individual services, and
+ * removing the version prefix for global services.
+ */
+export function truncateHcpOperationPath(path: string) {
```
zchsh, line 38: As mentioned on another file, some utils are copied over from the "old" view folder, with the intent of making the "old" view folder less entangled and easier to delete later.

zchsh:
@LeahMarieBush This PR is a bit chonky since I took a little detour to avoid using our complex and entangled  `SidebarNavMenuItem` components. I'm out early next week, but would be happy to sync on this sometime, or add more context in comments or the PR description, if that'd be helpful for review 🙇‍♂️

src/views/open-api-docs-preview-v2/index.tsx
LeahMarieBush, line 60: It looks like this sentence is missing a word at the beginning: `This page may unexpectedly after navigation due to differences in how this preview tool works relative to production contexts.`

LeahMarieBush (approved):
Code and preview looks good! Thanks for adding the extra comments about the changes, they were very helpful for the extra context needed. I left one comment about the message on the open api docs preview page but other than that it all looks good 👍

src/views/open-api-docs-preview-v2/index.tsx
EnMod, line 60: @zchsh Could you add a word here to clarify what may unexpectedly happen?

```suggestion
									{`This page may unexpectedly (break?|error out?|reload?) after navigation due to differences in how this preview tool works relative to production contexts. Reloading the page may resolve the issue. If reloading the page doesn't do anything, you may need to re-submit your spec file. If you're experiencing frequent issues requiring page reloads, reach out in #team-web-support in Slack.`}
```

src/views/open-api-docs-view-v2/components/sidebar-link/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+ * For example, an "external link" component could be something like
+ * `<SidebarLink><SidebarLinkText /><IconExternalLink16 /></SidebarLink>`.
+ *
+ * If composed patterns are repeated _exactly_, and in a way where we'd
+ * definitely want to update all instances at once, then it may be worth
+ * creating a re-usable component that captures the composition pattern.
+ *
+ * A component that captures a common pattern of composition should _not_ need
+ * any conditional statements. The props interface should be very simple - the
+ * point being to _reduce_ the complexity of repeated the identical composed
```
EnMod, line 39: Small typo perhaps? 🙂 

```suggestion
 * point being to _reduce_ the complexity of repeating the identical composed
```

EnMod (commented):
Had some comments as I gave this a glance!

src/views/open-api-docs-preview-v2/index.tsx
EnMod: Ah, I missed this comment! I noticed this too haha. Would it be something like `unexpectedly`:
- `break`?
- `reload`?

zchsh: Great catch, thanks y'all! Updated to `This page may appear unexpectedly after navigation rather the expected preview content due to differences in how this preview tool works...`

PR #2618, opened 2024-11-01
add operation content to OpenAPI v2 template

zchsh:
## 🔗 Relevant links

- [Preview link][preview] 🔎
- [Asana task][task] 🎟️

## 🗒️ What

Implements operation page content for the split-operations Open API template.

## 🤷 Why

Our current OpenAPI docs view renders all operations on a single page. For some spec files, the page data can get heavy enough that it becomes very slow and even error-causing to try to handle it.

This PR implements the operation page content for a new version of the Open API docs template, which splits operations to separate pages.

## 🛠️ How

- Pulls in components from the existing OpenAPI view
- Implements `<OperationContent />` component to match upstream setup
    - Only significant difference is operation name is now used as `<h1 />` rather than `<h2 />`

### HCP Vault Secrets

| Before | After |
| - | - |
| hcp-vault-secrets-before | hcp-vault-secrets-after |

### HCP Consul

| Before | After |
| - | - |
| hcp-consul-before | hcp-consul-after |

## 🧪 Testing

- [x] Visit the preview tool at [/open-api-docs-preview-v2]
- [x] Test out various specs, and compare to upstream
    - Example: compare an upload of the [HCP Consul Central][hcp-consul-central-spec] with the sidebar in the upstream [/hcp/api-docs/consul](https://developer.hashicorp.com/hcp/api-docs/consul)
    - Example: compare an upload of the [HCP Vault Secrets][hcp-vault-secrets-spec] with the sidebar in the upstream [/hcp/api-docs/vault-secrets](https://developer.hashicorp.com/hcp/api-docs/vault-secrets)

## 💭 Anything else?

Not at the moment!

[task]: https://app.asana.com/0/0/1208591364823440/f
[hcp-consul-central-spec]: https://github.com/hashicorp/hcp-specs/blob/main/specs/cloud-global-network-manager-service/preview/2023-10-10/hcp.swagger.json
[hcp-vault-secrets-spec]: https://github.com/hashicorp/hcp-specs/blob/main/specs/cloud-vault-secrets/stable/2023-06-13/hcp.swagger.json
[preview]: https://dev-portal-git-zsopen-api-v2-operation-content-hashicorp.vercel.app
[/open-api-docs-preview-v2]: https://dev-portal-git-zsopen-api-v2-operation-content-hashicorp.vercel.app/open-api-docs-preview-v2

### DIFF
--- a/src/views/open-api-docs-preview-v2/utils/get-props-from-preview-data.ts
+++ b/src/views/open-api-docs-preview-v2/utils/get-props-from-preview-data.ts
@@ -7,12 +7,15 @@ import { getStaticProps } from 'views/open-api-docs-view-v2/server'
 // Utils
 import { getOperationGroupKeyFromPath } from 'views/open-api-docs-view-v2/utils/get-operation-group-key-from-path'
 import { schemaTransformShortenHcp } from 'views/open-api-docs-view-v2/schema-transforms/schema-transform-shorten-hcp'
+import { schemaTransformComponent } from 'views/open-api-docs-view-v2/schema-transforms/schema-transform-component'
+import { shortenProtobufAnyDescription } from 'views/open-api-docs-view-v2/schema-transforms/shorten-protobuf-any-description'
 // Types
 import type {
 	OpenApiDocsViewV2Props,
 	OpenApiDocsViewV2Config,
 } from 'views/open-api-docs-view-v2/types'
 import type { OpenApiPreviewV2InputValues } from '../components/open-api-preview-inputs'
+import { schemaModComponent } from 'views/open-api-docs-view/utils/massage-schema-utils'
 
 /**
  * Given preview data submitted by the user, which includes OpenAPI JSON,
@@ -38,7 +41,23 @@ export default async function getPropsFromPreviewData(
 	// and prefer to have content updates made at the content source... but
 	// some shims are used often enough that they feel worth including in the
 	// preview too. Namely, shortening to `HCP` in the spec title.
-	const schemaTransforms = [schemaTransformShortenHcp]
+	const schemaTransforms = [
+		schemaTransformShortenHcp,
+		(schema) => {
+			return schemaTransformComponent(
+				schema,
+				'protobufAny',
+				shortenProtobufAnyDescription
+			)
+		},
+		(schema) => {
+			return schemaTransformComponent(
+				schema,
+				'google.protobuf.Any',
+				shortenProtobufAnyDescription
+			)
+		},
+	]
 	// Build page configuration based on the input values
 	const pageConfig: OpenApiDocsViewV2Config = {
 		basePath: '/open-api-docs-preview-v2',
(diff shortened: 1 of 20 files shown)


src/views/open-api-docs-preview-v2/utils/get-props-from-preview-data.ts
```diff
@@ -38,7 +41,23 @@ export default async function getPropsFromPreviewData(
 	// and prefer to have content updates made at the content source... but
 	// some shims are used often enough that they feel worth including in the
 	// preview too. Namely, shortening to `HCP` in the spec title.
-	const schemaTransforms = [schemaTransformShortenHcp]
+	const schemaTransforms = [
```
zchsh, line 44: I've added "shorten protobufAny" transforms to the preview tool, to make it easier to compare content. These could be removed later if they don't feel right.

src/views/open-api-docs-view-v2/components/content-with-permalink/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+ * text for screen readers. For simple use cases where `children` is a
+ * string, the `ariaLabel` can be passed the same value as `children`.
+ * For more complex use cases, the consumer should determine what text
+ * would be appropriate to pass to `ariaLabel`.
+ *
+ * Note that this component does _not_ handle placing the provided `id`
+ * in the DOM. It requires the consumer to place the `id` on an appropriate
+ * element, typically an element rendered in the provided `children`.
+ */
+export function ContentWithPermalink({
```
zchsh, line 27: All of the new files in this PR are existing components copied over from the previous view. As mentioned in previous PRs, rationale is that once we've migrated to this new view, we can delete the old view folder. Copying the necessary components into the new view, rather than importing them from the old folder or trying to build some abstraction, is intended to make it easier to clean up the old view folder once we've migrated away from it.

src/views/open-api-docs-view-v2/server.ts
```diff
@@ -51,16 +52,18 @@ export async function getStaticProps({
...
-	const rawSchemaData = await parseAndValidateOpenApiSchema(openApiJsonString)
-
-	/**
-	 * Apply any schema transforms.
-	 */
-	let schemaData = rawSchemaData
-	for (const schemaTransformFunction of schemaTransforms ?? []) {
-		schemaData = schemaTransformFunction(schemaData)
-	}
+	const schemaData = await parseAndValidateOpenApiSchema(
```
zchsh, line 57: Transforms need to be applied in the hook function argument provided by `parseAndValidateOpenApiSchema`, otherwise we end up dereferencing the schema _before_ applying transforms... which makes transforms that affect referenced definitions (such as shortening the `protobufAny` description) have no effect.

zchsh:
@nandereck Let me know if there's additional context I can add here! 🙇‍♂️

The main thing to note is probably that most of the added lines are copying components from the "old" view folder into the "new" view folder. The intent being to be able to easily delete the "old" view folder once we're done migrating away from it.

nandereck (approved):
LGTM - appears to work as expected!

PR #2620, opened 2024-11-05
Fix vagrant linux packages install, fix wget command based on feedback

im2nguyen:
## 🔗 Relevant links

Based on feedback by @chrisroberts:

> There are a couple issues with the content in the downloads page. In general, the `wget` command used for adding the hashicorp gpg key is incorrect.  It is currently `wget -O- https://apt.releases.hashicorp.com/gpg` but there needs to be a space added: `wget -O - https://apt.releases.hashicorp.com/gpg`.
Vagrant specific, the links on the downloads page are incorrect. Under linux, there should be only one binary download link for amd64 (which currently links to the "binary download"). The x86_64 links to an rpm package and the i686 link to a deb package.

- [Preview link](https://dev-portal-git-vagrant-update-install-binaries-wget-hashicorp.vercel.app/vagrant/install) 🔎
- [Asana task](url) 🎟️


chrisroberts (approved):
this looks great, thank you so much!

src/views/product-downloads-view/index.tsx
```diff
@@ -93,11 +93,20 @@ const ProductDownloadsViewContent = ({
 
 	// Group the selected release downloads by OS, for use in multiple places
 	const selectedRelease = releases.versions[currentVersion]
-	const downloadsByOS = useMemo(
+	let downloadsByOS = useMemo(
 		() => sortPlatforms(selectedRelease),
 		[selectedRelease]
 	)
 
+	if (currentProduct.slug === 'vagrant') {
```
heatlikeheatwave, line 101:
```suggestion
  // TODO: remove this L101-109 once metadata is corrected by rel eng
	if (currentProduct.slug === 'vagrant') {
```

src/views/product-downloads-view/index.tsx
```diff
@@ -93,11 +93,20 @@ const ProductDownloadsViewContent = ({
...
+		downloadsByOS = {
+			...downloadsByOS,
+			linux: {
+				amd64: downloadsByOS.linux.amd64,
```
heatlikeheatwave, line 105: this logic needs to have error handling. I selected an earlier version, and it caused an error
Screenshot 2024-11-05 at 12 05 28

heatlikeheatwave (changes requested):

src/views/product-downloads-view/index.tsx
im2nguyen: I wasn't able to replicate this locally or in the [deploy preview](https://dev-portal-git-vagrant-update-install-binaries-wget-hashicorp.vercel.app/vagrant/install)... but added a snippet that should address it

```
	if (currentProduct.slug === 'vagrant') {
		downloadsByOS = {
			...downloadsByOS,
			linux: {
				amd64: downloadsByOS?.linux?.amd64 || null,
			},
		}
	}
```

src/views/product-downloads-view/index.tsx
```diff
@@ -93,11 +93,20 @@ const ProductDownloadsViewContent = ({
...
 		[selectedRelease]
 	)
 
+	if (currentProduct.slug === 'vagrant') {
+		downloadsByOS = {
+			...downloadsByOS,
+			linux: {
+				amd64: downloadsByOS?.linux?.amd64 || null,
+			},
+		}
```
heatlikeheatwave, line 107: This will filter and prevent an error from being thrown when the user changes to a version that does not have a linux build.

```suggestion
		downloadsByOS = {
			...downloadsByOS,
			...(downloadsByOS.linux?.amd64
				? { linux: { amd64: downloadsByOS.linux.amd64 } }
				: {}),
		}
```

heatlikeheatwave (changes requested):

src/views/product-downloads-view/index.tsx
heatlikeheatwave: Unfortunately changing it `null` created another bug
Screenshot 2024-11-05 at 13 40 41

I added a comment with code that will work here https://github.com/hashicorp/dev-portal/pull/2620#discussion_r1829843021

If you change the version to `1.4.0` or any version other than the default, you should get the error. I'm using FF

tsconfig.json
heatlikeheatwave, line 1: What's the change for?

next-env.d.ts
```diff
@@ -1,5 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
+/// <reference types="next/navigation-types/compat/navigation" />
```
heatlikeheatwave, line 3: Was this intentional?

heatlikeheatwave (commented):
LGTM, just a couple questions about file changes that seem to have snuck in

PR #2645, opened 2024-12-10
chore: disable posthog default `$pageview` capture

prestonbourne:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Posthog is showing alot more page traffic than should be expected @anthonyvetter pointed out that there it seems to be roughly double which leads me to believe something is causing `pageview` events to fire twice.

Currently, we send pageview events manually, introduced in this PR

After following [this thread](https://posthog.com/questions/double-events-1) it seems we want to ensure we disable the default pageview event, which is what this PR does

### DIFF
--- a/src/hooks/use-posthog-analytics.ts
+++ b/src/hooks/use-posthog-analytics.ts
@@ -5,10 +5,14 @@
 
 import { useEffect } from 'react'
 import { useRouter } from 'next/router'
+import { type PostHog } from 'posthog-js'
 
 declare global {
+	/* 
+	Loaded via a posthog script tag in GTM based on datagrail consent
+	*/
 	interface Window {
-		posthog?: $TSFixMe
+		posthog?: PostHog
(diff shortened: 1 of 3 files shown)


src/hooks/use-posthog-analytics.ts
```diff
@@ -33,7 +37,9 @@ export default function usePostHogPageAnalytics(): void {
 
 	useEffect(() => {
 		// Ensures code only runs if PostHog has been initialized
-		if (!window?.posthog?.capture) return
+		if (!window?.posthog) return
+
+		window.posthog.config.capture_pageview = false
```
prestonbourne, line 42: This _should_ dedupe the events being sent to our posthog instance, we'll monitor after merging to see what the dataflow looks like

nandereck (approved):
LGTM!

PR #2654, opened 2024-12-13
feat: fetch paths from UDR all-docs-paths

LeahMarieBush:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-leah-featudr-all-docs-paths-hashicorp.vercel.app/server-sitemap.xml) 🔎
- [Asana task](https://app.asana.com/0/1206815131022336/1208939388171199/f) 🎟️

## 🗒️ What

This PR fetches the paths from the UDR `api/all-docs-paths` for generating the sitemap. It only does fetches these if there are migrated content repos in the config. After fetching the paths, it filters the paths from the content API that are duplicates and then merges both results together.

## 🧪 Testing

1. Go to [Preview link](https://dev-portal-git-leah-featudr-all-docs-paths-hashicorp.vercel.app/server-sitemap.xml)
2. Verify that there is not an entry for `https://developer.hashicorp.com/terraform` in the sitemap (Run command f with `https://developer.hashicorp.com/terraform<`)
3. Pull down this branch and run the command below. Verify that `https://developer.hashicorp.com/terraform` is in the sitemap. The UDR will return this as a path but the content API does not which is why it will only be included with the UDR results.
```
HASHI_ENV=unified-docs-sandbox npm start       
```

### DIFF
--- a/src/lib/__tests__/docs-content-fields.test.ts
+++ b/src/lib/__tests__/docs-content-fields.test.ts
@@ -1,10 +1,6 @@
 import { describe, it, expect, vi } from 'vitest'
 import { allDocsFields } from '../sitemap/docs-content-fields'
 
-// vi.mock('../sitemap/helpers', () => ({
-// 	makeSitemapField: vi.fn((args) => args),
-// }))
-
 describe('allDocsFields', () => {
 	beforeEach(() => {
 		vi.resetAllMocks()
--- a/src/lib/__tests__/docs-content-fields.test.ts
+++ b/src/lib/__tests__/docs-content-fields.test.ts
@@ -1,0 +1,83 @@
+import { describe, it, expect, vi } from 'vitest'
+import { allDocsFields } from '../sitemap/docs-content-fields'
+
+// vi.mock('../sitemap/helpers', () => ({
+// 	makeSitemapField: vi.fn((args) => args),
+// }))
+
+describe('allDocsFields', () => {
+	beforeEach(() => {
+		vi.resetAllMocks()
+	})
+
+	it('should fetch and return all docs fields from content API when unified-docs-sandbox is not set', async () => {
+		process.env.HASHI_ENV = 'production'
+		process.env.MKTG_CONTENT_DOCS_API = 'https://content-api.example.com'
+		const mockContentAPIDocsResult = [
+			{ path: 'doc1', created_at: '2025-01-07T18:44:51.431Z' },
+			{ path: 'doc2', created_at: '2025-01-07T18:44:51.431Z' },
+		]
+		global.fetch = vi.fn().mockResolvedValue({
+			json: vi.fn().mockResolvedValue({ result: mockContentAPIDocsResult }),
+		})
+
+		const result = await allDocsFields()
+
+		expect(fetch).toHaveBeenCalledWith(
+			'https://content-api.example.com/api/all-docs-paths'
+		)
+		expect(result).toEqual(
+			mockContentAPIDocsResult.map((page) => ({
+				loc: `https://developer.hashicorp.com/${page.path}`,
+				lastmod: page.created_at,
+				priority: 1,
+				changefreq: 'daily',
+			}))
+		)
+	})
+
+	it('should fetch and return all docs fields from both content API and UDR when unified-docs-sandbox is set', async () => {
+		process.env.HASHI_ENV = 'unified-docs-sandbox'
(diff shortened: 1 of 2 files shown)


src/lib/sitemap/docs-content-fields.ts
```diff
@@ -6,14 +6,45 @@
...
-	return docsResult.map((page: { path: string; created_at: string }) =>
-		makeSitemapField({
-			slug: `${page.path}`,
-			lastmodDate: page.created_at,
-		})
+	const { result: contentAPIDocsResult } = await getContentAPIDocsPaths.json()
+
+	// If there are docs that have been migrated to the unified docs repo, get the paths for those docs
+	// and merge them with the paths for the docs that haven't been migrated from the content API
+	if (__config.flags?.unified_docs_migrated_repos.length > 0) {
```
RubenSandwich, line 16: @LeahMarieBush can you also add a check if we are in the using the unified-docs-sandbox?

`process.env.HASHI_ENV === 'unified-docs-sandbox'`

RubenSandwich (changes requested):
Let's also check if they are in the `unified-docs-sandbox` before merging in the `unified_docs_migrated_repos` for right now.

src/lib/sitemap/docs-content-fields.ts
```diff
@@ -6,14 +6,45 @@
...
+		const getUDRDocsPaths = await fetch(
+			`${process.env.UNIFIED_DOCS_API}/api/all-docs-paths`
+		)
+		const { result: udrDocsResult } = await getUDRDocsPaths.json()
+
+		// Filter the result from the content API to take out the duplicate paths between
+		// the content API and UDR
+		const filteredContentAPIDocsResult = contentAPIDocsResult.filter(
```
RubenSandwich, line 24: @LeahMarieBush We should only merge in the paths from the products that are in `unified_docs_migrated_repos`. Cause `/api/all-docs-paths` might return products/paths that are not in `unified_docs_migrated_repos`. (Like when we are testing a new product repo in UDR.)

RubenSandwich: Also what do you think should we add `productSlugs` as am optional param to /api/all-docs-paths? That way we can do the filtering on the API server and then do a simple merge here.

LeahMarieBush: Yeah I think that adding that optional param to the endpoint would make this a lot easier, then we can just fetch the paths needed. I'll get a story written for that and put this on hold until that ticket is complete

src/lib/sitemap/docs-content-fields.ts
```diff
@@ -6,14 +6,54 @@
...
+				return `products=${repo}`
+			}
+		)
+		const productsFilterString = productsFilter.join('&')
+		const getUDRDocsPaths = await fetch(
+			`${process.env.UNIFIED_DOCS_API}/api/all-docs-paths?${productsFilterString}`
+		)
+		const { result: udrDocsResult } = await getUDRDocsPaths.json()
+
+		// Filter the result from the content API to take out the duplicate paths between
```
RubenSandwich, line 31: If we merge individual items we could get to a place where a doc that existed in the old content api but got removed in the new UDR api still shows up in the sitemap. 

For example lets say terraform migrated over, but `/terraform/docs/2024` exists in the old content api but it's name was updated to `/terraform/docs/2025` in the new UDR api. This check would allow both through because both API have to exist side by side for a while.

If any product is migrated over we really only want to accept paths from the new UDR api for that product. (So remove them from the old content api results before merging.)

Also this logic is now complicated enough that I believe it should be tested. Specifically the merging logic. It might be easier to test that logic if you pull out the merge logic into it's own pure function.

RubenSandwich (changes requested):
Lets:
* remove any product urls from the old content api if they have migrated
* add testing, as this logic has now gotten complicated

RubenSandwich (approved):
Looks great. 👍🏻

src/lib/__tests__/docs-content-fields.test.ts
```diff
@@ -0,0 +1,83 @@
+import { describe, it, expect, vi } from 'vitest'
+import { allDocsFields } from '../sitemap/docs-content-fields'
+
+// vi.mock('../sitemap/helpers', () => ({
```
RubenSandwich, line 4: @LeahMarieBush can you remove this comment?

PR #2669, opened 2025-01-18
Only return notfound if theres an error in retrieving the data (tutorials and collections)

im2nguyen:
## 🔗 Relevant links

Only return notfound when there's an error in retrieving the data for validated-patterns (doing it for tutorials and collections view)

- [Asana task](url) 🎟️

### DIFF
--- a/src/pages/validated-patterns/[...tutorialSlug]/index.tsx
+++ b/src/pages/validated-patterns/[...tutorialSlug]/index.tsx
@@ -24,12 +24,14 @@ export async function getStaticProps({
 	try {
 		const props = await getValidatedPatternsTutorialViewProps(params.tutorialSlug)
 
-	// If the tutorial doesn't exist, hit the 404
+		// If the tutorial doesn't exist, hit the 404
 		if (!props) {
+			console.error('Tutorial not found')
 			return { notFound: true }
 		}
 		return props
 	} catch (error) {
+		console.error(error)
 		return { notFound: true }
 	}
 }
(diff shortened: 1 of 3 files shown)


im2nguyen:
Currently, if you go to `/validated-designs`, the home page loads but the individual tutorials and collections do not load.

To test:

1. Build the `dev-portal` image locally
1. In this [PR](https://github.com/hashicorp/tutorials/pull/2430) use the local image. The individual tutorials and collections for validated designs should work

src/pages/validated-patterns/[...tutorialSlug]/index.tsx
```diff
@@ -21,14 +21,17 @@ export async function getStaticProps({
...
 	{ props: ValidatedPatternsTutorialViewProps } | { notFound: boolean }
 > {
-	return { notFound: true }
-	const props = await getValidatedPatternsTutorialViewProps(params.tutorialSlug)
+	try {
+		const props = await getValidatedPatternsTutorialViewProps(params.tutorialSlug)
 
 	// If the tutorial doesn't exist, hit the 404
-	if (!props) {
+		if (!props) {
```
Comment (RubenSandwich) [inline src/pages/validated-patterns/[...tutorialSlug]/index.tsx:28]: @im2nguyen let's log the error here, and also in the catch. Because these two shouldn't fail very often considering the use of `getStaticPaths`.

RubenSandwich (changes requested):
Let's just add some logging.

src/pages/validated-patterns/[...tutorialSlug]/index.tsx
Comment (im2nguyen) [reply src/pages/validated-patterns/[...tutorialSlug]/index.tsx:28]: just updated! 

this shouldn't fail at all (unless the directory doesn't exist in `/tutorials`, which is the state we're at right now)

RubenSandwich (approved):
Looks great. 👍🏻

PR #2691, opened 2025-02-19
Rn/fix udr docker image loading

RubenSandwich:
## 🔗 Relevant links

- [Asana task](https://app.asana.com/0/1209152747858598/1209402819381989) 🎟️

## 🗒️ What

When we are [fixing the MDX img links in dev-portal we are using the ${process.env.UNIFIED_DOCS_API} url](https://github.com/hashicorp/dev-portal/blob/d1ec2209615f2bb9b3f43afc53d372d3013b8faa/src/lib/remark-plugins/remark-rewrite-assets.ts#L38), which within docker is `http://experimental-docs-api:8080`. However that browser is making that request, so has no access to `http://experimental-docs-api:8080`  rather it should be: `http://localhost:8080`.

## 🧪 Testing

1. Checkout this PR and create a local docker image named `dev-portal:fix-docker-udr`.
> docker build -t dev-portal:fix-docker-udr .
2. Checkout `main` on web-unified-docs and [update docker-compose.yaml](https://github.com/hashicorp/web-unified-docs/blob/main/docker-compose.yaml#L3) to use `dev-portal:fix-docker-udr` instead of `hashicorp/dev-portal`.
3. Start UDR with `make`.
4. Check image loading in this url: http://localhost:3000/terraform/plugin/framework/internals/rpcs


src/lib/remark-plugins/remark-rewrite-assets.ts
```diff
@@ -34,9 +34,18 @@ export function remarkRewriteAssets(args: {
...
-					url = new URL(
-						`${process.env.UNIFIED_DOCS_API}/api/assets/${product}/${version}${node.url}`
-					)
+					let domain
+
+					if (
+						process.env.NODE_ENV === 'development' &&
+						process.env.HASHI_ENV === 'unified-docs-sandbox'
+					) {
+						domain = `http://localhost:${process.env.UNIFIED_DOCS_PORT}`
```
LeahMarieBush, line 43: Can you add UNIFIED_DOCS_PORT to the .env file as well? Otherwise running the API and dev-portal both locally makes the url `http://localhost:undefined....`

LeahMarieBush (approved):
LGTM 👍 tested locally and with a new dockerfile and both worked

PR #2693, opened 2025-02-25
Fix broken previous versions

im2nguyen:
## 🔗 Relevant links

This fixes the broken previous versions reported by the team. It checks if the path exists before it splits it.

Tested it successfully locally.

- [Asana task](url) 🎟️

## 🧪 Testing

1. Start dev-portal locally `npm run start`
2. Go to a product documentation ([Boundary](http://localhost:3000/boundary/docs))
3. Select a [previous version](http://localhost:3000/boundary/docs/v0.16.x) in the dropdown
4. Verify it loads 
image

### DIFF
--- a/src/views/docs-view/loaders/__tests__/remote-content.test.ts
+++ b/src/views/docs-view/loaders/__tests__/remote-content.test.ts
@@ -409,6 +409,49 @@ describe('RemoteContentLoader', () => {
 			path: 'v0.4.x/some-path',
 		})
 	})
+
+	test('should handle navData where first item has no path property', async () => {
+		// Create a modified version of navData with first item missing path property
+		const modifiedNavData = {
+			...navData_200,
+			result: {
+				...navData_200.result,
+				navData: [
+					// First item with no path property
+					{ title: 'Item with no path' },
+					// Rest of the items
+					...navData_200.result.navData.slice(1),
+				],
+			},
+		}
+
+		scope
+			.get('/api/content/waypoint/version-metadata')
+			.query({ partial: 'true' })
+			.reply(200, versionMetadata_200)
+		scope
+			.get('/api/content/waypoint/doc/v0.4.x/commands')
+			.reply(200, document_v4)
+		scope
+			.get('/api/content/waypoint/nav-data/v0.4.x/commands')
+			.reply(200, modifiedNavData)
+
+		const versionedDocsLoader = new RemoteContentLoader({
+			...loader.opts,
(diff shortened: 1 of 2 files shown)


src/views/docs-view/loaders/__tests__/remote-content.test.ts
```diff
@@ -409,6 +409,49 @@ describe('RemoteContentLoader', () => {
...
+		// This should not throw an error
+		const props = await versionedDocsLoader.loadStaticProps({
+			params: {
+				page: ['v0.4.x'],
+			},
+		})
+
+		// Verify that the function completed successfully
+		expect(props).toBeDefined()
+		expect(props.navData).toBeDefined()
```
rmainwork, line 453: Instead of asserting that `props` and `props.navData` are defined, I'd maybe suggest asserting some things about that first element. That way we're being more explicit about what we expect the outcome of the test to be.

```suggestion
		expect(props.navData[0].path).not.toBeDefined()
		expect(props.navData[0].title).toEqual('Item with no path')
```

What do you think?

src/views/docs-view/loaders/__tests__/remote-content.test.ts
```diff
@@ -409,6 +409,49 @@ describe('RemoteContentLoader', () => {
 			path: 'v0.4.x/some-path',
 		})
 	})
+
+	test('should handle navData where first item has no path property', async () => {
```
rmainwork, line 413:
```suggestion
	test('does not remove first navigation items with no path property', async () => {
```

src/views/docs-view/loaders/remote-content.ts
```diff
@@ -251,6 +251,7 @@ export default class RemoteContentLoader implements DataLoader {
 			// Remove item if it's an empty path (for non-latest versions, the version is prepended to the path)
 			if (
 				navData.navData.length > 1 &&
+				navData.navData[0].path &&
 				navData.navData[0].path.split('/').length > 1
```
rmainwork, line 255: We should be able to use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) here to get the same result

```suggestion
				navData.navData[0].path?.split('/').length > 1
```

rmainwork (changes requested):
Thanks for picking this up, Tu! A few small things but overall this is looking great! :)

rmainwork (approved):

PR #2704, opened 2025-03-05
Port the eslint config from UDR and next-hashicorp

rmainwork:
## 🔗 Relevant links

## 🗒️ What

This PR is the first of several aimed to rectify the broken linting setup in devdot which has been the cause of at least one incident recently. The intent is to bring together multiple PRs (starting with this one) on to an assembly branch (`rm/fix-linting`) and then PR from there into `main`. This helps to reduce the size of each PR as well as getting multiple eyes and reviews on the changes being made.

## 🤷 Why

This is an initial attempt to rectify the broken linting on dev dot by first bringing in the eslint config. 

## 🛠️ How

Porting the linting setup from UDR was deemed to be easier and faster than attempting to modify `hashicorp-next`, republish and then try to integrate the new package version.

The eslint config from UDR was written as an esmodule in order to use things like import - so a straight copy and paste wasn't possible. But with a little massaging (i.e: swapping out `import` for require) it seems to work in a rudimentary sense.

Additionally, some tweaks had to be made to package.json to uninstall `hashicorp-next` and remove references to it in pre-commit hooks(in order to be able to commit)

Finally the old rules from next-hashicorp were ported over in attempt to resolve lint issues and bring the lint rules in line with the current code style in devdot


eslint.config.js
```diff
@@ -0,0 +1,88 @@
...
+const pluginReact = require('eslint-plugin-react')
+const noBarrelFiles = require('eslint-plugin-no-barrel-files')
+
+module.exports = [
+	{
+		ignores: [
+			'.next',
+			'**/dist',
+			'eslint.config.js',
+			'content',
```
LeahMarieBush, line 13: dev-portal doesn't have a `content` directory so I think this can be removed

eslint.config.js
```diff
@@ -0,0 +1,88 @@
...
+			'public',
+			'vitest.config.mjs',
```
LeahMarieBush, line 15: The vitest config file in dev-portal ends in `mts` instead of `mjs`

eslint.config.js
rmainwork: Good point! I guess my rip was more shameless than I thought :p

rmainwork: Fixed that in [d3d638b](https://github.com/hashicorp/dev-portal/pull/2704/commits/d3d638bba6869e7f0dbc984317bcedf5f310db59)

eslint.config.js
rmainwork: Fixed in [a2b1e45](https://github.com/hashicorp/dev-portal/pull/2704/commits/a2b1e45e5f931ad146f9ade9808e849b959d507a)

PR #2707, opened 2025-03-06
@typescript-eslint/no-require-imports

rmainwork:
## 🔗 Relevant links

## 🗒️ What

This PR updates the eslint config file to allow the usage of inline `require` (thus resolving 13 linting issues). Under closer inspection of this (and of www) it seems as though (in the case of the latter) inline `require()` is used in various places. 

## 🤷 Why

## 💭 Anything else?
The significant commit can be found here: https://github.com/hashicorp/dev-portal/commit/7a9724b30a815f8b251065b48fcfeefd268aea27.


RubenSandwich (approved):
Makes sense and looks good. 👍🏻

PR #2710, opened 2025-03-10
no-prototype-builtins

rmainwork:
## 🔗 Relevant links

## 🗒️ What

This PR resolves linting issues related to the `no-prototype-builtins` linting rule.

## 🤷 Why

## 💭 Anything else?
Only to drop a link to [this](https://eslint.org/docs/latest/rules/no-prototype-builtins) page on `no-prototype-builtins` and some suggested ways to resolve that error.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -33,8 +33,6 @@ module.exports = [
 			'no-dupe-class-members': 'off',
 			// 'tsc' already handles this (https://github.com/typescript-eslint/typescript-eslint/issues/477)
 			'no-undef': 'off',
-			// Add TypeScript specific rules (and turn off ESLint equivalents)
-			// '@typescript-eslint/no-angle-bracket-type-assertion': 'warn',
 			'no-array-constructor': 'off',
 			'@typescript-eslint/no-array-constructor': 'warn',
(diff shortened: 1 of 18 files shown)


rmainwork:
Just to recap the conversation that @heatlikeheatwave and I had on slack:

- It looks like this rule comes from `eslint-recommended`: https://github.com/eslint/eslint/blob/main/lib/rules/no-prototype-builtins.js which is extended by `tseslint.configs.recommended` which is used [here](https://github.com/hashicorp/dev-portal/blob/rm/fix-linting/eslint.config.js#L20)
- It looks like there was previously some discussion [here](https://app.asana.com/0/1100423001970639/1199667739287945/f) about disabling this rule as it has the potential to lead to _"some funky code patterns"_
- A discussion topic was added to the meeting agenda (thanks @heatlikeheatwave!) to gather team consensus on this and raise awareness

rmainwork:
To recap a discussion we had in our meeting 3/17 - we decided to revert 3e67afd and disable this lint rule in line with our current coding conventions. Especially since this rule was actually disabled before in `web-platform-packages`

heatlikeheatwave (approved):
🚀

PR #2714, opened 2025-03-12
no-array-index-key

rmainwork:
## 🔗 Relevant links

## 🗒️ What

This PR resolves linting issues related to the `no-array-index-key` linting rule.

## 🤷 Why

## 💭 Anything else?
1. I would like to call out [this change](https://github.com/hashicorp/dev-portal/compare/rm/fix-linting...rm/no-array-index-key?expand=1#diff-0ee33eb12315f956f8e988aa744ca328320dc8e0b87700ef92db4185b3030006R9) where component typing was updated to use `PropsWithChildren` instead of manually typing `{ children: React.ReactChild[] }` - it wasn't  _directly_ in scope but it was showing as a react warning in my IDE and it seemed like a quick and easy fix at the time...
2. I'd also like to draw attention to [this file](https://github.com/hashicorp/dev-portal/pull/2714/files#diff-4a662c6db256155738bc8b6e0b6ab86cfb90e6b05511f17a6af09c7dbf351d97) and the changes in it. I tried to use various values (other than `index`) for the key, but in one case I had to just disable the lint rule for that line and in a couple of other places, I had to combine the fields in scope to get a (hopefully unique) value. Would love to hear if there's a better way of doing this.

### DIFF
--- a/src/components/author-primitives/packer/badges-header/index.tsx
+++ b/src/components/author-primitives/packer/badges-header/index.tsx
@@ -3,21 +3,17 @@
  * SPDX-License-Identifier: MPL-2.0
  */
 
-import React from 'react'
+import React, { PropsWithChildren } from 'react'
 import s from './style.module.css'
 
-function BadgesHeader({
-	children,
-}: {
-	children: React.ReactChild[]
-}): React.ReactElement {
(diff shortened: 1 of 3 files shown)


LeahMarieBush (approved):
LGTM!

PR #2716, opened 2025-03-13
Resolve remaining lint issues

rmainwork:
## 🔗 Relevant links

## 🗒️ What

This PR resolves the remaining lint issues - specifically(in decreasing number of errors): 

- `no-useless-escape`
- `valid-typeof`
- `react-hooks/exhaustive-deps`
- `@next/next/no-img-element`
- `import/no-anonymous-default-export`
- `react-hooks/rules-of-hooks`
- `@typescript-eslint/no-use-before-define`
- `@typescript-eslint/no-empty-object-type`
- `prefer-const`

Since each one of these rules only has one or two errors, one PR grouping them all together should be relatively digestible (in contrast to some of the other lint errors which had larger change sets). At this scale, separating each error out into it's own PR would've proved less efficient.

## 🤷 Why

The final PR to resolve existing lint issues on devdot


LeahMarieBush (commented):
Code looks good to me but it looks like the build and lint steps are failing. Can you look into the build and make sure it gets working before merging this? Also, if this is the remaining lint issue PR, should the lint check be passing as well?

LeahMarieBush (approved):
LGTM 👍

rmainwork:
Since the build is now passing (despite my not having changed anything). I'm going to merge this into the assembly branch and see what happens....

PR #2722, opened 2025-03-20
feat: zendesk search integration

LeahMarieBush:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-leah-featzendesk-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1204810504608081/1209669994962398/f) 🎟️

## 🗒️ What

This PR adds a 'Support' tab to the dev-portal search modal. The content from this tab is sourced from the algolia indexes for zendesk (dev_ZENDESK and prod_ZENDESK depending on environment). It also adds a little external link icon to the end of the url for support items.

Since zendesk articles don't have descriptions like the docs or tutorials articles do, I used the algolia snippet component to get a section of where the search term originated in the body of the zendesk article.

## 📸 Design Screenshots
Screenshot 2025-03-26 at 4 01 45 PM
Screenshot 2025-03-26 at 4 02 07 PM

## 🧪 Testing

1. Go to [preview link](https://dev-portal-git-leah-featzendesk-hashicorp.vercel.app/)
2. Open search modal by clicking search bar or command k
3. Type in a search query
4. Go to support tab and see the content from zendesk
5. Click on an item and verify it opens in a new tab and goes to the right article

---
  - https://app.asana.com/0/0/1209669994962398


LeahMarieBush:
Adding do not merge label until support team is ready for launch

src/components/command-bar/commands/search/unified-search/components/unified-hits-container/helpers/gather-search-tabs-content.tsx
```diff
@@ -23,7 +23,7 @@ const PRODUCT_SLUGS_WITH_INTEGRATIONS =
  * Each content type tab has a set of properties required for rendering.
  */
 export interface UnifiedSearchTabContent {
-	type: 'global' | 'docs' | 'tutorials' | 'integrations'
+	type: 'global' | 'docs' | 'tutorials' | 'integrations' | 'knowledgebase'
```
rmainwork, line 26:
```suggestion
	type: UnifiedSearchableContentType
```

src/components/command-bar/commands/search/helpers/hit-counts-provider.tsx
```diff
@@ -7,7 +7,10 @@ import { ReactNode, createContext, useContext, useState } from 'react'
 
 const SearchHitsContext = createContext([])
 
-type HitCounts = Record<'docs' | 'tutorials' | 'integrations', number>
+type HitCounts = Record<
+	'docs' | 'tutorials' | 'integrations' | 'knowledgebase',
```
rmainwork, line 11: What do you think about  using the existing `UnifiedSearchableContentType` to replace these unions (albeit with `Exclude` to yank 'global' out of there in this specific instance) instead?

```suggestion
	Exclude<UnifiedSearchableContentType, 'global'>
```

src/components/command-bar/commands/search/unified-search/types.ts
```diff
@@ -36,6 +36,7 @@ export type UnifiedSearchableContentType =
 	| 'docs'
 	| 'integration'
 	| 'tutorial'
+	| 'knowledgebase'
```
rmainwork, line 39: What do you think about rewriting this union type as an enum (as the source of truth for the values) and then a [template literal type](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) to generate the union type. 

```ts
export enum SearchContentTypes {
	GLOBAL = 'global',
	DOCS = 'docs',
	INTEGRATION = 'integration',
	TUTORIAL = 'tutorial',
	KNOWLEDGEBASE = 'knowledgebase',
}

/**
 * Derived union type from {@link SearchContentTypes}
 */
export type UnifiedSearchableContentType =`${SearchContentTypes}`
```
It _should_ be functionally the same

image

However, it would also provide some advantages

1. It would give us the advantages of using an enum (the ability to refer canonically to those values elsewhere in the code like [here](https://github.com/hashicorp/dev-portal/pull/2722/files#diff-38edeed014fa9ad4df05e7b63c5aaed3616f488d2a402df9ec27a1cbbf41273aR100)) so we could use `SearchContentTypes.GLOBAL` instead of just the string `'global'`. This seems trivial, but during the discovery phase of this ticket, I noticed there are a few typos in the existing code (we have `'global' | 'docs' | 'tutorials' | 'integrations'` [here](https://github.com/hashicorp/dev-portal/blob/917e3e780c05b2898851c80b4bc525fbbaa25244/src/components/command-bar/commands/search/unified-search/components/unified-hits-container/helpers/gather-search-tabs-content.tsx#L26C8-L26C56), [here](https://github.com/hashicorp/dev-portal/blob/917e3e780c05b2898851c80b4bc525fbbaa25244/src/components/command-bar/commands/search/unified-search/components/unified-hits-container/helpers/gather-search-tabs-content.tsx#L48C15-L48C66) and [here](https://github.com/hashicorp/dev-portal/blob/917e3e780c05b2898851c80b4bc525fbbaa25244/src/components/command-bar/commands/search/unified-search/components/unified-hits-container/helpers/gather-search-tabs-content.tsx#L86)......and then `hit.type !== 'integration'` [here](https://github.com/hashicorp/dev-portal/blob/917e3e780c05b2898851c80b4bc525fbbaa25244/src/components/command-bar/commands/search/unified-search/components/unified-hits-container/helpers/gather-search-tabs-content.tsx#L115))


2. It would also allow us to provide documentation for the enum values (if we wanted) like this:

    ```ts
    export enum SearchContentTypes {
	    GLOBAL = 'global',
	    DOCS = 'docs',
	    INTEGRATION = 'integration',
	    TUTORIAL = 'tutorial',
	    /**
	     * Data from zendesk
	     */
	    KNOWLEDGEBASE = 'knowledgebase'
    }
    ```

    image

What do you think?

src/components/command-bar/commands/search/unified-search/components/unified-hit/helpers/get-unified-hit-props.ts
```diff
@@ -16,7 +16,7 @@ import type { UnifiedHitProps } from '../types'
  */
 export function getUnifiedHitProps(hit: Hit): UnifiedHitProps {
 	// Content type, for icon
-	const type = hit.type as 'docs' | 'tutorial' | 'integration'
+	const type = hit.type as 'docs' | 'tutorial' | 'integration' | 'knowledgebase'
```
rmainwork, line 19:
```suggestion
	const type = hit.type as Exclude<SearchContentTypes, 'global'>
```

rmainwork (changes requested):

src/components/command-bar/commands/search/unified-search/types.ts
LeahMarieBush: I cleaned this up to use SearchContentTypes everywhere instead of the literal string values. I also decided to just use the enum as the type instead of creating the separate UnifiedSearchableContentType and it all seems to work as expected! Thanks for the feedback, it is much cleaner now 🎉

src/components/command-bar/commands/search/unified-search/types.ts
```diff
@@ -31,14 +31,16 @@ export type TutorialHitObject = Hit<{
 		level: CollectionLevelOption
 	}
 
-export type UnifiedSearchableContentType =
-	| 'global'
-	| 'docs'
-	| 'integration'
-	| 'tutorial'
+export enum SearchContentTypes {
```
RubenSandwich, line 34: Love this. ❤️

RubenSandwich (approved):
Looks good. 👍🏻

src/components/command-bar/commands/search/unified-search/types.ts
rmainwork: Sounds good - I wasn't sure how conservative you wanted to be with the types so in case you wanted to maintain the union, I figured I'd suggest a way to keep that if we wanted to. But I agree it's much neater to just have the enum.

rmainwork (changes requested):

src/components/command-bar/commands/search/unified-search/components/dialog-body/index.tsx
```diff
@@ -147,6 +153,19 @@ function SearchResults({
 						)
 					}
 				)}
+
+				<Index indexName={ZENDESK_ALGOLIA_INDEX_NAME} indexId="knowledgebase">
```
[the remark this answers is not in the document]
LeahMarieBush: Thanks for pointing this one out! I must have missed it

src/components/command-bar/commands/search/unified-search/components/dialog-body/index.tsx
```diff
@@ -136,8 +135,15 @@ function SearchResults({
...
 			    Maybe helpful to think of this as "the part that fetches results". */}
 			<InstantSearch indexName={ALGOLIA_INDEX_NAME} searchClient={searchClient}>
-				{['global', 'docs', 'integration', 'tutorial'].map(
-					(type: UnifiedSearchableContentType) => {
+				{[
+					SearchContentTypes.GLOBAL,
+					SearchContentTypes.DOCS,
+					SearchContentTypes.INTEGRATION,
+					SearchContentTypes.TUTORIAL,
+				].map(
```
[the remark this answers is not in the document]
LeahMarieBush: I like this idea but since we are not mapping over every value (SearchContentTypes.KNOWLEDGEBASE is not here), I'm going to keep this the same for now. When we switch over to using the omni index after launch I'll update this to use Object.values tho!

rmainwork (approved):

PR #2723, opened 2025-03-26
Add separate Fedora 40 and 41 install instructions

BrianMMcClain:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-fedora-40-41-hashicorp.vercel.app/terraform/install#linux) 🔎

## 🗒️ What

Add separate Fedora 40 and 41 install instructions

## 🤷 Why

According to our [Official Packaging Guide](https://www.hashicorp.com/en/official-packaging-guide), we currently support Fedora versions 40 and 41. In Fedora 41, the command to add a repository changed and did not implement backwards compatibility, meaning the install instructions are different for each version. [source](https://github.com/hashicorp/terraform/pull/35980)

Once we end support for Fedora 40, we won't need this differentiation between versions. We'll be able to revert this change.

I'm not sure the best approach to this, so I'm very open to another way to document this. This will also help us deprecate the following pages, which we've wanted to do to consolidate our install instructions into one source of truth:

- [Install Terraform Using Yum Packages for Linux Distributions](https://developer.hashicorp.com/terraform/cli/install/yum)
- [Install Terraform Using APT Packages for Debian and Ubuntu](https://developer.hashicorp.com/terraform/cli/install/apt)

## 🧪 Testing

Tested on Fedora 40
- [ ] docker run -it fedora:40 /bin/bash
- [ ] sudo dnf install -y dnf-plugins-core
- [ ] sudo dnf config-manager --add-repo https://rpm.releases.hashicorp.com/fedora/hashicorp.repo
- [ ] sudo dnf -y install terraform
- [ ] terraform -v

Tested on Fedora 41
- [ ] docker run -it fedora:41 /bin/bash
- [ ] sudo dnf install -y dnf-plugins-core
- [ ] sudo dnf config-manager addrepo --from-repofile=https://rpm.releases.hashicorp.com/fedora/hashicorp.repo
- [ ] sudo dnf -y install terraform
- [ ] terraform -v


LeahMarieBush (approved):
👍

PR #2732, opened 2025-04-03
feat: use omni algolia index for zendesk articles

LeahMarieBush:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

This PR update the zendesk algolia configuration to use the omni algolia index instead of the zendesk one. This will make the zendesk results returned in the 'All' tab without special configuration. The reason the separate index was used at first was so data in the dev and prod index could be tested without impacting the production site global results.

Since the algolia staging_DEVDOT_omni index settings were adjusted as part of this PR (and the prod ones will be adjusted after this merges as well), the search results across all tabs will be ordered differently from what they are on prod. They now are ordered based off the following criteria in order: page_title, description, headings. Now in the global search, results from different sections may be intermingled instead of all the results from one tab and then the next tab and so on.

## 🧪 Testing

1. Go to [preview link](https://dev-portal-git-leah-featzendesk-one-index-hashicorp.vercel.app/)
2. Open the search
3. Verify it behaves the same way it does in prod

### DIFF
--- a/config/base.json
+++ b/config/base.json
@@ -9,8 +9,7 @@
 			"description": "Explore {product} product documentation, tutorials, and examples."
 		},
 		"algolia": {
-			"unifiedIndexName": "prod_DEVDOT_omni",
-			"zendeskIndexName": "prod_ZENDESK"
+			"unifiedIndexName": "prod_DEVDOT_omni"
 		},
 		"analytics": {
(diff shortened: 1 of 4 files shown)


rmainwork (approved):
🚀

PR #2699, opened 2025-03-03
Posthog tracking for video embeds starts

im2nguyen:
## 🔗 Relevant links

This PR updates the VideoEmbed to send a custom PostHog event `video_start` whenever someone starts a video play. This works on both wistia and youtube videos. 

In addition, it adds `VideoEmbed` as an available MDX component for docs. This lets us use `VideoEmbed` instead of iframe for videos in documentation, so the behavior is consistent throughout the platform.

- [Preview link](https://dev-portal-git-video-embed-posthog-tracking-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/1202801212949828/1209530615984378/f) 🎟️

## 🧪 Testing

Related PR: https://github.com/hashicorp/terraform-docs-common/pull/866

1. Go to `dev-portal` and check out the `video-embed-posthog-tracking` branch
2. Build the image locally (`docker build . -t dev-portal-video-tracking`)
3. Update `DOCKER_IMAGE_LOCAL="dev-portal-local"` in this repo's `website/Makefile` file to `dev-portal-video-tracking`.
4. Start local preview (`make website/local`). You may need to add `.env.local` to `website`.
5. Navigate to `/terraform/cloud-docs/recommended-practices`
6. Verify the videoembed works as expected, check posthog to see if `video_play` event is triggered

image

PR #2738, opened 2025-04-23
Revert "Posthog tracking for video embeds starts"

rmainwork:
This PR reverts hashicorp/dev-portal#2699 due to a possible posthog configuration issues preventing content from rendering.

We believe there is a empty url in the posthog prod data that is causing a markdown rendering issue.


RubenSandwich (approved):
👍🏻

im2nguyen:
Hey, can you point to where the markdown rendering issue is so we can fix it asap? we need this posthog tracking

PR #2750, opened 2025-05-02
Provide UDR bypass token

rmainwork:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-rm-add-vercel-auth-bypass-token-111700-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/home/1207213190767773/1209731544373089) 🎟️

## 🗒️ What

Provide UDR Vercel auth bypass token(if present in the environment vars) when making requests to UDR. 

## 🤷 Why

To prevent bots from hitting our UDR API - we want to turn on deployment protection for preview deployments . However, doing so currently breaks UDR previews(which deploy a copy of dev portal from `main` alongside a preview of UDR)  when dev portal attempts to connect to the UDR preview endpoint to build pages.

## 🛠️ How

Using [github code search](https://github.com/search?q=repo%3Ahashicorp%2Fdev-portal+process.env.UNIFIED_DOCS_API&type=code) (as well as `grep`) I searched through the codebase for instances of `process.env.UNIFIED_DOCS_API`. This provided a starting point to find the instances where requests were being made to the UDR API. However, in some cases - [`getUDRDocsPaths`](https://github.com/hashicorp/dev-portal/blob/b526355311d83a3a5a4e23aa4471afee16dc823b/src/lib/sitemap/docs-content-fields.ts#L27) is used instead so another search had to be made to find all occurrences of _that_ too. Finally, as s cross-check a `grep` search was made locally for `fetch(` to find _all_ instances where _any_ HTTP calls were being made and make sure that any calls to the UDR API also passed the auth bypass token in the header

## 🧪 Testing

This may be tricky to test without merging to `main`, as it's functionality cuts across two systems (dev portal and UDR) so the only way to really test this (as far as I can see right now) is to:

- Merge this PR to `main` (after review of course!), 
- Configure the UDR frontend Vercel app with the relevant auth bypass token 
- Open a content change PR against UDR
- Watch the Vercel (and GitHub actions) logs to make sure that the build/deployment succeeds and the frontend can indeed communicate with the backend

Open to any other methods of testing prior to merge, however.

## 💭 Anything else?

I was curious to know what it would take to implement [this ticket](https://app.asana.com/1/90955849329269/project/1209314279859096/task/1210144435963336?focus=true) and did some exploratory work into that (wondered if it might make sense to knock that out at the same time as this since I'm working in that area). A `fetch` wrapper was going well initially, but that magic `getContentApiBaseUrl` function left us with a base URL that could be the old content API _or_ UDR so that may not have ended up working. In short, it would require some re-architecting of things that is likely outside of the scope of this ticket.... oh well.

### DIFF
--- a/src/data/vault.json
+++ b/src/data/vault.json
@@ -161,11 +161,6 @@
 			"iconName": "api",
 			"name": "APIs",
 			"path": "api-docs"
-		},
-		{
-			"iconName": "guide",
-			"name": "Tutorials",
-			"path": "tutorials"
 		}
 	],
 	"integrationsConfig": {


LeahMarieBush (approved):
👍 I'm not sure of any other ways to test this other than what you put in the PR description so this LGTM!

PR #2751, opened 2025-05-02
Change order of nav and sidebar items

RubenSandwich:
## 🔗 Relevant links

- [Asana task](url) 🎟️

### DIFF
--- a/src/data/vault.json
+++ b/src/data/vault.json
@@ -161,6 +161,11 @@
 			"iconName": "api",
 			"name": "APIs",
 			"path": "api-docs"
+		},
+		{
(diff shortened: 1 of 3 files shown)


src/data/vault.json
```diff
@@ -161,6 +161,11 @@
 			"iconName": "api",
 			"name": "APIs",
 			"path": "api-docs"
+		},
```
schavis, line 164: Can we put tutorials directly after Vault so it's second in the list? I know it was at the end before, but it makes more sense to get l group it together with docs instead of the reference tabs

schavis (approved):

PR #2756, opened 2025-05-20
Update video embed to also track video plays

im2nguyen:
## 🔗 Relevant links

The current implementation only tracks video starts rather than video plays (starts and when users restart the video after clicking pause). The historic way we've tracked video plays before the video embed was whenever users clicked the CSS selector that mapped to the wisita/youtube embed. This update makes it closer to the original implementation.

We still want to keep `video_start` as a metric. This newer `video_play` metric will let us reconcile with our historical data.

- [Asana task](url) 🎟️

## 🧪 Testing

1. Go to any page with a video in the deploy preview ([example](https://dev-portal-git-video-embed-ph-track-plays-hashicorp.vercel.app/terraform/tutorials/aws-get-started/infrastructure-as-code)). For event tracking to work, you must have adblocker disabled... also Arc disables tracking by default. Please use Safari or Chrome.
2. Start the video. Pause the video. Click the video again.
3. You should see two events for `video_play` in PostHog ([event details](https://eu.posthog.com/project/29987/data-management/events/0196efdd-2a0c-7631-b917-3c4e6330c57c))
  image

### DIFF
--- a/src/components/video-embed/index.tsx
+++ b/src/components/video-embed/index.tsx
@@ -15,7 +15,7 @@ import {
 	useMilestones,
 } from './helpers'
 import s from './video-embed.module.css'
-import { trackVideoStart } from 'lib/posthog-events'
+import { trackVideoStart, trackVideoPlay } from 'lib/posthog-events'
 
 /**
  * MAX_PLAYBACK_SPEED is based on max speeds for YouTube & Wistia.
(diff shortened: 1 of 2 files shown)


kaitlynnefuery (approved):
@im2nguyen this LGTM based on the test. I think it's good that @LeahMarieBush is tagged though. I think as a part of the PostHog experiments we _might_ be adjusting how PostHog is loaded to enable server side (rather than just client side) so she should be aware of this in relation to her work.

im2nguyen:
Sounds good! I'm going to wait for @LeahMarieBush to review before we merge, thank you! 😄

LeahMarieBush (approved):
LGTM 👍

PR #2762, opened 2025-05-28
Make UDR the source for terraform-docs-common and ptfe-releases

rmainwork:
## 🗒️ What

Update the dev portal config to make UDR the source for terraform-docs-common and terraform-entperirse (formerly ptfe-releases) docs.

## 💭 Anything else?

The (new) URL slug for `/content/ptfe-releases` is _actually_ `/content/terraform-enterprise`, but presumably (due to the work done in https://github.com/hashicorp/web-unified-docs/pull/419) `/content/ptfe-releases` is perfectly fine to use here (with a view to updating dev portal down the road to instead make requests to `/content/terraform-enterprise` instead?

PR #2765, opened 2025-06-02
Add PTFE and terraform-docs-common to UDR config to fix broken links

im2nguyen:
## 🔗 Relevant links

Since the TFE and terraform-docs-common repos are now in UDR, we also need to update the base and development config in dev-portal so it knows to use UDR for the content. We were seeing bugs in devdot as a result of this:

image

- [Preview link](https://dev-portal-git-migrate-ptfe-tf-docs-common-udr-hashicorp.vercel.app/terraform/docs/tools/mcp-server) 🔎

### DIFF
--- a/config/base.json
+++ b/config/base.json
@@ -45,6 +45,8 @@
 			"terraform-plugin-testing",
 			"terraform-docs-agents",
 			"terraform-cdk",
+			"ptfe-releases",
+			"terraform-docs-common",
 			"terraform"
 		]
 	}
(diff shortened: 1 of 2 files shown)


RubenSandwich:
@im2nguyen that was caused by a PR shipping that was supposed to happen after UDR shipped. We have a PR to turn on UDR for terraform-enterprise and HCP Terraform: https://github.com/hashicorp/dev-portal/pull/2762.

Closing.

PR #2782, opened 2025-07-09
Fix duplicate PostHog user_authenticated event due to layout re-renders

Younique98:
- Memoize last captured user Id using useRef
- Added guards to prevent duplicate firing on auth state changes
- Avoided structural changes to keep the patch low-risk and self-contained

## 🔗 Relevant links

- [Preview link](https://dev-portal-git-ericat-choreposthog-auth-event-dedupe-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/1/90955849329269/project/1210546436661486/task/1209865393442034?focus=true) 🎟️

## 🗒️ What

This PR fixes an issue where the user_authenticated PostHog event was firing mutliple times per session due to layout-level hook usage and rerenders triggered by auth transitions.

## 🤷 Why

The original uer_authenticated PostHog event was firing multiple times per session. This was caused by how the useAuthentication hook is invoked at layout level. It ran anytime the app layout mounts, rehydrates or reacts to changes in auth state (like session restoration or token refresh).

As a result, even though the user remained the same, the effect re-ran and sent duplicate tracking events which inflated analytics and obscuyred the true number of log ins. 

## 🛠️ How

- Memorized the last captured userId using useRef ensuring firing once per session per user
- Added guards to pervent duplicate PostHog capture/identify calls for the same user
- Left hook usage and layout structure unchanged to avoid risk or broader refactors

The issue stemmed from layout-level usage of the auth hook, which is valid and intentional but means side effects inside that hook (like analytics) need deduplication logic. This fix keeps everything self contained.

## 📸 Design Screenshots
In prod before
Screenshot 2025-07-08 at 4 34 17 PM

Locally - Dev before latest fix
Screenshot 2025-07-09 at 10 48 25 AM
Screenshot 2025-07-09 at 1 07 06 PM

 ###WIth latest Fix
Screenshot 2025-07-09 at 12 49 07 PM

## 🧪 Testing

- [ ] Step 1 Manually sign in and out mutlipe times with a console.log and console.count confirming event fires appropriately
- [ ] Step 2 Verify background reauths no longer trigger additional tracking
- [ ] Step 3 Confirm with PostHog & DevTools to identify and capture calls

### DIFF
--- a/src/hooks/use-authentication/helpers/make-sign-in.ts
+++ b/src/hooks/use-authentication/helpers/make-sign-in.ts
@@ -7,6 +7,7 @@ import { NextRouter } from 'next/router'
 import { signIn, SignInOptions as _SignInOptions } from 'next-auth/react'
 import { ValidAuthProviderId } from 'types/auth'
 import { DEFAULT_PROVIDER_ID } from '..'
+import posthog from 'posthog-js'
 
 interface MakeSignInOptions {
 	routerPath: NextRouter['asPath']
@@ -26,6 +27,12 @@ const makeSignIn = ({ routerPath }: MakeSignInOptions) => {
 		options: SignInOptions = {}
 	) => {
 		const { callbackUrl = routerPath, redirect = true } = options
+
+		posthog.capture('user_authenticated', {
+			source: 'sign_in',
+			trigger: 'manual_sign_in',
+			timestamp: new Date().toISOString(),
+		})
 
 		return signIn(provider, { callbackUrl, redirect })
 	}
--- a/src/hooks/use-authentication/index.ts
+++ b/src/hooks/use-authentication/index.ts
@@ -59,7 +59,6 @@ const useAuthentication = (
 ): UseAuthenticationResult => {
 	// Get router path for `signIn` and `signOut` `callbackUrl`s
 	const router = useRouter()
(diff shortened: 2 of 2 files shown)


LeahMarieBush (commented):
Hello! I checked out the preview and code and they both look good! However, I went to posthog to view the event activity and I'm still seeing multiple `user_authenticated` events when I log in (I only logged in once but I'm seeing 4 calls)
Screenshot 2025-07-09 at 10 46 30 AM

Younique98:
@LeahMarieBush Was that locally in dev or prod posthog?

LeahMarieBush:
> @LeahMarieBush Was that locally in dev or prod posthog?

This was from dev posthog

Younique98:
@LeahMarieBush Thank you so much for that!!I had placed it in the shared auth hook and noticed it was firing multiple times due to how layout-level rerenders and auth state hydration. Hmmm This time i moved it directly into the makesignIn() function. So this should guarantee it only fires once per actual sign-in event. Tested and looking good yay. Once again thanks!!! Still getting use to PostHog, so I truly appreciate it!

LeahMarieBush (approved):
LGTM! Thanks for fixing 🎉

PR #2841, opened 2025-09-17
Make URL parsing regex less greedy

rmainwork:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-rm-fix-malformed-urls-hashicorp.vercel.app/) 🔎
- [Asana task](https://app.asana.com/0/home/1207213190767773/1211156340338030) 🎟️

## 🗒️ What

Adds a `/` to the end of the regex, so that it stops matching at the first `/`. Otherwise, as well as removing a version number from the path, it also tries to remove other places in the URL (like the page slug) so `/upgrade-to-v.1.17.x` becomes `/upgrade-to-`

## 🧪 Testing

- [x] Navigate to https://developer.hashicorp.com/vault/docs/v1.13.x/upgrading/upgrade-to-1.13.x
- [x] Click one of the version dropdown links
- [x] Observe that the version is missing from the URL (making it malformed)
- [x] Navigate to https://dev-portal-git-rm-fix-malformed-urls-hashicorp.vercel.app/vault/docs/v1.13.x/upgrading/upgrade-to-1.13.x
- [x] Click one of the version dropdown links
- [x] Observe that the version is no longer missing from the URL
## 💭 Anything else?

---
  - https://app.asana.com/0/0/1211156340338030

### DIFF
--- a/src/data/alert-banner.json
+++ b/src/data/alert-banner.json
@@ -4,8 +4,8 @@
 		"product": "hashicorp",
 		"tag": "HashiConf 2025",
 		"url": "https://www.hashicorp.com/conferences/hashiconf#livestream",
-		"text": "Don't miss the live stream of HashiConf Day 1 happening now.",
+		"text": "Don't miss the live stream of HashiConf Day 2 happening now",
 		"linkText": "View live stream",
-		"expirationDate": "2025-09-25T14:30:00-05:00"
+		"expirationDate": "2025-09-26T14:30:00-05:00"
 	}
 }


RubenSandwich (changes requested):
@rmainwork any idea why our tests on failing on this change?

rmainwork:
@RubenSandwich not really - I've been investigating this today in among everything else

RubenSandwich:
> @RubenSandwich not really - I've been investigating this today in among everything else...my current frustration is actually that fixing the tests makes the prod code break and fixing the prod code makes the tests break. So there's clearly some mismatch between the two. A large part of that I think is versions without the `v` in the URL (i.e: `/docs/1.17.x/` and the regex I had in there is/was messing with that. The underlying _why_ is something I'm actively investigating though.

@rmainwork Okay. Let's wait till we understand that before merging this in then as we don't want prod to break; especially this close to Hashiconf.

RubenSandwich (approved):
Cool looks good and tests are passing.

@rmainwork I just ask that we don't merge this on a friday; just in case something goes wrong here.

PR #2853, opened 2025-09-24
UDR defined dev-portal product loading

RubenSandwich:
## 🔗 Relevant links

- [Use UDR all-docs-paths endpoint to determine products for preview dev-portal builds](https://app.asana.com/1/90955849329269/project/1205890531436279/task/1211025277045844?focus=true) 🎟️

## 🗒️ What

Changes the webpack loading process so that we can load async configurations. This is required because during dev and preview, but not prod, so that we can ask UDR which products it supports before attempting to load those products. We are doing this because it will make it so that Preview builds in UDR should not fail when new products are added on main.

This PR also makes it so that `getHashiConfig` is no longer exported from `config/index.js` as we want folks to use `loadHashiConfigForEnvironment` as it contains a lot of good default settings for loading configs so it cleans up a lot of code.

## 🧪 Testing

1. Checkout branch: `git checkout rn/udr-defined-product-loading`
2. Edit `.env` to load UDR locally e.g. `UNIFIED_DOCS_API=http://localhost:8080`
3. In your local UDR edit `productConfig.mjs` to change a products name or even comment out a product
4. Make sure that that product is updated in the now logged in dev-portal build e.g.
```
⚠️ Loading UDR Products: [
  ...
]
```

### DIFF
--- a/src/components/navigation-header/components/product-page-content/utils/get-nav-items.ts
+++ b/src/components/navigation-header/components/product-page-content/utils/get-nav-items.ts
@@ -63,6 +63,7 @@ export function getNavItems(currentProduct: ProductData): NavItem[] {
 						icon: navItem.icon as NavigationHeaderIcon,
 						label: navItem.label,
 						path: navItem.fullPath,
+						...(navItem.badge && { badge: navItem.badge }),
 					}
 				}),
 			},
--- a/src/components/sidebar/helpers/generate-product-landing-nav-items.ts
+++ b/src/components/sidebar/helpers/generate-product-landing-nav-items.ts
@@ -23,9 +23,10 @@ import { EnrichedNavItem, SidebarProps } from '../types'
 export const generateProductLandingSidebarMenuItems = (
 	product: ProductData
 ): EnrichedNavItem[] => {
-	const routes = getDocsNavItems(product).map(({ label, fullPath }) => ({
+	const routes = getDocsNavItems(product).map(({ label, fullPath, badge }) => ({
 		title: label,
 		fullPath,
+		...(badge && { badge }),
 	}))
 
 	let docsItems
--- a/src/data/terraform.json
+++ b/src/data/terraform.json
@@ -163,9 +163,13 @@
 		},
 		{
 			"iconName": "docs",
-			"name": "Terraform MCP Server <sup>BETA</sup>",
+			"name": "Terraform MCP Server",
 			"path": "mcp-server",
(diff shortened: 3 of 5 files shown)


.test/app-config.ts
```diff
@@ -1,13 +0,0 @@
...
- */
-
-import path from 'path'
-import { unflatten } from 'flat'
-import { getHashiConfig } from '../config'
-
-const env = process.env.HASHI_ENV || 'development'
-const envConfigPath = path.join(process.cwd(), 'config', `${env}.json`)
-
-export const __config = unflatten(getHashiConfig(envConfigPath))
```
RubenSandwich, line 13: This was only ever used in one test, [src/__tests__/e2e/seo.spec.ts](https://github.com/hashicorp/dev-portal/pull/2853/files#diff-65543d7399a3526230ded80d17ae2289cca6d669b8d8a74195a56a326159592d), so we moved it into that test.

config/base.json
```diff
@@ -36,7 +36,6 @@
 	"flags": {
 		"enable_datadog": true,
 		"enable_io_beta_cta": true,
-		"enable_hvd_on_preview_branch": false,
```
RubenSandwich, line 39: This hasn't been used for a while, so removing it

config/index.js
```diff
@@ -58,4 +104,4 @@ function getHashiConfig(configPath) {
 	}
 }
 
-module.exports = { loadHashiConfigForEnvironment, getHashiConfig }
+module.exports = { loadHashiConfigForEnvironment }
```
RubenSandwich, line 107: People should be using `loadHashiConfigForEnvironment` as it already has sane defaults, `getHashiConfig` needs _you_ to set the file loading and defaults so we should only be using it internally within this file.

.test/setup-vitest.js
```diff
@@ -4,11 +4,25 @@
...
+	if (url.includes(`${process.env.UNIFIED_DOCS_API}/api/supported-products`)) {
+		return Promise.resolve({
+			ok: true,
+			json: () => {
+				return Promise.resolve({
+					result: ['terraform', 'terraform-enterprise', 'well-architected-framework']
+				})
+			}
+		})
+	}
```
LeahMarieBush, line 22: Should this be mocked for the tests that need it instead of here?

LeahMarieBush (approved):
Looks good! Left one question

LeahMarieBush (approved):

PR #2985, opened 2026-04-14
Override sub deps to axios 1.15.0

RubenSandwich:
## 🗒️ What

Override any axios sub dependencies to  >= 1.15.0. It seems like `posthog-node` was the only dependency using axios.

## 🤷 Why

Current active CVE for axios <= 1.13.0.

## 🧪 Testing

1. Test to make sure posthog analytics, feature flags, etc still work.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15496,14 +15496,14 @@
 			}
 		},
 		"node_modules/axios": {
-			"version": "1.13.5",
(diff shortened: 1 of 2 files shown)


LeahMarieBush (approved):
LGTM 👍 Checked posthog and I am seeing event firing for this preview link

PR #2987, opened 2026-04-22
Version bump MDS packages to fix copyblock behavior

rmainwork:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

Fix a bug with the code block MDS component where only the first line of code was copied for heredoc blocks that ended with anything other than `EOF`.

## 🤷 Why

Two support tickets were raised regarding this behavior within a few weeks of each other. So, to provide a better experience to content authors, this bug has been fixed.

## 🧪 Testing

- [ ] Bring UDR up locally
- [ ] Bring devdot up locally (pointing at UDR)
- [ ] Add the following heredoc snippet in a codeblock  to a doc of your chosing:

```mdx
    $ kubectl apply -f - <<'OUTER'
    apiVersion: apps/v1
    kind: Deployment
    metadata:
                port: 9000
                initialDelaySeconds: 5
                periodSeconds: 5
    OUTER
```

- [ ] Click the "Copy" button at the top. 
- [ ] On `main` it should only copy the first line. 
- [ ] Check out this branch out (running `npm ci` to completely refresh node dependencies) and restart dev portal
- [ ] Click the copy button again on the same doc and the entire snippet should be copied to the clipboard


williamdalessandro (approved):
I see it working locally! Also an interesting thing I found is that if I copy the codeblock with a bunch of indented whitespace, it works on main and doesn't just copy the first line

rmainwork:
@williamdalessandro  yeah, i guess that breaks out of the regex matching behavior that it's using to look for `EOF`

PR #2991, opened 2026-05-08
fix sitemap generation

RubenSandwich:
## 🗒️ What

Our `/sitemap.xml` and `robots.txt` generation broke. (Lightly through an update of next-sitemap) This PR fixes that.

## 🧪 Testing

Current:

[/sitemap.xml](https://developer.hashicorp.com/sitemap.xml)
[robots.txt](https://developer.hashicorp.com/robots.txt)

Expected (This PR):

[/sitemap.xml](https://dev-portal-git-rn-fix-sitemaps-hashicorp.vercel.app/sitemap.xml)
[robots.txt](https://dev-portal-git-rn-fix-sitemaps-hashicorp.vercel.app/robots.txt)

## PCI review checklist

- [X] I have documented a clear reason for, and description of, the change I am making.

- [X] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [X] If applicable, I've documented the impact of any changes to security controls.


LeahMarieBush (approved):
Looks good! I noticed the `hcp-sitemap.xml` does not work when I was testing. It also doesn't work in prod so I'm not super concerned but is that something we need to fix/update? I'm not 100% sure what the hcp-sitemap should even be

PR #3007, opened 2026-06-10
Add Instana monitoring to devdot

rmainwork:
## 🔗 Relevant links

- [Preview link](https://dev-portal-git-rm-add-instana-rum-hashicorp.vercel.app/) 🔎
- [Monday Task](https://ibm.monday.com/boards/18408514043/pulses/12187692566) 🎟️

## 🗒️ What

Adds Instana EUM dashboard to devdot to (eventually) replace DataDog at the end of June. DataDog is not being replaced (yet) as the intent is to keep both going for an A/B comparison.

## 🤷 Why

IBM is not renewing the DataDog subscription and so it will lapse at the end of June.

## 🧪 Testing

- [ ] Navigate to any page in devdot(local or [preview](https://dev-portal-git-rm-add-instana-rum-hashicorp.vercel.app/)) and throw an exception in the console (for best results, wrap it in a `setInterval` to throw lots and light up the EUM dashboard like a Christmas tree).
- [ ] Go to [the instana sandbox devdot dashboard](https://test-hcp.instana.io/#/websiteMonitoring/website;websiteId=d-CdMGjjTKSIMaufhBqFlg/summary?timeline.to&timeline.fm&timeline.ar=true) and you should see the errors thrown appear in there.

image

image

image

image

PR #3014, opened 2026-06-17
Switch to TFC varset for instana

rmainwork:
Instead of keeping the values in state for this project, we can instead use org-wide varsets to point to production (or testing) instana instances

## 🗒️ What

Since the original PR (#3007) was closed, the iops team were kind enough to provide us with a terraform cloud varset we can use between projects and workspaces. 

## 🤷 Why

This will make it easier to add Instana/terraform to other apps (like UDR) going forward and keep all the credentials consistent.

## 🧪 Testing

- Kick off a new terraform run (or look at the run output attached to this PR)
- If the run passes, it worked

### DIFF
--- a/terraform/terraform.tf
+++ b/terraform/terraform.tf
@@ -18,6 +18,8 @@ terraform {
 }
 
 provider "instana" {
-  api_token = var.instana_production_endpoint_token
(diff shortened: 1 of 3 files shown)


RubenSandwich (approved):
🚢

PR #3021, opened 2026-06-25
Add slack notifications for RUM errors

rmainwork:
## 🗒️ What

Add Instana based slack alerts for RUM errors (JS errors) on devdot

## 🧪 Testing

- Comment out the `backend` stanza in `terraform.tf` to force the terraform to run using local state
- Run the terraform against test-hcp(you may have to grab [an API token](https://test-hcp.instana.io/#/config/securityAndAccess/accessControl/apiTokens?timeline.to=1782414253000&timeline.fm=1782414253000&timeline.ar=false&timeline.ws=1004653000) for test-hcp if you haven't already
- You may have to add an `import` block to your terraform code to import the existing website resource into the terraform code otherwise it will cause a naming conflict. You could also just delete or rename the existing website resource if you like.
- Run devdot locally as normal
- Throw a bunch of errors in the console
- You should see a smart alert fire

### DIFF
--- a/terraform/instana.tf
+++ b/terraform/instana.tf
@@ -2,3 +2,58 @@
 resource "instana_website_monitoring_config" "devdot" {
   name = var.website_name
 }
+
+resource "instana_alerting_channel" "slack" {
+  name = var.slack_channel_name
+  slack_app = {
+    app_id       = var.slack_app_id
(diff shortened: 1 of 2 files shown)


RubenSandwich (approved):
🚢

PR #3022, opened 2026-06-26
Add support for shouldNoIndex to OpenAPI pages, add Infragraph API page.

BrianMMcClain:
## 🔗 Relevant links

- [Asana task](url) 🎟️

## 🗒️ What

This PR adds a `shouldNoIndex` property to the OpenAPI view which renders the `no-index-tag-for-restricted-docs` component. This is separate from the `isVersionedUrl` property, which also adds the `noindex` tag but as I understand also would tread the API doc as versioned, even if the API itself is not versioned, adding unrelated dropdowns and alerts.

This PR also adds the page for the Infragraph API docs.

## 🤷 Why

As I understand it, the mechanism to add a `noindex` tag to an API page can only be done if you set `isVersionedUrl` to `true`, however this also adds other components related to versioned API docs. I don't believe you can _only_ add the `noindex` tag today.

This approach reuses the existing `no-index-tag-for-restricted-docs component used in our docs view to only add a `noindex` tag. This will let us support API documentation that we want public, but that we do not want crawled by search engines. 

## 🛠️ How

The implementation worked backwards from some of the existing mechanisms but I'm open to other approaches. Discovering the existing component pointed me in the direction of reusing as much as I can, but I know this was an implementation previously discussed so happy to make changes.

## 🧪 Testing

- [ ] Identify an API docs page that is not versioned (ie. Waypoint)
- [ ] Verify the rendered page source does not include a `noindex` tag
- [ ] Add `shouldNoIndex: true` to the `[[...page]].tsx` `PAGE_CONFIG`
- [ ] Reload the rendered page and verify it includes the expected `noindex` tag

## PCI review checklist

- [x] I have documented a clear reason for, and description of, the change I am making.

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [ ] If applicable, I've documented the impact of any changes to security controls.

### DIFF
--- a/src/pages/hcp/api-docs/infragraph/[[...page]].tsx
+++ b/src/pages/hcp/api-docs/infragraph/[[...page]].tsx
@@ -119,5 +119,4 @@ export const getStaticProps: GetStaticProps<
 	return await generateStaticPropsVersioned(PAGE_CONFIG, params?.page)
 }
 
-
 export default OpenApiDocsViewV2
--- a/src/pages/hcp/api-docs/infragraph/[[...page]].tsx
+++ b/src/pages/hcp/api-docs/infragraph/[[...page]].tsx
@@ -5,11 +5,9 @@
 
 // View
 import OpenApiDocsViewV2 from 'views/open-api-docs-view-v2'
-import {
-	generateStaticPaths,
-	generateStaticPropsVersioned,
-} from 'views/open-api-docs-view-v2/server'
+import { generateStaticPropsVersioned } from 'views/open-api-docs-view-v2/server'
 // Utils
+import { isDeployPreview } from 'lib/env-checks'
 import { getOperationGroupKeyFromPath } from 'views/open-api-docs-view-v2/utils/get-operation-group-key-from-path'
 // Schema transforms
 import { schemaTransformShortenHcp } from 'views/open-api-docs-view-v2/schema-transforms/schema-transform-shorten-hcp'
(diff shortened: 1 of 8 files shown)


BrianMMcClain:
Vercel build is failing because the spec does not currently exist. We are making the finial decisions on publishing this API, but we will ensure the spec is uploaded appropriately and rerun tests if we agree this is the appropriate path for this.

src/views/open-api-docs-view/types.ts
```diff
@@ -177,6 +177,12 @@ export interface OpenApiDocsViewProps {
 	 */
 	versionAlert?: ApiDocsVersionAlertProps
 
+	/**
+	 * Optional. If set to `true`, a `noindex` will be added to the page even
+	 * if `isVersionedUrl` is false.
+	 */
+	shouldNoIndex?: boolean
```
BrianMMcClain, line 184: This page is where I was the least confident on the changes and if we need to define this here as well as below. It seemed to be the case based on the error I received, but wanted to flag.

src/pages/hcp/api-docs/infragraph/[[...page]].tsx
```diff
@@ -0,0 +1,93 @@
...
+
+	/**
+	 * Massage the schema data a little bit
+	 */
+	massageSchemaForClient: (schemaData: OpenAPIV3.Document) => {
+		//  Replace "HashiCorp Cloud Platform" with "HCP" in the title
+		const withShortTitle = schemaModShortenHcp(schemaData)
+		// Return the schema data with modifications
+		return withShortTitle
+	},
```
Comment (LeahMarieBush) [inline src/pages/hcp/api-docs/infragraph/[[...page]].tsx]: I would recommend copying the `massageSchemaForClient` from [this file](https://github.com/hashicorp/dev-portal/blob/fd119f3c2fca1b412040a248644616e0b5ac47d7/src/pages/hcp/api-docs/consul/%5B%5B...page%5D%5D.tsx#L69). Although I don't know if this will be in the api-docs since the spec is not published, changing to the same as this file will cover if `protobufAny` is in the spec.

LeahMarieBush:
Other than my comment above this looks good! I can approve if you tag me once the spec is live

src/pages/hcp/api-docs/infragraph/[[...page]].tsx
Comment (BrianMMcClain) [reply src/pages/hcp/api-docs/infragraph/[[...page]].tsx]: Thanks! I've updated this method to match the recommendation and added the required imports in d17eac7

RubenSandwich (approved):

PR #3028, opened 2026-07-01
Add alert grace period

rmainwork:
## 🗒️ What

Updates the smart alert configuration to enable configuration of the cooldown/grace period of alerts. In order words, the amount of time before the metric is considered breached again. This helps prevent noise and alert fatigue so that (hopefully) only *new* issues trigger alerts. 

The default value is 5 minutes which is the same as the default value, but I prefer having a value explicitly set rather than just relying on default values to hopefully make it a bit more obvious how things are configured...

## 🤷 Why

Ideally this would've been covered by #3021 but the Instana provider did not support the configuration of this value at the time. In version 7.3.1 of the provider this is now supported (hence the provider version bump).

## 🧪 Testing

- Sign into HCP
- Read the plan output
- A value of 5 minutes (which is also the default) should now be explicitly set for the grace period.

### DIFF
--- a/terraform/instana.tf
+++ b/terraform/instana.tf
@@ -15,12 +15,12 @@ resource "instana_alerting_channel" "slack" {
 }
 
 resource "instana_website_alert_config" "js_errors" {
-  name        = "${var.website_name} new RUM error detected"
-  description = "${var.website_name} new RUM error detected"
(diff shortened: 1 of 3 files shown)


williamdalessandro (approved):
LGTM!

PR #3029, opened 2026-07-07
Add build metrics

RubenSandwich:
## Description

Adds Instana logging to the following things:
1. build metrics
2. repo sync GHA failures
3. doc content not found errors

## Testing

Check that metrics exist in the dashboard: https://prod-hashicorp.instana.io/#/customDashboards;query=deve/view;dashboardId=Ny24CKFwSBO2Kg4btaONHg;tagFilterExpression=!~?timeline.ws=86400000&timeline.to&timeline.fm&timeline.ar=false

### DIFF
--- a/.github/workflows/repo-sync.yml
+++ b/.github/workflows/repo-sync.yml
@@ -151,3 +151,18 @@ jobs:
             })
             // Error loud here, so no try/catch
             console.log('Merged the pull request successfully')
+
+      - name: Emit repo-sync failure span
+        if: failure()
+        run: npx hc-tools ./scripts/emit-otel-span-cli.ts
+        env:
+          # parameters
+          SPAN_NAME: repo-sync failed
+          SPAN_STATUS_MESSAGE: repo-sync workflow failed
+          SPAN_SCOPE_NAME: github-actions
+          SPAN_SERVICE_NAME: developer.hashicorp.com
+          SPAN_ATTRIBUTES: >-
+            {"ci.workflow":"${{ github.workflow }}","ci.run_id":"${{ github.run_id }}","ci.run_url":"${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}","git.ref":"${{ github.ref }}"}
+          # secrets
+          INSTANA_OTLP_ENDPOINT: ${{ secrets.INSTANA_OTLP_ENDPOINT }}
+          INSTANA_OTLP_API_TOKEN: ${{ secrets.INSTANA_OTLP_API_TOKEN }}
--- a/.github/workflows/repo-sync.yml
+++ b/.github/workflows/repo-sync.yml
@@ -34,6 +34,7 @@ jobs:
           github_token: ${{ secrets.PAT_HASHIBOT_WEB_DEV_PORTAL_INTERNAL_SYNC }}
 
       - name: Ship pull request
+        id: ship_pr
         uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
         with:
           github-token: ${{ secrets.PAT_HASHIBOT_WEB_DEV_PORTAL_INTERNAL_SYNC }}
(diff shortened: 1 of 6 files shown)


scripts/capture-build-metrics.ts
```diff
@@ -31,56 +35,172 @@ async function readNextTrace(cwd: string) {
...
+			.map((result) => {
+				return (result as PromiseRejectedResult).reason
+			})
+		if (failedSubmissions.length > 0) {
+			throw new AggregateError(failedSubmissions)
+		}
+	} catch (error) {
+		// Swallow errors
+		// we don't want to impact the build or make it seem like there's been
+		// an error in the actual app if something goes wrong when sending metrics
```
rmainwork: Minor thing, but we might want to either tweak or entirely remove this comment since we _are_ throwing errors....just not in prod. Either way - I wouldn't hold up the PR over it.

scripts/emit-otel-span.ts
```diff
@@ -0,0 +1,186 @@
...
+ * delivery when running in short-lived environments like CI. In fire-and-forget
+ * contexts the promise can be ignored.
+ */
+export function emitOtelSpan({
+	span,
+	serviceName = DEFAULT_SERVICE_NAME,
+	scopeName,
+	hostId = process.env.INSTANA_HOST_ID ?? serviceName,
+	endpoint = process.env.INSTANA_OTLP_ENDPOINT,
+	apiToken = process.env.INSTANA_OTLP_API_TOKEN,
```
rmainwork: I would maybe suggest renaming this to `agentToken`/`INSTANA_OTLP_AGENT_TOKEN` or something to that effect as api keys and agent tokens aren't the same thing...

When you've done that, I'll go ahead and update #3032 with the new env var name

RubenSandwich: @rmainwork done

.github/workflows/repo-sync.yml
```diff
@@ -151,3 +154,25 @@ jobs:
...
+                "ci.run_url": "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}",
+                "git.ref": "${{ github.ref }}",
+                "pr.url": "${{ steps.ship_pr.outputs.pr_url }}"
+              }
+            }' \
+            --scope-name '.github/workflows/repo-sync.yml' \
+            --service-name 'developer.hashicorp.com' \
+            --host-id 'github-actions'
+        env:
+          INSTANA_OTLP_ENDPOINT: ${{ secrets.INSTANA_OTLP_ENDPOINT }}
```
rmainwork, line 177: Does this value _need_ to be a secret or could we just just use a var here?  Right now I've got this value as a var in #3032, but if it needs to be a secret let me know and I can update my PR so we can meet in the middle.

RubenSandwich: It does not need to be, but our team has generally only used secrets. I guess to keep everything in one tab? 🤷🏻‍♂️

rmainwork: My only reservation is that it makes it hard to see what the current value is. Obviously for secrets we don't want that but for something like an endpoint URL it might make sense. It can just make it easier when things change to know what the expected value looks like. What do you think?

scripts/emit-otel-span.ts
```diff
@@ -0,0 +1,186 @@
...
+ * delivery when running in short-lived environments like CI. In fire-and-forget
+ * contexts the promise can be ignored.
+ */
+export function emitOtelSpan({
+	span,
+	serviceName = DEFAULT_SERVICE_NAME,
+	scopeName,
+	hostId = process.env.INSTANA_HOST_ID ?? serviceName,
+	endpoint = process.env.INSTANA_OTLP_ENDPOINT,
+	apiToken = process.env.INSTANA_OTLP_AGENT_TOKEN,
```
rmainwork, line 139:
```suggestion
	agentToken = process.env.INSTANA_OTLP_AGENT_TOKEN,
```

We might want to rename this too, just to keep it consistent with the env var. I wouldn't bother over it normally, but Instana does have API tokens that behave slightly differently to agent tokens...

rmainwork (approved):
Looks good - a small note about var naming, but I'm not going to hold the review up over it.


Also let me know if you want to review and merge #3032 into this or want me to just wait and rebase it off `main`.

PR #3037, opened 2026-07-10
Remove old import blocks

rmainwork:
## 🗒️ What

Removes old terraform import blocks to clean up the code

## 🤷 Why

The import blocks only existed to bring existing resources under Terraform management. Now that this has been done, those import blocks are basically a no-op. There's no harm keeping them, but they clutter up the config, so it's best to just get rid of them

## 🧪 Testing

- Open the terraform plan and make sure no changes are indicated (there should be no changes planned)

### DIFF
--- a/terraform/instana.tf
+++ b/terraform/instana.tf
@@ -60,12 +60,6 @@ resource "instana_website_alert_config" "js_errors" {
   }
 }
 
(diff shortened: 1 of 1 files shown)


RubenSandwich (approved):
👍🏻

PR #3070, opened 2026-07-31
Terraform housekeeping

rmainwork:
Add a terraform ignore file to prevent terraform uploading the entire project to HCP (rather than just terraform/). This helps speed up plan times quite a bit.

Also unignores `.terraformignore` so that we can commit it


ryan-yu-ibm (approved):
lgtm!

PR #3079, opened 2026-08-11
Remove "HCP Button" from Consul

rmainwork:
HCP Consul was EOL last November, so we don't want the "Try HCP Consul"
button in the navbar anymore


RubenSandwich (approved):
🚢

PR #3098, opened 2026-09-08
feat: Replaced process.env vars with env-var

ryan-yu-ibm:
## 🔗 Relevant links

- [Monday task](https://ibm.monday.com/boards/18402251594/views/240427671/pulses/10910804320) 🎟️

## 🗒️ What

This PR reinforces our conditional checks made throughout dev-portal when involving environment variables. An instance of this would be the string `'true'` versus the boolean `true`. It is very easy to switch between the two and could unintentionally change our dev environment.

## 🤷 Why

Future proofing our configs and functions that does environmental variable comparisons.

>[!NOTE]
I chose to not replace every `process.env` variable out with `env-var` since I interpreted the task as patching the areas that could be potential points of failures (i.e. places with conditional checks). Though, if desired, I can revamp all `process.env` variables to use `env-var`.

## 🛠️ How

1. Installed env-var with npm; Looked into [env-var documentation
](https://github.com/evanshortiss/env-var/blob/HEAD/API.md)
2. Replaced `process.var.VERCEL_ENV` to the env-var equivalent within `next.config.js` to test functionality
3. Queried Bob to comb through repo for instances of process.env that do a direct comparison
4. Provided example of how to change those instances and had it revise with me before applying any changes
5. Made sure the local build still started after env changes

## 🧪 Testing

1. Open the [preview](https://dev-portal-k5dmd0bpy-hashicorp.vercel.app/) and make sure the site still functions
2. Note if anything breaks or is off

## PCI review checklist

- [x] I have documented a clear reason for, and description of, the change I am making.

- [x] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [x] If applicable, I've documented the impact of any changes to security controls.


src/pages/_app.tsx
```diff
@@ -50,7 +51,10 @@ import { ConditionalPostHogProvider } from 'components/posthog/posthog-provider'
 // Local imports
 import './style.css'
 
-if (typeof window !== 'undefined' && process.env.AXE_ENABLED === 'true') {
+const AXE_ENABLED = env.get('AXE_ENABLED').asBool()
+const NODE_ENV = env.get('NODE_ENV').asString()
```
williamdalessandro, line 55:
```suggestion
const clientEnv = from({
	AXE_ENABLED: process.env.AXE_ENABLED,
	NODE_ENV: process.env.NODE_ENV,
})
const AXE_ENABLED = clientEnv.get('AXE_ENABLED').asBool()
const NODE_ENV = clientEnv.get('NODE_ENV').asString()
```

I found an issue here when it come to how the value for `AXE_ENABLED` is set at runtime. Due to how env-var actually evaluates/injects env variables and interacts with nextjs, sometimes certain ways values are passed in can trip it up (specifically how `AXE_ENABLED` is coming from an object in [next.config.js](https://github.com/hashicorp/dev-portal/blob/main/next.config.js#L113)) so right now it is being set as `undefined` and not actually working as we normally expect. Here from the env-var documentation it describes a use-case to get around this shown [here](https://github.com/evanshortiss/env-var/blob/master/README.md#webapp-example). There may be other situations like this that arise from using env-var in client-side code to grab values from next.config.js, so just look for more of those.

The exact situation going on is this:
`AXE_ENABLED` (and other values from `next.config.js`) is injected into the browser bundle by Next.js (via webpack) and does static text substitution at build time. Next.js' webpack bundler finds the _**exact string**_ `process.env.AXE_ENABLED` in source files and replaces it with its value. env-var on the other hand internally reads from `process.env` using bracket notation ([process.env['AXE_ENABLED']](https://github.com/evanshortiss/env-var/blob/master/lib/variable.js#L56)). Since webpack is specifically looking for `process.env.AXE_ENABLED` and NOT `process.env['AXE_ENABLED']`- webpack doesn't recognize this so the substitution never fires and the browser sees the value as `undefined`.

src/pages/_app.tsx
```diff
@@ -4,6 +4,7 @@
  */
 
 // Third-party imports
+import env from 'env-var'
```
williamdalessandro, line 7:
```suggestion
import { from } from 'env-var'
```

williamdalessandro (changes requested):
Hey @ryan-yu-ibm good job on this so far! I agree with not replacing the use of process.env everywhere and found an edge case with env-var and next.js's webpack bundler that I detailed below.

At first I wasn't really sure if we really wanted to use env-var just because it adds another package to the project and it hasn't been updated in [2 years](https://github.com/evanshortiss/env-var/releases). However after looking into it, it is extremely lightweight and doesn't add any runtime dependencies so I don't mind it hanging around in that case

PR #3104, opened 2026-09-16
Remove support articles from search

rmainwork:
## 🗒️ What

Remove support tab from search box

## 🤷 Why

Zendesk has been sunset but the associated redirect simply takes users to the IBM support home page. Down the road we will add IBM knowledgebase support but for the time being it's a better UX to simply remove the tab entirely so that users aren't just sent to the IBM support home page.

This PR reverts #2732 

---

Some additional cleanup work is probably needed post-merge to go turn off the Zendesk/Algolia connector and remove the records in there


src/components/command-bar/commands/search/unified-search/components/unified-hit/index.tsx
williamdalessandro, line 8:
```suggestion
```

williamdalessandro, line 14:
```suggestion
```

williamdalessandro, line 22:
```suggestion
```

williamdalessandro (changes requested):
Just requesting to remove these unused imports to silence the linter- otherwise it all looks good!

PR #3106, opened 2026-09-18
fixing pr closing bug for dev-portal-internal

williamdalessandro:
## 🗒️ What

Hashibot kept closing open PRs in dev-portal internal every time there was a repo sync. Turns out it was because the command to grab the list of open **_repo-sync_** prs was actually grabbing **_every_** open PR, so it was closing every open PR in dev-portal-internal specifically. Changed the behavior to match exactly what we're doing in UDR. Also removed some logic that locks the PR conversations because I don't see much use for it. 

Specifically with the github api `github.rest.pulls.list({ owner, repo, head, base })`- `head` is used as a filter and the api requires `owner:branch` format for it to work. Without the owner prefix, it ignores the filter and returns all open PRs.

## 🤷 Why

Hashibot bullying me
image

## 🛠️ How

[Matched the logic from UDR pretty much 1:1](https://github.com/hashicorp/web-unified-docs/blob/main/.github/workflows/repo-sync.yml#L56-L57)

## 🧪 Testing

If you want to see exactly what was happening, you can run these CURL commands to see it

- First grab your gh auth token by running in terminal:
`gh auth token`

- Take that token and paste them into the curl commands below

With owner:branch format (correct)

```
curl -H "Authorization: Bearer <insert your gh auth token here>" \
  "https://api.github.com/repos/hashicorp/dev-portal/pulls?state=open&head=hashicorp:repo-sync&base=main"
```

Without owner: prefix (what the broken code was doing)

```
curl -H "Authorization: Bearer <insert your gh auth token here>" \
  "https://api.github.com/repos/hashicorp/dev-portal/pulls?state=open&head=repo-sync&base=main"
```

- Since there arent any open repo-sync prs right now, hte first command should return an empty list, however since the second one is using the incorrect command it will return every open PR

## PCI review checklist

- [ ] I have documented a clear reason for, and description of, the change I am making.

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [ ] If applicable, I've documented the impact of any changes to security controls.


RubenSandwich (changes requested):
~~I don't believe that this PR fixes this problem, here is the start of code that I believe is causing it:~~

https://github.com/hashicorp/dev-portal/blob/49cac02894a06281c8b918dcdde7d2b4ca281880/.github/workflows/repo-sync.yml#L59

Edit: Ignore me, I was wrong.

RubenSandwich (approved):
🚢

